PR #5, opened 2016-09-17
Update the README

cdrage:
A bunch of things are happening in this PR:
  - Moving from .rst to markdown for the root directory
  - Updating the README with the relevant fork information
  - Removing LICENSE.flashmedia and LICENSE.pbs (these are imported from
    upstream packages)
  - Removing CONTRIBUTING.rst, we'll write a new manifest for this in
    the future that is more leniant than the past one.


gravyboat:
I believe we need to keep the associated licenses since we're still using those pieces of software in the project correct? Otherwise this looks good to me.

cdrage:
@gravyboat Yeah, the licenses stay the same, but LICENSE.bsd and LICENSE.flash are redundant (both use the same licenses upstream anyways and are in the references packages)

cdrage:
@fishscene maybe we should remove the CHANGELOG since we'll be doing our own? I'll update the PR.

cdrage:
Usually I'd wait for another ACK... But I've got to make an exception with this.

Going to merge this in since we have many people going to github.com/streamlink/streamlink with no info at the moment that it's a fork.

Feel free to open up another PR to revert the change if you're unhappy with it and/or want to update the README!

gravyboat:
I'm fine with it, changes looked good and my concerns regarding the licenses were addressed.

fishscene:
Sorry for the late reply. Weekend family time. :) If it's possible for now to keep a pre-fork legacy changelog, that might be useful to reference for now. If we can also keep giving credit to previous contributors, I think that'd be a great way to honor the fork. Maybe have a legacy/license folder where we can stuff these things in to? Just an idea.

PR #6, opened 2016-09-17
Add some python releases

gravyboat:
Add in some Python releases for Travis and clean a few things up, more modifications are needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,23 @@
 language: python
 sudo: false
 python:
-- 2.6
-- 2.7
-- 3.3
+  - "2.6"
+  - "2.7"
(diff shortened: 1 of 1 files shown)


gravyboat:
Looks like the travis ci artifacts stuff is no longer supported in the way it is used here (https://docs.travis-ci.com/user/uploading-artifacts/). For the time being since we want this to remain open and they only support S3 I'm going to remove the artifacts. We're also several releases behind on faraday. Will modify that later once we get tests passing again.

gravyboat:
Tests are passing again (though there are some errors in Travis), this should be good to merge.

cdrage:
LGTM :+1:

PR #10, opened 2016-09-19
add py35 to tox

SpootDev:


cdrage:
Hey @SpootDev do you mind adding a description and such before doing any changes? also, adding this to master would be best ^^.

Having at least a description / one ACK from someone else would work best.

Issue #2, opened 2016-09-16
Set up Coveralls


gravyboat:
I'm a big fan of Coveralls for test coverage details (https://coveralls.io/), would be nice if we set that up on this repository.

fishscene:
I don't have an opinion one way or the other, but if @gravyboat thinks this will help and can take ownership of this, let's do it.
Does anyone have further thoughts on this?

gravyboat:
I'll take care of it if no one has concerns, was going to do it this weekend and didn't review the docs enough to figure out how to handle it for an org.

cdrage:
@gravyboat if you set it up with third-party support, i'll get a notification about allowing it and I can accept it. +1 for coveralls support

cdrage:
@gravyboat there we go! i set it up, should get notifications on each PR opened now

gravyboat:
@cdrage Good stuff, I looked in the coveralls interface this weekend and it wasn't showing up for some reason. Glad to see you got it!

gravyboat:
This is now working as per the example PR here: https://github.com/streamlink/streamlink/pull/13#event-794486121, closing as resolved.

gravyboat:
Looks like there is a minor configuration issue causing the coverage to show as no data even though it is being sent, there are several errors within Travis which I'll investigate. Opening back up for right now.

cdrage:
@gravyboat I think `--cov livestreamer` needs to be added

gravyboat:
@cdrage More than likely that is the issue, previously I ran it through nosetests so it operated a bit differently. I'll prep a PR later today, as well as adding in a `.coveragerc` file so we don't get so much garbage.

gravyboat:
Here are some pretty good instructions, will implement this later today: https://github.com/audreyr/how-to/blob/master/python/use_coverage_with_unittest.rst#set-up-coveralls

cdrage:
@gravyboat found the issue, here's the PR: https://github.com/streamlink/streamlink/pull/16

PR #16, opened 2016-09-19
Fix travis

cdrage:


cdrage:
Closes #2

gravyboat:
Good stuff, thanks!

gravyboat:
I don't think this will work as those options are only supported by pytest, and this repo is using the standard unittest (not a huge fan myself but let's not stray into those weeds for now). I think modifying your line to be `script: coverage run --source streamlink setup.py test` should work.

cdrage:
@gravyboat what about installing pytest? I quite like the output that pytest uses in comparison to unittest.

I've updated the PR, I'll see if this one passes this time.

gravyboat:
I'm fine with it, haven't used pytest a ton so as long as it supports the existing unit tests (which I believe it should) then it's cool with me.

sbstp:
The build failed, I guess you can't use sudo in there. I think you can install pip packages without sudo.

cdrage:
argh, i forgot about that, i'll update it again. 3rd times the charm, eh?

gravyboat:
Awesome job, thanks @cdrage!

PR #24, opened 2016-09-20
Update plugin.douyutv

steven7851:
make https as default


sbstp:
I tested it and it works.

PR #26, opened 2016-09-20
Add script to build & copy the docs

sbstp:
This script is pretty simple, it builds the docs, then copies them in streamlink.github.io repo, replacing the old files.


cdrage:
Yay! Thanks for the hard work @sbstp 

Btw, what about the README? (making sure README.md in streamlink/streamlink is the same as the docs site?

sbstp:
@cdrage yea, could do that, unless we want to put a custom readme indicating that the docs are generated and should be modified in the other repo.

gravyboat:
I think we should have a custom readme indicating the docs are generated in the main repo, :+1: to that idea. Then we don't have people confused. Better yet, if the readme had instructions on how to contribute to the docs that linked to the main repo it would be awesome.

sbstp:
@gravyboat Yea, I thought about it but it's slightly complex. We need to encrypt a github access token with travis' public key so that it can decrypt it and push to the repo.

sbstp:
Also, we have like 5-6 builds running for every commit, but we only need one to generate the docs. If they all try to push it's gonna be a race condition.

gravyboat:
Good point, for the time being I'm definitely fine with this, it would be nice at some point if we automated the whole process at the very least for branched/tagged releases!

sbstp:
I found a  [page](https://docs.travis-ci.com/user/customizing-the-build#Explicity-Including-Jobs) that could help us doing what we want. We could pick a specific version and tell it to build the docs. But it should only be executed on the master branch, so that's another isssue.

sbstp:
This is a draft of what it might look like. It uses the Python 3.5 build to publish the docs. So the `publish-docs.sh` script is only executed on Python 3.5 and on the master branch.
```yaml
language: python

sudo: false

matrix:
  include:
  - python: "2.6"
  - python: "2.7"
  - python: "3.3"
  - python: "3.4"
  - python: "3.5"
    env: BUILD_DOCS=yes


before_install:
  - pip install pytest pytest-cov coveralls coverage
  - travis_retry gem install --version 0.8.9 faraday

install:
  - python setup.py install

script:
  - python -m pytest -vv --cov streamlink

deploy:
  provider: script
  script: publish-docs.sh
  on:
    branch: master
    condition: $BUILD_DOCS=yes

after_success:
  - sh .travis/build-win32.sh
  - coveralls

```

sbstp:
I think we should get the docs to auto publish everytime something is pushed to master and do the releases manually. I don't think there's a safe way for travis to push to pypi.

cdrage:
@sbstp +1 for me for this, wanna merge this in and test it out with some documentation changes? I was just looking at our docs and some of them need to be edited :)

src/streamlink/plugins/douyutv.py
```diff
@@ -94,9 +94,9 @@ def _get_streams(self):
         match = _url_re.match(self.url)
         channel = match.group("channel")
 
-        http.headers.update({"User-Agent": USER_AGENT})
+        http.headers.update({'User-Agent': USER_AGENT})
         http.verify=False
-        http.mount('http://', HTTPAdapter(max_retries=99))
+        http.mount('https://', HTTPAdapter(max_retries=99))
```
cdrage: dont know how this got in here, you''ll need to rebase against master ^^

cdrage:
nvm, there's a commit conflict ^^.

If you could open up a separate PR for the travis modified script that would be great! That means that any changes in this repo will replicate to the docs repo.

sbstp:
I'm not sure how that commit made it in here, I'll rebase. The travis script isn't ready yet, we need to setup an ssh key pair. I'll test that stuff in another repo.

sbstp:
@cdrage ready for another review.

cdrage:
LGTM! Lets merge.

sbstp:
Neat. I've just finished testing the automatic publish. If you want to take a look at the demo:
Source repo (streamlink) : https://github.com/sbstp/travis-test
Docs repo (streamlink.github.io) : https://github.com/sbstp/travis-test-target

cdrage:
@sbstp dude, you're awesome. this looks great! :dancers:

PR #28, opened 2016-09-21
Add script to generate authors list / update authors

cdrage:
This adds a script to generate authors list based upon the git log. This
also updates the AUTHORS list to a more up-to-date version with everyone
whom has ever committed to the project.


gravyboat:
Good stuff!

PR #30, opened 2016-09-21
Add release script

cdrage:
This is a script I use in order to release something on Github as well
as push onto Pypi. I use this with a similar project called "kubeshift".

This release script will automatically process things such as renaming
the versioning, creating changelogs as well as uploading to Pypi.

After this is merged, we'll go ahead and do our first release of
Streamlink!

### DIFF
--- a/script/release.sh
+++ b/script/release.sh
@@ -1,0 +1,303 @@
+#!/bin/bash
+UPSTREAM_REPO="streamlink"
+CLI="streamlink"
+
+usage() {
+  echo "This will prepare $CLI for release!"
+  echo ""
+  echo "Requirements:"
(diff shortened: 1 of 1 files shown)


gravyboat:
Looks good to me, I would like to see this added to the Travis build at some point so we can do all the builds automatically going forward (and packaging and all that nonsense).

cdrage:
@gravyboat may be a bit difficult in terms of releases, especially since the changelog is manually written!

Heads up, I should have the release out by tmrw

gravyboat:
We'll definitely still have to write release notes. I'm primarily talking about packaging building as it could be part of this at some point in the future.

PR #36, opened 2016-09-23
Automated docs updates via travis-ci

sbstp:
> Uses the Python 3.5 build to generate the docs. Then, if there are any changes, it pushes them to the streamlink.github.io repo using a key pair. The deploy-key.enc file here is the private key of that pair. The public key was added to the streamlink.github.io repo as a deploy key.

This is pretty hard to test on the live repo, so we might have to try a couple of times.


cdrage:
Hey @sbstp do you need more admin privileges in order to test this functionality? Good job man on the changes, although I didn't see any updates to streamlink.github.io repo through this PR.

sbstp:
The script only runs when something is pushed to master, otherwsie someone could just send a pull request modifying the `.travis.yml` and `cat` the private key.

I can't test it because the secure information is only available on pushes to master. But I've tested all of this in my dummy repo and it worked. So if there's any error it would be very minor, like a typo.

sbstp:
Don't merge! I just found an error.

cdrage:
@sbstp Ah okay.

Well, I'd like to update a few of the docs on here. Should we merge this in now and then see what happens when we make a doc change?

cdrage:
@sbstp Np. Fix the error and merge when you're ready, I trust you!

script/pushdocs.sh
```diff
@@ -0,0 +1,32 @@
...
+git clone "$DOCS_REPO_URL" "$DOCS_REPO_NAME"
+bash script/makedocs.sh "$DOCS_REPO_NAME"
+
+# git config
+cd "$DOCS_REPO_NAME"
+git config user.name "$DOCS_USER"
+git config user.email "<>"
+git add --all
+# Check if anythhing changed, and if it's the case, push to origin/master.
+if git commit -m 'update docs' -m "Commit: https://github.com/sbstp/travis-test/commit/$TRAVIS_COMMIT" ; then
```
cdrage: i think this needs to be fixed to streamlink.github.io ?

sbstp: Yep that's the error.

sbstp:
I didn't update the plugin matrix when I contributed those last 2 plugins, so we can use this to test it.

script/pushdocs.sh
```diff
@@ -0,0 +1,32 @@
...
+git clone "$DOCS_REPO_URL" "$DOCS_REPO_NAME"
+bash script/makedocs.sh "$DOCS_REPO_NAME"
+
+# git config
+cd "$DOCS_REPO_NAME"
+git config user.name "$DOCS_USER"
+git config user.email "<>"
+git add --all
+# Check if anythhing changed, and if it's the case, push to origin/master.
+if git commit -m 'update docs' -m "Commit: https://github.com/streamlink/streamlink/commit/$TRAVIS_COMMIT" ; then
```
cdrage: errrrr, still wrong? should be pushing to https://github.com/streamlink/streamlink.github.io ?

sbstp: No, that's not the target repo, it's the commit message with a pointer to the commit that triggered this change. Like [this](https://github.com/sbstp/travis-test-target/commit/59764033e02f64fc79b721ce02c01f1fb82c4aea), see the 3rd line of the commit. The remote is configured automatically when the repo is cloned.

cdrage: Ahhhh that makes sense

sbstp:
Alright let's see if this works.

sbstp:
It [worked](https://github.com/streamlink/streamlink.github.io/commit/9d508376ab2da9d598e1b31eb764b8b8ba20b8ab). I'm not sure why the search index changed though.

PR #39, opened 2016-09-24
Disable docs on pull requests

sbstp:
This is a tiny tweak to the `pushdocs.sh` script, if a pull request was made from a branch inside this repo, it would have built and pushed the docs from that PR. We need to wait a merge to master before updating the docs!


cdrage:
lgtm

PR #21, opened 2016-09-20
Set some basic things to ignore in coverage

gravyboat:
This simply adds some items for coverage to ignore, namely Python releases, and inits.

PR #48, opened 2016-09-27
Remove tox

gravyboat:


gravyboat:
I'm proposing that we remove Tox unless someone has a good reason to continue using it. At this time all it does is run the test scripts in multiple environments, this is something we don't need as our travis.yml is already configured to do so in clean environments. This also helps address some conflicts with how coverage is handled which I encountered here: https://github.com/streamlink/streamlink/pull/21

sbstp:
I'm not exactly sure what tox does, but it does seem to overlap with the travis config.

martinth:
@sbstp Yes and no. It overlaps in the sense that both Travis and Tox run tests in multiple environments. What is different is that tox does so locally which is somewhat handy because you can use it during development as part as the edit-compile-test cycle so you can spot problems earlier before doing a PR. 

I have used tox in the past when converting stuff from Python 2 to 3. However if you are *not* in the middle of such a task it's not really that beneficial because you won't have many environment related problems inside the scope of a normally sized PR. 

Because of that I don't see any real problem with removing it, especially since the code as a whole is already relatively mature so changes will not become really large in the near future.

cdrage:
LGTM!

PR #45, opened 2016-09-27
Update copyright author

cdrage:
Sorry Chris!

PR #52, opened 2016-09-29
Update LICENSE

ethanhlc:
This seems to have been missed in Pull Request #45.


cdrage:
@ethanhlc just to note as well, as cannot change the author's name due to the license agreement / fork. that has to stay.

PR #60, opened 2016-10-01
Picarto plugin: multistream workaround (fixes #50)

### LINKED ISSUE #50: Picarto Multi Streams
Seems that a fix for the Picarto multi streams has been provided here: 
https://github.com/chrippa/livestreamer/pull/1469 

Thanks!

Latent-Logic:
This works for both multistream and individual streams as long as
it isn't a multistream where the primary streamer is offline.

If you go to a page linked to a multistream but the primary streamer
is offline, then this will incorrectly match the secondary or tertiary
stream and record them instead of (correctly) reporting the stream
offline.


gravyboat:
Good stuff, thank you!

PR #62, opened 2016-10-01
Don't delete everything if docs build fail

sbstp:
In the last merge, the docs' build just failed randomly, and all the docs got removed. This is an attempt to fix it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,14 @@
 language: python
-
-python:
-  - 2.6
-  - 2.7
-  - 3.3
-  - 3.4
-  - 3.5
-
 sudo: false
 
-env:
-  - BUILD_DOCS=yes
+matrix:
+  include:
+  - python: "2.6"
+  - python: "2.7"
+  - python: "3.3"
+  - python: "3.4"
+  - python: "3.5"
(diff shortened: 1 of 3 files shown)


sbstp:
Merged because it's the only way to test it.

gravyboat:
Hmm, I'm confused as to why the changes to the python call broke the docs, sorry about that.

sbstp:
There's a difference between
```yml
python:
  - 2.6
  - 2.7
  - 3.3
  - 3.4
  - 3.5
env:
  - BUILD_DOCS=yes
```

And 
```yml
matrix:
  include:
  - python: "2.6"
  - python: "2.7"
  - python: "3.3"
  - python: "3.4"
  - python: "3.5"
    env: BUILD_DOCS=yes
```
The first one sets the BUILD_DOCS=yes variable on all the builds. The second one only sets it for Python 3.5. The `env: BUILD_DOCS=yes` variable is only applied to the last object.

gravyboat:
Gotcha, I didn't notice the yaml structure specifically for the env line. We can either keep it that way, or add a conditional to the `env` section to set it so docs are only true for python 3.5

PR #64, opened 2016-10-03
Fix instances of chrippa/streamlink to streamlink/streamlink in docs

### LINKED ISSUE #63: Links in Streamlink docs occasionally point to chrippa/streamlink, not streamlink/streamlink
Some links in docs point to the chrippa/livestreamer repo on GitHub. As all instances of "livestreamer" were changed en masse to "streamlink", those links will stay broken even when releases etc are up.

ethanhlc:
Fixed #63.

Only instances of the problem seemed to be in install.rst


sbstp:
Thanks!

PR #71, opened 2016-10-06
TVPlayer plugin supports the latest version of the website

beardypig:
A fix for the TVPlayer.com plugin, hasn't worked for a while and this supports the new page layout and API.


gravyboat:
Good stuff, thank you!

PR #75, opened 2016-10-07
itvplayer: updated the productionId extraction method

beardypig:
The page layout had changed, this updates the productionId extraction method to support the new page layout.


gravyboat:
Thanks.

PR #79, opened 2016-10-09
fixed periscope live streaming and allowed url re

boda2004:
This PR basically allows to handle URLs of kind https://www.periscope.tv/<username>/<broadcast_id> (before it was https://www.periscope.tv/w/<broadcast_id> only)
And it handles live streams same way as recorded.


gravyboat:
Good stuff, thanks!

Issue #72, opened 2016-10-07
Build a package and get it upstreamed into Debian


gravyboat:
Currently livestreamer is available as a package which can be easily installed on Debian/Ubuntu, this is not the case for streamlink. If we want to get this packaged up and ready for the next release we need to move quickly. as @javiercantero mentions [here]https://github.com/chrippa/livestreamer/issues/1427#issuecomment-249653892) they'll be freezing the the release in the beginning of January. While I'm familiar with building RPMs and am okay at building debs I'm not very familiar with building debs that go into the upstream repo. If we have anyone who is an expert working on the project I would love to get your help with this. I don't think we'll be able to get this sort of coverage (http://docs.livestreamer.io/install.html#linux-and-bsd-packages), but having it be available for at least Debian and Ubuntu would be nice.

There is some existing stuff related to the builds that chrippa put in place (https://packages.debian.org/sid/video/livestreamer), but I think we'll have to do some clean up to get everything prepared.

 Is this something that we want to do? Or should we wait on a package till we've rebuilt the CLI, then hope we can get it in at a later date?

javiercantero:
Maybe it's more straightforward to open a RFP (Request for Package) first and see if any Debian maintainer is willing to do the packaging for you. Ubuntu also [recommends to talk to a Debian maintainer teams](https://wiki.ubuntu.com/UbuntuDevelopment/NewPackages) in order to accelerate the process (the relevant one here would be [Python Applications Packaging Team](https://wiki.debian.org/Teams/PythonAppsPackagingTeam))

amurzeau:
Hi,

I managed to get an updated Debian package working in Debian without additional dependencies to package (but that's my first real Debian package).
I used a separate Debian repository with its very own history as it seemed easier to handle (for example, the released source is not just a git archive of the tag).

The source package is at https://github.com/amurzeau/streamlink-debian and based on @breunigs work in #80. Changes I have made are available here: https://github.com/amurzeau/streamlink-debian/compare/95712f7d71417079fde126fc210352f9c259f475...master

Some indications :
* I choose to only make python3 packages as some of the python-backport dependencies are not available at all in Debian (like `backports.shutil_which`)
* I removed copyright information about non existing file `src/streamlink/packages/pbs.py` and added one for `src/streamlink/packages/requests_file.py` and some specific files in the sphinx rtd theme
* I updated quilt patches to the 0.5.0 streamlink version and changed a bit the patches headers so gbp pq keep them
[...]

PR #80, opened 2016-10-09
Debian packaging base

breunigs:
Hey everyone,

I guess this is a start to solve #72. 

Given the module name changed, simply renaming the Debian package and start tracking upstream doesn't seem that simple. Also, using livestreamer's most recent devel version and applying the 20 something PRs @javiercantero suggested is a lot of effort, so I got the impression helping you here is a more sensible use of my time.

Given the volatile nature of a tool such as this and the very rigid Debian process means that you'd either have to debianize "just the fixes" to the Debian packaged to version to keep it useful **or**, provide a recent version via the backports repository. Only the latter makes sense, really, for streamlink's case. So, first take away: there's no rush, if you want a working streamlink you will have to put it into backports anyway. I was asked more than once to do that for livestreamer, but declined because of lack of time.

This brings me to the next point: maybe you're better off providing it in a separate apt repository. That way you won't have to find a sponsor who will upload your package, if you decide to package on your own. And you would have to send that email every time you release a new version. There's stuff like the [openSUSE Build Service](https://en.opensuse.org/openSUSE:Build_Service) which might be of use here, but I have never used any of those.

Anyway, the patch here inherits from the livestreamer ones, which worked mostly out of the box if the renaming does not count. Of course, the `debian/changelog` would need to be cleaned and reflect the actual streamlink version `0.0.1` before it could be put to Debian at all. Similarly, you might need to update the `debian/copyright`.

The Debian packaging/building is a bit peculiar. As such, the flow I used for livestreamer more or less requires this to be maintained in a separate repository, to make it easy to build the tool in a clean environment. That is mostly useful to ensure one specified all build dependencies and the like. You can build on your Debian/Ubuntu system, using:
```
sudo apt build-dep livestreamer # or install build deps manually
git clean -fd
git reset --hard
dpkg-buildpackage -us -uc -b
```
Which will result in four packages: just the library/module for py2 and py3, the documentation and the actual program. It probably makes sense to just include everything in one packages, but I couldn't get that to work back then. Debian packaging is unfortunately not exactly straight forward. The built packages are clean:
```
lintian --pedantic --display-experimental --display-info
# I: streamlink: spelling-error-in-manpage usr/share/man/man1/streamlink.1.gz specifiying specifying
```

If you are interested, I can provide my notes on my update/build flow for livestreamer. My sponsor for new versions was/is [Piotr Ożarowski](https://people.debian.org/~piotr/sponsor), so maybe ask him to sponsor livestreamer, too, or rather, instead?

Hope that helps,
Stefan


javiercantero:
> Given the volatile nature of a tool such as this and the very rigid Debian process means that you'd either have to debianize "just the fixes" to the Debian packaged to version to keep it useful or, provide a recent version via the backports repository.

Well, I don't think Debian users expect for `stable` packages to be up-to-date. I suspect that the vast majority of us that want the most recent versions have chosen to use `testing` or even `unstable` instead. I think that just maintaining streamlink reasonably up to date in `unstable` (and therefore `testing`) should be enough to keep everyone happy.

debian/copyright
```diff
@@ -0,0 +1,72 @@
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Contact: Christopher Rosell <chrippa@tanuki.se>
```
bastimeyer: This line still refers to the maintainer of Livestreamer and needs to be changed.

debian/copyright
```diff
@@ -0,0 +1,72 @@
...
+Source: https://github.com/streamlink/streamlink
+Upstream-Name: streamlink
+
+Files: *
+Copyright: 2011-2016 Christopher Rosell <chrippa@tanuki.se>
```
bastimeyer: Another copyright line needs to be added here in addition to this one. Is 2011-2016 correct? Or is it 2011-2015?

debian/copyright
```diff
@@ -0,0 +1,72 @@
...
+License: BSD-2-clause
+
+Files: src/streamlink/packages/pbs.py
+Copyright: 2011-2012 Andrew Moffat
+License: Expat
+
+Files: docs/_themes/sphinx_rtd_theme_violet/*
+Copyright: 2013 Dave Snider
+           2014 Christopher Rosell <chrippa@tanuki.se>
```
bastimeyer: This can stay, right?

debian/python-streamlink-doc.doc-base
```diff
@@ -0,0 +1,9 @@
+Document: python-streamlink-doc
+Title: streamlink documentation
+Author: Christopher Rosell
```
bastimeyer: Another reference to the Livestreamer's maintainer.

breunigs:
I have amended the names for copyright. Since streamlink doesn't have a shared email address, I opted to use @gravyboat's with a custom suffix. The rest is copyrighted to the team, hopefully making it future proof.

breunigs:
@javiercantero Considering the rather tiny group of users of livestreamer ([cmp. debian popcontest statistics](https://qa.debian.org/popcon-graph.php?packages=livestreamer&show_installed=on&want_legend=on&want_ticks=on&date_fmt=%25Y-%25m&beenhere=1)) and that I have been asked a couple of times to update livestreamer on stable, I beg to differ. If I remember correctly, it was mostly people in the channel or bug reports that were saying "I have Debian and Twitch doesn't work with livestreamer". So, to some extent they expect recent versions or at least working ones.

sbstp:
Yea I'm not sure about having this project packaged in Debian stable. This is the kind of project that needs frequent updates and can't stay frozen for 2 years. I'd rather have it in backports and sid.

gravyboat:
@sbstp I agree, we need more regular updates than what stable can provide. Also I'm not sure if I want to be the contact point for people having issues with the package, I'll have to think about it. I wish there was a better way to do a public email that just created issues...

fishscene:
@gravyboat I had no idea such a thing existed, but it apparently does:
https://www.google.com/#q=create+github+issue+from+e-mail

https://zapier.com/zapbook/zaps/10314/create-github-issues-from-new-emails-on-gmail/

gravyboat:
@fishscene Yeah I saw that actually. I was looking for something where I didn't have to give a third party app email access. I wish there was a GitHub endpoint for it.

gravyboat:
I'll just bite the bullet and we can use my email for now.

javiercantero:
Note that Debian Stretch has entered the soft freeze [(announce)](https://lists.debian.org/debian-devel-announce/2017/01/msg00002.html), so it's no longer possible to ship streamlink in it (if anyone wanted to).

gravyboat:
@javiercantero Thanks for the heads up, I don't think we had any luck talking to the existing sponsor (I think someone approached them but maybe it was a different distro).

stepshal:
@sbstp @gravyboat backports is just packages from testing so it's gonna end up in stable anyway.

PR #95, opened 2016-10-18
Updated connectcast plugin to support RTMP streams

beardypig:
Connectcast.tv seems to have switched to RTMP streams. I have updated the plugin to reflect this.


gravyboat:
Thanks!

Issue #89, opened 2016-10-15
Stream quality shortcuts unavailable?


Vangelis66:
Hello 

I am currently testing v0.0.2 of the python script on Windows Vista SP2 x86, 
using the "portable" windows package kindly offered by @RosadinTV 

In livestreamer, I could issue: 

`livestreamer -l debug --http-proxy "proxyhost:80" -o "itv1-1800k.flv" "http://www.itv.com/hub/itv" best` 

and it would work. In fact, issuing: 

`livestreamer -l debug --http-proxy "proxyhost:80" -o "itv1-1800k.flv" "http://www.itv.com/hub/itv"`

would inform me that: 
```
[cli][info] Found matching plugin itvhub for URL http://www.itv.com/hub/itv
Available streams: 1200k (worst), 1500k, 1800k (best)
```
indicating the **worst** and **best** stream quality variants. 
With streamlink, I only get available bitrates: 
```
[cli][info] Found matching plugin itvplayer for URL http://www.itv.com/hub/itv
Available streams: 1200, 1500, 1800
```
so specifying "best" beforehand in my command doesn't work... 

```
[cli][info] Found matching plugin itvplayer for URL http://www.itv.com/hub/itv
error: The specified stream(s) 'best' could not be found.
       Available streams: 1500, 1200, 1800
```

This means I have to run streamlink twice: 
Once to be informed of available bitrates, 
a second time specifying the highest bitrate (indicated by first run):  

`streamlink -l debug --http-proxy "proxyhost:80" -o "itv1-1800k.flv" "http://www.itv.com/hub/itv" 1800`

Am I getting something wrong here?

Thanks for your efforts in resurrecting what appears to be a staled project!

RosadinTV:
I tried and yes, you are right.
For some reason in Python 3.5.2 the stream qualities didnt get the k ending.
I mean with this stream in Python 2.7 returns: Available streams: 1200k (worst), 1500k, 1800k (best) 
In Python 3.5.2 returns: Available streams: 1800, 1500, 1200
Maybe somebody of the staff know a fix for this issue.

Vangelis66:
@RosadinTV

Many thanks for looking into this. 

> with this stream in Python 2.7 returns: 
> Available streams: 1200k (worst), 1500k, 1800k (best) 

Pending an intervention from the main devs, 
could you be so kind as to compile a win32 "portable" package 
(like your second one, where Python and Streamlink dirs are separated), 
[...]

PR #96, opened 2016-10-18
Fix stream quality naming issue with py2 vs. py3, fixing #89

beardypig:
py2 and py3 handle integer division differently, in py2 1800000 / 1000 would be 1800, however in py3 it would be 1800.0. This means that for some plugins (itv, and livestream at least) the quality could be named 1800.0k, which is not handled as expected.


gravyboat:

Issue #94, opened 2016-10-18
Option --version-check outputs invalid input response


404NetworkError:
### Expected Behavior

`[cli][info] Your Streamline version (0.0.2) is up to date!`
### Actual Behavior

```
usage: streamlink [OPTIONS] [URL] [STREAM]

Use -h/--help to see the available options or read the manual at http://docs.streamlink.io/
```
### Steps to reproduce the behavior

`streamlink --version-check`

TheSneakySniper:
Please use the correct command, --version.

scottbernstein:
--version-check is different than --version

--version-check calls back to the server to see if you have the latest version or not.
________________________________

beardypig:
This issue is caused by an Invalid version number in the PyPi json information endpoint for streamlink: https://pypi.python.org/pypi/streamlink/json
The offending version field...
```json
"version": "1.14.0rc1"
```
The `version_check` method uses `StrictVersion`, which disallows suffixes such as `rc1` which generates an `Exception`, which is silenced... Even if you change it to `LooseVersion`, which does permit `rc1`, `--version-check` will tell you that the latest version of streamlink is `1.14.0rc1` (lies).

```
[cli][info] A new version of Streamlink (1.14.0rc1) is available!
```

As to why the version is set to `1.14.0rc1`, one of the streamlink team may know and might be able to fix it :)

beardypig:
I have created a work-around that will get the latest version that has release information - it works, but the better solution would be to fix the PyPi entry if possible.

gravyboat:
There's a discussion going on over in #97 to discuss getting this fixed.

beardypig:
This is fixed now :)

PR #97, opened 2016-10-18
Work-around to fix broken --version-check option

beardypig:
This PR provides a work-around for #94. 

The version number is incorrect in the PyPi JSON info endpoint: https://pypi.python.org/pypi/streamlink/json. It reports the version as `1.14.0rc1`, this could be related to and old publish attempt before the `setup.py` had been fully fixed? 

This is only intended to be merged if a solution cannot be found for fixing the version number in PyPi. Another solution would be to release a version greater than `1.14.0rc1`, which is probably not what is wanted given the version number reset before :)

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -851,8 +851,14 @@ def check_version(force=False):
     if force or not latest_version:
         res = requests.get("https://pypi.python.org/pypi/streamlink/json")
         data = res.json()
-        latest_version = data.get("info").get("version")
-        cache.set("latest_version", latest_version, (60 * 60 * 24))
+        versions = sorted([StrictVersion(version)
+                           for version, info
+                           in data.get("releases", {}).items()
+                           if len(info)],
+                          reverse=True)
+        if len(versions):
+            latest_version = str(versions[0])
+            cache.set("latest_version", latest_version, (60 * 60 * 24))
 
     version_info_printed = cache.get("version_info_printed")
     if not force and version_info_printed:


gravyboat:
Thanks for catching this @beardypig. I don't think we want to release a version number like that. @cdrage Can you regenerate the json data? Or can we contact the PyPi team and get it removed?

beardypig:
It might just be as simple as removing the offending release in PyPi. 

I just tested with a dummy package and when you login to PyPi, in the list of releases, there is a hide option and a remove option. Even if you hide the bad version it will still show up as the latest version in the JSON API, It has to be removed for the correct latest version to show. If this doesn't work, I suggest a Streamlink PyPi admin (@cdrage, I guess) contacts the PyPi team and asks for some assistance.

cdrage:
@beardypig @gravyboat Sorry guys, for some reason I didn't get pinged on these notifications.

If I merge this and do a release, I'm assuming it'll be fixed? Hence the newest version will have the updated JSON. The previous versions I can either deprecate the old releases and/or contact the PyPi team to get it fixed.

beardypig:
You should be able to just delete the funky version (it looks like it hidden at the moment) in the PyPi interface, which should fix it. If you can avoid merging this patch it would be better, as it's kind of a hack :-)

cdrage:
@beardypig there we go, I removed 1.14.0rc1 you were right. it was "hidden" and still on there. Since it was a higher version number, it appeared under the json data.

Let me know if this is good! :) I checked and it appears to be gone! https://pypi.python.org/pypi/streamlink/json

beardypig:
Awesome, looks good from here :)

Are you the only one with administrative access on PyPi? Might be worth adding @gravyboat or someone else for when you're not around :-)

gravyboat:
@beardypig Are you cool with closing this if everything looks good?

PR #98, opened 2016-10-20
Finish the installer

sbstp:


gravyboat:
Good stuff, thanks @sbstp.

gravyboat:
Are we good to go now with the Python3 installer or does anything else need to be done before we can put it on the releases page?

bastimeyer:
Is there a way to change the shebang in `streamlink-script.py` from `python.exe` to `pythonw.exe` or is it possible to create two separate streamlink executables that invoke two different scripts?

`python.exe` always opens a cmd window and doesn't return stdout/stderr, which makes it impossible to use from somewhere else other than the Windows shell. `pythonw.exe` on the other hand doesn't work when calling streamlink from the shell.

I had the same issues with my Twitch GUI and Livestreamer when it was installed via pip, but the installer worked fine.

Btw, what's the status of building the installer via TravisCI, etc? Afaik, the current one was created and uploaded manually.

sbstp:
You might want to direct your first question at the [pynsist repo](https://github.com/takluyver/pynsist). I don't know if this can be done.

> Btw, what's the status of building the installer via TravisCI, etc? Afaik, the current one was created and uploaded manually.

Just need someone to do the work, requires a bit of travis/bash knowledge.

Issue #103, opened 2016-10-26
Twitch.tv: use https only


stevek123:
Since they support https now, I think the plugin should use the new url: https://usher.ttvnw.net instead of http://usher.twitch.tv (line 131), so every part of the connection would go securely.

gravyboat:
https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugins/twitch.py#L131 is the line in question.

stevek123:
Yes, that's the one. After replacing the URL, the plugin still works fine.

stevek123:
Thank you!

PR #104, opened 2016-10-27
Changed the twitch plugin to use https instead of http

Erk-:
as discussed in #103


gravyboat:
Were you able to build this and confirm everything worked correctly locally?

Erk-:
Yes it worked with the streams I tried I could not get the audio option to work but, it does not work in the normal version either.

gravyboat:
I'll merge this for now then, thanks!

stevek123:
Audio only also works great for me with:
`streamlink https://www.twitch.tv/monstercat audio`

Erk-:
It must be something with my setup then.

Issue #92, opened 2016-10-16
Changelog in docs empty


ethanhlc:
Clicking the changelog link in the github.io page leads to an empty Changelog page.

PR #107, opened 2016-10-28
Modify the changelog link

gravyboat:
Resolves https://github.com/streamlink/streamlink/issues/92

Unfortunately the index is in rst, and the changelog is in markdown so I just made a quick and dirty change to get this fixed for now. We'll probably want to fix this in a better way at some point.


sbstp:
Looks good.

PR #113, opened 2016-11-01
Fix ORF TVthek plugin

sqrt2:
The way stream metadata (such as HLS playlist URLs) is embedded into the ORF TVthek website has changed. As a result, streamlink was unable to extract any stream sources. This PR adapts the plugin to these changes and restores previous functionality.

Additionally, tabs have been converted to spaces as spaces seem customary across the project.


gravyboat:
Thank you!

Issue #19, opened 2016-09-20
Support for non-ASCII display names


lewisje:
I've been using an OAuth token in Livestreamer for well over a year, ever since I learned it could be done, to help out the stream viewer counts; today, after seeing the latest beta of Chatty could handle localized display names accurately, I decided to finally get one for myself, and then whenever I tried to launch Livestreamer, I got this error:

```
error: 'ascii' codec can't encode characters in position 0-6: ordinal not in range(128)
```

I couldn't go back to the old completely anonymous way, because Twitch no longer allows anonymous API requests, so instead I tried this workaround that has been floating around, the following line in my livestreamerrc:

```
http-header=Client-ID=jzkbprff40iqj646a697cyrvl0zt2m6
```

It allowed me to watch streams, but I would like to be able to be logged in via the stream-viewer again.

sbstp:
Was the error message bigger than this? Was there a stack trace? That would help us trace where the issue was. Also were you using Python 2 or Python 3?

lewisje:
I believe I was using Python 2, and I didn't see a stack trace, just that error followed by termination.

gravyboat:
This appears to have been fixed in #122, I'm going to close it for now, if it's still a problem please re-open the issue.

steveooo9:
This still exists in default DigitalOcean 18.04 with Python2 and Python3

beardypig:
@steveooo9 can you create a new issue with all the details?

PR #122, opened 2016-11-03
Automatically determine the encoding when parsing JSON with HTTPSession

### LINKED ISSUE #120: Non unicode escaped API breaks the plugin
I have a problem on Streamlink 0.0.2 with Twitch plugin. Luckily I figured out what's going on and fixed my own copy of code. So I would like to share my fix though I am not sure with my decision.

My question is:
> Is it safe to assume that JSON APIs other than Twitch, use particular text encoding?

Here is some error log.
```
> streamlink.exe --player <some my player> --twitch-oauth-token <redacted> https://twitch.tv/<readacted> source
[cli][info] Found matching plugin twitch for URL https://twitch.tv/<readacted>
[plugin.twitch][info] Attempting to authenticate using OAuth token
error: 'cp949' codec can't encode character '\u011b' in position 48: illegal multibyte sequence
```

Little backgrounds.

1. Twitch API does not escape 'non-ascii' characters.
2. Twitch API does not state `charset` in the `Content-type` header. Only states MIME type `application/json`.
3. I am running Korean version of Windows 10, and the plugin tries to decode response body with OS default encoding(presumably) which is CP949 in my case.

Let me follow the code flow.

1. Twitch plugin tries to authenticate with my OAuth token, in https://github.com/streamlink/streamlink/blob/0.0.2/src/streamlink/plugins/twitch.py#L264
2. ...which is `user` method in helper class, in https://github.com/streamlink/streamlink/blob/0.0.2/src/streamlink/plugins/twitch.py#L208
3. ...which requests API with GET method, in https://github.com/streamlink/streamlink/blob/0.0.2/src/streamlink/plugins/twitch.py#L167
4. I know Twitch API returns in JSON format, so going https://github.com/streamlink/streamlink/blob/0.0.2/src/streamlink/plugins/twitch.py#L182
5. ...which tries to parse response body text in JSON, implicitly tries to decode response body from bytes to str, in https://github.com/streamlink/streamlink/blob/0.0.2/src/streamlink/plugin/api/http_session.py#L70

I could simply edit `http_session.py` and fix it by stating 'known' charset before accessing `text` property, something like:
```
+        if res.encoding is None:
+            res.encoding = 'utf8'
         return parse_json(res.text, *args, **kwargs)
```

beardypig:
This should fix #120 and if I understand #19 it should fix that too. 

According to [RFC4627: The application/json Media Type for JavaScript Object Notation (JSON)
](http://www.ietf.org/rfc/rfc4627.txt) JSON is always encoded as Unicode (see section 3.  Encoding) and the specific Unicode encoding can be determined by looking at the pattern of NULL bytes in the first 4 bytes. 

I have added a method that implements this pattern matching and added a couple of tests for it, the tests use `unittest.mock` which is not available in Python 2 - I added `mock` to the `before_install` command for travis. 

Props to @cyluss 👍

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
     env: BUILD_DOCS=yes
 
 before_install:
-  - pip install pytest pytest-cov codecov coverage
+  - pip install pytest pytest-cov codecov coverage mock
 
 install:
   - python setup.py install
--- a/src/streamlink/plugin/api/http_session.py
+++ b/src/streamlink/plugin/api/http_session.py
@@ -67,8 +67,33 @@ class HTTPSession(Session):
             self.mount("https://", HTTPAdapterWithReadTimeout())
 
     @classmethod
+    def determine_json_encoding(cls, sample):
+        """
+        Determine which Unicode encoding the JSON text sample is encoded with
+
+        RFC4627 (http://www.ietf.org/rfc/rfc4627.txt) suggests that the encoding of JSON text can be determined
+        by checking the pattern of NULL bytes in first 4 octets of the text.
+        :param sample: a sample of at least 4 bytes of the JSON text
+        :return: the most likely encoding of the JSON text
+        """
+        nulls_at = [i for i, j in enumerate(bytearray(sample[:4])) if j == 0]
+        if nulls_at == [0, 1, 2]:
+            return "UTF-32BE"
+        elif nulls_at == [0, 2]:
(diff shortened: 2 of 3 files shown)


cyluss:
Awesome solution! I should have read RFC thoroughly.

gravyboat:
Looks good to me, thanks for working on this @beardypig and @cyluss.

PR #123, opened 2016-11-03
TVPlayer.com: fix for 400 error, correctly set the platform parameter

beardypig:
The User-Agent header needs to be sent for the page to fill in the correct platform variable.


gravyboat:
Thank you.

gravyboat:
@beardypig do you want an invite to the Streamlink org? You've contributed a good amount and I think we're pretty open to adding people if they are interested (@cdrage please confirm).

beardypig:
Sure, I'm interested :)

beardypig:
@gravyboat thanks :-D

cdrage:
@beardypig add you mannn

gravyboat:
@cdrage That was me, I figured you would be fine with it since there wasn't an immediate 'no' response. @beardypig No problem!

PR #125, opened 2016-11-04
Retrying hosted streams with hosting disabled will always retry

### LINKED ISSUE #82: retry-streams overrides twitch-disable-hosting
When using both arguments `--retry-streams DELAY` and `--twitch-disable-hosting`, on the first retry streams are switched to the hosted stream anyway.  Instead, streamlink should keep retrying without streaming the hosted stream.

beardypig:
This PR fixes #82

There was a bug in the way that `Twitch._check_for_host` worked in that it would update the `channel_id` to the hosted channel (if the requested channel was hosting), and as such on the next retry it would see that hosted channel (now stored in `channel_id`) was not hosting and would play the hosted channel.

This fix updates `Twitch._check_for_host`so that it does not modify the object state and moves that logic in to the calling method `Twitch._get_hls_streams`.

It also fixes a bug where if you had one channel hosting another channel, that was hosting a different channel, eg. channel_a -> channel_b -> channel_c, you would not find the streams for channel_c. It will also detect cycles and stop eg. channel_a -> channel_b -> channel_a would create an infinite loop :)

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -1,3 +1,4 @@
+# coding=utf-8
 import re
 
 import requests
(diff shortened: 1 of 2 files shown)


gravyboat:
Good stuff, thanks.

Issue #58, opened 2016-09-30
Please fix the ustream plugin


TheSneakySniper:
The plugin seems to not be able to detect desktop streams... (This was ported directly from livestreamer, where it was never fixed. using librmtp never worked to download desktop streams, i am linking the post)
https://github.com/chrippa/livestreamer/issues/1122

Some streams that are in 720p don't show up, due to it only detecting mobile versions of the streams. 
See here:

`[cli][info] Found matching plugin ustreamtv for URL http://www.ustream.tv/nasahdtv
Available streams: mobile_240p (worst), mobile_360p, mobile_480p, mobile_720p (best)`
While this specific stream has 720p (not sure why), other streams do not. 

This is a problem with the plugin only detecting mobile versions of streams.

Ustream Livestreamer plugin history: 
https://github.com/chrippa/livestreamer/commits/develop/src/livestreamer/plugins/ustreamtv.py

TheSneakySniper:
I have asked stackoverflow if they know anything.
http://stackoverflow.com/questions/39798859/why-does-this-plugin-for-ustream-only-do-mobile-streams-how-can-i-also-support

gravyboat:
Why are you posting something like that on Stackoverflow? It's not their job to investigate code,and you will most likely be downvoted.

TheSneakySniper:
¯\_(ツ)_/¯

TheSneakySniper:
I will remove it...

TheSneakySniper:
From the streamlink ustream plugin:
`HLS_PLAYLIST_URL = (
    "http://iphone-streaming.ustream.tv"
    "/uhls/{0}/streams/live/iphone/playlist.m3u8"
)`
https://raw.githubusercontent.com/spaceplumber/livestreamer/06d9a2688dddf96b0694851647c19f6877b42557/src/livestreamer/plugins/ustreamtv.py 
This is my proof for saying that it only supports mobile streams.

tp0:
Have the non-mobile streams ever worked in past <2 years? I remember they opened and stream played, but there was constant buffering (segments) = unwatchable.

TheSneakySniper:
No, they have not. From what I have read in the code (I may be wrong), all it does is get the ustream channel ID and then plug that into a pre constructed text;
`_url_re = re.compile("""
    http(s)?://(www\.)?ustream.tv
    (?:
        (/embed/|/channel/id/)(?P<channel_id>\d+)
    )?
    (?:
[...]

PR #132, opened 2016-11-06
Ustream Partial Fix

RosadinTV:
-To avoid errors now only mobile streams are displayed
-Mobile prefix removed
-VOD not supported warning added


gravyboat:
It seems like this might conflict with the work @beardypig has started for https://github.com/streamlink/streamlink/issues/58. Could you please weigh in @beardypig? This PR looks good but I don't want to limit people to mobile streams if the original problem is addressed.

RosadinTV:
@gravyboat Also notice this, the ideal thing is fix all part of the Plugin (i mean support desktop streams and VOD too) this PR is something like a bypass until someone who know more about Python (in fact, i almost didnt know anything) fix the whole thing.

gravyboat:
@RosadinTV There's a PR over here: https://github.com/streamlink/streamlink/pull/137 that takes care of the Ustream issues, I have a few concerns for merging it since it adds additional dependencies to Streamlink, but it would be great if you could try it out.

RosadinTV:
@gravyboat Checked, seems promising but a bit unstable for now (and dont support VOD too), however is going in the right way.
Also i like the new dependency pymp4, maybe can be usefull for upcoming or updating plugins, i mean another streams providers can change to the same distribution method that ustream is using.
On the other hand i didnt agree with the idea of dropping the python-librtmp dependency, it can be usefull in the future.

beardypig:
@RosadinTV the problem with `python-librtmp` is that it is no longer maintained and had become quite hard to install on Windows machines using newer versions of Python. It is only an optional dependency and dropping it just means not using it any more, for this plugin.

RosadinTV:
@beardypig 
With dropping the dependency i mean dont include it by default in release, but you are right, tried installing Python-librtmp in Python 3.5.2 and didnt finish, some errors ocurred .. knowing this the position is hard to take 😕

Issue #126, opened 2016-11-05
Adulst Swim: support for live streams


stepshal:
http://www.adultswim.com/videos/streams/
http://www.adultswim.com/videos/streams/toonami - subtitles available,
http://www.adultswim.com/videos/streams/toonami-pre-flight
http://www.adultswim.com/videos/streams/marathon
http://www.adultswim.com/videos/streams/last-stream-on-the-left

stepshal:
Ahh, i couldn't watch them now, they probably geo-restircted now.

beardypig:
Fixed in #139 :)

PR #139, opened 2016-11-07
plugins.adultswim: added adultswim.com plugin

beardypig:
As requested by @stepshal in #126.

Only the live streams (http://www.adultswim.com/videos/streams/) are supported, the VOD streams appear to be encrypted HDS streams.


gravyboat:
Good stuff, thanks!

PR #163, opened 2016-11-14
plugins.cybergame: support the new streaming url for live streams

### LINKED ISSUE #147: cybergame.tv plugin not work for livestreams
`[cli][info] Found matching plugin cybergame for URL http://cybergame.tv/sharkisfinetoo/`
`[cli][info] Available streams: 720p (worst, best)`
`[cli][info] Opening stream: 720p (rtmp)`
`[cli][error] No data returned from stream`
For stream http://cybergame.tv/<streamername> Internet Download Manager  produces a link like rtmp://cybergametv.cdnvideo.ru/cybergame//<streamername>.VODs fine.

beardypig:
This PR fixes #147. 

cybergame.tv has switched to using a fixed RTMP url for the live streams, there doesn't appear to be an easy way to automatically get that url. They could change the URL in the future, and if they do that often I will work out a better method to get the RTMP URL, but for now it works.

### DIFF
--- a/src/streamlink/plugins/cybergame.py
+++ b/src/streamlink/plugins/cybergame.py
@@ -5,6 +5,7 @@ from streamlink.plugin import Plugin
 from streamlink.plugin.api import http, validate
 from streamlink.stream import RTMPStream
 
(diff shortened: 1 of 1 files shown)


gravyboat:
Thanks.

PR #167, opened 2016-11-15
Automated Windows nightly builds

beardypig:
As requested, this PR adds automated nightly (actually build on each commit to master) for the Streamlink NSIS Windows installer, and pushes them to S3. 

The releases are listed [on S3](https://streamlink-builds.s3.amazonaws.com/index.html), and the latest build is always available here [streamlink-latest.exe](https://streamlink-builds.s3.amazonaws.com/nightly/windows/streamlink-latest.exe).

Nightly builds are automatically removed from S3 after 60 days. 

For nightly builds, the version number is constructed out of the current release version (eg. 0.0.2) and the current short commit hash (eg. ba92efb), so they become `0.0.2+ba92efb`. The file name is generated with the build number also (`streamlink-0.0.2-34-ba92efb.exe`), this is so the most recent file can easily be identified and the files will sort in a useful order. 
For release builds (tagged commits) the version number is constructed just using the current release version, so the version would be `0.0.2` and the file name would be `streamlink-0.0.2.exe`. 

@cdrage If you want I can set travis to automatically add the result of tagged builds (ie. only when the commit has a tag) to the GitHub releases to make releasing a bit easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,18 @@
 language: python
-sudo: false
+sudo: required
 
 matrix:
   include:
-  - python: "2.6"
-  - python: "2.7"
-  - python: "3.3"
-  - python: "3.4"
-  - python: "3.5"
-    env: BUILD_DOCS=yes
+  - python: '2.6'
+  - python: '2.7'
+  - python: '3.3'
+  - python: '3.4'
+  - python: '3.5'
+    env: BUILD_DOCS=yes BUILD_INSTALLER=yes STREAMLINK_INSTALLER_DIST_DIR=$TRAVIS_BUILD_DIR/dist/nsis
 
 before_install:
-  - pip install pytest pytest-cov codecov coverage mock
+  - pip install pytest pytest-cov codecov coverage mock pynsist
+  - sudo pip install s3cmd
 
 install:
   - python setup.py install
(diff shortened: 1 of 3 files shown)


bastimeyer:
You could add the `nsis` package to the [`addons/apt/packages`](https://docs.travis-ci.com/user/installing-dependencies/#Adding-APT-Packages) list, which makes it a bit clearer that it is a dependency of the build.

Other than that, it looks pretty good. 👍 

Little side note: The `before_deploy` stuff needs be moved to `script` once Travis gets set up to also deploy to github-releases, because `before_deploy` gets executed for each deploy config and it would build the installer multiple times.

bastimeyer:
Another thing, or a maybe question:
Does using `sudo: true` work as intended? The docs say that `required` is the choice here: https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

beardypig:
Yep, my bad... was reading the columns wrong :D

beardypig:
Nice, I didn't know about the apt addon :)

beardypig:
I thought building the installer using `before_deploy` was cleaner than putting a guard in the `script` or `after_success` section to only build with Python 3.5. But maybe doing that is better, to ensure that it does only get build once.

.travis.yml
```diff
@@ -23,17 +25,17 @@ script:
 
 after_success:
   - codecov
+  - if [[ $BUILD_INSTALLER == 'yes' ]]; then pip install pynsist; ./script/makeinstaller.sh; fi
```
bastimeyer: Sorry for nitpicking here, but the `after_success` life cycle hook ignores failures:
https://docs.travis-ci.com/user/customizing-the-build/#Breaking-the-Build

> The exit code of after_success, after_failure, after_script and subsequent stages do not affect the build result. However, if one of these stages times out, the build is marked as a failure.

What I meant was adding the ./script/makeinstaller.sh call to the `script` hook.

Regarding the pip dependencies:
Wouldn't it make much more sense adding this to `before_install` hook, since there's already a list of pip dependencies?

beardypig:
OK. Cleaned a few things up.

.travis.yml
```diff
@@ -23,17 +25,17 @@ script:
...
+addons:
+  apt:
+    packages:
+    - nsis
 
 deploy:
   provider: script
   skip_cleanup: true
-  script: if [ -n "${TRAVIS_TAG}" ]; then s3cmd put build/nsis/streamlink-*.exe s3://streamlink-builds/release/windows/; else s3cmd put build/nsis/streamlink-*.exe s3://streamlink-builds/nightly/windows/; s3cmd put build/nsis/streamlink-*.exe s3://streamlink-builds/nightly/windows/streamlink-latest.exe; fi
+  script: sudo pip install s3cmd; if [ -n "${TRAVIS_TAG}" ]; then s3cmd put build/nsis/streamlink-*.exe s3://streamlink-builds/release/windows/; else s3cmd put build/nsis/streamlink-*.exe s3://streamlink-builds/nightly/windows/; s3cmd put build/nsis/streamlink-*.exe s3://streamlink-builds/nightly/windows/streamlink-latest.exe; fi
```
bastimeyer: Since the installer gets built for every commit, dependency installations should be bundles together, I think.
Just being curious here, doesn't the already integrated s3 deployment config of Travis support conditional files?

.travis.yml
beardypig: Fair enough, been a while since I did much travis work. Yeah `pynsist` is better in the `before_install`, you're right.

.travis.yml
beardypig: `s3cmd` doesn't work with Python 3, and is only required when there is a deploy to S3, in this case I thought it better to install it only for the deploy. But, I don't mind moving it to  `before_install` with the others. I would have used the version in `apt` but it's a bit outdated and doesn't support loading the secret id/key from environment variables. 
The Travis S3 deploy wants to upload all of the files, I couldn't get it to upload only small set of files, also I wanted to to copy the current installer to `streamlink-latest.exe`.

beardypig: Maybe it can be cleaned up some more anyway...

cdrage:
@beardypig I think it's great that it's added to s3 as a nightly build.

But IMO, the "releases" should be done manually. Once the release is out, it's out. No way to change it. 

Releases should be built manually, but I'm happy to have code in Travis CI for Windows nightly binaries to be generated.

beardypig:
@cdrage, fair enough. It would be only done on the tagged commit, so unless you remove the tag and add it again it would not be changed, it would be tied to the tags listed on the GitHub releases page.

.travis.yml
bastimeyer: > s3cmd doesn't work with Python 3, and is only required when there is a deploy to S3

Yes, right, but the installer gets built for every commit, so it's better to bundle all deps IMO.

> I couldn't get it to upload only small set of files

Isn't it possible to use the [`local_dir`](https://docs.travis-ci.com/user/deployment/s3/#Deploy-From-Only-One-Folder) s3 deploy property? I haven't used s3 before, though, so maybe I'm wrong.

> also I wanted to to copy the current installer to streamlink-latest.exe

Files could me moved/renamed by the bash script

beardypig: `local_dir` might work. But that would mean moving the resulting `exe` as the directory it outputs to at the moment gets all of temp files, etc. Would be a lot cleaner though :)

beardypig:
@cdrage how do you feel about moving the output from the NSIS install builder from `build/nsis/` to `dist/`?

beardypig:
@bastimeyer Thanks for all the suggestions, the travis file is much cleaner now...

beardypig:
I used an environment variable to give the option to override the output directory for the NSIS installers, which is also used as the `local_dir` in the S3 deploy. Running the `makeinstaller.sh` not on travis should give the same output as it did before.

cdrage:
@beardypig that move sounds better, far cleaner.

once this is merged, i'll modify the release script to also build the windows .exe and add it to each release on github.

then our next release will not only include linux, but also windows support.

beardypig:
@cdrage cool, did you check out my other PR (#168) for pycrypto support (in the form of pycryptodome)?

gravyboat:
This looks good to me, I'm going to merge. Thanks @beardypig!

beardypig:
Awesome, nightly (kinda) builds done!

cdrage:
hey @beardypig 

so what files would I actually need for the release? I thought it was just one .exe?

I ran `./script/makeinstaller.sh` and I have these files:

```sh
streamlink/build/nsis  master ✔                                                                                                                                                                                                                                           3h29m  
▶ ls
bin  glossyorb.ico  installer.nsi  msvcrt  pkgs  Python  _rewrite_shebangs.py  rtmpdump  Streamlink.launch.pyw  _system_path.py
```

beardypig:
That's weird... Do you have `nsis` installed?

beardypig:
@cdrage did you get it working? What was the console output?

cdrage:
@beardypig yeah, nsis wasn't installed. 

Are you able to add a check for when building streamlink to check that pyinst as well as nsis are installed?

beardypig:
Yep. I already added in #172 but I can pull it out in to a separate PR if you like.

PR #168, opened 2016-11-15
Use pycryptodome inplace of pyCrypto

beardypig:
`pycryptodome` is a fork of `pyCrypto` and can be used as a drop-in replacement for `pyCrypto`, either package can be used. 

The advantage of `pycryptodome` over `pyCrypto` is that it has wheels available on PyPi for Windows, which makes it easier for Windows users to install and for us to include in the NSIS installer. I have updated the installer script to include `pycryptodome`.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -147,7 +147,7 @@ Name                                 Notes
 **Optional**
 --------------------------------------------------------------------------------
 `RTMPDump`_                          Required to play RTMP streams.
-`PyCrypto`_                          Required to play some encrypted streams.
+`PyCrypto`_                          Required to play some encrypted streams (`pycryptodome`_ is a compatible alternative).
 `python-librtmp`_                    Required by the *ustreamtv* plugin to be
                                      able to use non-mobile streams.
 ==================================== ===========================================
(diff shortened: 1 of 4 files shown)


gravyboat:
Looks good to me. We probably shouldn't make it an option package for the Windows build, I'm already tired of seeing questions related to it on the Windows release.

beardypig:
Yeah, pyCrypto is pain and pyCryptodome just works, I tested that MTV case but wasn't able to test the PSVue case.

gravyboat:
Definitely. I mean that we should include this in the Windows releases by default (unless this does that and I'm just missing it, if so we should clarify the documentation).

beardypig:
That's exactly what it does, it will include it in the Windows build, but not the Linux or macOS install via pip :-)

beardypig:
Can update the docs specifically for the Windows related things. I'm not sure exactly where all that stuff is referenced. 

I would also suggest we include it in the standard pip/setup.py install, but that should be a separate issue :-)

gravyboat:
Agreed. I'm going to merge this for now, we can discuss doc updates or other changes in another issue. Thanks!

Issue #155, opened 2016-11-12
Streamlink for Windows (Release)


RosadinTV:
With this program you can build a ready to work Streamlink release for Windows with a few clicks.
The generated release is always the latest available because the source is master.zip
Download: https://github.com/RosadinTV/Streamlink-for-Windows/archive/master.zip
Look at README.txt for requirements.

DragonzMaster:
Good Job. 👍 
The next step is to package streamlink into exe.

RosadinTV:
@DragonzMaster Maybe in a future add a option to create a Standalone EXE.
I don't get along with python to exe compilers, so probably it will be a VB.NET approach.
But of course if you can use those compilers is fine, feel free to do what you want 😄

DragonzMaster:
I mess around with them but unfortuntally facing some problems. Can't find the right way yet to do that.

RosadinTV:
@DragonzMaster Yeah, its some tricky, my plan is generate a .exe on the fly with embeded resources (streamlink+python) so when the exe opens it unpacks all to the temp directory (if required) and then execute the script.
Maybe dont work, its a thesis.

Vangelis66:
¡Hola! @RosadinTV 
I spent the last hour testing your latest offering!
 
 Since I am on Vista SP2 x86 currently, I barely meet the minimum OS requirements; 
actually, the last version of .NET Framework 4.x officially supported by Microsoft on 
Vista is 4.6, but I have managed to install manually .NET Framework 4.6.1 without 
any issues so far (updates for .NetFW 4.6.1 must also be installed manually from Microsoft®Update 
Catalog site) - on the contrary, latest .NET Framework 4.6.2 won't install on Vista. :-{ 

Your app loads and runs fine on my setup (VistaSP2 x86, .NetFW 4.6.1) :  
![sfw](https://cloud.githubusercontent.com/assets/9669492/20243369/4c31c456-a95d-11e6-8da8-81f4e2c9f54c.jpg)

I am only nit picking here, but you could, of course, improve a bit the English wording, 
for instance it did not dawn on me at first that I had to press the "Not completed" string 
to initiate the download of streamlink source - "Start/Begin" would have been more 
obvious, but again this is just me being peculiar... The same goes for starting the build process 
of the end package...
 
Anyhow, the compilation of the package went along fine, the message on 
the popup could be tailored further: 

[...]

PR #171, opened 2016-11-16
Support for relative paths for rtmpdump

beardypig:
This patch allows the path to `rtmpdump(.exe)` to be specified as a relative path or as an absolute path. 

FYI @RosadinTV, this patch might break the current version of your portable builder (see my [comment](https://github.com/streamlink/streamlink/issues/155#issuecomment-260892115) on #155).

### DIFF
--- a/src/streamlink/stream/rtmpdump.py
+++ b/src/streamlink/stream/rtmpdump.py
@@ -1,4 +1,5 @@
 import re
+import os.path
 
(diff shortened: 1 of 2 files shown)


gravyboat:
Good stuff, thanks.

beardypig:
Yeah, this is broken on Linux... I'll fix it.

Issue #116, opened 2016-11-02
0.0.3 release


cdrage:
Been 21 days since the last release. Down for another one?

Now that we have rtmp updated + windows docs + debian deb package installer thingy, it makes sense.

Let me know if there any blockers on this, otherwise, I can do a release later tomorrow.

gravyboat:
We aren't building the Windows installer automatically yet is my only concern for creating more manual work. @sbstp opinions?

RangerCoder99:
Thank you so much for updating this project, and all your hard work on this!

cdrage:
@gravyboat I should add this to the release / build script we have.

sbstp:
I'd like to add some polish if you guys don't mind. Shouldn't be too long, it's just the config file thingy on Windows.

gravyboat:
I'm fine with waiting @sbstp! Having the config file taken care of would be good for typical users (I am assuming that is what you are referring to, if not let me know).

karlo2105:
Could you do something like chrippa did with automatically nightly build : http://livestreamer-builds.s3.amazonaws.com/livestreamer-latest-win32.zip

sbstp:
Yeah, the first time you run streamlink, it would create the file in %APPDATA%, and I might also add a switch such as `streamlink --edit-config` that starts an editor (windows notepad, unix nano/vim/emacs/$EDITOR) with the config file. You're right though, it can wait.

gravyboat:
@sbstp The editor idea is a good idea! Thanks for prioritizing the configuration file, it's super helpful to have that available. @karlo2105 personally I am not a fan of nightly builds, people use them and then ask for support when instead they could just compile it themselves if they want the latest release since everything is available.

thinkpad4:
Sbstp, are you gonna make another windows .exe installer for 0.0.3 like you did for 0.0.2?

sbstp:
Yes I will.

thinkpad4:
<3

karlo2105:
@gravyboat
Nightlybuild is interesting when you want to update streamlink locally altogether on Windows. 
Otherwise you are obliged to update step by step every part of streamlink. 
I like to have choice, nobody has problems with nightly builds with Livestreamer and no one is obliged to use nightlybuilds.

dehesselle:
Hey guys,
[...]

PR #172, opened 2016-11-16
Install the streamlinkrc file in to the users %APPDATA% directory

### LINKED ISSUE #81: On Windows, no "streamlinkrc" file found after installation
Despite [it said so in the document](https://streamlink.github.io/cli.html#configuration-file).

beardypig:
I believe this PR fixes #81 and thus unblocks #116. 

This patch adds an extended nsi template that includes the `streamlinkrc` config file in the installer. The config file is only installed _if_ there is no pre-existing config file in the users `%APPDATA%\streamlink` directory. After the config file has been installed the path to `rtmpdump` is _always_ updated to point to the `rtmpdump.exe` included with the `streamlink` install.

I replaced the newlines in the `win32/streamlinkrc` file with Windows line endings, so that it is more friendly for Windows users to edit in notepad :)

### DIFF
--- a/script/makeinstaller.sh
+++ b/script/makeinstaller.sh
@@ -3,6 +3,9 @@
 
 set -e # stop on error
 
+command -v makensis > /dev/null 2>&1 || { echo >&2 "makensis is required to build the installer. Aborting."; exit 1; }
+command -v pynsist > /dev/null 2>&1 || { echo >&2 "pynsist is required to build the installer. Aborting."; exit 1; }
+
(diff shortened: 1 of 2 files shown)


RosadinTV:
@beardypig Good to know that streamlinkrc file is now notepad ready 😄

gravyboat:
This looks good to me. @sbstp any concerns?

sbstp:
As long as it works, no concerns.

beardypig:
Be good to have some other people build and test it. I have only tested it on Windows 10.

gravyboat:
Thanks!

PR #174, opened 2016-11-17
Improve NSIS installer script

bastimeyer:
- Remove unnecessary start menu shortcut creation
- Don't automatically close the copy/remove file section
- Add checkbox for opening the documentation in the web browser
- Add checkbox for editing the config file

### DIFF
--- a/script/makeinstaller.sh
+++ b/script/makeinstaller.sh
@@ -54,6 +54,37 @@ cat >"${build_dir}/installer_tmpl.nsi" <<EOF
 !include "TextFunc.nsh"
 [% extends "pyapp.nsi" %]
 
+[% block modernui %]
+    ; let the user review all changes being made to the system first
+    !define MUI_FINISHPAGE_NOAUTOCLOSE
+    !define MUI_UNFINISHPAGE_NOAUTOCLOSE
+
+    ; add checkbox for opening the documentation in the user's default web browser
+    !define MUI_FINISHPAGE_RUN
+    !define MUI_FINISHPAGE_RUN_TEXT "Open online manual in web browser"
+    !define MUI_FINISHPAGE_RUN_FUNCTION "OpenDocs"
+    !define MUI_FINISHPAGE_RUN_NOTCHECKED
+
+    Function OpenDocs
+        ExecShell "" "https://streamlink.github.io/cli.html"
+    FunctionEnd
+
+    ; add checkbox for editing the configuration file
+    !define MUI_FINISHPAGE_SHOWREADME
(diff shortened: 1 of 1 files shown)


bastimeyer:
Publisher data for the uninstaller info in the Windows control center is still missing, as I mentioned here: https://github.com/streamlink/streamlink/issues/116#issuecomment-261114313

The installer file does have the default nsis icon right now. This should be changed once we have an icon for streamlink.

gravyboat:
This looks good to me! Any other input @sbstp, @beardypig?

beardypig:
Looks good, but I haven't had chance to test it out yet.

sbstp:
I just tested it, works well (Windows 10). The copy/remove window automatically closing really bothered me, thanks for fixing this. Nothing blocking a merge from my side.

gravyboat:
I'm going to merge this in. I think with this and https://github.com/streamlink/streamlink/pull/178 we're ready for the 0.0.3 release.

sbstp:
Yep I agree. We should setup a milestone for 0.0.4

gravyboat:
@sbstp You mean for 0.0.3? Or you want to create a new milestone for 0.0.4?

beardypig:
I also tested it out just now :) Worked fine for me too (Windows 10 VM). 
Uninstalling also removed the shortcut that was created by the current installer :)

bastimeyer:
> Uninstalling also removed the shortcut that was created by the current installer :)

Yep, the `uninstall_shortcuts` block also needs to be replaced with an empty one.
Is this negligible or do you want another PR?

beardypig:
I think removing the shortcut is suitable, I wonder if it could be removed at install time.

bastimeyer:
> I wonder if it could be removed at install time

Shortcuts are not being created anymore by the installer, so why do you want to remove shortcuts at install time?

beardypig:
When upgrading, at least it should remove the shortcut from an old install. Otherwise you might have an orphan shortcut.

bastimeyer:
Is it really the job of a newer installer to remove obsolete stuff from previous releases? I'd say that this is what the uninstaller of older releases is for. Removing stuff during the installation is a bit unexpected IMO. 

The startmenu shortcut could also be used-created and could have custom parameters, so that is actually useful. Upgrading would then always delete it.

gravyboat:
@bastimeyer I think the issue is that many users won't remove the old version before installing the new version. What does it look like if I do an installation of 0.0.3, then simply install 0.0.4? Will the old release and the associated content be removed? After a bunch of releases it would be a bummer if old stuff wasn't removed (similar to the trash that Nvidia does with their installer).

PR #178, opened 2016-11-19
stream.rtmpdump: fixed the rtmpdump path issue, introduced in 6bf7fd7

beardypig:
This fixes the issue reported by @mmetak.

That was my bad, missed testing on linux :| 

This patch uses `shutil.which` (introduced in Python 3.3, with included backport) to find the command on the PATH

### DIFF
--- a/src/streamlink/packages/shutil_which.py
+++ b/src/streamlink/packages/shutil_which.py
@@ -1,0 +1,77 @@
+"""Backport of shutil.which from Python 3.5
+
+The function is included unmodified from Python stdlib 3.5.1,
+and is (C) Python
+"""
+
+import os
+import sys
+
+__version__ = '3.5.1'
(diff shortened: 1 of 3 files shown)


gravyboat:
Thanks!

sbstp:
Just wondering if the code needs to be this complex. On Windows the path to rtmpdump needs to be absolute, but on Linux, it can just be `rtmpdump`, and subprocess.Popen will find it in the PATH by itself.

gravyboat:
I'm pretty sure this was required so we could properly support Python 2.7 sadly since shutil.which doesn't exist there.

beardypig:
@sbstp, @gravyboat is correct, it's just for `shutil.which`. In Windows, the current directory is also included at the front of the PATH.

PR #181, opened 2016-11-19
Remove shortcut from previous releases on Windows

bastimeyer:
Follow up of #174 

- Remove old start menu shortcut when installing streamlink >= 0.0.3
- Don't try to remove shortcuts when uninstalling (shortcuts are not being created anymore)

### DIFF
--- a/script/makeinstaller.sh
+++ b/script/makeinstaller.sh
@@ -100,7 +100,12 @@ cat >"${build_dir}/installer_tmpl.nsi" <<EOF
 [% endblock %]
 
 [% block install_shortcuts %]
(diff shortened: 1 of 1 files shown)


gravyboat:
Thanks!

beardypig:
@bastimeyer I think you were right before about not removing shortcuts that could have been created by the user. It's a tricky thing to solve, on the one hand you don't want the installer to leave junk around. But on the other hand, you don't want it to overstep and delete things the user created. 
But I suppose it's reasonable to remove the shortcut that it created. Maybe it should be noted somewhere in the docs or the installer intro... Thoughts?

PR #183, opened 2016-11-20
fix livestream thanks to @intact

### LINKED ISSUE #182: Plugin for Livestream.com not working right?  exit's quickly for hls and not at all for normal streams
I am trying to get a live stream on livestreamer.com to work and i can't get it to play more then about 35 seconds...

When I run this command:
streamlink "http://livestream.com/Miraclenet/events/5004281" 270p --fifo --player omxplayer

it gives me an error about an swf being needed.  When I run this command:
streamlink "http://livestream.com/Miraclenet/events/5004281" 270p_hls --fifo --player omxplayer

it will play the stream but just for about 35 seconds or so...  I kinda don't want to have to restart it every 35 seconds to watch this stream...  I'd like it to run until I stop it myself...

Any help for this non-python, non-linux guy would be very helpful...

btw, this is running on a Raspberry Pi.  Just got a nice little 7 inch lcd for it and set it up on my desk to be able to watch it while I work, but can't get it to play for long at a time...

(edited to correct commands used)

Brainzyy:
Just need @intact 's blessing

fixes #182

### DIFF
--- a/src/streamlink/plugins/livestream.py
+++ b/src/streamlink/plugins/livestream.py
@@ -22,7 +22,10 @@ _stream_config_schema = validate.Schema({
             ),
         }, None)
     },
-    validate.optional("playerUri"): validate.text
+    validate.optional("playerUri"): validate.text,
+    validate.optional("viewerPlusSwfUrl"): validate.url(scheme="http"),
+    validate.optional("lsPlayerSwfUrl"): validate.text,
+    validate.optional("hdPlayerSwfUrl"): validate.text
 })
 _smil_schema = validate.Schema(validate.union({
     "http_base": validate.all(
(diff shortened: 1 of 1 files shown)


beardypig:
For future reference you can set the author of a commit by doing `git commit --author "name <email>"` :-)

Brainzyy:
@beardypig Thanks for the info, but I'm a noob when it comes to command line git. I just commited this with github web.

gravyboat:
Are we good to merge this in? Any concerns past what @intact brought up? I would rather have it work most of the time than die all of the time.

beardypig:
As @intact pointed out, this reverts a previous fix for the naming of stream qualities in py2 vs. py3 (#96).

beardypig:
I created a separate PR with the commit from @intact which doesn't squash the previous fix in #96.

Brainzyy:
Thank you @beardypig, I'll close this now, sorry for my failure.

beardypig:
Hardly a failure :) Easy to miss :D

PR #184, opened 2016-11-21
Change icon

sbstp:
I tested it and it works (Windows 10). The only place where it shows is in the uninstall list in the control panel.


beardypig:
Works for me (Windows 10 VM).

cdrage:
LGTM

PR #185, opened 2016-11-21
Fix for plugin.livestream courtesy of @intact 

### LINKED ISSUE #182: Plugin for Livestream.com not working right?  exit's quickly for hls and not at all for normal streams

beardypig:
Latest fix from @intact for the livestream plugin (chrippa/livestreamer/pull/1277), without squashing previous `streamlink` fixes. Could fix #182, but needs to be tested on a Raspberry Pi.

### DIFF
--- a/src/streamlink/plugins/livestream.py
+++ b/src/streamlink/plugins/livestream.py
@@ -22,7 +22,10 @@ _stream_config_schema = validate.Schema({
             ),
         }, None)
     },
-    validate.optional("playerUri"): validate.text
+    validate.optional("playerUri"): validate.text,
+    validate.optional("viewerPlusSwfUrl"): validate.url(scheme="http"),
+    validate.optional("lsPlayerSwfUrl"): validate.text,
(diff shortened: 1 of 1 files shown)


gravyboat:
Thanks!

Issue #169, opened 2016-11-16
[Youtube] URL has an invalid label


StarWolf3000:
With a lot of Youtube URLs I only get
`[cli][error] Could not open stream: Unable to open URL: http://r4---sn-4g5e6n7k.googlevideo.com/videoplayback?id=o-AMx00u_Hy1utcNFZMyq4fLygBtKL09bG6RGnc83-F8Af&dur=14472.394&signature=977F6BDF0E09E5C488AA419336106DF2EBAEDF84.326F564D004C949EC33D69C1AB82A47553651372&mm=31&mn=sn-4g5e6n7k&upn=dyJorCtyKNU&initcwndbps=697500&ip=<IP address>&key=yt6&lmt=1477844257912727&mime=video%2Fmp4&ms=au&source=youtube&sparams=dur%2Cid%2Cinitcwndbps%2Cip%2Cipbits%2Citag%2Clmt%2Cmime%2Cmm%2Cmn%2Cms%2Cmv%2Cnh%2Cpl%2Cratebypass%2Csource%2Cupn%2Cexpire&mv=m&ratebypass=yes&pl=17&ipbits=0&mt=1479283246&itag=22&expire=1479304889&nh=IgpwZjAxLmZyYTE1Kg04Ny4xMjguMjM2LjUz (URL has an invalid label.)`

For example, this one works: https://www.youtube.com/watch?v=n6gI_RszUgY
while this one does not: https://www.youtube.com/watch?v=iVogxITG97k

Also why are the URLs HTTP and not HTTPS when fetching? With a fetch script I use in my browser, the URLs are always HTTPS.

beardypig:
I could not recreate this on Windows 10... It is possibly a bug in requests, did you use the `streamlink` installer? If you did, could you run the following commands in a command prompt and tell me what it prints? 

```bat
cd "C:\Program Files\Streamlink\pkgs"
"..\Python\python.exe" -c "import requests; print(requests.__version__)"
```

StarWolf3000:
The output is: 2.11.1
Thats the version that is installed with the 0.0.2 Windows setup.

Strangely it seems to work now and then, I will investigate this further on my system, so its maybe a bug on my side, because I do have other Python versions installed on my system, but they are not in the paths, and streamlink was installed on my second HDD with letter E.

Joseph65:
im also having the same error but in Linux
when i watch this live stream https://www.youtube.com/watch?v=HZR0E8TzCsM
[...]

PR #186, opened 2016-11-21
Pin requests to <2.12.0 to avoid the strict IDNA2008 validation

beardypig:
Currently `requests>=2.12.0` has strict validation for IDNA2008. This means that some URIs that are in use (specifically by YouTube) will not validate and thus will not work in `streamlink`. This PR pins `requests` to a version before `2.12.0`. We can use a later version if YouTube fix their non-compliant URIs or `requests` gives to option to use a less strict IDNA parer. 

This is documented in #169.

> If I understand it correctly, IDNA was extended to reverse xx--. UTS-46 defines the criteria for valid IDNA encoded unicode characters.
> 
> > The first 5 criteria are from [IDNA2008]. Criterion #2 in particular is meant to allow for future label extensions beyond just xn--, such as for future versions of IDNA. Some implementations appear to consider such extentions unlikely, and allow labels such as "r3---sn-apo3qvuoxuxbt-j5pe".

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,8 @@ if version_info[0] == 2 or (version_info[0] == 3 and version_info[1] < 4):
 if (version_info[0] == 2 and version_info[1] == 6 and version_info[2] < 3):
     deps.append("requests>=1.0,<2.0")
 else:
-    deps.append("requests>=1.0,<3.0")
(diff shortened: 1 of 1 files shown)


gravyboat:
Good catch.

PR #187, opened 2016-11-21
Update release scripts

cdrage:
Update the release script to include Windows binary generation as well.


gravyboat:
Good stuff! For the next release we should see about getting a new repo made for @RosadinTV's content then triggering that build from this one.

PR #189, opened 2016-11-21
Spelling error in release script

cdrage:
Encountered this minor issue in the release script. :) Damn capitals.


gravyboat:
Good catch.

PR #191, opened 2016-11-22
Modify the browser redirect

### LINKED ISSUE #190: Error with using --twitch-oauth-authenticate
Using that causes a browser to open to http://livestreamer.tanuki.se/en/develop/twitch_oauth.html?error=redirect_mismatch&error_description=Parameter+redirect_uri+does+not+match+registered+URI   In the main.py file this link is show http://streamlink.tanuki.se/en/develop/twitch_oauth.html when it should link to this https://streamlink.github.io/twitch_oauth.html  https://github.com/streamlink/streamlink/blob/master/src/streamlink_cli/main.py#L536

https://github.com/streamlink/streamlink/issues/4#issuecomment-261737748

gravyboat:
This simply ensures we're doing the redirect to https://streamlink.github.io/twitch_oauth.html instead of Chrippa's site. Fixes #190.

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -533,7 +533,7 @@ def authenticate_twitch_oauth():
        access to their Twitch account."""
 
     client_id = TWITCH_CLIENT_ID
-    redirect_uri = "http://streamlink.tanuki.se/en/develop/twitch_oauth.html"
+    redirect_uri = "https://streamlink.github.io/twitch_oauth.html"
     url = ("https://api.twitch.tv/kraken/oauth2/authorize/"
            "?response_type=token&client_id={0}&redirect_uri="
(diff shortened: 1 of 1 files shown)


sbstp:
Yeah, we'll probably get a few more of those. LGTM

thinkpad4:
I just tested the latest nightly with this fix in it, still being redirected to http://livestreamer.tanuki.se/en/develop/twitch_oauth.html?error=redirect_mismatch&error_description=Parameter+redirect_uri+does+not+match+registered+URI  and I checked main.py to be sure https://streamlink.github.io/twitch_oauth.html is listed, it is.

beardypig:
Unfortunately this won't work. The redirect comes from the twitch side, you need to create a new twitch app and set the redirect URL accordingly. Then you need to update the client id in the twitch plugin.

PR #197, opened 2016-11-23
Use ConsoleOutput for run-as-root warning

ioblank:
Right now when outputting the stream to stdout (-O option) as root the new warning will end up being printed alongside the stream's content.
It should work like other console messages. This way it will respect the loglevel set and it will be printed on stderr when outputting the stream to stdout.


beardypig:
Looks good to me :-)

gravyboat:

Issue #202, opened 2016-11-23
Can we get a plugin for the website 9anime.to?


Victolabs:

beardypig:
Sure! I created a PR with a plugin to support 9anime.to :)

gravyboat:
Closing this since it's good as soon as we merge #207

PR #207, opened 2016-11-24
Add support for 9anime.to

beardypig:
This PR adds support for the site 9anime.to as requested by @Victolabs in #202 :)

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -63,6 +63,7 @@ meerkat             meerkatapp.co        Yes   --
 mips                mips.tv              Yes   --    Requires rtmpdump with K-S-V patches.
 mlgtv               mlg.tv               Yes   --
 nhkworld            nhk.or.jp/nhkworld   Yes   No
+nineanime           9anime.to            --    Yes
 nos                 nos.nl               Yes   Yes   Streams may be geo-restricted to Netherlands.
 npo                 npo.nl               Yes   Yes   Streams may be geo-restricted to Netherlands.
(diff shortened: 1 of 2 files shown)


beardypig:
@Victolabs If you are able to test this, that would be great. I tested quite a few episodes, but I didn't try every episode so some might not work. If you find any that don't work, let me know and I'll try and fix them.

Victolabs:
@beardypig How would I go about downloading the file itself? And once I do that, does it go in this directory? `C:\Program Files (x86)\Streamlink\pkgs\streamlink\plugins`

beardypig:
@Victolabs Here's a link to the [nineanime.py](https://github.com/beardypig/streamlink/blob/e81977dee466c6eb8b0b4d6a890ed6f3ed74a2c6/src/streamlink/plugins/nineanime.py) file. Yep that's where you need to copy it to :)

Victolabs:
@beardypig thats just the code. Not the actual file itself. Nevertheless I got the file by downloading the "master" if you will (I'm still new to github. Sorry) and extracting the .py file there. 

So I have played around with it and it seems to be working perfectly! I just cant believe that you actually went ahead and made the plugin for me/everyone else! This has definitely hooked me on board with streamlink. 

And for a simple answer: Yes, the plugin works perfectly.

gravyboat:
Thanks @beardypig!

beardypig:
True @Victolabs, there is a raw button that will send you to actual file https://raw.githubusercontent.com/beardypig/streamlink/e81977dee466c6eb8b0b4d6a890ed6f3ed74a2c6/src/streamlink/plugins/nineanime.py :-)

Issue #146, opened 2016-11-09
Support for ATV Webtv


Rotarum:
Hey there,

this is from the official ATV (FTA Channel) Livestream:

http://www.atv.com.tr/webtv/canli-yayin

Can you add a Plugin for that ? :)

Greetings

gravyboat:
Support for this has been added.

PR #208, opened 2016-11-24
Add support for the live stream on atv.com.tr

beardypig:
This PR adds support for the live stream on http://www.atv.com.tr/webtv/canli-yayin as requested by @Rotarum in #146. Unfortunately I cannot read much of the site, so I don't really know what else can be streamed :)


gravyboat:
Thanks @beardypig!

PR #209, opened 2016-11-24
Include omxplayer in the list of players in the docs

beardypig:
I have added OMXPlayer to the list of players in the docs, and I added some notes about the workarounds for getting some streams to work as discussed in #182.

Comments? :)


gravyboat:
Looks good to me, thanks.

PR #152, opened 2016-11-11
plugins.connectcast: fix regex issue with python 3

beardypig:
Fixes #151 ... missed this one...

PR #217, opened 2016-11-25
More updates to the player docs about OMXPlayer

beardypig:
A small update to the player docs from #152 :)


gravyboat:
Good stuff, thanks!

PR #221, opened 2016-11-28
Added tigerdile.com support.

thatlinuxfur:
Credit to skulblakka for original implementation in livestreamer. See: https://github.com/skulblakka/livestreamer/commit/ec9ff879d06dd7db089a22234202a0f48daf9970

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -94,6 +94,7 @@ svtplay             - svtplay.se         Yes   Yes   Streams may be geo-restrict
                     - oppetarkiv.se
 tga                 - star.plu.cn        Yes   No
                     - star.tga.plu.cn
+tigerdile           tigerdile.com        Yes   --
 tv3cat              tv3.cat              Yes   Yes   Streams may be geo-restricted to Spain.
 tv4play             - tv4play.se         Yes   Yes   Streams may be geo-restricted to Sweden.
                                                      Only non-premium streams currently supported.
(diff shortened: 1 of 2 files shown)


beardypig:
FYI I haven't seen an online stream for this yet so I haven't been able to test it, but I think there are a couple of possible issues :-)

thatlinuxfur:
I took a stab at regex-ifying it. Was that what you had in mind?

beardypig:
Yep :-) I did pretty much the same, e040a808100c3ef24dec1fab139527f6fa675ddc. I also noticed that the subdomain can be swf, so I modified the regex to match any subdomain :-)

thatlinuxfur:
Cool. Anything else jump out at you that needs fixing?

beardypig:
Only to support the SFW urls, eg. https://sfw.tigerdile.com/stream/zannah/ 

The other changes I made were mostly just style changes.

thatlinuxfur:
Oh, OK. I misunderstood what you meant with the sfw URLs. I hadn't seen those before. Updated to match either www. or sfw.

beardypig:
Nice :)

I've still never seen a stream online though :)

beardypig:
@thatlinuxfur yeah, I made a typo :D swf vs sfw ... looks very similar ;)

thatlinuxfur:
Cool, I didn't know about that non-capturing group trick.

And yes, it's quite a small site. I think there are only ~100 members or so.

beardypig:
:) Well I'm happy to merge this. It would be nice to test it... but if it doesn't work, you can fix it :p

thatlinuxfur:
Thanks!

beardypig:
You're welcome. Thanks for your work :-)

Issue #216, opened 2016-11-25
plugin support request for Bigo Live Streaming Social Network ?


joneelee81:
Here is the homepage: https://www.bigo.tv/ It is a Live Streaming Social Network, similar to Twitter Periscope.

There are some other websites that appear to pull the streams from bigo live, here are some examples:
http://bigoweb.co/country/us
http://www.bigospace.com/lives/GB/

I can't seem to find the m3u8 sources, but they also use FlashVars:
`param name="FlashVars" value="sid=1000000&uid=22222&ip=183.129.210.18&port=7778"`

Seems like the parameters are generated from `http://live.bigo.tv/0123456789` where `0123456789` might be the `sid`

beardypig:
Hi @joneelee81, I have created a PR which adds support for bigo.tv (live.bigo.tv and bigoweb.co). If you want to test it out you can download the plugin file [bigo.py](https://raw.githubusercontent.com/beardypig/streamlink/92533d2666dd98da9e373662f59ac1868630a28b/src/streamlink/plugins/bigo.py) and copy it to your `streamlink` plugins directory.

Where the plugins directory is depends on your OS and how you installed `streamlink`. If you are using Windows you can put extra plugins in `%APPDATA%\streamlink\plugins` and if you are using a POSIX-like OS (Linux, macOS) you can put the plugins here: `~/.config/streamlink/plugins` (you might need to create the directory first).

beardypig:
Fixed by #222.

joneelee81:
@beardypig - just tested the `bigo.py` file, by placing it in `~/.config/streamlink/plugins` - I'm on macOS Sierra 10.12.1 & streamlink 0.1.0.

First impressions, all appears good.  I will report back if I find any issues.

THANK YOU - your support for this project is very much appreciated.

Rotarum:
Bigo Live broken

"error: No plugin can handle URL: http://www.bigo.tv/sid...."

Can someone have a look?

bastimeyer:
This thread has been closed more than two years ago.
Please open a [new plugin issue](https://github.com/streamlink/streamlink/issues/new?template=plugin_issue.md) and provide the necessary details as described by the template.

PR #222, opened 2016-11-28
Add support for bigo.tv

beardypig:
As requested by @joneelee81 in #216. 

This one was quite interesting, it uses a custom streaming protocol. To start the stream you have to connect to the ip and port (found in the flashVars) and then send both the `uid` and `sid` as little-endian unsigned longs. Then the server responses with an FLV :)


gravyboat:
Seems like a weird process to get streaming, but if it works it works! Thanks.

beardypig:
I guess one of their engineers wanted to make their own custom streaming server/protocol :-) Maybe they thought it would be harder to reverse, bwhahah...

PR #223, opened 2016-11-28
Small update to the install docs

beardypig:
Updated the install docs to reflect the reality of installing `pycryptodome` automatically


gravyboat:
Thanks.

Issue #227, opened 2016-11-28
Support for DingIt.tv


Brainzyy:
This is a continuation of https://github.com/chrippa/livestreamer/issues/974
I'm usually pretty good at figuring out stream links and all that stuff, but this one had me scratching my head.
All the talk about a extension being needed to play their streams is not true anymore.

beardypig:
Sounds like a fun challenge :-)

Brainzyy:
@beardypig Yeah, I thought so too ;) From what I could find out last time, they use a service which obfuscates all the trafic and makes it (for me at least ;) ) unsniffable.

beardypig:
It think it was quite simple to find the streams, maybe I missed something??
All the highlights or previous live events I tried were available in HLS. @Brainzyy do you know any streams that were ocshield (octoshape) only?

beardypig:
I created a PR which supports HLS, I was unable to find a stream that did not work. If you find one, let me know :)

Brainzyy:
Thank you very much @beardypig. Is there any way you could explain how you did this? Last time I tried, everything was obfuscated with ocshield.

beardypig:
Just fired up the good ol' Chrome developer tools... When looking at those streaming sites, I normally search in the network list for `m3u8`, `mpd`, etc. `master.m3u8` pops right out :) Working back from there, it was the `api/get_player_flashvars` call to get the stream url path. Poking around in the swf just confirmed that the HLS urls were hardcoded. I did a bit of mitmproxy foo to get the client to request the the `akaurl` based stream, by deleting the `pereakaurl` key from the response :) 

They might have changed it recently to move away from the octoshape stuff :)

Brainzyy:
Yeah, I'm sure I checked all that stuff last time :p It's been a while. Thank you for checking again and doing all that effort :) @beardypig

gravyboat:
Hey guys feel free to keep the discussion going but I'm going to close this since I merged the associated PR.

PR #228, opened 2016-11-29
Add support for dingit.tv

beardypig:
As requested by @Brainzyy in #227 and other in chrippa/livestreamer#974. 

I have not tested live streams as there were none to test.

It is possible that some stream will not work, in addition to HLS [dingit.tv](http://www.dingit.tv) also use [Octoshape](http://octoshape.com) which provides a proprietary streaming server and flash client. I had a bit of a poke around trying to reverse it but it is quite complex, and if HLS works I see no point trying to support the Octoshape client. 

The HLS URLs are hardcoded in the swf player, so could be changed with a new version of their client. 

If anyone tests this plugin out and finds a stream that does not work, please report it :)


gravyboat:
Thanks @beardypig!

PR #238, opened 2016-12-01
Add twitch_oauth to index.rst

mmetak:


docs/index.rst
```diff
@@ -33,7 +33,7 @@ such as:
 on the :ref:`plugin_matrix` page.
 
 Quickstart
------------
+----------
```
gravyboat: Thanks for catching this, I'm not sure why we're indenting by 1 to begin with, but this is a good find.

docs/changelog.rst
```diff
@@ -3,4 +3,4 @@
 Changelog
 =========
 
-.. include:: ../CHANGELOG.rst
+.. include:: ../CHANGELOG.md
```
gravyboat: Did you build the docs on this one to confirm it will look okay? I considered updating this a while back but when I searched online it seems like the rst file won't render the markdown correctly.

mmetak: What would then be best approach? Convert `CHANGELOG.md` to rst and put it in `docs/changelog.rst`, Convert `CHANGELOG.md` to `CHANGELOG.rst` in root dir and leave `docs/changelog.rst` as it is or something else?

gravyboat: @mmetak I'm not sure to be honest, when @cdrage builds the changelog it's in md format, but we use RST for the docs. We might have to see how much effort it would take to modify the format of the changelog. I would rather do that than update all the docs.

gravyboat:
@mmetak I think we'll have to modify the way the setup script does the generation as opposed to moving it to an RST file so it gets generated accurately on every build. Otherwise it would be really tedious to have to do this every time!

gravyboat:
@cdrage What do you think about generating the changelog as an rst file instead of a markdown file? Right now the issue is since all the docs are RST the changelog can't be easily included without formatting changes.

mmetak:
No problem @gravyboat, I'm fairly new to all this so thank you for bearing up with me. ;-)

gravyboat:
@mmetak No worries, I appreciate you working on this. Having the docs set up properly is extremely important!

cdrage:
@gravyboat fine by me, I'll make a manual change + update the release script for this.

gravyboat:
@mmetak Do you want to close this PR out, then open a new one without the changelog included? Then we'll get your other improvements while keeping the automated generation of the changelog. I hate leaving an issue open for this long when there are great changes to be had.

mmetak:
@gravyboat if I commit just the twitch without changelog it should be fine right?

docs/index.rst
```diff
@@ -64,3 +64,4 @@ See their respective sections for more information on how to use them.
     issues
     api_guide
     api
+    twitch_oauth
```
beardypig: The other change was still good :)

beardypig: OK. Cool :)

PR #241, opened 2016-12-03
Update client ID

gravyboat:
Fixes https://github.com/streamlink/streamlink/issues/190


beardypig:
Works like a charm. Thanks ;)

PR #249, opened 2016-12-05
plugins.crunchyroll: support ultra quality for subscribers

### LINKED ISSUE #243: Crunchyroll can't find 1080p streams
The current plugin seems to only find streams up to 720p.
However I tested the old plugin and it seems to be working again I tested the links I used in #70 and a few more but couldn't reproduce the issue with the old plugin.
More testing may be needed but it's more than likely safe to revert back.

beardypig:
Missed this last time, didn't test the subscriber streams. 

Should resolve #243, also renames the streams from bitrate to low, mid, high so that all the names are consistent :)


gravyboat:
Thanks @beardypig!

Issue #246, opened 2016-12-05
"failed to create process" when launching Streamlink in Chocolatey test environment


scowalt:
Ran into this trying to validate the `streamlink-twitch-gui` Chocolatey package. The [Chocolatey test environment](https://github.com/chocolatey/chocolatey-test-environment) is a Windows Server 2012 virtual machine that allows package maintainers to test packages in an isolated environment.

After running the windows installer in this environment, trying to execute `streamlink.exe` gives the following error:
```
failed to create process
```

As a work-around, I tried running `Python\python.exe bin\streamlink-script.py`. This gave me the following error:
![image](https://cloud.githubusercontent.com/assets/1100550/20870286/1f050284-ba3a-11e6-84ad-ab43f17680c1.png)

Some quick research led me to believe that the VC++ 2015 redistributable might be necessary to run streamlink. After installing this redistributable and restarting the virtual machine, I get a different error running `Python\python.exe bin\streamlink-script.py`:
```
C:\Program Files (x86)\Streamlink>Python\python.exe bin\streamlink-script.py
Failed to load plugin viagame:
  File "imp.py", line 234, in load_module
  File "imp.py", line 172, in load_source
  File "<frozen importlib._bootstrap>", line 693, in _load
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 665, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "pkgs\streamlink\plugins\viagame.py", line 7, in <module>
    from streamlink.plugin.api.support_plugin import viasat
  File "pkgs\streamlink\plugin\api\__init__.py", line 27, in __getattr__
    return load_support_plugin(name)
  File "pkgs\streamlink\plugin\api\support_plugin.py", line 39, in load_support_
plugin
    raise ImportError("No module named '{0}'".format(name))
ImportError: No module named 'viasat'


usage: streamlink-script.py [OPTIONS] [URL] [STREAM]

Use -h/--help to see the available options or read the manual at http://docs.streamlink.io/
```

I'm not sure why this is happening. Perhaps someone more knowledgeable knows what's going on here. If you need to get a local repro going, the [Chocolatey test environment](https://github.com/chocolatey/chocolatey-test-environment#requirements) is pretty easy to set up and is cross platform.

Vangelis66:
@scowalt wrote: 

> Some quick research led me to believe that 
[...]

PR #250, opened 2016-12-05
Include Visual Studio 2015 Runtime in the installer

beardypig:
As pointed out by @scowalt in #246, the Visual Studio 2015 Runtimes are required for some versions Windows. I have modified the installer to install the VC runtimes if they aren't already installed. The [Python Embedded Distribution](https://docs.python.org/3.5/using/windows.html#embedded-distribution) page also points out that the runtimes are not included and should be installed by the installer. This is possibly an oversight of `pynsist`, but we can fix it ourselves :)

I tested this out of a vanilla Windows Vista 32bit VM, it would be good if @scowalt could test it out on the Chocolatey test environment and if some other people could test it out :)

I have build a version of the installer that people can use to test this out: [streamlink-0.1.0-vcredist-7a50314.exe](https://s3.amazonaws.com/streamlink-builds/beardypig/streamlink-0.1.0-vcredist-7a50314.exe).

### DIFF
--- a/script/makeinstaller.sh
+++ b/script/makeinstaller.sh
@@ -55,7 +55,7 @@ EOF
 
 cat >"${build_dir}/installer_tmpl.nsi" <<EOF
 !include "TextFunc.nsh"
-[% extends "pyapp.nsi" %]
+[% extends "pyapp_msvcrt.nsi" %]
 
 [% block modernui %]
     ; let the user review all changes being made to the system first
@@ -90,7 +90,6 @@ cat >"${build_dir}/installer_tmpl.nsi" <<EOF
 
 [% block install_files %]
     [[ super() ]]
-
     ; Install config file
     SetShellVarContext current # install the config file for the current user
     SetOverwrite off # config file we don't want to overwrite
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -176,6 +176,11 @@ instead.
 Windows binaries
 ----------------
 
+.. important::
+
+    Windows XP is not supported.
+    Windows Vista requires at least SP2 to be installed. 
+
 You can download the latest stable Windows installer `here <https://github.com/streamlink/streamlink/releases>`_.
 
 You can download the latest nightly Windows installer `here <https://streamlink-builds.s3.amazonaws.com/nightly/windows/streamlink-latest.exe>`_.


script/makeinstaller.sh
```diff
@@ -89,6 +91,25 @@ cat >"${build_dir}/installer_tmpl.nsi" <<EOF
...
+    \${If} \${RunningX64}
+        ReadRegStr \$1 HKLM "SOFTWARE\Wow6432Node\Microsoft\VisualStudio\14.0\VC\Runtimes\x86" "Installed"
+        ; if installed, jump past the installer
+        StrCmp \$1 1 installed
+    \${Else}
+        ReadRegStr \$1 HKLM "SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x86" "Installed"
+        StrCmp \$1 1 installed
+    \${EndIf}
+
+    ; If not installed run the installer
```
scowalt: Should we also run the x64 installer on 64-bit systems? Using the `vcredist2015` chocolatey package as reference, it looks like we may want to install both x86 and x64 on 64-bit systems

beardypig: As far as I understand it, we use the 32 bit Python and thus only require the x86 version of the VC runtime.

beardypig: As for pulling it from a remote server, I considered it but thought it would cause less problems to just include the binary, less chance of it not working :-)

gravyboat:
This looks fine to me, have you guys finished all the testing you want to do?

scowalt:
Having trouble building the binary for this. If @beardypig could share a the installer built with this script I will test it in the Chocolatey environment.

beardypig:
Sure I'll build one in a bit a link to it in the top post.

beardypig:
@scowalt I have added a link to a build of the installer in the first post :)

scowalt:
Looks like I still get the same error after running the executable and restarting the machine:
image
image

My only thought is that the vcredist2015 chocolatey package seems to install some windows updates that could be important in getting everything to run.

Either way, I'm not sure how much you want to block on this. I could always add vcredist2015 to the dependencies in the Chocolatey package. This would potentially cause us to contribute to the VC++ wall of doom, but at least we'll know it's working for everyone:
wall of doom

beardypig:
OK @scowalt, maybe the x64 installer does need to be included as well... I will try some stuff on the chocolatey test env :)

beardypig:
Yeah, sounds like you also need some Windows update for Server 2012 and Windows 8.1 for it to work...

beardypig:
@scowalt I'd go with adding vcredist2015 as a dependency to the Chocolatey package. This could turn in to a bit of rabbit warren, with Windows updates, etc. 

It might just be better to tell people they need to install the VC 2015 Runtimes and make sure all the Windows updates are installed. 

What about if we give a warning yes/no dialog before the installer if the VC2015 runtimes are not installed? Something like:
```
Visual C++ Runtime x86 is required! Streamlink will not work without it. Do you still wish to continue?
```

scowalt:
I'll add `vcredist2015` as a Chocolatey dependency.

I'm fine with a pre-install warning. Maybe a link to the runtime if we're not going to bundle it? Also, listing it as a requirement in the README would be nice.

I would just confirm that whatever warning you add will still allow the installation to run silently. I'm running the installer silently using the command `<installer>.exe /S`. See https://github.com/streamlink/streamlink-chocolatey/blob/master/Streamlink/tools/chocolateyinstall.ps1#L19

scottbernstein:
@Scott Walters --


I have seen this "failed to create process" error on streamlink on Windows -- the issue relates to where the Python install directory is located.  If it has spaces in the path you will see this error, for some reason.

beardypig:
I feel pretty dumb as `pynsist` actually supports this ... 😓 

I have updated the installer script, and also the docs to say that XP is not supported 😃

bastimeyer:
Wouldn't it be better to use `[% extends "pyapp_msvcrt.nsi" %]` at the head of our custom template instead of `[% extends "pyapp.nsi" %]`?
https://github.com/streamlink/streamlink/blob/0.1.0/script/makeinstaller.sh#L58
https://github.com/takluyver/pynsist/blob/1.9/nsist/pyapp_msvcrt.nsi#L1
This way you don't need to copy its content.

beardypig:
@bastimeyer didn't work when I tried it...

beardypig:
Must have been a temporal anomaly, cos it works fine now .... :|

beardypig:
@scowalt I've updated the installer in the top post, and I tested it on the Chocolatey vagrant machine and it worked as expected. Do you want to give it a go. 

I think this is good to merge unless anyone has any objections :-)

gravyboat:
@beardypig I'm fine with it, I'll leave it unmerged for a bit in case there are any other comments, if not I'll merge it tonight.

beardypig:
No objections? It seems to work, see #255 and #267.

gravyboat:
Thanks @beardypig!

thinkpad4:
Is Visual Studio 2015 Runtime included in the nightly builds?

scowalt:
Thanks for getting this done @beardypig!

This will get picked up in Chocolatey when the executable gets built for the next release (0.1.1 or 0.2.0 😄). Until then, I have a Chocolatey package in the moderation queue that adds `vcredist2015` as a dependency.

beardypig:
@thinkpad4 sure is, did you have a problem with it?

thinkpad4:
@beardypig Nope just making sure.  Thanks

beardypig:
@RosadinTV for your portable installer, you might need to add the UCRT DLLs (if you don't already, I'm not sure if you do), so that users don't run in to the same problem when they don't have KB2999226 (Windows 10 Universal C Runtime) installed.

RosadinTV:
@beardypig If i use your "streamlink\python\" folder is right? (I see it contains api-ms-xx dll's)

beardypig:
@RosadinTV there are two zip files that you can use an x86 one and an x64 one, might be cleaner for you to use those :-)

RosadinTV:
@beardypig I will use the x86 one, thanks ❤️

PR #257, opened 2016-12-06
Update Archlinux and Ubuntu install instructions, add Solus and Void

mmetak:


beardypig:
If you wanna tidy up the rest of the installers go ahead. Most of them do not work, as far as I can tell these work:

- Arch Linux (aur)
- Arch Linux (aur, git)
- Gentoo Linux
- Solus _the one you added_
- Ubuntu _the one you added_

We should remove the ones that don't work, so that the ones that do work don't get missed :) 

And you can remove the Warning :)

beardypig:
Cool. Looks good to me :)

mmetak:
@beardypig just a sec, have 'void' linux to add ;-)

beardypig:
Nice. Some of the packages are out of date, I'll drop an email to the maintainers and ask them if they will update to the latest version :)

beardypig:
Thanks @mmetak, really useful :)

PR #259, opened 2016-12-06
Update URL for docs to point to the github.io page

beardypig:
The CLI was giving out invalid URLs for the documentation.


gravyboat:

Issue #166, opened 2016-11-15
Start playing stream immediately


RosadinTV:
It would be cool if stream start playing before waiting to download a whole segment, i mean if you directly play a stream using a player in about 2 seconds loads but with livestreamer/streamlink we have to wait until the first segment (i guess) is downloaded and this can take a time in some connections.
In summary it will be cool to start playing when some data is available instead of waiting for whole download, the same happens with downloads, the state report only refresh when a whole segment is downloaded (maybe like #165 reported)

agentqwe:
Exactly! 
for examp twitch broadcasting mpeg-ts video, so what is problem play segment immidiatly?(im not programmer?so maybe my words sound stupid...))

RosadinTV:
@agentqwe I think the problem is the program wait for download the whole segment before start playing or refreshing download.
For example if a stream using HLS have 10 segments of 15 mb each, instead of playing the segment when 1 mb is downloaded Streamlink wait until the 15 mb is downloaded.
The same goes for the cache, i mean .. the cache is only added to the player when the whole next segment is downloaded.

agentqwe:
for example when i use ff for twitch and this - https://addons.mozilla.org/ru/firefox/addon/twitch_5/ in statistics of addon i saw segments of 2 and 4sec with bitrait 3-4mbps(forexpl) so its about 1-1.5mb every segment, i set in twitch addon playerbiffer 1sec, it was like in player sometimes buffer left less than 1sec )), its not bad coz it let see stream with almost no delay to broadcaster, but if next downloading segment will play immidiatly - buffer will never clear even buffer length will be 1 or 0 sec  :) am i right?

RosadinTV:
@agentqwe I dont know if i understand you but you if you have a good internet connection you can watch a stream almost without delay because the buffer can maintain in a few seconds from live, with livestreamer you can use --hls-live-edge 1 to stay as close as possible to live.
The difference is that if the segment is for example 6 seconds each you cant stay 1 second close to buffer with streamlink like with playing directly with a player.
In my case i dont care about stay close to live, in fact the best thing of the Project for me is the ability to store good amount of cache without issues, only that i will preffer to dont wait to whole segment download to start viewing a stream and the thing goes worst when you use --hls-live-edge with a big number because the stream only start when download the x specified ammount of segments.

agentqwe:
yes, i use --hls-live-edge 1, yep for up to 5mbps streams my inet work good
[...]

PR #260, opened 2016-12-06
Start playing HLS Streams as soon as any data has been download

beardypig:
As pointed out in #166, HLS streams do not start playing right away. 

Instead of waiting for the full segment to be downloaded, start writing out the segments as they are downloaded. This will improve the delay starting HLS streams and the delay from "live" in live streams.


gravyboat:
Nice this is a good change! Thanks!

Issue #234, opened 2016-11-30
Bigo live plugin not working


eccccco:
hey guys,
I am not able to open streams on Bigo Live.
I tried these urls with streamlink: 
http://bigoweb.co/show/[web ID]
live.bigo.tv/[web ID]

btw. thanks for continuing the livestreamer project!

beardypig:
@eccccco can you give the URLs you tried and what error it gave?

beardypig:
Do you have the latest nightly or installed from git? The bigo plugin is not available in the latest stable version (0.1.0).

eccccco:
i was using the stable version. the message streamlink gave me was: "no plugin can handle the url ..." 
i am going to try the nightly version (windows) and report back

beardypig:
@eccccco have you tried enabling caching in your player? eg. If you are using VLC try adding `--player 'vlc --file-caching 1000'` to the streamlink command (see https://streamlink.github.io/issues.html#streams-are-buffering-lagging).

clagnuts:
Hi,

I also can't now get this particular plugin to work. I'm using the latest nightly build and entering urls from bigoweb.co (e.g. http://bigoweb.co/show/2123012099 per previous poster). The correct plugin is found okay, but then I get the error "No streams found on this URL."

I believe that Bigo may recently may altered how their streams work, since some third party websites hosting their streams have also stopped loading them properly in recent days.

Many thanks for a great app!

gravyboat:
Could you please try the new nightly when it gets built? We just merged #273.

clagnuts:
Working now, thanks for your help.

DarkParousia:
Hello, is this still up? I got a problem with bigo. The video isnt loading for me. I uploaded a video of it so it would be alot clearer what my problem is. Do you guys have any solution for it? It would be a huge help if yes, im facing this issue for  weeks now :((( btw this is the video of my issue. https://www.youtube.com/watch?v=OWsNmM0YLL8&t=95s

bastimeyer:
No, this thread has been closed almost a year ago and was opened for a different reason.

Please use the issue tracker's search function next time before posting into dead threads:
https://github.com/streamlink/streamlink/issues?q=is%3Aissue%20is%3Aopen%20bigo
I'm not sure though, why there are two different threads.

There's also a pull request (#1053), but the code submitter seems to have abandoned their work.

youstarvideochatvlogsapp:
Hey Guys,
[...]

PR #273, opened 2016-12-08
Updated bigo plugin to support site changes

beardypig:
Should fix the issues reported in #234. The site was updated and the FlashVars were changed.


gravyboat:
Good stuff, thanks!

PR #274, opened 2016-12-08
License notices for bundled libraries/software 

beardypig:
I compiled some license notices to keep us a bit more legit :)


gravyboat:
These are good to have, thanks!

PR #278, opened 2016-12-08
Add metadata to Windows installer

### LINKED ISSUE #277: Version and Publisher in Windows Control Panel?
Not an issue in itself, rather an enhancement suggestion...

When the stable or nightly NSIS installer is executed on Windows 
(Vista SP2 x86 in my case), the "version" and "Publisher" columns 
in Control Panel's "**Programs and Features**" (or its iteration in 
other Windows versions) are not being populated by the installer. 
You can observe that in this attachment: 

prog feat

As you can see, some other installed apps do follow the same route, 
but the acceptable standard is to include both, version at least...
E.g., current version is **0.1.0** and Publisher could be set to 
**Streamlink**/ **Streamlink Organisation** or whatever string 
the devs deem appropriate...
I hope you do consider it... :smile_cat:

bastimeyer:
Resolves #277 

Added version, publisher, about-url, help-url and estimated-size data.

virtualbox_win7_08_12_2016_23_07_04

### DIFF
--- a/script/makeinstaller.sh
+++ b/script/makeinstaller.sh
@@ -55,6 +55,7 @@ installer_name=${dist_dir}/${STREAMLINK_INSTALLER}.exe
 EOF
 
 cat >"${build_dir}/installer_tmpl.nsi" <<EOF
+!include "FileFunc.nsh"
 !include "TextFunc.nsh"
 [% extends "pyapp_msvcrt.nsi" %]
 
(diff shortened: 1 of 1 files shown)


gravyboat:
Good stuff! Thanks @bastimeyer!

beardypig:
Cool, that means the nightlies should get fancy versions ...

Vangelis66:
Thanks to all :+1: 

I've just installed latest nightly, it **does** incorporate this PR: 


p f

Nothing fancy in the version string, IMHO :wink: :smile:

One small nitpick though; the size column reports it to be ca. 67.5 MB, 
while Windows Explorer reports that "%ProgramFiles%\Streamlink" is sized only 
half of that (31.3MB/33.5MB on disk); something particular to my system? 

pfslsize

beardypig:
@Vangelis66 it calculates the size at install time and it doesn't update. Did you remove anything from the Streamlink folder since installing?

Vangelis66:
@beardypig asked: 

> Did you remove anything from the Streamlink folder since installing?

Absolutely not! It was a clean install, i.e. previous nightly version was uninstalled, 
laptop rebooted, ccleaner used to remove any registry leftovers, %ProgramFiles% 
did not contain anything streamlink related, as expected "%APPDATA%\streamlink" 
contained streamlinkrc (which I didn't remove manually). 
It was only after above steps that I proceeded with the new nightly installer setup...

I understand "size" is calculated only approximately, but the huge difference 
between calculated and actual size was it that made me ask...

beardypig:
Not sure what would cause that... I cannot reproduce it :(

Vangelis66:
@beardypig said

> Not sure what would cause that... I cannot reproduce it :(

Well, I ain't sure what fixed it, but this is what I did: 
1. Uninstalled **streamlink 0.1.0+4b52a65** from Control Panel => Programs and Features
2. Rebooted laptop
3. Used CCleaner to remove any remnants of previous installation
4. Re-installed **streamlink-0.1.0-379-4b52a65.exe**

Now size column reports a size figure identical to the real one 
[Size on disk: 33,5 MB (35.155.968 bytes)]: 

p f2

All's well that ends well!

beardypig:
Interesting. Does CCleaner report which things it removed?

RuKeBo:
I just installed latest nightly Windows build and mine says 86.6 MB in Programs & Features and 89.4 MB on disk in Explorer.  Pretty close but much higher than others.

beardypig:
@RuKeBo do you have the ffmpeg stuff installed?

RuKeBo:
Yes, ffmpeg installed.  54 MB.

PR #294, opened 2016-12-14
Support for the live channels on trt.net.tr

beardypig:
Adds support for live streams on http://www.trt.net.tr.


gravyboat:
Thanks @beardypig!

Issue #293, opened 2016-12-14
Parameter parsing bug on Windows


bastimeyer:
The parameter parser doesn't handle escaped characters correctly on Windows.

Running this works on Linux/MacOS, but fails on Windows:
```
streamlink -p /path/to/vlc -a "--input-title-format \"\\\"\" {filename}" STREAM best
```

This is related to chrippa/livestreamer#1337
(reported at streamlink/streamlink-twitch-gui#256 and streamlink/streamlink-twitch-gui#362)

I can't remember if Linux/MacOS were affected by this in Livestreamer. Didn't test it...

The error log:
```
[cli][info] Starting player: C:/Program Files (x86)/VideoLAN/VLC/vlc.exe
Traceback (most recent call last):
  File "C:\Program Files (x86)\Streamlink\bin\streamlink-script.py", line 12, in <module>
    main()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 899, in main
    handle_url()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 499, in handle_url
    handle_stream(plugin, streams, stream_name)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 382, in handle_stream
    success = output_stream(stream)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 259, in output_stream
    output.open()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\output.py", line 21, in open
    self._open()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\output.py", line 123, in _open
    self._open_subprocess()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\output.py", line 138, in _open_subprocess
    self.player = subprocess.Popen(self._create_arguments(),
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\output.py", line 116, in _create_arguments
    return shlex.split(cmd) + shlex.split(args)
  File "shlex.py", line 273, in split
  File "shlex.py", line 263, in __next__
  File "shlex.py", line 90, in get_token
  File "shlex.py", line 166, in read_token
ValueError: No closing quotation
```

beardypig:
[...]

PR #300, opened 2016-12-15
Windows command line parsing fix

beardypig:
This PR does a couple of things. 

Firstly, it fixes the issue reported by @bastimeyer in #293, with some tests (see below). `subprocess.Popen` can be called with a string for the command line under Windows (same as `shell=True` under POSIX). Calling `subprocess.Popen` with the string command line means that it doesn't have to be parsed using `shlex` or quoted, which avoids the parsing issue. Parsing is still required on POSIX as `shell=True` is not advised. 

I tested in `cmd` and `powershell` on Vista and Windows 10:
```powershell
streamlink -a "--input-title-format \"Poker \\\"Stars\\\"\" {filename}" twitch.tv/pokerstars worst
streamlink -p "C:\Program Files\VideoLAN\VLC\vlc.exe --input-title-format \"Poker \\\"Stars\\\"\"" twitch.tv/pokerstars worst
streamlink twitch.tv/pokerstars worst
```
Also, tested in `streamlink-twitch-gui`, I disabled the custom window title and then added `--input-title-format "\"{name}\" {game}"` to the custom parameters. VLC displayed `"Stream Name" Game` in the title :)

Secondly, it adds some automated testing on Windows via AppVeyor, the test script only installs and runs the tests. I moved the dev/test dependencies to `dev-requirements.txt` so they could be shared between Travis and AppVeyor easily. It might need some extra tweaking after the merge though :)

### DIFF
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -1,0 +1,108 @@
+import os.path
+import unittest
+
+import streamlink_cli.main
+from mock import patch, ANY
+from streamlink import Streamlink
+from streamlink_cli.compat import is_win32
+
+PluginPath = os.path.join(os.path.dirname(__file__), "plugins")
+
+
+def setup_streamlink():
+    streamlink_cli.main.streamlink = Streamlink()
+    streamlink_cli.main.streamlink.load_plugins(PluginPath)
+    return streamlink_cli.main.streamlink
+
+
+class TestCommandLineInvocation(unittest.TestCase):
+    """
+    Test that when invoked for the command line arguments are parsed as expected
+    """
+
+    @patch('streamlink_cli.main.setup_streamlink', side_effect=setup_streamlink)
+    @patch('subprocess.Popen')
+    @patch('sys.argv')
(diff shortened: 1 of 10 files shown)


gravyboat:
Nice @beardypig! This looks good to me. I'm going to leave it open in case anyone else has feedback (especially @bastimeyer).

tests/test_cmdline.py
```diff
@@ -0,0 +1,112 @@
...
+        if not passthrough:
+            mock_popen.assert_called_with(commandline, stderr=ANY, stdout=ANY, bufsize = ANY, stdin = ANY)
+        else:
+            mock_popen.assert_called_with(commandline, stderr=ANY, stdout=ANY)
+
+    #
+    # POSIX tests
+    #
+
+    @unittest.skipIf(is_win32, "test only applicable in a POISX OS")
```
bastimeyer: Found a typo 😅 
Also the copied lines below...

beardypig: It actually refers to a French OS PoisX ...

gravyboat:
Good catch I totally missed that, merging! Thanks for reviewing @bastimeyer!

PR #302, opened 2016-12-15
Small tweak to the coverage config

beardypig:
`streamlink_cli` wasn't included before, it is now. 
The vendored packages don't need to be covered...

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,9 +1,13 @@
 [run]
-source = streamlink
+source =
+    streamlink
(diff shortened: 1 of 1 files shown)


gravyboat:
Good catch, thanks!

beardypig:
The real question is, what does it do to the coverage....

gravyboat:
@beardypig +6.97%, not bad!

beardypig:
When it's above 50% we can have a party...

PR #303, opened 2016-12-15
plugins.atresplayer: add support for live streams on atresplayer.com

beardypig:
Geo-restricted to spain.


gravyboat:
Thanks @beardypig, you're the champion of obscure geo-restricted streaming sites!

beardypig:
Got some more Turkish ones coming up too ;)

Issue #253, opened 2016-12-05
Stream from local m3u8 playlist file or stdin?


zacklb:
Is it possible to give streamlink as input an m3u8 on disk or piped in from stdin?

The following didn't work
```
streamlink hls://file.m3u8 best -o out.ts
streamlink hls://file://file.m3u8 best -o out.ts
streamlink hls:///mnt/g/file.m3u8 best -o out.ts
```
It would also be nice if you could pipe in the url to an m3u8 or it's contents directory via stdin.
For example
```
echo $some_url | streamlink hls://stdin best -o out.ts
streamlink hls://stdin best -o out.ts < file.m3u8
curl ... | streamlink hls://stdin best | ffmpeg ...
```

beardypig:
Neither are currently possible, but local file opening would be a cool addition. 

Would you want to pipe the contents of the file or the URL. If you want to do the URL you can do it easily with a bit of scripting. I think piping the contents would be more useful, and would go hand in hand with local file operations. 

Related: chrippa/livestreamer/issues/1514

zacklb:
> I think piping the contents would be more useful, and would go hand in hand with local file operations.

I agree. Maybe a unified `file://` handler, where if it's empty (or `-`), the handler looks at stdin, otherwise it interprets the rest as a filepath. Although, streamlink would then have to figure out what protocol the stream is. Or one could nest the protocol prefixes on input: `hls://file://...`, but this is pretty messy.

beardypig:
It would have to include the protocol prefix, the default would still be http though. I think it might be doable with some relatively small changes to regular http functions, which are wrappers around `requests`. Possibly something along the lines of https://pypi.python.org/pypi/requests file.

beardypig:
@zacklb do you have an test cases that you could provide?

beardypig:
@zacklb I have a WIP if you want to check it out. I have not tested it on Windows, but it should support relative paths, absolute paths and reading from stdin. https://github.com/beardypig/streamlink/tree/local-files

For example
```
streamlink hlsvariant://file://./playlist.m3u8 best
streamlink hlsvariant://file://../playlist.m3u8 best
streamlink hlsvariant://file:///path/to/playlist.m3u8 best
cat playlist.m3u8 | streamlink hlsvariant://file://- best
```

Not the most attractive URIs but it works.

zacklb:
This looks great. I can test it on Windows in a few hours. I'll find a way to mock an environment to test this new functionality.

[...]

PR #304, opened 2016-12-15
Support for local files

beardypig:
As requested in #253 and chrippa/livestreamer#1514. 

Support for local files via the `file://` protocol.

`hls`, `hlsvariant` and `httpstream` are supported. Files can be loaded from relative or absolute paths, or even read from `stdin`.

eg.
```sh
streamlink hlsvariant://file://./playlist.m3u8 best
cat playlist.m3u8 | streamlink hlsvariant://file://- best
# httpstream works, but is mostly pointless :)
streamlink httpstream://file://./hadoken.mp4
cat hadoken.mp4 | streamlink httpstream://file://- best
```

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -37,6 +37,18 @@ This command will tell Streamlink to attempt to extract streams from the URL
 specified, and if it's successful, print out a list of available streams to choose
 from.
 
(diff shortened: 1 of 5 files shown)


gravyboat:

PR #305, opened 2016-12-16
Turkish Streams IV

beardypig:
Support for these sites:

- powerapp.com.tr
- tv360.com.tr
- kanal7.com


gravyboat:
Thanks @beardypig!

PR #311, opened 2016-12-17
Fix Vaughn Live plugin

msikma:
This does two things:

* Updates the `PLAYER_VERSION`—the old one does not work anymore.
* Adds the ability to use URLs with `/embed/video` in them, which come from embedded players.

This seems to work for me.


beardypig:
Looks good, but I think the current release needs to be sorted before we merge anything else.

gravyboat:
@beardypig Agreed.

gravyboat:
Merging now that the `0.2.0` issues are resolved. Thanks @msikma!

PR #329, opened 2016-12-27
Added support for bongacams.com streams

xkbd:
This adds support for desktop (rtmp) streams on bongacams.com

### DIFF
--- a/src/streamlink/plugins/bongacams.py
+++ b/src/streamlink/plugins/bongacams.py
@@ -1,6 +1,3 @@
-"""Streamlink plugin for bongacams.com"""
-
-import json
 import re
 import time
 
@@ -14,7 +11,6 @@ from streamlink.packages.flashmedia import AMFPacket, AMFMessage
 from streamlink.plugin.api import validate, http
 from streamlink.plugin import Plugin
 from streamlink.stream import RTMPStream
-from streamlink.utils import parse_json
 
 
 CONST_FLASH_VER = "WIN 24,0,0,186"
--- a/src/streamlink/plugins/bongacams.py
+++ b/src/streamlink/plugins/bongacams.py
@@ -114,7 +114,7 @@ class bongacams(Plugin):
         if len(amf_response.messages) != 1 or amf_response.messages[0].target_uri != "/1/onResult":
             raise PluginError("unexpected response from amf gate")
 
-        stream_source_info = parse_json(json.dumps(amf_response.messages[0].value), schema=amf_msg_schema)
+        stream_source_info = amf_msg_schema.validate(amf_response.messages[0].value)
         self.logger.debug("source stream info:\n{}", stream_source_info)
 
         stream_params = {
--- a/src/streamlink/plugins/bongacams.py
+++ b/src/streamlink/plugins/bongacams.py
@@ -4,7 +4,7 @@ import time
 from io import BytesIO
 from hashlib import md5
 
-from streamlink.compat import urljoin, urlparse, urlunparse
+from streamlink.compat import bytes, urljoin, urlparse, urlunparse
 from streamlink.exceptions import PluginError, NoStreamsError
 from streamlink.packages.flashmedia.types import AMF0String, AMF0Value, U32BE
 from streamlink.packages.flashmedia import AMFPacket, AMFMessage
@@ -66,6 +66,7 @@ class bongacams(Plugin):
         stream_page_domain = match.group(4)
         stream_page_path = match.group(5)
         country_code = CONST_DEFAULT_COUNTRY_CODE
+        is_paid_show = False
 
         # create http session and set headers
         http_session = http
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/bongacams.py
```diff
@@ -0,0 +1,145 @@
...
+        match = url_re.match(self.url)
+
+        baseurl = match.group(1)
+        chathost = match.group(2)
+        chathost_url = match.group()
+        amf_gateway_url = urljoin(baseurl, CONST_AMF_GATEWAY_LOCATION)
+        country = urlsplit(baseurl).netloc[:2].lower()
+
+        # create http session and set headers
+        http_session = requests.Session()
```
beardypig: You should use `from streamlink.plugin.api import http` instead of creating a Session as it uses global options.

src/streamlink/plugins/bongacams.py
```diff
@@ -0,0 +1,145 @@
...
+    "performerData" : {
+        "username" : validate.text,
+    }
+})
+
+
+class AMFMessage2(AMFMessage):
+    @property
+    def size(self):
+        size = AMF0Value.size(self.value)
```
beardypig: `size` is the total size of the packet, you should override `_serialise` instead.

src/streamlink/plugins/bongacams.py
```diff
@@ -0,0 +1,145 @@
...
+                username=stream_source_info['userData']['username'],
+                access_key=stream_source_info['localData']['NC_AccessKey'],
+                data_key=stream_source_info['localData']['dataKey']
+            )
+        }
+
+        self.logger.debug("Stream params:\n{}", stream_params)
+        stream = RTMPStream(self.session, stream_params)
+
+        return {'best': stream, 'high': stream}
```
beardypig: `best` is an implicit quality you only need to return `high`

src/streamlink/plugins/bongacams.py
```diff
@@ -0,0 +1,145 @@
...
+
+        if r.status_code != 200:
+            raise PluginError("unexpected status code for {}: {}", r.url, r.status_code)
+
+        amf_response = AMFPacket.deserialize(BytesIO(r.content))
+
+        if len(amf_response.messages) != 1 or amf_response.messages[0].target_uri != "/1/onResult":
+            raise PluginError("unexpected response from amf gate")
+
+        stream_source_info = parse_json(json.dumps(amf_response.messages[0].value), schema=amf_msg_schema)
```
beardypig: FYI, I think you could do `amf_msg_schema.validate(amf_msg_schema)` instead of dumping and parsing again :-)

beardypig:
Ah, one thing, can you add it to the list of supported sites in the docs? https://github.com/streamlink/streamlink/blob/master/docs/plugin_matrix.rst. 

Cheers

src/streamlink/plugins/bongacams.py
xkbd: Thaks, fixed.

xkbd:
Ahaha :-D I tested it.
All comments fixed.

beardypig:
Cool :) 

Doesn't seem to work with https://bongacams.com/test, ie. With no subdomain.

xkbd:
Oops.
Apparently, subdomains are not used for IP-addresses from the English-speaking countries.
For all other request will be redirect to COUNTRY_CODE.baseurl.
Didn't know it.

I can check redirect to subdomain in first request and if is not just will add subdomain `en`. 
Then it should work.

xkbd:
Fixed.

xkbd:
Sorry for last commit, can you revert/discard/delete it?

bastimeyer:
> Sorry for last commit, can you revert/discard/delete it?

You can always [rebase](https://git-scm.com/docs/git-rebase) and force push the branch of your own forked repo. This automatically updates the PR. You should also combine multiple logical commits and only push the ones that are necessary.

xkbd:
Thank you. I understand. Sorry again.

gravyboat:
@bastimeyer @xkbd We can always squash and merge at the end as well so don't worry about it too much!

src/streamlink/plugins/bongacams.py
```diff
@@ -0,0 +1,153 @@
...
+                username=stream_source_info['userData']['username'],
+                access_key=stream_source_info['localData']['NC_AccessKey'],
+                data_key=stream_source_info['localData']['dataKey']
+            )
+        }
+
+        self.logger.debug("Stream params:\n{}", stream_params)
+        stream = RTMPStream(self.session, stream_params)
+
+        return {'high': stream}
```
beardypig: The convention for a single live stream is to call the stream "live" :)

xkbd:
Ok, fixed.

beardypig:
Thanks for the contribution @xkbd :-)

Issue #331, opened 2016-12-27
Bug with --can-handle-url


RosadinTV:
For example if we use:
Streamlink.exe --can-handle-url "http://ThisLinkMagic.hahaha/TEST" 
The program tries to HEAD the url, i think with --can-handle-url the only thing the program must do is check if the url matches with an associated plugin regex, with current method a long latency can be produced by waiting for HEAD task on a non-streamlink compatible url, what is really painfull in external excripting.
I mean for example i have a program that check if a URL is compatible with Streamlink, if not the link is played directly with a player, if HEAD for non-compatible urls is removed, the waiting will be really lower.
Edit: HEAD can be usefull in redirect cases, like this: http://cor.to/1RVk (that redirects to vaughnlive)
But i think this a really less common used way to enter urls, also an alternative parameter like --can-handle-url-nohead can be really cool if preserve of head function is wanted.

RosadinTV:
I fixed this, soon i'll post a PR

gravyboat:
Do you want to get this into the main repo? Right now it's just in the portable repo.

PR #333, opened 2016-12-28
--can-handle-url-no-redirect parameter added

RosadinTV:
Do the same as --can-handle-url but without HTTP operations involved as issued in https://github.com/streamlink/streamlink/issues/331

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -213,6 +213,13 @@ general.add_argument(
     """
 )
 general.add_argument(
+    "--can-handle-url-nohead",
+    metavar="URL",
+    help="""
+    Same as --can-handle-url but without HTTP operations involved.
+    """
+)
+general.add_argument(
     "--config",
     action="append",
     metavar="FILENAME",
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -213,10 +213,10 @@ general.add_argument(
     """
 )
 general.add_argument(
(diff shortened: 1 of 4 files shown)


RosadinTV:
@beardypig What do you think about this PR?
(I mean you are the PR master here haha)

beardypig:
Looks good. I'd think hard about the name, once it's there we're stuck with it :-) Maybe `--can-handle-url-no-redirect` describes it better? @gravyboat, wanna weigh in here?

src/streamlink_cli/argparser.py
```diff
@@ -213,6 +213,13 @@ def keyvalue(value):
     """
 )
 general.add_argument(
+    "--can-handle-url-nohead",
+    metavar="URL",
+    help="""
+    Same as --can-handle-url but without HTTP operations involved.
```
beardypig: Maybe `Same as --can-handle-url but without following redirects when looking up the URL` is more clear about the indented function?

beardypig:
If you feel up to the challenge you can try and add a test for the `resolve_url_nohead` method :-) Should be quite straightforward to add it around here [test_session.py#L31](https://github.com/streamlink/streamlink/blob/master/tests/test_session.py#L3). 

Feliz 2017.

RosadinTV:
@beardypig Thanks for your review, i will apply your suggestions.
Happy 2017 too, best wishes :clinking_glasses:

beardypig:
Looks good to me.

beardypig:
Thanks @RosadinTV

PR #340, opened 2016-12-30
Don't reject future pycryptodome versions

coypoop:
I'm not sure about the exact details of the effect, but I think specifying a single version might create an issue in the future, if one wants to update cryptodome and streamlink hasn't been updated yet.

I'm fairly confident there isn't a problem with future cryptodome versions yet, as 3.4.3 is the latest! :-)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ else:
     deps.append("requests>=1.0,!=2.12.0,!=2.12.1,<3.0")
 
 # this version of pycryptodome is known to work and has a Windows wheel for py2.7, py3.3-3.5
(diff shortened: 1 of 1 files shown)


beardypig:
Sounds fine to me, but I would limit it to less than 4.0 ie. `>=3.4.3,<4.0.0` in case there is a breaking change release (assuming they use sensible version numbers).

beardypig:
Sorry @coypoop, I included this in some other packaging changes and didn't credit you.

coypoop:
NP! thanks for committing

PR #349, opened 2017-01-03
plugins.trtspor: added support for trtspor.com

beardypig:
Adds support for http://www.trtspor.com.

Supports TV and radio streams, the TV streams are geo-locked to Turkey, so tricky to test without being in Turkey or without a Turkish proxy. 

Example URLs:
- http://www.trtspor.com/canli-yayin-izle/trtspor-159/ (geo-locked to Turkey)
- http://www.trtspor.com/canli-yayin-izle/trtspor-160/


gravyboat:

Issue #20, opened 2016-09-20
Move Twitch plugin to use Twitch API v3


thinkpad4:
Or merge javiercantero's v3-api branch https://github.com/javiercantero/livestreamer/commits/v3-api

javiercantero:
Note that some of the API used by livestreamer to get the stream is not documented and apparently not public, so it can change without prior warning, be deprecated or not comply with the rest of the specifications. For example, the API call efbc85ea613fd0a200ca65a6f6320be0c775b771/src/livestreamer/plugins/twitch.py#L186 which doesn't work with the kraken API format (api.twitch.tv/kraken/). So I'm a little afraid of changing anything that it's not documented even when I've checked that the change works right now.

thinkpad4:
Jariercantero, well I'm using your version of twitch.py and it is working perfect, I would chance to say it seems to load the stream faster then before

cdrage:
Any reason to use the undocumented API v3 *yet*? Hold off until it's 100% documented and available.

javiercantero:
@cdrage The API v3 is totally documented and working (see [here](https://github.com/justintv/Twitch-API)). What is not documented (with any API version) or **supported** is to [download Twitch content outside their player](https://www.reddit.com/r/Twitch/comments/2p0zka/tardsplaya_stopped_working_for_everyone/cmsjqgj)). livestreamer and other clients work because they use the same calls that their client uses to get the stream, but they could modify it at will.

bastimeyer:
> undocumented API v3

You're mixing things up here a bit... Both public and private Twitch APIs are being used.

There are two supported versions of the public `/kraken` namespace right now, namely v2 and v3. When no API version is specified in the request, the latest one is always assumed. Setting an explicit version header will ensure that nothing will break once a newer version gets released. See https://github.com/justintv/Twitch-API

The private namespace that's being used is `/api` and also `tmi.twitch.tv` for checking for hosted channels. See https://github.com/streamlink/streamlink/blob/ef3476952fa9e6fa66f50e0fa515cabe070f7360/src/streamlink/plugins/twitch.py#L193-L218

I think there are a couple of API calls that could be changed so that the public namespace is used instead of the private one. I might be wrong, though.

javiercantero:
> I think there are a couple of API calls that could be changed so that the public namespace is used instead of the private one. I might be wrong, though.

No, you are not: `/api/viewer/*` also works as `/kraken/viewer/*` (but it's not part of the public API)

[...]

PR #350, opened 2017-01-03
Upgrade Twitch plugin to use API v3

javiercantero:
Current Twitch API documentation[1] recommends for every API request to specify the API version by setting the Accept HTTP header to the appropriate value. This patch:

* sets the value to v3 (older versions of the API are going to be removed).
* uses the base URL`https://api.twitch.tv/kraken` to access (public) API resources

  [1]: https://github.com/justintv/Twitch-API

See #20 for details

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -198,8 +198,11 @@ class TwitchAPI(object):
         else:
             url = "https://{0}.twitch.tv{1}".format(self.subdomain, path)
 
+        headers = { 'Accept': 'application/vnd.twitchtv.v3+json',
+                    'Client-ID': TWITCH_CLIENT_ID }
+
         # The certificate used by Twitch cannot be verified on some OpenSSL versions.
-        res = http.get(url, params=params, verify=False, headers={'Client-ID': TWITCH_CLIENT_ID})
+        res = http.get(url, params=params, verify=False, headers=headers)
 
         if format == "json":
             return http.json(res, schema=schema)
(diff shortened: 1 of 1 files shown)


beardypig:
Cool. Do you the methods you removed no longer work? We should try not to change APIs, we aren't using those methods internally but it's possible (unlikely) that someone external is using the `TwitchAPI` class.

javiercantero:
@beardypig The calls to `/api/channels/<channel>/subscription` and `/api/channels/<channel>/viewer` aren't part of the public API[1], so not a good idea to try to provide them to third party apps in the first place.

  [1]: https://github.com/justintv/Twitch-API/blob/master/v3_resources/channels.md

beardypig:
That's fair enough, I just don't like removing methods. The channel_info ones is legitimate though. 
I'd be tempted to stick a [warning](https://docs.python.org/2/library/warnings.html) in those undocumented methods incase someone is using it. They'll get a warning instead of a `AttributeError` :) 

Maybe like
```python
def channel_subscription(self, channel, **params):
    warning.warn("channel_subscription is an unsupported API call")
    return self.call("/api/channels/{0}/subscription".format(channel), **params)
```

If it doesn't work in the future, too bad - we can change it to return `None`.

javiercantero:
@beardypig Ok, I have removed that commit.

gravyboat:
@beardypig Any other concerns on this one?

beardypig:
Looks good to me.

gravyboat:
Thanks @javiercantero!

Issue #327, opened 2016-12-26
Feature Request. Ability to specify subprocess errorlog path


xkbd:
Is it possible?

beardypig:
Which subprocess? The player or stream subprocesses? I don't think either is possible at the moment, but it is possible to add an option.

xkbd:
Stream subprocess. Option `--errorlog`.
I'd like to set the path but this parameter is a boolean.

beardypig:
It is possible to add an option like `--errorlog-path` or similar. Would probably be worth while adding a path option for `--verbose-player`.

xkbd:
By the way, it would be great to store logs in one place.

when i use `--errorlog` dest path for log file: %tmp%
In %appdata% i see `Streamlink.launch.pyw.log`
Also at `%appdata%` exists direcory `streamlink`
Why not store all the logs by default in the %appdata%\streamlink\logs?

Vangelis66:
@xkbd wrote: 

> Why not store all the logs by default in 
> the %appdata%\streamlink\logs?

This is a sensible request, I am "+1"ing it for the installed versions 
of streamlink; however, I am mostly running portable installations 
myself, so I wouldn't want additional leftovers on the host machine; 
some prevision should be taken for the portable flavours (currently 
two exist, one from @beardypig and another from @RosadinTV), 
so that the logs folder is created inside the portable installation directory...

PR #351, opened 2017-01-03
Specify the subprocess errorlog path

beardypig:
As requested in #327 :)

Adds the option `--subprocess-errorlog-path` (`--errorlog-path`) to specify the path of the subprocess error log.

eg.
```
streamlink --errorlog-path rtmpdump.log itv.com/hub/itv best 
```

### DIFF
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -58,7 +58,8 @@ class Streamlink(object):
             "stream-segment-threads": 1,
             "stream-segment-timeout": 10.0,
             "stream-timeout": 60.0,
-            "subprocess-errorlog": False
+            "subprocess-errorlog": False,
+            "subprocess-errorlog-path": None
         })
         self.plugins = {}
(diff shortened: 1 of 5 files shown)


gravyboat:
Good stuff, any idea why we keep seeing this check against codecov/patch? I'm pretty sure that wasn't happening before and it always fails.

beardypig:
Nope, I think it's only started recently... Maybe there is a setting in codecov.io for patch coverage diffs... It's kind of a pain :-)

beardypig:
I think they might have changed a default on codecov.io, I think we need to create their yml file (apparently we don't have one) to tweak the settings a bit: https://codecov.io/gh/streamlink/streamlink/settings/yaml

beardypig:
@gravyboat how does this look to merge?

gravyboat:
Looks good to me, I'll merge it now. Thanks!

Issue #288, opened 2016-12-12
Picarto 'live' flash support (sometimes) broken


mjevans:
During the failed case, the value of the match group is:

"REDACTED",1,0, 0, 1, "public", 1

the value of channel_server_res.text is:

edge4-us-east.picarto.tv

Unfortunately I can't collect any more data at this time.

streamlink -l debug https://picarto.tv/REDACTED best
[cli][info] Found matching plugin picarto for URL https://picarto.tv/REDACTED
[cli][info] Available streams: live (worst, best)
[cli][info] Opening stream: live (rtmp)
[cli][debug] Pre-buffering 8192 bytes
[cli][error] No data returned from stream
Fatal Python error: could not acquire lock for <_io.BufferedReader name=7> at interpreter shutdown, possibly due to daemon threads

Thread 0x00007f0bc1002700 (most recent call first):
  File "/usr/lib/python3.5/site-packages/streamlink/stream/wrappers.py", line 85 in stop
  File "/usr/lib/python3.5/site-packages/streamlink/stream/wrappers.py", line 77 in run
  File "/usr/lib/python3.5/threading.py", line 914 in _bootstrap_inner
  File "/usr/lib/python3.5/threading.py", line 882 in _bootstrap

Current thread 0x00007f0bca470400 (most recent call first):
  File "/usr/lib/python3.5/site-packages/streamlink/stream/wrappers.py", line 85 in stop
  File "/usr/lib/python3.5/site-packages/streamlink/stream/wrappers.py", line 105 in close
  File "/usr/lib/python3.5/site-packages/streamlink/stream/streamprocess.py", line 27 in close
Aborted (core dumped)

---

However some other random adult/clean streams I tried do work...

[cli][info] Found matching plugin picarto for URL https://picarto.tv/B
[cli][info] Available streams: 1295k (worst, best)
[cli][info] Opening stream: 1295k (hls)
[stream.hls][debug] Reloading playlist
[stream.hls][debug] Adding segment 624 to queue
[stream.hls][debug] Adding segment 625 to queue
[stream.hls][debug] Adding segment 626 to queue
[cli][debug] Pre-buffering 8192 bytes
[cli][info] Starting player: mpv --cache 256
[cli][debug] Writing stream to output
[stream.hls][debug] Download of segment 624 complete
[stream.hls][debug] Download of segment 625 complete
[stream.hls][debug] Download of segment 626 complete
[stream.hls][debug] Reloading playlist
[...]

PR #353, opened 2017-01-04
plugins.picarto: Remove the unreliable rtmp stream

beardypig:
This PR fixes the issue reported in #288.


gravyboat:
Good stuff! Thanks @beardypig.

PR #355, opened 2017-01-05
No longer include copies of the which/get_terminal_size backports

beardypig:
It is unnecessary to include `backports.shutil_which` and `backports.shutil_get_terminal_size` in the source tree when they are easily included as dependencies. I suspect it was originally done to ease packaging for Windows when streamlink had a Python 2 Windows installer. I also updated the package list to be discovered using `setuptools.find_packages` which is the preferred method :)


gravyboat:
Good catch, thanks @beardypig!

Issue #332, opened 2016-12-28
'Read timed out' appears to not be handled properly as of 0.2.0


CyberBotX:
I use streamlink under FreeBSD (I've made a port for it and submitted it, but I need to update it for the latest version) to watch streams on Twitch. With 0.1.0 and earlier, whenever a read timeout seemed to occur, I'd get a message such as this:

`[stream.hls][error] Failed to open segment 4370: Unable to open URL: https://video-edge-835ef0.ord02.hls.ttvnw.net/transcode-x2-835b70/sinister1_24051668848_571899805/chunked/index-0000004371-tN3E.ts (HTTPSConnectionPool(host='video-edge-835ef0.ord02.hls.ttvnw.net', port=443): Read timed out.)`

With 0.2.0, though, I instead see an exception occurring, like so:

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/requests/packages/urllib3/response.py", line 232, in _error_catcher
    yield
  File "/usr/local/lib/python3.5/site-packages/requests/packages/urllib3/response.py", line 314, in read
    data = self._fp.read(amt)
  File "/usr/local/lib/python3.5/http/client.py", line 448, in read
    n = self.readinto(b)
  File "/usr/local/lib/python3.5/http/client.py", line 488, in readinto
    n = self.fp.readinto(b)
  File "/usr/local/lib/python3.5/socket.py", line 575, in readinto
    return self._sock.recv_into(b)
  File "/usr/local/lib/python3.5/ssl.py", line 929, in recv_into
    return self.read(nbytes, buffer)
  File "/usr/local/lib/python3.5/ssl.py", line 791, in read
    return self._sslobj.read(len, buffer)
  File "/usr/local/lib/python3.5/ssl.py", line 575, in read
    v = self._sslobj.read(len, buffer)
socket.timeout: The read operation timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/requests/models.py", line 676, in generate
    for chunk in self.raw.stream(chunk_size, decode_content=True):
  File "/usr/local/lib/python3.5/site-packages/requests/packages/urllib3/response.py", line 357, in stream
    data = self.read(amt=amt, decode_content=decode_content)
  File "/usr/local/lib/python3.5/site-packages/requests/packages/urllib3/response.py", line 324, in read
    flush_decoder = True
[...]

PR #357, opened 2017-01-05
stream.hls: retry stream segments

beardypig:
This PR fixes the issue reported in #332. I did some tests which seemed to work, unplugged the network cable and set a really low timeout, the stream recovered correctly.

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -87,29 +87,40 @@ class HLSStreamWriter(SegmentedStreamWriter):
             self.logger.error("Failed to open segment {0}: {1}", sequence.num, err)
             return self.fetch(sequence, retries - 1)
 
-    def write(self, sequence, res, chunk_size=8192):
-        if sequence.segment.key and sequence.segment.key.method != "NONE":
-            try:
-                decryptor = self.create_decryptor(sequence.segment.key,
-                                                  sequence.num)
-            except StreamError as err:
-                self.logger.error("Failed to create decryptor: {0}", err)
-                self.close()
-                return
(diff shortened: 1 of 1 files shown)


gravyboat:
Looks good to me, thanks!

RosadinTV:
Awesome work @gravyboat

RosadinTV:
@gravyboat Yes, my mistake 😛

scottbernstein:
@beardypig -- is there a corresponding CLI parameter to control the value for this setting?

beardypig:
@scottbernstein yep, there has been for a long time https://streamlink.github.io/cli.html#cmdoption-hls-segment-attempts

scottbernstein:
Thanks, @beardypig!  Yes I utilize this CLI option frequently, but it was previously ignored?  Or was it just a new implementation of the code that was ignoring it?


beardypig:
Yep, the new implementation was not properly retrying. It's also the case for hds streams, the retry option doesn't actually make any difference. I think the design is a bit poor and could do with some improvements.

scottbernstein:
Thanks for the clarification). While examining the implementation I also realized why a couple of times I have gotten stack overflow errors after extreme amounts of retries -- the recursive nature of the error handling is limited by the amount of recursion allowed in your system's version of Python, which in my Windows environment is just under 1000. Will keep my retry settings below that....

On Jan 25, 2017, at 1:56 AM, beardypig <notifications@github.com<mailto:notifications@github.com>> wrote:


Yep, the new implementation was not properly retrying. It's also the case for hds streams, the retry option doesn't actually make any difference. I think the design is a bit poor and could do with some improvements.

-

PR #359, opened 2017-01-05
packages.flashmedia: fix bug in AMFMessage

beardypig:
`flashmedia.amf.AMFMessage` sets a field incorrectly when serialising, this PR rectifies that issue.

As this package is no longer maintained by its author, and already has some `streamlink` specific patches, I will continue with small bug fixes for `flashmedia` until a suitable replacement can be found...

### DIFF
--- a/src/streamlink/packages/flashmedia/amf.py
+++ b/src/streamlink/packages/flashmedia/amf.py
@@ -55,7 +55,7 @@ class AMFMessage(Packet):
     def _serialize(self, packet):
         packet += AMF0String(self.target_uri)
         packet += AMF0String(self.response_uri)
-        packet += U32BE(self.size)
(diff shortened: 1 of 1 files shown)


gravyboat:
Thanks for this. How important is flashmedia support wise? Looking through the repo it doesn't seem like THAT many plugins/streams are using it, with the biggest being ustream.

beardypig:
It's mainly for flv stuff, but no, not many do use it. I've have written most of the parsing stuff for AMF etc. in construct already (for fun). I wanted to make a pure python rtmp client, sort of works :-) Might be able to make a flashutils package and stick it in a repo in the streamlink org. I haven't tried to port the flv concat stuff, but from what I remember it's quite straightforward :-)

PR #361, opened 2017-01-05
plugins: all regex string should be raw

beardypig:
Support for non-standard escape characters has been deprecated in Python 3.6 so all the regex string that contained regex style escape sequences (eg. `\w`) have to be raw strings.


gravyboat:
Good fix, thanks @beardypig.

PR #362, opened 2017-01-05
Boost the test coverage a bit

beardypig:
Added some tests for a few things. Including one meta test that loads each of the plugins to ensure that there are no syntax errors :)


beardypig:
Seems like the new tests found a bug in one of the plugins for Python2.6 :)

gravyboat:
Nice! Thanks @beardypig!

Issue #364, opened 2017-01-06
The Picarto arms race continues - HTML5 HLS is now a thing for (some ?) streams.


mjevans:
The Picarto arms race continues - HTML5 HLS is now a thing for (some ?) streams.

gravyboat:
Again?! This is getting ridiculous, if they want to keep us from handling their apps is it even worth doing? Jeez what a pain...

mjevans:
They've actually changed from an integer to text for the stream type.

Oddly the place stream option has a /different/ value in the popout for the page than in this snip.

Flash is now literally 'flash'.  In the popout the value returned is 'hls'.


This method needs one line changed to fix for the current interface change.

@staticmethod
    def _get_stream_arguments(page):
...
        player_id, product, offline_image, online, is_flash = \
            map(lambda a: bool(int(a)), [player_id, product, offline_image, online, True if is_flash == "flash" else False])







                <div class="flexPlayerOuter">

                        <div class="centerFixed playerBackdrop flexPlayerInnerFull">
                                <div class="posRel">
                                        <video playsinline id="playerHolder1" class="video-js vjs-default-skin vjs-big-play-centered vjs-16-9 overflowHidden"></video>
                                        
                    <div class="colBGWW pd10 roundCorners ontop maininfoFeatureMenu boxShadow zI1 text-left fontSize1 techMenu-XXX" style="display: none;">
                        <div class="cutLine0 dark pdbot5 bot5 fontSize1-2">Playback options</div>
                        <div class="pdtop5">
                            <div class="displayBlock btnBase btn-sm topbot5 techBtn tech_switch" data-channel="XXX" data-playerid="1" data-token="public" data-tech="flash">
                                FLASH
                            </div>
                                                        <div class="displayBlock btnBase btn-sm topbot5 techBtn tech_switch" data-channel="XXX" data-playerid="1" data-token="public" data-tech="hls">
                                HTML5 (HLS)
                            </div>
                        </div>
                    </div>

                                        <div class="colBGWW pd10 roundCorners ontop maininfoFeatureMenu boxShadow zI1 text-left fontSize1 stnMenu-XXX" style="display: none;">
                                                <div class="cutLine0 dark pdbot5 bot5 fontSize1-2">Player Options</div>
                                                <div class="pdtopbot5">
                                                        <div class="bot5 fontSize0-8">Please give us Feedback.</div>
                                                        <div class="bot5 displayBlock btnBase btn-sm Green settingsBtn"
[...]

PR #365, opened 2017-01-06
Picarto update to support new site layout

beardypig:
The page layout changed very slightly, they changed one of the arguments in their `placeStream` method so that it takes a name for the default streaming technology (flash or hls), where as before it was an integer. 

While the site change was small I decided to update the plugin a bit, it now has better logging messages, detects multi-channel streams, and detects non-existent stream pages.

It also fixes the issue reported in #364.


gravyboat:
Thanks!

PR #367, opened 2017-01-06
plugins.mitele: support for live streams on mitele.es

beardypig:
Most of the time the streams are geo-blocked to Spain, and sometimes (but not always) some of the streams have DRM. 

Example URLs:
- mitele.es/directo/telecinco
- mitele.es/directo/cuatro
- mitele.es/directo/boing
- mitele.es/directo/bemad


gravyboat:
Thanks @beardypig!

Issue #335, opened 2016-12-28
Help install streamlink on fedora


Koisell:
I just discover and install streamlink on my fedora 25(I used livestreamer before).
Because pip didn't directly work, I put here instruction for fedora users.

For the installation on a fedora OS you need to install the python-devel package before ;)
$ sudo dnf install python-devel
$ sudo pip install -U streamlink

I see their is a package project in WIP: #262 
If I can give my help :)

Koisell:
I just see that a guy melmorabity make a build in the fedora Copr. I uninstall streamlink with pip and re install it using:
$ sudo dnf copr enable melmorabity/streamlink
$ sudo dnf install streamlink

It works fine.
So if your using fedora  23, 24, 25 I recommand you this way.
But the way explained above should work on fedora, CentOS and REHL (replace dnf by yum).

beardypig:
Strange that `pip` is installed without `python-devel`, on debian `pip` depends on `python-dev` :-)

Koisell:
pip is installed by default on fedora ;)

melmorabity:
I submitted streamlink for inclusion in Fedora:
https://bugzilla.redhat.com/show_bug.cgi?id=1405271

gravyboat:
I'm going to close this out since it's resolved.

PR #368, opened 2017-01-06
Add a note about python-devel needing to be installed in some cases

beardypig:
As per the discussion in #335


gravyboat:
This is a good note to have, thanks.

Issue #271, opened 2016-12-08
Convert changelog to rst


beardypig:
As discussed in #238, we should convert the changelog.md file to rst so that it plays nice with the rest of the docs.

mmetak:
@beardypig I kinda already did that in that #238 PR? Or are you looking for a way to automatically generate rst changelog beside the md with the release?

beardypig:
Yeah. It needs to be automated, which requires changing the release script, which @cdrage volunteered to do ;) 

This is blocking your PR #238 :)

mmetak:
`pandoc CHANGELOG.md --from markdown --to rst -s -o docs/changelog.rst`

BTW is 0.2.0 any closer to release? ;)

beardypig:
Easier and more reliable to just generate rst, rather than converting from my :-P

I think there are a few PRs that people want to be merged before the release of 0.2.0 :)

joshka:
I noticed that the changelog contains a lot of information that for an end user to use (and that a dev can usually find more easily by hitting github). Would you maybe consider using a format somewhat like: http://keepachangelog.com/ suggests?

gravyboat:
The suggested items based on keepachangelog.com versus what we have here: https://github.com/streamlink/streamlink/releases:

* It’s made for humans, not machines, so legibility is crucial. - We already do this when we note the major features of the release so non-technical users can easily review them. 
* Easy to link to any section (hence Markdown over plain text). - We don't currently do this, and the changelog isn't usually such a long list that we currently need this
* One sub-section per version. - We fulfill this though not perfectly, each release has a different area on the page (though this could be condensed into a version specific changelog that has multiple releases).
* List releases in reverse-chronological order (newest on top). - We already do this.
* Write all dates in YYYY-MM-DD format. (Example: 2012-06-02 for June 2nd, 2012.) It’s international, sensible, and language-independent. - We do this automatically via the GitHub release details noting when the build was released, though the formatting isn't perfect.
* Explicitly mention whether the project follows Semantic Versioning. - We already follow semantic versioning, I don't really feel like this needs to be explicitly noted, those who don't follow semantic versioning should probably note this instead. ;)
* Each version should:
  * List its release date in the above format. - As noted above we kind of do this.
  * Group changes to describe their impact on the project, as follows: - Currently we simply group all of these together as we don't have enough changes to really justify categories. If we did that we would have more categories than changes in many cases.
    * Added for new features.
    * Changed for changes in existing functionality.
[...]

PR #328, opened 2016-12-27
Automated GitHub releases

bastimeyer:
See https://github.com/streamlink/streamlink/issues/261#issuecomment-268732799 and https://github.com/streamlink/streamlink/pull/310#issuecomment-267738808

This adds support for automated Github releases once a new new tag gets pushed to the repository.
The .tar.gz source archive and the PGP signature file are currently not being uploaded and need to be added in another commit...

When a new tag gets pushed, TravisCI automatically creates a new Github release (the `RELEASES_API_KEY` env var needs to be set first in the TravisCI settings, didn't want to include this in the `.travis.yml` file), uploads the Windows installer and then executes a python script that updates the release `name` and `body` based on the tag and format of the `CHANGELOG.md` file. This means that the format can't be changed, or the script needs to be altered. The script is based on [this](https://github.com/streamlink/streamlink-twitch-gui/blob/v1.0.1/build/travis/update_release.js) (which has been working great so far for me). I'm not a python dev and this was my first attempt, so things will very likely not be perfect. 😱

Tag validation can also be added in the future, so that it only creates releases when a signed tag was successfully validated.

See a successful upload on my test account/repo:
https://github.com/bastimeyer-test/my-test-repo2/compare/0.2.0...0.2.1
https://travis-ci.org/bastimeyer-test/my-test-repo2/jobs/186876690
https://github.com/bastimeyer-test/my-test-repo2/releases/tag/0.2.1

PR #369, opened 2017-01-06
docs/release: generate the changelog as rst instead of md

beardypig:
I took a stab at modifying the release script so that the change log is generated as rst as discussed in #238 and #271. 

@cdrage should take a look at this.


gravyboat:
With this PR I believe we're good to merge: https://github.com/streamlink/streamlink/pull/328, any concerns? If not I'll merge that one in, then we can get this one merged once others have reviewed it.

beardypig:
I think #328 conflicts with the release script a bit, the release script creates a github Release page as well. I think just the part where it add the built installer from travis to the github tag is sufficient for now.

beardypig:
I think this portion of the `.travis.yml` c42f96d77437a2f46e0279bae75f3571c825b06c will be enough for now, just to get the Windows installer automatically uploaded.

CHANGELOG.rst
```diff
@@ -1,4 +1,4 @@
-streamlink 0.2.0 (12-16-2016)
+streamlink 0.2.0 (2106-12-16)
```
bastimeyer: 2016

beardypig: Damn that time capsule

gravyboat:
@bastimeyer That would be great, then we can get both of these merged in and be up to date across the board.

beardypig:
@gravyboat @bastimeyer we'll have to make sure @cdrage is up-to-date (as he's the de-facto release manager) with the release process if the release page is created automatically, as the current release script makes the release page - maybe we need to update that script at the same time. We could do the Windows installer upload and the other changes separately, so we get the advantages of the automatic Windows builds but don't have to update the release script. 

@bastimeyer it would be good to update the release script to support rst :)

cdrage:
This looks good to me :) Next release I can test this out and see if it generates it correctly! But good job :+1:

gravyboat:
Closing as requested here: https://github.com/streamlink/streamlink/pull/328#issuecomment-272094312

PR #371, opened 2017-01-06
plugins.adultswim: support https urls

beardypig:
The plugin erroneously reported that it does not support https://www.adultswim.com URLs.


gravyboat:
Thanks @beardypig.

PR #379, opened 2017-01-08
Links on crunchy's rss no longer contain the show name in the url

asermax:
I don't know since when, but rss links on crunchy no longer contain the show name slug on the url, so the current regex fails to match against those links.
I didn't remove the group since some people get the link directly from the page, which still use the previous format.

### DIFF
--- a/src/streamlink/plugins/crunchyroll.py
+++ b/src/streamlink/plugins/crunchyroll.py
@@ -46,7 +46,7 @@ _url_re = re.compile(r"""
     (?:
         com|de|es|fr|co.jp
     )
-    /[^/&?]+
+    (?:/[^/&?]+)?
     /[^/&?]+-(?P<media_id>\d+)
 """, re.VERBOSE)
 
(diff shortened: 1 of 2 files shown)


beardypig:
Thanks! Looks good, do you have some example URLs so I can verify and get it merged :-)

If you feel like it you could add some tests for the URL matching for crunchyroll similar to this [test_plugin_adultswim.py](https://github.com/streamlink/streamlink/blob/master/tests/test_plugin_adultswim.py). :-)

asermax:
Added some tests, but just in case here are some examples of valid urls:

```
http://www.crunchyroll.com/march-comes-in-like-a-lion/episode-115-omnibus-special-episode-728151
http://www.crunchyroll.com/march-comes-in-like-a-lion/media-728151
http://www.crunchyroll.com/media-728151
```

Let me know if there's anything else I should do 👍

beardypig:
Awesome, looks great! Thanks @asermax :)

PR #380, opened 2017-01-08
added chocolatey to the documentation

lawli3t:
Opted for just adding a row in the Other platforms instead of putting a separate subheading under Windows binaries.


beardypig:
Perfect, thanks :)

Issue #376, opened 2017-01-07
Mytf1.fr : TF 1, TMC, NT 1, HD 1, LCI


karlo2105:
http://www.tf1.fr/direct
Channels on TF 1 site require only a free login and French proxy. They use HDS and HLS. 
LCi doesn't require login just proxy : http://www.lci.fr/direct/

But it would be of course better if you can bypass login.

beardypig:
Can take a look, do you know any good proxy for lci? I can use unlocator for tf1 but I don't know any good http proxies for France.

karlo2105:
You can try here : 
http://spys.ru/free-proxy-list/FR/
http://www.freeproxylists.net/?c=FR&pt=&pr=&a%5B%5D=0&a%5B%5D=1&a%5B%5D=2&u=0
https://www.proxynova.com/proxy-server-list/country-fr/

Vangelis66:
@karlo2105 wrote: 

> LCi doesn't require login, just proxy: 
> http://www.lci.fr/direct/

@beardypig wrote: 

> do you know any good proxy for lci?

I am not inside France, but I find accessing the live (direct) stream from Greece 
does not need a French proxy; the "HDS Link detector" Firefox addon reveals 
the following HDS manifest:

`http://lcilivhlshdslive-lh.akamaihd.net/z/lci_1@301585/manifest.f4m?hdnea=st=1483822658~exp=1483824458~acl=/*~hmac=1bd7f429b37a778f80a7654e2f2db1d7a813fe5c0d050acb5279451fbd108c82&hdcore=2.11.3&g=MHIRXZLQFQZI` 

which, of course, has a limited lifespan; it contains five quality variants, with bitrates: 
240k, 496k, 896k, 1296k & 2596k

@beardypig 
I've been meaning to ask, you, of course, are not obliged in any way to reply, 
but what is your physical location?

Regards

karlo2105:
I think through HLS, they provide 720p stream.

Vangelis66:
@karlo2105 wrote: 

> I think through HLS, they provide 720p stream

The 2596k AdobeHDS stream is **720p**, too. 
But I have stumbled upon a problem; 
AdobeHDS.php would download streams 
from the manifest fine, 
but streamlink fails to, citing: 
`error: This manifest requires the 'pvswf' parameter to verify the SWF`

beardypig:
```
[...]

PR #381, opened 2017-01-08
plugins.tf1: added plugin to support tf1.fr and lci.fr

beardypig:
As requested in #376. 

Example URLs:
- tf1.fr/ft1/direct
- tf1.fr/nt1/direct
- tf1.fr/tmc/direct
- tf1.fr/lci/direct
- lci.fr/direct


gravyboat:
Thanks!

Issue #375, opened 2017-01-07
Support to raiplay.it


karlo2105:
They have several channels mostly geoblocked to Italy some work outside too. 
http://www.raiplay.it/dirette/
I think they provide HDS and HLS, it would be nice to cover both systems.

beardypig:
Looks pretty straight forward and I have a suitable proxy, should be able to whip something up in a couple of days :-)

karlo2105:
Many thanks for your contribution. ;)

beardypig:
Closing this as #382 has be merged.

PR #382, opened 2017-01-08
plugins.raiplay: added plugin to support raiplay.it

beardypig:
As requested in #375.

Example URLs:
- raiplay.it/dirette/rai1
- raiplay.it/dirette/rai3
- raiplay.it/dirette/rainews24
- raiplay.it/dirette/raisport1


gravyboat:

Issue #378, opened 2017-01-08
vaughnlive "error: Unable to validate response text: Minimum length is 3 but value is 1"


JourneyOver:
Dunno when it happened exactly, but at some point today Vaughnlive must of updated something and broken streamlink, as any link to a stream I try always just results in it erroring out..

Example - 

``` 
streamlink http://vaughnlive.tv/chrisdave2 best
[cli][info] Found matching plugin vaughnlive for URL http://vaughnlive.tv/chrisdave2
error: Unable to validate response text: Minimum length is 3 but value is 1
```
it pretty much does this for any stream url I use from vaughnlive, even tested with the nightly build and it does the same thing.

gravyboat:
Unless this is a new change I believe it was fixed over here: https://github.com/streamlink/streamlink/pull/311, give the nightly build a shot: https://streamlink-builds.s3.amazonaws.com/nightly/windows/streamlink-latest.exe

JourneyOver:
think one of the main reasons they do it is because it circumvents their dumb ad thing that shows up on livestreams when ever you go to a livestream since as I recall livestreamer (streamlink) are the only two options that are able to show the livestreams without said ad since you can't place the ads on any media players and ad blockers don't seem to be able to block the ad without also blocking the whole video in the process. So they decide to break any functions to get around it whenever they can. :s

It's pretty much the same reason http://www.arconaitv.me/ stopped functioning in livestreamer because arconai decided he wanted to break everything so people have to view his dumb ads to watch the livestreams as well.

thinkpad4:
There has to be a way to compensate for Vaughnlive changing the player version all the time

Vangelis66:
Surely related, is @fadster actually proposing a fix?

https://github.com/chrippa/livestreamer/issues/1566

beardypig:
@Vangelis66 yep, that is the new info url :)

gravyboat:
I'm going to close this since #383 is merged. Thanks for the report @JourneyOver!

JourneyOver:
@gravyboat yep looks like everything is working now in the nightly build *. so this can now be closed for now (going to go ahead and close it since you seemed to forget about closing it) >..< and it's not a problem at all, love that you guys have decided to fork livestreamer and keep things going ^^

*until Vaughn decides to go and break things again..

beardypig:
@JourneyOver I'm working on a more robust fix, but I need to wait to see if it will actually work when the player version or info URL changes :-)

[...]

PR #383, opened 2017-01-08
plugins.vaughnlive: updated player version and info URL

beardypig:
Fixes the bug reported in #378. 

The Info URL and player version have been updated, I added some extra debug logging while updating it. 

The version and info url are hardcoded in the swf, is possible that we could decompress the swf and try to extract the version and info url.


gravyboat:
Thanks @beardypig!

PR #389, opened 2017-01-09
Fix TF1.fr plugin

### LINKED ISSUE #376: Mytf1.fr : TF 1, TMC, NT 1, HD 1, LCI

beardypig:
Seems the streams only worked some of the time with the existing plugin. I have updated the plugin to use the correct `pvswf` url and the correct stream info urls. 

I also added the `g` (`guid`) parameter to the Akamai hdcore URLs for manifests which was required for these streams. 

Fixes #376 correctly this time.

### DIFF
--- a/src/streamlink/plugins/tf1.py
+++ b/src/streamlink/plugins/tf1.py
@@ -43,8 +43,11 @@ class TF1(Plugin):
         if m:
             hls_stream_url = m.group(1)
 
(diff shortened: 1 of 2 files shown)


gravyboat:
Thanks @beardypig!

PR #390, opened 2017-01-09
plugins.tv8cat: added support for tv8.cat live stream

beardypig:
Adds support for http://tv8.cat/directe


gravyboat:
Thanks @beardypig!

PR #391, opened 2017-01-09
Add basic tests for stream.me plugin

Brainzyy:
I don't know if this is useful or if my test cases are useful, so any feedback is welcomed.


beardypig:
Any test is better than no test :D Thanks :)

PR #393, opened 2017-01-09
plugins.stream: fix a default scheme handling for urls

beardypig:
Due to a bug introduced in 707df4f27bf144f4e33dc76ffafbcf2fac2519b8 URLs with scheme (eg. no http) but with a port number would not get http:// appended by default, as urlparse parses anything before the colon as the scheme.


gravyboat:

PR #394, opened 2017-01-10
Add support for a Bulgarian sports channel

beardypig:
Adds support for sportal.bg as requested by @roshavagarga.

Pretty much just supports one URL:
- sportal.bg/sportal_live_tv.php?str=15


gravyboat:
Thanks @beardypig.

PR #395, opened 2017-01-10
plugins.bnt: fix key request user agent

beardypig:
When loading the HLS streams some User Agents were rejected, specifically Firefox. Updating the User Agent string to Chrome resolves the issue.


gravyboat:
That hot test coverage improvement! Thanks @beardypig.

PR #396, opened 2017-01-10
Update to support new SSH101.com site layout

beardypig:
Dunno if there are many streams left on this site, but I have updated the plugin to support any that are left.


gravyboat:
Awesome, thanks @beardypig!

Issue #401, opened 2017-01-11
Windows installer VersionInfo


xkbd:
Hi. Can someone add to makeinstaller.sh installer version information? (windows shows this information on the Version tab of the file properties)
Example: http://nsis.sourceforge.net/Examples/VersionInfo.nsi

beardypig:
Good idea :)

xkbd:
Thanks.
```
>sigcheck -nobanner streamlink-latest.exe

        Verified:       Unsigned
        Link date:      20:01 19.02.2012
        Publisher:      n/a
        Company:        Streamlink
        Description:    Streamlink Installer
        Product:        Streamlink
        Prod version:   n/a
        File version:   0.2.0+af386a7
        MachineType:    32-bit
```

PR #402, opened 2017-01-11
windows-installer: add the version info to the installer file

beardypig:
As suggested in #401 :)


gravyboat:
Nice, thanks @beardypig.

PR #404, opened 2017-01-12
include CHANGELOG.rst instead of .md in the egg

beardypig:
Two minutes after merging #328 I notice this.. D'oh


gravyboat:
Glad you caught it!

PR #405, opened 2017-01-12
Include alt streams when parsing HLS playlists

### LINKED ISSUE #370: Adult Swim doesn't play any streams.
https://www.adultswim.com/videos/streams
https://www.adultswim.com/videos/streams/toonami
https://www.adultswim.com/videos/streams/williams-stream

beardypig:
This fixes #370 but also fixes a wider bug in HLS playlist parsing. 

When parsing an HLS playlist only one stream was output for each quality, if multiple streams existed at the same quality the ones appearing later in the playlist would be dropped. This patch fixes this behaviour and will output "alt" streams. eg. `360p_alt, 720p_alt, 360p (worst), 720p (best)` which allows auto-switching if the primary stream fails (as is in the case with the Toonami stream). 💥

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -320,8 +320,17 @@ class HLSStream(HTTPStream):
             stream_name = (names.get(name_key) or names.get("name") or
                            names.get("pixels") or names.get("bitrate"))
 
-            if not stream_name or stream_name in streams:
+            if not stream_name:
                 continue
+            if stream_name in streams:  # rename duplicate streams
+                stream_name = "{0}_alt".format(stream_name)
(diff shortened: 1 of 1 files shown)


scottbernstein:
@beardypig -- why limit the alts to 2?  Why not offer all "alts" that are available?

beardypig:
To match the logic from the plugins `streams` API. 
If you have more than 2 alts in a playlist something might be screwy, but we could update the other logic to not limit it. I don't know the original thinking that led to it being limited to 2, perhaps there was an issue with 100s of alts in some stream.

gravyboat:
Thanks @beardypig!

PR #406, opened 2017-01-12
Adultswin VOD support

beardypig:
This PR adds support for VOD on adultswim.com as mentioned in #370. 

Some of the content is encrypted (adultswim.com/videos/specials/the-adult-swim-golf-classic-extended) and thus not viewable, and content some is not encrypted and works fine. 

There was a bug in the `HDSStream` class that selected the wrong segment for the last fragment in the segment, it should actually be the first fragment in the next segment ...

### DIFF
--- a/src/streamlink/plugins/adultswim.py
+++ b/src/streamlink/plugins/adultswim.py
@@ -4,8 +4,10 @@ from pprint import pprint
 
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import http, validate
+from streamlink.stream import HDSStream
 from streamlink.stream import HLSStream
-from streamlink.utils import parse_json
+from streamlink.stream import HTTPStream
+from streamlink.utils import parse_json, parse_xml
 
 
 class AdultSwim(Plugin):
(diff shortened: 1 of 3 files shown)


gravyboat:
Should a check be added to this so when it detects the stream is encrypted it outputs an error saying so (and thus we cannot stream it)? I have a feeling we'll see issues saying people can't watch that specific stream.

beardypig:
It already output an error when trying to play an encrypted stream that it can't decrypt, but the wording of the error  could probably be improved :-)

If you try this one: adultswim.com/videos/specials/the-adult-swim-golf-classic-extended it gives the error.

gravyboat:
@beardypig Yeah I saw that it uses the 'no plugin can handle this url' error and that's actually why I commented about it in the first place, it's not really clear that the issue is the site is encrypted, it just makes it seem like the plugin doesn't support it, which while true isn't the main reason.

beardypig:
The plugin should support that URL. You should get an output similar to this:
```sh
(streamlink-python3.6) ↪ streamlink "adultswim.com/videos/specials/the-adult-swim-golf-classic-extended" best
[cli][info] Found matching plugin adultswim for URL adultswim.com/videos/specials/the-adult-swim-golf-classic-extended
[cli][info] Available streams: 720p_alt, 234p (worst), 664k, 270p, 360p, 432p, 1296k, 540p, 720p, 2796k, 1080p, 3596k, 4596k (best)
[cli][info] Opening stream: 4596k (hds)
[stream.hds][error] Error reading fragment 1-1: Tag has filter flag set, probably encrypted
[cli][info] Starting player: /usr/local/bin/VLC
[stream.hds][error] Error reading fragment 1-2: Tag has filter flag set, probably encrypted
[stream.hds][error] Error reading fragment 1-3: Tag has filter flag set, probably encrypted
[stream.hds][error] Error reading fragment 1-4: Tag has filter flag set, probably encrypted
[stream.hds][error] Error reading fragment 1-5: Tag has filter flag set, probably encrypted
[stream.hds][error] Error reading fragment 1-6: Tag has filter flag set, probably encrypted
[stream.hds][error] Error reading fragment 2-7: Tag has filter flag set, probably encrypted
[stream.hds][error] Error reading fragment 2-8: Tag has filter flag set, probably encrypted
^CInterrupted! Closing currently open stream...
```

gravyboat:
@beardypig I ran it with my current version of the plugin like an idiot, I'm fine with this, sorry about that, busy with work.

beardypig:
No problem :) I think will update the wording of the message so say that it's probably encrypted and therefore not supported.

gravyboat:
Sounds good, thanks @beardypig.

stepshal:
@beardypig it's PR is for specials support, and not for actual VODs like this one: 
http://www.adultswim.com/videos/streams/toonami-pre-flight/friday-january-6th-2017

stepshal:
@beardypig @gravyboat but ecnrypted is only streams with bitrate like `4596k`, `1080p` is not encrypted.

beardypig:
@stepshal you're right about some of the VOD URLs, I'll fix those now. 
Other VODs are supported, like episode that don't require a cable subscription eg. adultswim.com/videos/aqua-teen-hunger-force/vampirus/

PR #407, opened 2017-01-12
Use AVconv if available and FFmpeg is not available

beardypig:
There was a bug that prevented `avconv` from being used if `ffmpeg` was not available. I have fixed the bug and `avconv` can be used instead of `ffmpeg` (for example on Debian).


gravyboat:
This is nice to have. Thanks @beardypig!

PR #411, opened 2017-01-12
Move streamlink_cli.utils.named_pipe in to streamlink.utils

beardypig:
Refactor of `streamlink_cli.utils.named_pipe` so that the `streamlink` module can remain independent from the `streamlink_cli` module, ie. `streamlink` has no dependency on `streamlink_cli`, only the other way around.

The `streamlink_cli.utils` API is not affected by this change.


gravyboat:
Nice! This is a really good enhancement that sets us up for future dev work, thanks @beardypig!

PR #417, opened 2017-01-13
stream.hds: omit HDS streams that are protected by DRM

beardypig:
As discussed in https://github.com/streamlink/streamlink/pull/406#issuecomment-272246813; I have found a better way to deal with the encrypted HDS streams, by simply omitting them when parsing the manifest file.

If the `drmAdditionalHeader` element has some content then it's protected by DRM, so that stream is skipped.


stepshal:
@beardypig so no need in pycryptodome then?

gravyboat:
Thanks @beardypig!

PR #427, opened 2017-01-15
streamlink_cli.main: close stream_fd on exit

### LINKED ISSUE #337: python interpreter crash
Environment: 
windows 8.1 x64
streamlink: latest nightly build for windows

- First, i get a message "rtmpdump.exe has stopped working"
Like this:
error img

- Then "Python has stopped working"

streamlink output:
```
[cli][info] Opening stream: high (rtmp)
[cli][error] Failed to read data from stream: Read timeout
Fatal Python error: could not acquire lock for <_io.BufferedReader name=5> at in
terpreter shutdown, possibly due to daemon threads

Thread 0x000014bc (most recent call first):
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\stream\wrappers.py", line 67 in run
  File "threading.py", line 914 in _bootstrap_inner
  File "threading.py", line 882 in _bootstrap

Current thread 0x000012f8 (most recent call first):
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\stream\wrappers.py", line 85 in stop
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\stream\wrappers.py", line 105 in close
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\stream\streamprocess.py", line 27 in close

```

I also have windows 7 x32 and error sometimes reproduced on win 8 only.

xkbd:
One more attempt to resolve #337

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -935,7 +935,7 @@ def main():
         finally:
             if stream_fd:
                 try:
-                    console.msg("Closing currently open stream...")
+                    console.logger.info("Closing currently open stream...")
                     stream_fd.close()
                 except KeyboardInterrupt:
                     sys.exit()
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -928,13 +928,14 @@ def main():
             setup_plugin_options()
             handle_url()
         except KeyboardInterrupt:
+            # Close output
+            if output:
+                output.close()
             console.msg("Interrupted! Exiting...")
         finally:
             if hasattr(stream_fd, "close"):
                 console.msg("Closing currently open stream...")
                 stream_fd.close()
-            if hasattr(output, "close"):
-                output.close()
     elif args.twitch_oauth_authenticate:
         authenticate_twitch_oauth()
     elif args.help:
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -933,9 +933,12 @@ def main():
                 output.close()
             console.msg("Interrupted! Exiting...")
         finally:
-            if hasattr(stream_fd, "close"):
-                console.msg("Closing currently open stream...")
-                stream_fd.close()
+            if stream_fd:
+                try:
+                    console.msg("Closing currently open stream...")
+                    stream_fd.close()
+                except KeyboardInterrupt:
+                    sys.exit()
     elif args.twitch_oauth_authenticate:
         authenticate_twitch_oauth()
     elif args.help:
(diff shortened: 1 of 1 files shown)


xkbd:
I think is better to use a console.logger.info() instead console.msg(). Or nothing.

src/streamlink_cli/main.py
```diff
@@ -931,16 +931,11 @@ def main():
...
-                console.msg("Interrupted! Closing currently open stream...")
-                try:
-                    stream_fd.close()
-                except KeyboardInterrupt:
-                    sys.exit()
-            else:
-                console.msg("Interrupted! Exiting...")
+            console.msg("Interrupted! Exiting...")
+        finally:
+            if hasattr(stream_fd, "close"):
```
beardypig: You should check that `stream_fd is not None` too.

xkbd: If stream_fd is None hasattr returns False and close() will not be called. Is not it?

beardypig: You're quite right :-)

src/streamlink_cli/main.py
```diff
@@ -931,16 +931,11 @@ def main():
...
+                console.msg("Closing currently open stream...")
+                stream_fd.close()
```
beardypig: Originally there was a `try..catch` here to catch `KeyboardException`, this should go back in. If `close` takes a little time and user hits ctrl-c they'll see the `KeyboardException` rather than just exiting.

xkbd: As a result, I just need to add finally block:

```python
finally:
    if stream_fd:
        stream_fd.close()

```
Right?

beardypig: You'll need 
```py
try:
    stream_fd.close()
except KeyboardInterrupt:
    sys.exit()
```

Inside the finally block to replace the existing `stream_fd.close()`. It is assumed if the user is pressing ctrl-c they want the process to end, there may be other exceptions caused by that, but that is acceptable...

xkbd:
I would like to remove `console.msg("Closing currently open stream...")`
It does not look good for me:
```
[cli][info] Opening stream: live (rtmp)
[cli][info] Starting player: not-exists-file
error: Failed to start player: not-exists-file ([WinError 2] Не удается найти указанный файл)
Closing currently open stream...
```

```
[cli][info] Player closed
[cli][info] Stream ended
Closing currently open stream...
```

beardypig:
Yeah, make it an info message like you said before :-)

beardypig:
Good work @xkbd :) Looks good to merge.

gravyboat:
Thanks @xkbd for the PR and @beardypig for reviewing.

PR #428, opened 2017-01-15
Fix swf_url in livestream.com plugin

sqrt2:
(At least some) livestream.com streams currently won't play because streamlink attempts to open a URL that starts with "http:////" (four slashes). This PR fixes this issue.


beardypig:
Looks good to me, thanks :-)

PR #429, opened 2017-01-15
plugins.bongacams: a few small changes

xkbd:
* class AMFMessage2 removed because AMFMessage fixed
* rtmpdump option 'verbose' removed
* added basic can_handle_url() tests


src/streamlink/plugins/bongacams.py
```diff
@@ -138,13 +129,18 @@ def _get_streams(self):
...
             # Details: https://github.com/streamlink/streamlink/issues/321
-            "conn" : "S:{username} --conn=S:{access_key} --conn=B:0 --conn=S:{data_key}".format(
+            "conn": "S:{username} --conn=S:{access_key} --conn=B:0 --conn=S:{data_key}".format(
                 username=stream_source_info['userData']['username'],
                 access_key=stream_source_info['localData']['NC_AccessKey'],
                 data_key=stream_source_info['localData']['dataKey']
             )
         }
 
+        subprocess_errorlog_path = self.session.options.options.get('subprocess-errorlog-path')
```
beardypig: Might be worth putting this in to the RTMPStrram class as it seems generally useful :-)

beardypig:
Looks good. The tests are a bit complex, better to keep tests simple as it's easier to verify that the test cases are as you intended. Otherwise you need to write tests for your tests ;-)

src/streamlink/plugins/bongacams.py
xkbd: it sounds reasonable

xkbd: I think better to add a separate option. `--subprocess-errorlog-debug` or something like that.
Because debug output very BIG: "Extremely verbose, including hex dumps of all packet data."

And in the case with `subprocess-errorlog-path`/`subprocess-errorlog` would be great to set `self.params["verbose"] = True`

beardypig: This option is specific to rtmpdump, so would affect RTMPStrram only. Debug is very verbose, verbose would be better by default. 

Perhaps you could look at the global logging level and decide which option to use based on that. If it's set to info then set verbose, if it's to debug then set debug, and for error/warning don't set either. Then you don't have to add an extra option.

beardypig: `self.session.logger.level` should have the level. There is no method to map level number to name, or any constants for levels. If you do want to compare log levels I would definitely add some constants to the logging module. 

The logging module is not very good, I'm not sure why the standard python logging library is not used as it's far superior. It's on my TODO list to change all the logging to use python logging, but I haven't got round to it yet :-) Now that I think about it, I should really spend the time and do it...

xkbd: As a result, it would be acceptable following:
```python
if self.session.options.get('subprocess-errorlog-path') or self.session.options.get('subprocess-errorlog'):
    if self.session.logger.level > 3 and self._supports_param("debug"): # don't think '_supports_param' worth checking
        self.params["debug"] = True
    elif self._supports_param("verbos"):
        self.params["verbose"] = True
```

xkbd: > self.session.logger.level should have the level. There is no method to map level number to name, or any constants for levels.

`self.session.logger.Levels[self.session.logger.level]` :)

beardypig: Yeah but I'd do `self.session.logger.level > 2` for verbose, so it's normal rtmpdump logging for error/warning levels. But I really don't like hard coding those numbers, in case they change... I'm going to look at the logging stuff over the next few days. 

Do you want to add this to RTMPStream and create a new PR just for that?

xkbd:
```python
if self.session.options.get('subprocess-errorlog-path') \
or self.session.options.get('subprocess-errorlog'):
    # in a `try/except ValueError` does not wrap intentionally because exception in this case is not a normal situation
    if self.session.logger.Levels[self.session.logger.level] == 'debug':
        self.params["debug"] = True
    else:
        self.params["verbose"] = True
```

That's all :)

xkbd: > Do you want to add this to RTMPStream and create a new PR just for that?

yep sure

src/streamlink/plugins/bongacams.py
```diff
@@ -138,13 +129,18 @@ def _get_streams(self):
                                        self._get_stream_uid(stream_source_info['userData']['username'])),
             # Multiple args with same name not supported.
             # Details: https://github.com/streamlink/streamlink/issues/321
-            "conn" : "S:{username} --conn=S:{access_key} --conn=B:0 --conn=S:{data_key}".format(
+            "conn": "S:{username} --conn=S:{access_key} --conn=B:0 --conn=S:{data_key}".format(
```
beardypig: Are you going to update this to use the latest features from the new process stream class? :-)

gravyboat:
@beardypig It needs a rebase then it should be good.

gravyboat:
@xkbd Could you address those two merge conflicts real quick? Should be good to go once that's done.

xkbd:
@beardypig @gravyboat hi.
Done :)

gravyboat:
@xkbd Awesome thanks!

Issue #412, opened 2017-01-13
Windows Installer adds streamlink to PATH when it already exists 


thinkpad4:
The last 2 times I installed the latest nightly build something bugged my shortcuts that have %windir% in it.  For example MSPaint, the shortcut is %windir% Paint.exe and after installing the nightly, all shortcuts that have %windir% in the target of the shortcut break.  Nothing is deleted, it just seems to lose what %windir% means, and only restarting fixes it.  Now I still need to do more testing on other computers but I wanted to create a issue message before I forgot.  This is Windows 7 SP1 if that helps

gravyboat:
Any relation to this: http://superuser.com/questions/355594/windows-7s-path-and-environment-variables-are-corrupted seems like there are quite a few instances of these types of problems.

thinkpad4:
I tested it last night, I was able to open the shortcuts, then installed the nightly then they stopped working, and I made sure to check it was ONLY the shortcuts with %windir% in the target listing in the shortcut properties.  I then rebooted, all the shortcuts worked again, then I reinstalled the nightly again and only the nightly was used like 1 min after windows loaded so nothing else was changed and yep those shortcuts only containing %windir% stopped being able to be loaded.

thinkpad4:
100% it is something to do with the Streamlink Nightly, I just testing it again under controlled conditions.  It removes the PATH %windir%

bastimeyer:
There is no logic that removes the `windir` env var...
- https://github.com/takluyver/pynsist/blob/1.9/nsist/pyapp.nsi#L89
- https://github.com/takluyver/pynsist/blob/1.9/nsist/_system_path.py#L110-L174
- https://github.com/streamlink/streamlink/blob/8b710ba7f642e0906413157e60e602ce41cfe1cf/script/makeinstaller.sh#L30-L187

Just check the keys in the registry if you want to be sure:
`HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Session Manager\Environment`

thinkpad4:
I figured it out, it WAS the PATH having too many characters in it, seems everytime I install a new nightly of Streamlink, a new "Program Files (x86)\Streamlink\bin" gets added, so the PATH entry was filled up with that

skulblakka:
>seems everytime I install a new nightly of Streamlink, a new "Program Files (x86)\Streamlink\bin" gets added

For me this only happens if I don't uninstall streamlink before installing it again.

[...]

PR #434, opened 2017-01-16
installer: remove the streamlink bin dir from %PATH% before installing

beardypig:
Fix the duplicate streamlink bin directories in `%PATH%` as reported in #412 by @thinkpad4


gravyboat:
Thanks @beardypig!

Issue #432, opened 2017-01-16
--twitch-disable-hosting breaks VOD playback


GreyAsteroid:
Using --twitch-disable-hosting while someone is hosting another channel causes VOD playback to break for all their VODs.

```
streamlink https://www.twitch.tv/talutha/v/114091621
[cli][info] Found matching plugin twitch for URL https://www.twitch.tv/talutha/v/114091621
[plugin.twitch][debug] Getting video HLS streams for talutha
[plugin.twitch][info] Attempting to authenticate using OAuth token
[plugin.twitch][info] Successfully logged in as GreyAsteroid
[plugin.twitch][info] talutha is hosting bwana
[plugin.twitch][info] hosting was disabled by command line option
error: No streams found on this URL: https://www.twitch.tv/talutha/v/114091621
```

GreyAsteroid:
On another note, even after enabling hosting, streamlink seems confused and spouts out this

```
[plugin.twitch][info] cryaotic is hosting effinslowbeef
[plugin.twitch][info] switching to effinslowbeef
[plugin.twitch][debug] Getting video HLS streams for effinslowbeef
```
Then proceeds to play the VOD from the original streamer

gravyboat:
That's odd, when I do that I get this:
```
streamlink twitch.tv/cryaotic best
[cli][info] Found matching plugin twitch for URL twitch.tv/cryaotic
[plugin.twitch][info] cryaotic is hosting effinslowbeef
[plugin.twitch][info] switching to effinslowbeef
[cli][info] Available streams: audio, mobile (worst), low, medium, high, source (best)
[cli][info] Opening stream: source (hls)
[cli][info] Starting player: /usr/bin/vlc
```

For your original issue I get this:
```
 streamlink https://www.twitch.tv/talutha/v/114091621 best
[cli][info] Found matching plugin twitch for URL https://www.twitch.tv/talutha/v/114091621
[cli][info] Available streams: audio, mobile (worst), low, medium, high, source (best)
[cli][info] Opening stream: source (hls)
[cli][info] Starting player: /usr/bin/vlc
```
Though that may be because the user isn't hosting. What does the command you're running for the second comment look like, and what version of Streamlink are you running? Do you have any custom configuration set?

GreyAsteroid:
The context of my second comment is
```
streamlink twitch.tv/cryaotic/v/114852831 best
[cli][info] Found matching plugin twitch for URL twitch.tv/cryaotic/v/114852831
[...]

PR #435, opened 2017-01-16
plugins.twitch: only check hosted channels when playing a live stream

beardypig:
A reported in #432 by @GreyAsteroid (cool avatar btw). 

This PR changes the `_get_hls_streams` so that it only checks for hosted channels when playing a live stream, and ignores hosted channels when playing VOD.


gravyboat:

PR #436, opened 2017-01-16
General tweaks to the docs

beardypig:
Minor tweaks to the docs. Fixed all the warnings and enabled "Fail on warning" option so that problems with the docs will cause the build to fail on warning, the docs are also built for every commit and PR.
Restructured the travis script/after_success so that the docs are only pushed if the build was successful (only master branch and no PRs).


gravyboat:

beardypig:
Looks that way :-)

PR #438, opened 2017-01-16
stream.rtmpdump: set rtmpdump option "verbose" or "debug"

xkbd:
As we discussed with @beardypig at #429 

Set  rtmpdump option `verbose` or `debug` (depending on `self.session` logging level) if streamlink option `subprocess-errorlog`/`subprocess-errorlog-path` enabled.

### DIFF
--- a/src/streamlink/stream/rtmpdump.py
+++ b/src/streamlink/stream/rtmpdump.py
@@ -41,6 +41,17 @@ class RTMPStream(StreamProcess):
 
             self.params["socks"] = self.session.options.get("rtmp-proxy")
 
+        # set rtmpdump logging level
+        if self.session.options.get("subprocess-errorlog-path") \
+        or self.session.options.get("subprocess-errorlog"):
+            for i in ("quiet", "verbose", "debug"):
+                if i in self.params:
+                    del self.params[i]
+            if self.session.logger.Levels[self.session.logger.level] == "debug":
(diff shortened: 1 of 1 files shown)


xkbd:
I also wrote simple tool for testing: `fake-rtmpdump`. Is an executable file that reads fake_stream.flv (previously dumped with rtmpdump) and printed it to stdout.
If replace the original rtmpdump.exe on this is convenient to test some things without having to present rtmp-stream.
It supports some simple options that can be passed from plugins.

`--help` screen:

```
D:\TEMP\>fake-rtmpdump.exe /?
Args: '/?'

fake-rtmpdump.exe: Fake rtmpdump.exe, test tool for Streamlink (github.com/streamlink/streamlink)

--help|h|?                print this screen and exit
--flv|o                   FLV output file name, if the file name is '-' print stream to stdout
--crash                   raise exception
--no-exit                 sleep 60 seconds after EOF
--no-data                 don't read input file just sleep 60 seconds and exit
--read-8k                 read only first 8192 bytes

* input stream - perviously dumped with rtmpdump file "fake_stream.flv" at current directory
* args passed to the app always printed to stderr
```

Its located here: https://github.com/xkbd/tools/tree/master/fake-rtmpdump/fake-rtmpdump/bin/Release

xkbd:
#441 included it

PR #440, opened 2017-01-17
Fix deprecation warnings in Python 3.6

beardypig:
Regular expressions should be raw strings (`r""`) when they contain regex escape sequences (eg. `\w`).


gravyboat:

PR #442, opened 2017-01-17
streamlink_cli.utils.progress: write new line at finish

xkbd:
without it:
```
[download][test.flv] Written 129.2 MB (5s @ 7.5 MB/s)
cli][info] Stream ended
```

with it:
```
[download][test.flv] Written 129.2 MB (5s @ 7.5 MB/s)
[cli][info] Stream ended
```


beardypig:
Seems to only be a problem on Windows. 

Looks good to merge @xkbd, thanks 👍

PR #443, opened 2017-01-17
Update installation instructions for Fedora

melmorabity:
Streamlink is now available in Fedora from official repositories. This PR updates installation instructions for Fedora.


beardypig:
Very cool. Thanks @melmorabity :-)

PR #445, opened 2017-01-18
plugins.adultswim: correct behaviour for archived streams

beardypig:
Fixes the remaining issue with adult swim VOD support.


gravyboat:

Issue #447, opened 2017-01-18
9anime.to not working again :(


bobvargas:
Hey guys I have found another plugin bug on 9anime.to this time when I want to watch doraemon 2005 it gives me an error like this:

C:\Users\SUS>streamlink -p C:\Users\SUS\Downloads\VideoLAN\VLC\vlc.exe 9anime.to
/watch/doraemon-2005.834o/kko61v 720p
[cli][info] Found matching plugin nineanime for URL 9anime.to/watch/doraemon-200
5.834o/kko61v
error: Unable to open URL: //9anime.to/grabber-api/ (Invalid URL '//9anime.to/gr
abber-api/': No schema supplied. Perhaps you meant http:////9anime.to/grabber-ap
i/?)
what that it mean? Please help me.

beardypig:
Looks like something has changed on the site, but it's probably an easy fix.

bobvargas:
@beardypig I know sorry to said again that the plugin is dealing with issues with site changes but hopefully you fix the api on 9anime.to @beardypig

beardypig:
Should be easily fixed :-)

gravyboat:
Closing this since #449 is merged.

sucht:
sorry for highjacking, but this is about 9anime not working so...

anyway, i'm getting

streamlink https://9anime.to/watch/detective-conan.myz/9jo760 best
[cli][info] Found matching plugin nineanime for URL https://9anime.to/watch/detective-conan.myz/9jo760
error: Unable to validate JSON: Key 'grabber' not found in {u'error': u'token'}

gravyboat:
@sucht Can you please open a new issue?

sucht:
done, and sorry.

PR #449, opened 2017-01-19
plugins.nineanime: add scheme to grabber api url if not present

beardypig:
This PR fixes the issue reported in #447.

The API calls have been updated to use the same URL scheme as the stream URL requested. If the URL requests use http then the API requests also use http, if the stream URL uses https then the API requests use https. 

I suspect 9anime.to is making a transistion to https only.


gravyboat:
Good stuff, thanks @beardypig. HTTPS only is definitely worth updating for!

PR #450, opened 2017-01-19
session: add an option to disable Diffie Hellman key exchange

beardypig:
This is useful if an old server has a DH key that is too short for modern implementations of ssl.

(As far as I understand it): 
Due to the logjam attack (CVE-2015-4000) DH primes that are too short will be rejected by ssl. If the stream you are trying to load uses a weak DH key, this option allows you to disable Diffie Hellman as a possible key exchange method and use an alternative. This may or may not be a good idea, and should be used with caution.

### DIFF
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -4,6 +4,8 @@ import re
 import sys
 import traceback
 
+import requests
+
 from . import plugins, __version__
 from .compat import urlparse, is_win32
 from .exceptions import NoPluginError, PluginError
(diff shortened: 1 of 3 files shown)


gravyboat:
Do we really want to support servers that have less security? What streams are having this issue? I'm not a fan of enabling services to continue being insecure if it forces their users to put pressure on them to get with the times.

beardypig:
It's similar to the `--http-ssl-no-verify` option in that respect. It doesn't actually lower security as a different key exchange algorithm will be used, but it does offer a work around for any server not updated with a larger dh prime. 

The stream is some Greek IPTV stream.

beardypig:
Maybe a more generic option to augment the the cipher list would be more suitable?

gravyboat:
Hmm, I think I would prefer your method as opposed to an option that 'disables all the things' since we'll have to back it out at some point or modify it I'm sure. I'm going to merge this in, good to know the reasoning, thanks!

Issue #403, opened 2017-01-11
SRG SSR plugin


karlo2105:
Hello

Another challenge for beardypig. ;)
Swiss public TV includes channels in : 
- German : SRF 1, SRF 2, SRF Info : `http://www.srf.ch/play/tv/live`
- French : RTS Un, RTS Deux, RTS Info : `http://www.rts.ch/play/tv/direct`
- Italian : RSI La 1, RSI La 2 : `http://www.rsi.ch/play/tv/live`

Sometimes they air live sports events exclusively online on their website. 
Channels are of course geoblocked to CH but it can be resolved with proxy. 

Example address source for RTS Un can be found here :
`http://il.srgssr.ch/integrationlayer/1.0/ue/rts/video/play/3608506.json`

They propose HDS and HLS.

beardypig:
I'll take a look, if I can find a decent Swiss proxy :-)

beardypig:
@karlo2105 I had a go, I wasn't able to test it as I couldn't find a working Swiss proxy ...

gravyboat:
This is added, if you could test it @karlo2105 and confirm we can close this.

karlo2105:
Thanks very much I will give a try if i manage with working proxy. ;)

karlo2105:
When I use proxy, it buffers a bit but with streamlink I got error on all channels. 
Take a look : 

`[cli][info] Found matching plugin srgssr for URL http://www.srf.ch/play/tv/live#
?tvLiveId=c4927fcf-e1a0-0001-7edd-1ef01d441651
error: Unable to open URL: http://il.srgssr.ch/integrationlayer/1.0/ue/c4927fcf-
e1a0-0001-7edd-1ef01d441651/video/play/srf.json (404 Client Error: Not Found for
 url: http://il.srgssr.ch/integrationlayer/1.0/ue/c4927fcf-e1a0-0001-7edd-1ef01d
441651/video/play/srf.json)
[End of Streamlink for Windows]`

Some additional addresses : 
----------------------------
SRF.ch
------
`http://www.srf.ch/livestream/player/srf-1`
`http://www.srf.ch/livestream/player/srf-2`
`http://www.srf.ch/livestream/player/srf-info`

Live sports events : you can find below events marked with small cam which will be broadcasted on their website
`http://www.srf.ch/sport/resultcenter/results`

RTS.ch
------
Live sports events 
`http://www.rts.ch/sport/programmes/`

RSI.ch
------
Live sports events
[...]

PR #451, opened 2017-01-19
Support for SRG SSR sites

beardypig:
This PR adds the plugin requested in #403.

Testing is required as I couldn't find a Swiss proxy that would work :)

The following URLs would be good test cases:
- srf.ch/play/tv/live
- rsi.ch/play/tv/live#?tvLiveId=livestream_La1
- rsi.ch/play/tv/live?tvLiveId=livestream_La1
- rtr.ch/play/tv/live
- rts.ch/play/tv/direct#?tvLiveId=3608506
- srf.ch/play/tv/live#?tvLiveId=c49c1d64-9f60-0001-1c36-43c288c01a10


gravyboat:
Thanks @beardypig!

PR #453, opened 2017-01-20
cli: try to terminate the player process before killing it

### LINKED ISSUE #276: Less violent way of closing player when stream ends
Currently streamlink uses SIGKILL to close the player when a stream ends. This prevents the player from doing its own cleanup. For example, mpv leaves DPMS/screensaver disabled because of this. I know there is --player-no-close option, but that has an unwanted side-effect of not closing the player immediately in some situations.

I suggest fixing it by using SIGTERM instead:
```diff
diff -bur streamlink-0.1.0-orig/src/streamlink_cli/output.py streamlink-0.1.0/src/streamlink_cli/output.py
--- streamlink-0.1.0-orig/src/streamlink_cli/output.py  2016-11-21 21:56:29.000000000 +0200
+++ streamlink-0.1.0/src/streamlink_cli/output.py       2016-12-08 22:08:23.000000000 +0200
@@ -161,7 +161,7 @@
 
         if self.kill:
             with ignored(Exception):
-                self.player.kill()
+                self.player.terminate()
         self.player.wait()
 
     def _write(self, data):
```

beardypig:
Only `kill` the player process if the process fails to stop from a `terminate` (after a timeout of 10 seconds). 
On Windows `terminate` is the same as `kill`.

Fixes #276

### DIFF
--- a/src/streamlink_cli/output.py
+++ b/src/streamlink_cli/output.py
@@ -67,6 +67,8 @@ class FileOutput(Output):
 
 
 class PlayerOutput(Output):
+    PLAYER_TERMINATE_TIMEOUT = 10.0
+
     def __init__(self, cmd, args=DEFAULT_PLAYER_ARGUMENTS, filename=None, quiet=True, kill=True, call=False, http=False,
                  namedpipe=None):
         super(PlayerOutput, self).__init__()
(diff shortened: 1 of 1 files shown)


gravyboat:

src/streamlink_cli/output.py
```diff
@@ -160,7 +162,15 @@ def _close(self):
 
         if self.kill:
             with ignored(Exception):
-                self.player.kill()
+                self.player.terminate()
+                if not is_win32:
+                    t, timeout = 0.0, self.PLAYER_TERMINATE_TIMEOUT
+                    while not self.player.poll() and t < timeout:
```
LoneFox78: Shouldn't this be `while self.player.poll() is None` ?
Now it waits if the player returned 0, for example if I close mpv by pressing Q...

beardypig: You're damn right it should be!

PR #454, opened 2017-01-20
Add a general locale setting which can be used by plugins

beardypig:
The `crunchyroll-locale` option was useful, but too specific to crunchyroll. There is a general usefulness for users to be able to control their desired locale settings. For example, in streams with multiple audio languages the correct language could be selected based on the locale setting. (I am working on this for HLS Streams). 

I have added a `Localization` class that provides some useful methods for compare languages using different length codes (eg. 'en' vs. 'eng') and for determining the default locale settings. 

The Crunchyroll plugin has been updated to use the new locale setting and will benefit from automatically switching to the users locale, much like it would on the web version of crunchyroll. 

Two dependencies were added `iso-639` and `iso3166`, which provide information and lookup for countries and languages.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -156,6 +156,8 @@ Name                                 Notes
 `python-requests`_                   At least version **1.0**.
 `python-singledispatch`_             Only needed on Python versions older than **3.4**.
 `pycryptodome`_                      Required to play some encrypted streams
+`iso-639`_                           Used for localization settings, provides language information
+`iso3166`_                           Used for localization settings, provides country information
 
 **Optional**
 --------------------------------------------------------------------------------
@@ -176,6 +178,8 @@ Name                                 Notes
 .. _pycryptodome: https://pycryptodome.readthedocs.io/en/latest/
 .. _python-librtmp: https://github.com/chrippa/python-librtmp
 .. _ffmpeg: https://www.ffmpeg.org/
+.. _iso-639: https://pypi.python.org/pypi/iso-639
+.. _iso3166: https://pypi.python.org/pypi/iso3166
 
 
 Installing without root permissions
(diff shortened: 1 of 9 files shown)


gravyboat:
This looks like a great addition to me. I'm going to leave it open in case anyone else wants to weigh in, but I like this a lot.

beardypig:
Cool. I've got a working addition for hls parsing too that picks external audio tracks based on the locale preference. It's pretty cool :-) Just need to do some more testing, it's tricky to find examples of hls playlists with multiple external audio tracks :-)

fishscene:
Excellent idea. I fully support this.

gravyboat:
@beardypig Looks like Travis failed out, any concerns there or are you good with the merge anyways?

beardypig:
It's python crashing, not reassure what's causing it... I haven't looked properly. I'm pretty sure it's good to go, I almost exclusively use 3.6 now so I'm confident that it does work in 3.6 despite the build failure. 

I'll take a look at the crash data and try to work it out.

bastimeyer:
You can always trigger a rebuild in travis if you think that a failed build was caused by something else...

beardypig:
I was finally able to repro the issue, by using docker and getting the [travis-python](https://quay.io/repository/travisci/travis-python) container and basically duplicating the build process that travis runs. 

```sh
pytest tests/ # crashes
python -m pytest tests/ # crashes
pytest tests # crashes
python -m pytest tests # crashes

pytest # works fine
python -m pytest # works fine 
pytest tests/* # works fine
python -m pytest tests/* # works fine
```

It seems to only affect the travis VM, if I run the same process in a clean precise VM it runs fine.

beardypig:
The fix appears to be to use the built in `mock` module.

gravyboat:
Good stuff @beardypig!

PR #457, opened 2017-01-21
plugins.chaturbate: new regex

rinukkusu:
Since yesterday Chaturbate seems to have a new way of setting the .m3u8 playlist for their player.


beardypig:
Thanks for the fix @rinukkusu.

PR #460, opened 2017-01-22
Add a github issue template

bastimeyer:
I'm not sure if an issue template is really necessary here or if this specific template fits the project, but I thought that we could at least have a discussion about it.

👎
- It might be overwhelming or intimidating for casual users or bug reporters
- Not all information is always necessary

👍
- People *usually* read the template before submitting, resulting in clearer reports
- Reduces the number of duplicates
- Environment details will *most likely* be included without asking first (reduces number of responses)
- Posts are formatted correctly *most of the time*

My suggested template can ofc be changed if it's too much or if sth is missing... 😜

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,36 @@
+*Thanks for reporting an issue!*
+*Please read the contribution guidelines first!*
+
+*Feel free to use the following template. Be as detailed as possible.*
+*Don't forget to remove this text before submitting.*
+
+----
+
+### Checklist
+
+- [x] This is a bug report.
+- [ ] This is a plugin request.
+- [ ] This is a feature request.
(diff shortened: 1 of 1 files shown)


gravyboat:
This looks good to me, fail2ban and many other projects have similar templates and it's helpful.

beardypig:
I'm in favour of a template, for bugs I always end up asking the same questions... Which version of which OS are you using, how did you install streamlink link, have you tried the nightly, etc...

bastimeyer:
> have you tried the nightly

Additional stuff like that could be added to the CONTRIBUTING.md file (among plugin guidelines, etc.). I think it's too much to include this in the issue template.

gravyboat:
I'm going to merge this as is for now, we can add more to it later if we want. Thanks @bastimeyer!

PR #466, opened 2017-01-22
plugins.pluzz: fix HLS streams

melmorabity:
HLS streams on pluzz.francetv.fr now require token generation, as for HDS streams.


beardypig:
Looks good to me @melmorabity, thanks!

Issue #471, opened 2017-01-24
Error when playing large twitch archives


kg:
- [x] This is a bug report.

### Description

```'NoneType' object is not subscriptable``` error occurs when playing a large twitch video

### Expected / Actual behavior

A clearer playback failure reason would be good, alternately it could work. I'm guessing something is wrong with the HLS playlist for this archived stream.

### Reproduction steps / Stream URLs to test

1. https://www.twitch.tv/carcinogensda/v/116807345
Plays fine in the browser, but if played using streamlink (or livestreamer, fwiw) you get:
```
Traceback (most recent call last):
  File "C:\Program Files (x86)\Streamlink\bin\streamlink-script.py", line 12, in <module>
    main()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 899, in main
    handle_url()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 499, in handle_url
    handle_stream(plugin, streams, stream_name)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 382, in handle_stream
    success = output_stream(stream)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 249, in output_stream
    stream_fd, prebuffer = open_stream(stream)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 225, in open_stream
    stream_fd = stream.open()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\stream\hls.py", line 261, in open
    reader.open()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\stream\segmented.py", line 185, in open
    self.worker = self.__worker__(self)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\stream\hls.py", line 128, in __init__
    self.reload_playlist()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\stream\hls.py", line 141, in reload_playlist
    playlist = hls_playlist.load(res.text, res.url)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\stream\hls_playlist.py", line 276, in load
    return parser(base_uri).parse(data)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\stream\hls_playlist.py", line 242, in parse
    self.parse_line(lineno, line)
[...]

PR #473, opened 2017-01-24
Read negative segment duration as 0

beardypig:
If an invalid duration for a segment is encountered eg. `#EXTINF:-0.016,` it is ignored and read as `0` duration, the duration is not that important for our use. The baviour for most players seems to be to ingore the invalid duration and round it to 0, and I don't think it's too bad for us to do the same.

This occurs in some Twitch VOD streams, possibly due to very long duration (the example in #471 is 12 hours long), although I was unable to recreate the issue with any other very long stream (twitch.tv/dotapit/v/116506478 18h works fine).


gravyboat:

PR #477, opened 2017-01-24
Play twitch VOD stream from the beginning even if is still being recorded

### LINKED ISSUE #476: Twitch: Playing vods of still recording streams dont start at the beginning
Hi.
Basicially when you want to watch a vod via
streamlink twitch.tv/someguy/v/12344321 high
and the vod you are referencing is still getting recorded aka the streamer is still live and the stream is recorded to that vod ( 12344321 ) streamlink will just stream whatever is currently streamed by the streamer.
I expected that streamlink would start from the vod beginning tho. If I wanted to watch the current stream, i wouldnt specifiy the vod.

beardypig:
This PR fixes #476, an option has been added to `HLSStream` to force the playlist to be played from the start. 

This new option is generally useful for any live stream that has a VOD/DVR feature where you want to start from the beginning of the stream rather than the "live edge".

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -496,7 +496,10 @@ class Twitch(Plugin):
             return {}
 
         try:
-            streams = HLSStream.parse_variant_playlist(self.session, url)
+            # If the stream is a VOD that is still being recorded the stream should start at the
+            # beginning of the recording
+            streams = HLSStream.parse_variant_playlist(self.session, url,
+                                                       force_restart=not stream_type == "live")
         except IOError as err:
             err = str(err)
             if "404 Client Error" in err or "Failed to parse playlist" in err:
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -189,7 +189,7 @@ class HLSStreamWorker(SegmentedStreamWorker):
             self.playlist_end = last_sequence.num
 
         if self.playlist_sequence < 0:
-            if self.playlist_end is None:
+            if self.playlist_end is None and not self.stream.force_restart:
                 edge_index = -(min(len(sequences), max(int(self.live_edge), 1)))
                 edge_sequence = sequences[edge_index]
                 self.playlist_sequence = edge_sequence.num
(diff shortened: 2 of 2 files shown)


gravyboat:
Thanks @beardypig.

Wishbringer:
This breaks watching live content, for me, not sure why.

beardypig:
@Wishbringer do you have an example? It should only affect VOD content.

Wishbringer:
Seems it may have been related to the localization stuff that was also added with this.

Wishbringer:
Here is the trace (hope it makes sense):
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\plugins\twitch.py", line 557, in _get_streams
    return self._get_hls_streams("live")
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\plugins\twitch.py", line 521, in _get_hls_streams
    force_restart=not stream_type == "live")
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\stream\hls.py", line 345, in parse_variant_playlist streams[name_prefix + stream_name] = stream
UnboundLocalError: local variable 'stream' referenced before assignment

NeoFahrenheit:
I don't know if this was adressed before. if so, I am sorry. :)

I would enjoy using a VOD-link (or just twitch.tv/user) of a STILL BEING RECORDED livestream and, when the download comes to the current **live**stream point, the streamlink continues to download if I was just used streamlink twitch.tv/user.

Example:
https://i.imgur.com/w5u7IEg.png

Is that possible?
Sorry for asking here, I searched [here](https://github.com/streamlink/streamlink/issues?utf8=%E2%9C%93&q=Start+live+stream+from+beginning) and I think it would not be good to create a new topic to ask this. I already tried a bunch of different things suggested by some users, but nothing worked.

gravyboat:
@NeoFahrenheit No it can't work like that because the VOD that is written while the streamer is streaming is different than the livestream and we'd somehow have to support smashing the two different files together which is outside the focus of this project. What you probably want to do (and keep in mind sometimes the recorded VOD is behind the live VOD in terms of length so it might not match up) is write a small wrapper to run a Streamlink command to grab the VOD, then another to grab the livestream then you can combine them in post.

PR #480, opened 2017-01-24
Updating the release script.

cdrage:
When doing the 0.3.0 release I encountered some errors generating the
header of the RST file.

Also, we now automatically build the Windows binary as part of the
release progress through Travis, thus, removing the binary from being
uploaded / generated.

### DIFF
--- a/script/release.sh
+++ b/script/release.sh
@@ -43,13 +43,6 @@ requirements() {
   fi
 }
 
-windows_binary() {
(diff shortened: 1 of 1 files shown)


gravyboat:
Good stuff, I'm going to leave this open in case anyone else wants to review, thanks @cdrage.

gravyboat:
No other reviews, merging.

PR #482, opened 2017-01-25
cli: wait for process to exit, not exit with non-0 error code

beardypig:
As pointed out by @LoneFox78. The loop should wait for the process to exit regardless of the error code, rather than waiting for a non-0 error code. Currently it waits for a truthy response, which means it will wait even if the process ended with exit code 0. 

A gross error on my part this one ...


gravyboat:
Good catch, thanks @LoneFox78 and thanks for the fix @beardypig!

Issue #481, opened 2017-01-24
Add "path to ffmpeg" in config file


Farfie:
### Checklist
- [x] This is a feature request.

### Description
Add "path to ffmpeg" in config file for people who already have it and dont want it installed again by streamlink's installer.

stevek123:
It should be there.

```
# FFMPEG is used to mux separate video and audio streams in to a single
# stream so that they can be played. The pull or relative path to ffmpeg
# or avconv should be specified here.
#ffmpeg-ffmpeg=C:\Program Files (x86)\Streamlink\ffmpeg\ffmpeg.exe
ffmpeg-ffmpeg=
```

wlerin:
Aye but does it make any attempt to locate an existing ffmpeg? It looks like the user still has to fill in the path themselves after install before ffmpeg and rtmpdump will work.

beardypig:
On Mac, Linux and Windows it will search the `PATH` to find `ffmpeg(.exe)` or `avconv`. However, if you do deselect the `ffmpeg` or `rtmpdump` sections during the Windows installer you _will_ need to update the streamlinkrc file to point to `ffmpeg` or `rtmpdump`. I'm working on a fix for this right now... 

fyi, for VLC is searches the `PATH` and some default install locations for `vlc.exe`

karlo2105:
I have issues to open multiaudio stream with ffmpeg. 
Streamlink and ffmpeg are located here : `C:\Program Files\Python27\Lib\site-packages`

I copied ffmpeg folder from Streamlink portable and moved it to path above. 
I just installed VLC 3.0. Take a look on output. 
`streamlink -l debug --ffmpeg-ffmpeg "C:\Program Files\Python27\Lib\site-packages\ffmpeg\ffmpeg" "hlsvariant://https://streamingrmp-1479.kxcdn.com/vod/smil:hls-maudios-prod.smil/playlist.m3u8" best
[cli][info] Found matching plugin stream for URL hlsvariant://https://streamingr
mp-1479.kxcdn.com/vod/smil:hls-maudios-prod.smil/playlist.m3u8
[hls.parse_variant_playlist][debug] Using external audio track for stream 360p (
language=en)
[hls.parse_variant_playlist][debug] Using external audio track for stream 270p (
language=en)
[cli][info] Available streams: 270p (worst), 360p (best)
[cli][info] Opening stream: 360p (hls-multi)
[stream.hls][debug] Reloading playlist
[...]

PR #483, opened 2017-01-25
Fix bug in customized Windows install

beardypig:
As discussed in #481, the config file previously needed to be updated if the user chose not to install ffmpeg. With this patch, if `rtmpdump.exe` or `ffmpeg.exe` are in the `%PATH%` they will be located automatically, if they are not found then the default behaviour still occurs (which is to omit streams that require `ffmpeg` or `rtmpdump`). 

Should preempt a few issues.


gravyboat:

Issue #377, opened 2017-01-07
Move Twitch plugin to use Twitch API V5


gravyboat:
Moving discussion from https://github.com/streamlink/streamlink/issues/20 to here as there is a patch applied for V3, but we should stay ahead of the game by moving directly to V5.

thinkpad4:
I would say this should be fast tracked, just incase Twitch pushes v5 early

gravyboat:
It's already been updated to v3 so we should be okay for a bit.

stevek123:
V3 is scheduled to be deprecated (2017) and removed altogether (Feb. 13, 2018)

thinkpad4:
All the more reason to start testing V5 to be the default API streamlink uses for twitch and keep V3 as a fallback until it is removed

beardypig:
Looks like v5 changes the `channels` api endpoints and all requests require the channel ID, instead of the channel shortname/username as in v3. It seems like the best way to do this is to use the channel search api and find the channel with the matching name. 

Most of the API calls made by the plugin are actually to the private `/api` prefix, which hasn't changed on the site. If they are upgraded their public API it's probably safe to assume that they will update their private API soon as well. But moving to use the v5 public API won't change that :)

beardypig:
The move to v5 API was done as part of #484.

stevek123:
On a side note @beardypig: Do you think it would be possible to echo the node that is being used to the command line? It's part of the m3u8 file, and the same as the stream server domain.

beardypig:
The cluster ID? Like `lhr03`?

stevek123:
Yeah, or the whole node part like `video-edge-835028.sfo01`. Sometimes it takes a while for a stream to populate to a closer node, and it would be nice to see which one is used when you start the stream

beardypig:
Can add it as a debug message, printing it all the time is probably not useful to most people.

stevek123:
Yeah, it can be hidden, but it would be nice to have. Would it be possible to turn it on with a command line parameter instead of turning on the whole debugging? Debugging level generates a lot of log, but something like `--twitch-server` could only output this info as a plus if it's added to the command

stevek123:
Should I open a new issue if I want this to get implemented?

beardypig:
@stevek123, yeah. I don't really want to add an extra command line argument for something like that though. Maybe an info level message is suitable, not sure...

PR #484, opened 2017-01-25
plugins.twitch: update public API calls to use v5 API

beardypig:
The main change from v3 was to look up the channel ID, the user endpoint is used to find the channel ID instead of the channel endpoint (thanks @bastimeyer). 

I grouped the functions in to private/public/unused so that it's easier to tell which ones need to work. I also added some warnings to the unused/unsupported methods in case anyone is importing them (unlikely, I know). 

As discussed in #377.

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -1,10 +1,9 @@
 # coding=utf-8
 import re
-
-import requests
-
+import warnings
 from random import random
 
+import requests
 from streamlink.compat import urlparse
 from streamlink.exceptions import NoStreamsError, PluginError, StreamError
 from streamlink.plugin import Plugin, PluginOptions
@@ -180,9 +179,10 @@ class UsherService(object):
 
 
 class TwitchAPI(object):
-    def __init__(self, beta=False):
+    def __init__(self, beta=False, version=3):
         self.oauth_token = None
         self.subdomain = beta and "betaapi" or "api"
+        self.version = version
 
     def add_cookies(self, cookies):
         http.parse_cookies(cookies, domain="twitch.tv")
@@ -198,7 +198,7 @@ class TwitchAPI(object):
         else:
             url = "https://{0}.twitch.tv{1}".format(self.subdomain, path)
 
-        headers = {'Accept': 'application/vnd.twitchtv.v3+json',
+        headers = {'Accept': 'application/vnd.twitchtv.v{0}+json'.format(self.version),
                    'Client-ID': TWITCH_CLIENT_ID}
 
         # The certificate used by Twitch cannot be verified on some OpenSSL versions.
@@ -216,26 +216,27 @@ class TwitchAPI(object):
         self.subdomain = subdomain_buffer
         return response
 
-    def access_token(self, endpoint, asset, **params):
-        return self.call("/api/{0}/{1}/access_token".format(endpoint, asset), **params)
-
-    def channel_info(self, channel, **params):
-        return self.call("/kraken/channels/{0}".format(channel), **params)
-
-    def channel_subscription(self, channel, **params):
-        return self.call("/api/channels/{0}/subscription".format(channel), **params)
-
-    def channel_viewer_info(self, channel, **params):
-        return self.call("/api/channels/{0}/viewer".format(channel), **params)
-
-    def token(self, **params):
-        return self.call("/api/viewer/token", **params)
+    # Public API calls
 
     def user(self, **params):
         return self.call("/kraken/user", **params)
 
+    def users(self, **params):
+        return self.call("/kraken/users", **params)
+
     def videos(self, video_id, **params):
         return self.call("/kraken/videos/{0}".format(video_id), **params)
+
+    def channel_info(self, channel, **params):
+        return self.call("/kraken/channels/{0}".format(channel), **params)
+
+    # Private API calls
+
+    def access_token(self, endpoint, asset, **params):
+        return self.call("/api/{0}/{1}/access_token".format(endpoint, asset), **params)
+
+    def token(self, **params):
+        return self.call("/api/viewer/token", **params)
 
     def viewer_info(self, **params):
         return self.call("/api/viewer/info", **params)
@@ -245,6 +246,17 @@ class TwitchAPI(object):
 
     def clip_status(self, channel, clip_name, schema):
         return http.json(self.call_subdomain("clips", "/api/v1/clips/" + channel + "/" + clip_name + "/status", format=""), schema=schema)
+
+    # Unsupported/Removed private API calls
+
+    def channel_viewer_info(self, channel, **params):
+        warnings.warn("The channel_viewer_info API call is unsupported and may stop working at any time")
+        return self.call("/api/channels/{0}/viewer".format(channel), **params)
+
+    def channel_subscription(self, channel, **params):
+        warnings.warn("The channel_subscription API call has been removed and no longer works",
+                      category=DeprecationWarning)
+        return self.call("/api/channels/{0}/subscription".format(channel), **params)
 
 
 class Twitch(Plugin):
@@ -279,7 +291,7 @@ class Twitch(Plugin):
         parsed = urlparse(url)
         self.params = parse_query(parsed.query)
 
-        self.api = TwitchAPI(beta=self.subdomain == "beta")
+        self.api = TwitchAPI(beta=self.subdomain == "beta", version=5)
         self.usher = UsherService()
 
     def _authenticate(self):
@@ -374,7 +386,7 @@ class Twitch(Plugin):
             chunk_stop = chunk_start + chunk_length
             chunk_stream = HTTPStream(self.session, chunk_url)
 
-            if start_offset >= chunk_start and start_offset <= chunk_stop:
+            if chunk_start <= start_offset <= chunk_stop:
                 try:
                     headers = extract_flv_header_tags(chunk_stream)
                 except IOError as err:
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/twitch.py
```diff
@@ -216,27 +216,28 @@ def call_subdomain(self, subdomain, path, format="json", schema=None, **extra_pa
...
+    def videos(self, video_id, **params):
+        return self.call("/kraken/videos/{0}".format(video_id), **params)
 
     def channel_info(self, channel, **params):
         return self.call("/kraken/channels/{0}".format(channel), **params)
 
-    def channel_subscription(self, channel, **params):
-        return self.call("/api/channels/{0}/subscription".format(channel), **params)
+    def channel_search(self, channel, **params):
+        return self.call("/kraken/search/channels", query=channel, **params)
```
bastimeyer: the `offset=0` and `limit=1` query parameters could be added here

src/streamlink/plugins/twitch.py
```diff
@@ -462,8 +474,15 @@ def _access_token(self, type="live"):
 
         return sig, token
 
+    def _get_channel_id(self, channel):
+        results = self.api.channel_search(channel)
+        for ch in results["channels"]:
```
bastimeyer: instead of iterating multiple search results

beardypig:
Good point about `--twitch-disable-hosting`, that could definitely be optimised. 
Can do the other change, it doesn't really reduce the complexity, but probably one result is sufficient, it might be a slightly faster request though.

bastimeyer:
> My original thinking was the channel you wanted might not be first, I'm not sure exactly what it searches and in what order it sorts the results..

Hm, I think you're right. It doesn't always return the expected data. This is weird. See my examples:

Returns the correct channel data of twitch.tv/diablo (even though there are more popular channels)
```
curl -H "Accept: application/vnd.twitchtv.v5+json" -H "Client-ID: phiay4sq36lfv9zu7cbqwz2ndnesfd8" "https://api.twitch.tv/kraken/search/channels?query=diablo&offset=0&limit=1"
```

Doesn't return any channels, although there's twitch.tv/dota (weird)
```
curl -H "Accept: application/vnd.twitchtv.v5+json" -H "Client-ID: phiay4sq36lfv9zu7cbqwz2ndnesfd8" "https://api.twitch.tv/kraken/search/channels?query=dota&offset=0&limit=1"
```

Returns a different channel other than twitch.tv/arma (weird)
```
curl -H "Accept: application/vnd.twitchtv.v5+json" -H "Client-ID: phiay4sq36lfv9zu7cbqwz2ndnesfd8" "https://api.twitch.tv/kraken/search/channels?query=arma&offset=0&limit=1"
```

bastimeyer:
Okay, had a look at the Twitch dev forums...
The correct way of getting a channel id in the v5 API is this:

```
curl -H "Accept: application/vnd.twitchtv.v5+json" -H "Client-ID: phiay4sq36lfv9zu7cbqwz2ndnesfd8" "https://api.twitch.tv/kraken/users?login=dota"
```

This returns a `user` record instead of a channel...

beardypig:
Fair enough... I guess channel is equivalent to user, so that makes sense :)

beardypig:
Removing the hosted channel call when hosting is disabled changes the output slightly. 

If `--twitch-disable-hosting` is enabled; currently the channel is always tested to see if it is hosting (minimum 2 API calls), and if it's not hosting there is an HTTP call for the stream m3u8 (3 HTTP requests total). If the channel is hosting, then no streams are returned (2 HTTP requests).

The updated version doesn't make the hosting check, but always makes the m3u8 request, even if the channel is hosting (1 HTTP request, always). As the general case is that the channel is not hosting, it is a reduction of 2 HTTP calls. 

Additionally the stream token/signature request has been moved to after resolution of the hosted channels has been done, this reduced the number of requests if the channel is hosting.

src/streamlink/plugins/twitch.py
```diff
@@ -216,27 +216,28 @@ def call_subdomain(self, subdomain, path, format="json", schema=None, **extra_pa
         self.subdomain = subdomain_buffer
         return response
 
-    def access_token(self, endpoint, asset, **params):
-        return self.call("/api/{0}/{1}/access_token".format(endpoint, asset), **params)
+    # Public API calls
+
+    def user(self, **params):
+        return self.call("/kraken/user", **params)
```
bastimeyer: The API endpoint is `/kraken/users` for retrieving the channel id...

src/streamlink/plugins/twitch.py
bastimeyer: Do we still need the channel_search method now that we are using the other API call?

src/streamlink/plugins/twitch.py
```diff
@@ -462,34 +474,42 @@ def _access_token(self, type="live"):
...
         self._authenticate()
-        sig, token = self._access_token(stream_type)
+        self._hosted_chain.append(self.channel)
+
         if stream_type == "live":
-            hosted_channel = self._check_for_host()
-            if hosted_channel and self.options.get("disable_hosting"):
+            hosted_channel = not self.options.get("disable_hosting") and self._check_for_host()
+            if self.options.get("disable_hosting"):
                 self.logger.info("hosting was disabled by command line option")
```
bastimeyer: Is it necessary to always print this message, even though a channel is not hosting another one?

src/streamlink/plugins/twitch.py
beardypig: Hmm, `user` worked fine... I'll update it.

src/streamlink/plugins/twitch.py
beardypig: Nah, can remove it.

src/streamlink/plugins/twitch.py
bastimeyer: > user worked fine

I'm getting a 401 response...
```
$curl -H "Accept: application/vnd.twitchtv.v5+json" -H "Client-ID: phiay4sq36lfv9zu7cbqwz2ndnesfd8" "https://api.twitch.tv/kraken/user?login=dota"
{"error":"Unauthorized","status":401,"message":"Token invalid or missing required scope"}

$ curl -H "Accept: application/vnd.twitchtv.v5+json" -H "Client-ID: phiay4sq36lfv9zu7cbqwz2ndnesfd8" "https://api.twitch.tv/kraken/users?login=dota"
{"_total":1,"users":[{"display_name":"dota","_id":"715054","name":"dota","type":"user","bio":null,"created_at":"2008-06-08T11:10:39.802846Z","updated_at":"2017-01-26T16:32:19.834402Z","logo":null}]}
```

src/streamlink/plugins/twitch.py
beardypig: Only because it was there before, to check if the channel is hosting to disable the message conditionally is going back to checking for hosting every time. 
My reasoning was that it's a nice reminder to user that they turned hosting off (in the config) and they might not get any streams back because of it. Wouldn't be too bad to remove it... Or maybe it just better to check to see if the channel is hosting regardless.

thinkpad4: I agree with @beardypig the reminder is useful to remind me I have it disabled

beardypig: Back to always testing for hosting then :-)

src/streamlink/plugins/twitch.py
beardypig: True that, `user` only works if you're logged in, ie. include an oauth token :)

beardypig:
I've made some changes based on what has been discussed. I also fixed the new VOD URLs as reported in #496, but we can skip that commit and create new PR if that is preferred :)

gravyboat:
I'm fine with this one, opinions @bastimeyer and @thinkpad4?

src/streamlink/plugins/twitch.py
bastimeyer: Hm... We're checking for hosted channels, although an explicit parameter can be set to disable this behavior (switching to a different channel at least). IMO, there's no need to print this message if the parameter was set, at least in the info channel. Maybe it could be moved to the debug log output.
If you want to keep it, it's also fine I guess, but the API requests are unnecessary. I'm just wondering if this could confuse any users, because automatically switching to hosted channels is disabled in the [Twitch GUI](https://github.com/streamlink/streamlink-twitch-gui/blob/v1.1.1/src/app/models/stream/parameters.js#L61-L64) and having such a log output is unexpected for users who don't ever use the CLI.

beardypig: It's been this way since the `--twitch-disable-hosting` flag was added, if users have been confused they haven't mentioned it (afaik). It's more of a reminder to users that have it set in their `streamlinkrc` file. Having said that, if the log message wasn't there already I wouldn't add it ;) I can't decide if to remove it or not, I like the idea of making less requests, but I'm not sure about removing the existing log message... 

The most optimal way to do it (in the case of no hosting) would be to request the channel stream first, then if it 404s check to see if it's hosting another channel.

bastimeyer: I'm talking about having a log output at all times when using the parameter, which was the case in your previous commits before you've force-pushed new commits...

> The most optimal way to do it

If there's a nice, non-messy solution for this, so that stream launch times can be reduced, I would welcome it.

thinkpad4:
Unless the channel is hosting someone it shouldn't be displayed, example: 
[cli][info] Found matching plugin twitch for URL twitch.tv/crream
[plugin.twitch][info] Attempting to authenticate using OAuth token
[plugin.twitch][info] Successfully logged in as Thinkpad4
[plugin.twitch][info] hosting was disabled by command line option
error: No streams found on this URL: twitch.tv/crream

See, they aren't hosting someone so informing me that I disabled hosting in the command line is not needed.  Only when I load a channel that IS hosting someone should the [plugin.twitch][info] hosting was disabled by command line option be displayed informing me that I won't see the hosted channel.

bastimeyer:
I think we should merge this. The purpose of this PR was upgrading to v5. Everything else like reducing the API calls or changing the log output is just an enhancement and a different "issue". Also, the Twitch VOD url fix has been added here and needs to be merged as well.

gravyboat:
Agreed @bastimeyer, merging.

PR #485, opened 2017-01-25
Support for external audio streams in HLS

### LINKED ISSUE #111: Please add external audio track feature
I dig up here some interesting question feature from original project : https://github.com/chrippa/livestreamer/issues/1464

There are more and more streams with separated audio track(s) from playlist which results in a picture without audio. It would be great it this feature could be implemented in streamlink project.

beardypig:
This PR requires #454 be merged first. 

It adds the requested external audio stream feature to the `hls`  module (fixed #111). The `locale` setting is used to pick an appropriate audio stream when there are multiple available. If an audio stream with the language matching your requested language (or system language) then that stream will be picked over the others, if none are found then the default stream is used. 

Example stream: https://streamingrmp-1479.kxcdn.com/vod/smil:hls-maudios-prod.smil/playlist.m3u8

This HLS Stream has 3 audio streams, in 3 different languages, english, french and german (not that they are actually in those languages, there are just 2 extra random audio streams). If you select any of the available languages (`--locale en_US`, `--locale de_DE` or `--locale fr_FR`) you will get the corresponding audio stream. If you select a different locale (eg. `es_ES`) you will get the default language, which in this case in English. 

Subtitles are _not_ supported, it's not really very easy to add support for subtitles (I tried), as `ffmpeg` will not mux subtitles from a stream. ie. you need to download the full subtitle stream before any output is given. If you have a live stream with a separate subtitle track the stream will never begin, as the subtitle stream will never end.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -156,6 +156,8 @@ Name                                 Notes
 `python-requests`_                   At least version **1.0**.
 `python-singledispatch`_             Only needed on Python versions older than **3.4**.
 `pycryptodome`_                      Required to play some encrypted streams
+`iso-639`_                           Used for localization settings, provides language information
+`iso3166`_                           Used for localization settings, provides country information
 
 **Optional**
 --------------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


gravyboat:
#454 is merged, thanks @beardypig!

PR #491, opened 2017-01-26
Localization/HLS fixes

beardypig:
Some fixes for some issues that have come up with wider testing of the Localization and HLS external audio streams.


gravyboat:

PR #498, opened 2017-01-27
plugins.zhanqi: use new api

steven7851:
With the new api does not need to download the entire page.
Delete useless code "stream_weight", and make http as live for best quality.

### DIFF
--- a/src/streamlink/plugins/zhanqi.py
+++ b/src/streamlink/plugins/zhanqi.py
@@ -6,6 +6,9 @@ from streamlink.stream import HTTPStream, HLSStream
 
 API_URL = "https://www.zhanqi.tv/api/static/v2.1/room/domain/{0}.json"
 
+STATUS_ONLINE = 4
+STATUS_OFFLINE = 0
+
 _url_re = re.compile(r"""
     http(s)?://(www\.)?zhanqi.tv
     /(?P<channel>[^/]+)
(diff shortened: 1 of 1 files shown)


gravyboat:
Can we get some tests to go along with this? These are pretty big changes to not have any.

src/streamlink/plugins/zhanqi.py
```diff
@@ -62,25 +30,17 @@ class Zhanqitv(Plugin):
...
-
-        res = http.get(API_URL.format(roomID=roomID))
+        res = http.get(API_URL.format(channel))
         room = http.json(res, schema=_room_schema)
         if not room:
             self.logger.info("Not a valid room url.")
             return
 
-        if room["status"] != STATUS_ONLINE:
+        if room["status"] != 4:
```
beardypig: Why remove the constant? Seems to me that having the code read `room["status"] != STATUS_ONLINE` is more understandable than `room["status"] != 4`.

gravyboat:
@beardypig Are we good to merge this?

gravyboat:
Unless there are further concerns I would like to get this merged today.

beardypig:
It looks OK. I haven't tested it though.

gravyboat:
I haven't either. I'm going to merge it since I don't feel like doing the work to set it up so I can connect to a Chinese stream. Plus I trust that @steven7851 wouldn't open a PR that doesn't work.

gravyboat:
Thanks @steven7851!

PR #504, opened 2017-01-29
plugins.pluzz: ignore DRM-protected streams + log messages

melmorabity:
This PR improved the Pluzz plugin:

* DRM-protected streams are ignored (for example: http://pluzz.francetv.fr/videos/castle_20h35_,152224484.html)
* print log messages when no stream is available (geo-restricted content, DRM, expired content...).


beardypig:
Looks good to me :)

gravyboat:
Thanks @melmorabity!

PR #512, opened 2017-01-31
Fixes and additions to the dogus plugin

beardypig:
Fixed support for the ntv.com.tr live stream and added support for eurostartv.com.tr/canli-izle.


gravyboat:
Good stuff, thanks!

PR #514, opened 2017-02-01
plugins.douyutv: fix room id regex

steven7851:
from https://github.com/soimort/you-get/issues/1657

### DIFF
--- a/src/streamlink/plugins/douyutv.py
+++ b/src/streamlink/plugins/douyutv.py
@@ -12,8 +12,6 @@ from streamlink.stream import HTTPStream
 MAPI_URL = "https://m.douyu.com/html5/live?roomId={0}"
 LAPI_URL = "https://www.douyu.com/lapi/live/getPlay/{0}"
 LAPI_SECRET = "A12Svb&%1UUmf@hC"
-SHOW_STATUS_ONLINE = 1
-SHOW_STATUS_OFFLINE = 2
 STREAM_WEIGHTS = {
     "low": 540,
     "middle": 720,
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/douyutv.py
```diff
@@ -109,9 +108,11 @@ def _get_streams(self):
         res = http.get(MAPI_URL.format(channel))
         room = http.json(res, schema=_room_schema)
         if not room:
+            self.logger.info("API error")
```
gravyboat: Can we be more descriptive with this error? `API error` isn't very helpful for users.

src/streamlink/plugins/douyutv.py
```diff
@@ -109,9 +108,11 @@ def _get_streams(self):
...
             return
 
         if room["show_status"] != SHOW_STATUS_ONLINE:
+            self.logger.info("Channel offline now!")
```
gravyboat: Can we make this more neutral? `Channel is offline` or something would be fine.

src/streamlink/plugins/douyutv.py
```diff
@@ -109,9 +106,11 @@ def _get_streams(self):
         res = http.get(MAPI_URL.format(channel))
         room = http.json(res, schema=_room_schema)
         if not room:
+            self.logger.info("Not a valid room url.")
             return
 
-        if room["show_status"] != SHOW_STATUS_ONLINE:
+        if room["show_status"] != 1:
+            self.logger.info("Stream current unavailable.")
```
beardypig: You should change this to `Stream currently unavailable`

beardypig:
Looks good, just one suggested change for grammar :)

beardypig:
Looks good to me now :)

beardypig:
What d'ya think @gravyboat ?

gravyboat:
Yeah I'm good with this, thanks @steven7851.

PR #515, opened 2017-02-01
Update license and debian copyright

gravyboat:


gravyboat:
@bastimeyer If you're fine with these please go ahead and do the merge, not much to review here.

bastimeyer:
I was on my phone and couldn't merge (I don't like using Github for this kind of stuff because it uses a different committer name). Now it looks like @beardypig and I merged at the same time 😞

beardypig:
Haha github says I merged, git says you merged :) No harm :)

bastimeyer:
That's the issue with Github merges. It does its own thing and creates new SHAs when a PR gets squashed, rebased or committer/author names get changed. That's not ideal when a PR branch gets reused and I personally don't like it.

I must have pulled the master branch right after you merged via github, because I didn't see this in my browser after I read the comment and added the remote repo url. I could have checked my local log of the master branch, but I didn't, because I didn't expect this merge in this little time frame and pushed a normal no-ff merge commit of the remote branch. I didn't want to reset the HEAD of the master branch and then force-push, so now it's redundant. But yeah, very small issue and no harm done. 😝

Issue #518, opened 2017-02-02
Ceskatelevize.cz plugin merging


karlo2105:
### Checklist

- [ ] This is a bug report.
- [ ] This is a plugin request.
- [ ] This is a feature request.
- [x] I used the search function to find already opened/closed issues or pull requests.

### Description

Ceskatelevize is Czech public TV service with 6 channels 👍 All channels but CT24 are geoblocked to Czechia. 

1. CT1 - `http://www.ceskatelevize.cz/ct1/zive/`
2. CT2 - `http://www.ceskatelevize.cz/ct2/zive/`
3. CT24 - `http://www.ceskatelevize.cz/ct24/`
4. CT sport - `http://www.ceskatelevize.cz/sport/zive-vysilani/`
5. CT Decko - `http://decko.ceskatelevize.cz/zive/`
6. CT Art - `http://www.ceskatelevize.cz/art/zive/`

You can find live content by channels here too : `http://www.ceskatelevize.cz/ivysilani/zive`

### Expected / Actual behavior

This plugin has been created by [rubickcz](https://github.com/rubickcz) for livestreamer. 
It was proposed as [pull request](https://github.com/chrippa/livestreamer/pull/1508) to livestreamer. 
Currently it displays some errors while I exit geoblocked stream. 

```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Ddr\Downloads\streamlink-portable-latest\streamlink\\streamlink
-script.py", line 18, in <module>
    main()
  File "C:\Users\Ddr\Downloads\streamlink-portable-latest\streamlink\packages\st
reamlink_cli\main.py", line 947, in main
    handle_url()
  File "C:\Users\Ddr\Downloads\streamlink-portable-latest\streamlink\packages\st
reamlink_cli\main.py", line 484, in handle_url
    console.exit(u"{0}", err)
  File "C:\Users\Ddr\Downloads\streamlink-portable-latest\streamlink\packages\st
reamlink_cli\console.py", line 69, in exit
    self.msg("{0}", msg)
  File "C:\Users\Ddr\Downloads\streamlink-portable-latest\streamlink\packages\st
reamlink_cli\console.py", line 45, in msg
    self.output.write(formatted)
  File "encodings\cp850.py", line 19, in encode
UnicodeEncodeError: 'charmap' codec can't encode character '\u2019' in position
398: character maps to <undefined>

```
### Environment details (operating system, python version, etc.)

[...]

PR #520, opened 2017-02-02
Support for Ceskatelevize streams

beardypig:
As per the request in #518, originally copied from chrippa/livestreamer/pull/1508, renamed the plugin and added some better geoblocked error notification.


gravyboat:

Issue #463, opened 2017-01-22
retries on the HLS playlist_reload?


mrawji:
First off, THANK YOU! I was pulling HLS streams via FFMPEG before and they'd behave poorly after about 2 hours. Streamlink has been perfect on that front!

Here's my question.. I was streaming and the reload_playlist method received an timeout about 1 hour into the stream. It's most likely a transient issue with the server / network / internet / etc... this caused streamlink to basically stop. Is there a way to set retries on that?

Thanks!

beardypig:
There is no retry code in the `reload_playlist` method, but there should be!

scottbernstein:
+1 on adding retry code around the reload_playlist method, and making the number of retires configurable from the command line like other retries.

________________________________

beardypig:
Since #522 this should be fixed.

Issue #446, opened 2017-01-18
Read timeout on all attempts when downloading from Twitch


Razunter:
Trying to download stream from Twitch, getting "read timeout" every time at about 5-50mb.
Can watch on Twitch without issues.

Stream url: https://www.twitch.tv/ognglobal/v/115348644
Quality: best

```
d:\Soft\Streamlink\bin>streamlink -f -o "H:\temp\[ENG] OVERWACH APEX S2 ENERGIZED BY HOT6.mp4" https://www.twitch.tv/ognglobal/v/115348644 best --hls-segment-threads 3
[cli][info] Found matching plugin twitch for URL https://www.twitch.tv/ognglobal/v/115348644
[cli][info] Available streams: audio, mobile (worst), low, medium, high, source (best)
[cli][info] Opening stream: source (hls)
[download][..2 ENERGIZED BY HOT6.mp4] Written 7.5 MB (6s @ 1.1 MB/s)                                                   Exception in thread Thread-1:
Traceback (most recent call last):
  File "d:\Soft\Streamlink\pkgs\requests\packages\urllib3\response.py", line 298, in _error_catcher
    yield
  File "d:\Soft\Streamlink\pkgs\requests\packages\urllib3\response.py", line 380, in read
    data = self._fp.read(amt)
  File "http\client.py", line 448, in read
  File "http\client.py", line 488, in readinto
  File "socket.py", line 575, in readinto
  File "ssl.py", line 929, in recv_into
  File "ssl.py", line 791, in read
  File "ssl.py", line 575, in read
socket.timeout: The read operation timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "d:\Soft\Streamlink\pkgs\requests\models.py", line 703, in generate
    for chunk in self.raw.stream(chunk_size, decode_content=True):
  File "d:\Soft\Streamlink\pkgs\requests\packages\urllib3\response.py", line 432, in stream
    data = self.read(amt=amt, decode_content=decode_content)
  File "d:\Soft\Streamlink\pkgs\requests\packages\urllib3\response.py", line 397, in read
    raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
  File "contextlib.py", line 77, in __exit__
  File "d:\Soft\Streamlink\pkgs\requests\packages\urllib3\response.py", line 303, in _error_catcher
    raise ReadTimeoutError(self._pool, None, 'Read timed out.')
[...]

PR #522, opened 2017-02-03
Ensure retries with HLS Streams

beardypig:
This should fix both #446 and #463. 

I fixed the retries on the HLS segments as well as adding retries (via a general HTTPSession mechanism) to the `playlist_reload` and `create_decryptor` methods. A new `--hls-reload-playlist-attempts` option has been added to control the retries for the playlist. 

I reverted most of the changes made for #166/#260 as it has caused some unforeseen issues. Mainly due to the exception handling of `HTTPSession.request` and `iter_content` being quite different. (The exception handling in livestreamer/streamlink is generally a mess.)

We'll have to reopen #166 until we have a more suitable solution.

### DIFF
--- a/src/streamlink/plugin/api/http_session.py
+++ b/src/streamlink/plugin/api/http_session.py
@@ -1,10 +1,13 @@
+import time
 from requests import Session, __build__ as requests_version
 from requests.adapters import HTTPAdapter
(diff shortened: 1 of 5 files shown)


gravyboat:

PR #525, opened 2017-02-03
Update release script again to include sdist

cdrage:
Another update to the release script to include sdist
upload instead of that 50MB binary :)


gravyboat:
Awesome, the smaller we can make this the better! I'll leave this open in case anyone wants to weigh in, but I'm always for smaller file sizes.

PR #528, opened 2017-02-03
Localization enhancements 

beardypig:
Added `Country` and `Language` classes. Optionally use `pycountry` if `iso639` or `iso3166` are not available.


gravyboat:
Thanks @beardypig!

Coacher:
@beardypig, thank you very much!

PR #535, opened 2017-02-05
Option to set crunchyroll session ID

### LINKED ISSUE #505: Support providing a custom Crunchyroll session ID
The documentation suggests using an HTTP proxy to access region-locked content in Crunchyroll. However this isn't necessary, as all you need to do so is a session ID belonging to the region you want. Now I'm not asking for streamlink to generate one for you, as nice as that would be -- simply that it lets you specify it with an argument (it's only a 32 character string).

beardypig:
Adds the option `--crunchyroll-session-id` which allows the user to specify a specific API session ID, useful for bypassing the region restrictions. Fixes #505

### DIFF
--- a/src/streamlink/plugins/crunchyroll.py
+++ b/src/streamlink/plugins/crunchyroll.py
@@ -205,7 +205,8 @@ class Crunchyroll(Plugin):
         "username": None,
         "password": None,
         "purge_credentials": None,
-        "locale": None
+        "locale": None,
+        "session_id": None,
     })
 
     @classmethod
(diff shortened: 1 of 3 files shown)


gravyboat:

theincognito-inc:
@beardypig This option is not present in the current youtube_dl? I am using the latest git pull and directly running it via python.

I got the following message:

```
Usage: __main__.py [OPTIONS] URL [URL...]

__main__.py: error: no such option: --crunchyroll-session-id
```

Issue #534, opened 2017-02-04
Plugin for ludo.fr


Surfoo:
Hello,

Could you add a new plugin for ludo.fr, it's almost the same plugin like the pluzz plugin (same french TV)

Thanks!

gravyboat:
Can we see some examples URLs and so forth?

Surfoo:
for example: 
  - http://www.ludo.fr/heros/il-etait-une-fois-la-vie

gravyboat:
Closing since the associated PR is merged.

PR #536, opened 2017-02-05
plugins.pluzz: add support for ludo.fr and zouzous.fr

melmorabity:
This PR improves the pluzz plugin to add support for the following VOD platforms for youth, provided by France Télévisions:

* ludo.fr (as requested in https://github.com/streamlink/streamlink/issues/534)
* zouzous.fr

As for the streams on pluzz.francetv.fr, videos on these two websites are georestricted to France, Andorra and Monaco.

### DIFF
--- a/src/streamlink/plugins/pluzz.py
+++ b/src/streamlink/plugins/pluzz.py
@@ -101,6 +101,7 @@ class Pluzz(Plugin):
         geolocked = False
         drm = False
         expired = False
+
         for video in videos['videos']:
             video_url = video['url']
 
@@ -156,7 +157,9 @@ class Pluzz(Plugin):
                 if match is not None:
                     bitrate = match.group('bitrate')
                 else:
-                    bitrate = video['format']
+                    # Fallback bitrate (seems all France Televisions MP4 videos
+                    # seem have such bitrate)
+                    bitrate = '1500k'
                 yield bitrate, HTTPStream(self.session, video_url)
 
         if offline:
--- a/src/streamlink/plugins/pluzz.py
+++ b/src/streamlink/plugins/pluzz.py
@@ -13,9 +13,9 @@ class Pluzz(Plugin):
     API_URL = 'http://sivideo.webservices.francetelevisions.fr/tools/getInfosOeuvre/v2/?idDiffusion={0}&catalogue={1}'
     TOKEN_URL = 'http://hdfauthftv-a.akamaihd.net/esi/TA?url={0}'
 
-    _url_re = re.compile(r'http://(pluzz\.francetv\.fr/(videos/.+\.html|[\w-]+)|www\.ludo\.fr/heros/[\w-]+)')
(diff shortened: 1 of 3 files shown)


src/streamlink/plugins/pluzz.py
```diff
@@ -140,6 +150,13 @@ def _get_streams(self):
             elif '.m3u8' in video_url:
                 for stream in HLSStream.parse_variant_playlist(self.session, video_url).items():
                     yield stream
+            elif '.mp4' in video_url:
+                match = self._mp4_bitrate_re.match(video_url)
+                if match is not None:
+                    bitrate = match.group('bitrate')
+                else:
+                    bitrate = video['format']
+                yield bitrate, HTTPStream(self.session, video_url)
```
beardypig: I think when the quality is bitrate it should have the `k` suffix. Streamlink does some conversion to bitrates so they are more comparable to resolution. Might not be a problem in this case, just an FYI.

melmorabity: thanks for the tip ;). I will update my PR accordingly.

Surfoo:
Thanks for the PR @melmorabity!

src/streamlink/plugins/pluzz.py
melmorabity: In fact video['format'] is a fallback when the bitrate is not available in the stream URL (the bitrate is not provided by the Pluzz API). In such a case, video['format'] corresponds to the video type (something like 'mp4-dp' for HTTP MP4 streams here).

beardypig: So for the HTTP streams they'll have a quality name like `mp4-dp`? Is there a way to turn that into something more useful? or replace it with something more useful? `live` is the fallback option, it will get sorted in the list of streams. 
I think `mp4-dp` will not work if it's the only stream available and you try to play `best` (or `worst`). It looks like you'll only ever get one stream when it's `mp4`? Maybe change the name of the stream to `live` as a fallback, or add a `stream_weight` method so they get sorted :)

gravyboat:
With the latest change this looks good to go to me, any other concerns @beardypig?

beardypig:
Just some possible issues with the stream weights for vod streams.

src/streamlink/plugins/pluzz.py
melmorabity: After checking as many http 'mp4-dp' videos as I could find on pluzz, ludo and zouzous, it looks like all of them have a bitrate of about 1500k. Maybe it's reasonable to use '1500k' as a fallback bitrate if it's not available in the video URL.

beardypig: That sounds like a reasonable thing to do. I think it would be best if all the bitrates are suffixed with `k` as well, for consistency.

melmorabity: Thanks for your advice ;). Notice that if the bitrate is available in the MP4 filename, it always includes the 'k' suffix

gravyboat:
I think we're good to go here now with 2bab33d, any further comments/concerns?

beardypig:
Looks good to merge.

gravyboat:
Thanks for the review @beardypig and thanks for the PR @melmorabity!

Issue #541, opened 2017-02-05
Openrec.tv wont recognize SOURCE quality


jstrat231:
*Thanks for reporting an issue!*
*Please read the contribution guidelines first!*

*Feel free to use the following template. Be as detailed as possible.*
*Don't forget to remove this text before submitting.*

----

### Checklist

- [x] This is a bug report.
- [x] This is a plugin request.
- [x ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description

I'm new to this website and coding. I downloaded the windows installer and when I open up CMD and loaded up a stream with the following command:
"streamlink openrec.tv/live/saYffsafdY source"   it keeps telling me that the best quality it can find is 720p. It doesnt get the source quality which is normally always 720p60 at a higher bit rate. 

 Could you please help me? Thank you
...

### Expected / Actual behavior

...

### Reproduction steps / Stream URLs to test

1. ...
2. ...
3. ...

### Environment details (operating system, python version, etc.)

...

### Comments, logs, screenshots, etc.

...

beardypig:
Created a PR to add the `source` quality :)

jstrat231:
Thank u so much sir but I am a total noob when it comes to this stuff. Is there any chance you could do a step by step on how I'm supposed to add this to streamlink? Thank you very much!

(I installed streamlink with the windows.exe installer)

beardypig:
The easiest thing is to wait until the PR has been merged, then you can install the latest nightly/dev version. But if you really want to get it right now...
Download https://raw.githubusercontent.com/beardypig/streamlink/bdf5e8b939bf0e3dc622eb2dff866bee5370f87e/src/streamlink/plugins/openrectv.py and copy it to `C:\Program Files\Streamlink\pkgs\streamlink\plugins\` (depends on where you installed Streamlink)

jstrat231:
thank u very much :D :D :D

gravyboat:
Closing since the associated PR is merged.

PR #542, opened 2017-02-06
plugins.openrectv: add source quality for openrectv

beardypig:
Added `source` quality to the `OPENRECtv` plugin, as requested in #541 :)


gravyboat:
Thanks @beardypig!

Issue #538, opened 2017-02-05
Crunchyroll plugin fails on Python 2.6 (string.format)


scottbuettner:
### Checklist

- [x] This is a bug report.
- [x] This is a plugin request.
- [ ] This is a feature request.
- [x] I used the search function to find already opened/closed issues or pull requests.
(didn't actually find any, but I looked)

### Description
The Crunchyroll plugin contains the following format string:
`API_USER_AGENT = "Mozilla/5.0 (iPhone; iPhone OS 8.3.0; {})"`
which fails on Python 2.6, which requires format placeholders to be explicitly numbered. See https://docs.python.org/2.7/library/string.html#formatstrings : 

> Changed in version 2.7: The positional argument specifiers can be omitted, so '{} {}' is equivalent to '{0} {1}'.

### Expected / Actual behavior
Expected: Crunchyroll plugin is able to load stream
Actual:
```
>>> streamlink.streams("http://www.crunchyroll.com/jojos-bizarre-adventure/episode-1-the-man-possessed-by-an-evil-spirit-652155")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.6/site-packages/streamlink/api.py", line 13, in streams
    return session.streams(url, **params)
  File "/usr/lib/python2.6/site-packages/streamlink/session.py", line 423, in streams
    return plugin.streams(**params)
  File "/usr/lib/python2.6/site-packages/streamlink/plugin/plugin.py", line 240, in streams
    raise PluginError(err)
streamlink.exceptions.PluginError: zero length field name in format
```

### Reproduction steps / Stream URLs to test
```
>>> import streamlink
>>> streamlink.streams("http://www.crunchyroll.com/jojos-bizarre-adventure/episode-1-the-man-possessed-by-an-evil-spirit-652155")
```
(Also fails on CLI)
My testing failed identically with or without my Crunchyroll credentials, so I believe they are not required to verify this.

### Environment details (operating system, python version, etc.)
CentOS 6, Python 2.6

> Python 2.6.6 (r266:84292, Aug 18 2016, 15:13:37)
> [GCC 4.4.7 20120313 (Red Hat 4.4.7-17)] on linux2

### Comments, logs, screenshots, etc.
I'll be submitting a (very trivial) pull request shortly. If I have time I'll go through some other plugins and see if they have similar issues.

scottbuettner:
[...]

PR #544, opened 2017-02-06
Drop Python 2.6 support

beardypig:
As pointed out by @scottbuettner in #538/#539, we don't really have good support for Python 2.6. `requests` is currently pinned to `<2.0` for Python 2.6 and as such lacks SNI support. A lot of plugins use Python 2.6 incompatible format strings (I'll admit a lot of those were added by me:). 

Python 2.6 is no longer supported by the core Python team and is not getting any more updates. A lot of packages are starting to deprecate Python 2.6, and will soon stop supporting Python 2.6. 

I am proposing that deprecate Python 2.6 support, remove it from CI and from the docs as a supported Python version. We can continue to support it in a best-effort capacity, but I think "official" support puts an extra drain on the development work. However, if anyone does wish to continue Python 2.6 support they are more than welcome.


gravyboat:
I see no reason to continue Python 2.6 support. The only users I can think of who might be impacted are those running very old Linux distros, is there anyone else? I agree, it's too much work to be worth supporting.

Brainzyy:
I agree with all the above. Since there's no official support for it, it would be better to drop it.

cdrage:
I agree too to drop 2.6 support.

beardypig:
If this PR is merged it will remove Python 2.6 from the CI builds (travis, appveyor doesn't do 2.6) and will update the docs noting that Python 2.7 is minimum Python version supported. Nothing is done to explicitly stop people from installing `streamlink` on Python 2.6.

Most of the obvious Python 2.6 issues are also fixed by this PR.

cdrage:
LGTM :+1:

gravyboat:
Then I'm merging it, next up let's drop 2.7!

PR #556, opened 2017-02-09
stream.hls: last chance fallback audio

beardypig:
As a last ditch effort, fallback on the first audio stream presented as the audio stream to use when an external audio stream is required.


karlo2105:
This has solved auto detection for default external audio, thanks buddy.

gravyboat:
Thanks @beardypig!

PR #557, opened 2017-02-09
AppVeyor build fix

beardypig:
This fixes the build issues we've been having for the last couple of days, since the new version of `pycryptodome` doesn't have wheels for some versions of Python...


cdrage:
Passes, so LGTM :+1:

PR #560, opened 2017-02-10
stream: make Stream responsible for generating the stream_url

beardypig:
The current method (`streamlink_cli.utils.stream.stream_to_url`) converts a stream to a URL. The responsibility of converting a stream to a URL should be that of the `Stream` class.

A new `to_url` method has been added to the `Stream` class to facilitate this. This makes is much easier for `Stream` subclasses to generate URLs for streams if possible. For example, in the FilmOn specific HLS stream class, `FilmOnHLS` a subclass of `HLSStream`, it is not always possible to generate a suitable URL, as some streams expire almost immediately.

### DIFF
--- a/src/streamlink/plugins/filmon.py
+++ b/src/streamlink/plugins/filmon.py
@@ -48,6 +48,13 @@ class FilmOnHLS(HLSStream):
         else:
             return self._url
 
+    def to_url(self):
+        url = self.url
+        expires = self.watch_timeout - time.time()
+        if expires < 0:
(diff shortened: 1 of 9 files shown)


gravyboat:
This looks good to me but since it's a big chance I'll leave it open in case others want to review. :+1: @beardypig.

beardypig:
Cool, sounds good. I added a bunch of tests too :)

gravyboat:
No one else is weighing in so I'm merging this. Thanks!

PR #561, opened 2017-02-10
utils.l10n: fix bug in iso3166 country lookup

beardypig:
The `pycountry` vs. `iso639+iso3166` l10n tests were broken so that the only the `pycountry` tests were being run. This has now been fixed, which uncovered a bug in the `iso3166` country lookup.


gravyboat:
Awesome, thanks @beardypig!

PR #563, opened 2017-02-10
Update underscores title release script

cdrage:
Update the release script again since I forgot to remove the underscores
for dashes..


beardypig:
Like the `tr` wizardry :)

PR #565, opened 2017-02-10
tests: speed up the cmdline tests

beardypig:
When #453 was merged it added some possible delay in killing the player process, this delay was also experienced in the tests. By mocking the expected return values of `poll()`, the delay can be avoided and the runtime of the tests reduced from around 60 seconds to 6 seconds.

### DIFF
--- a/src/streamlink_cli/output.py
+++ b/src/streamlink_cli/output.py
@@ -97,8 +97,7 @@ class PlayerOutput(Output):
     @property
     def running(self):
         sleep(0.5)
-        self.player.poll()
-        return self.player.returncode is None
+        return self.player.poll() is None
 
     def _create_arguments(self):
         if self.namedpipe:
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -32,8 +32,14 @@ class TestCommandLineInvocation(unittest.TestCase):
     @patch('sys.argv')
     def _test_args(self, args, commandline, mock_argv, mock_popen, mock_setup_streamlink, passthrough=False):
         mock_argv.__getitem__.side_effect = lambda x: args[x]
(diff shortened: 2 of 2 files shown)


beardypig:
Travis: total build time down to ~14 minutes from ~24 minutes.
ApyVeyor: jobs stuck in queue...

gravyboat:
Thanks @beardypig, tests were definitely taking a while to run!

beardypig:
They're still not quick, but there isn't much we can do about that. AppVeyor is sooooo slooooow.

gravyboat:
Yeah they could definitely be faster but I don't really feel like spending too much time working on improving them. AppVeyor however is absolutely brutal, and a LOT of that is us just waiting in queue to get our code built. Travis has spoiled me.

beardypig:
Yeah, only a few seconds are actually spent running the tests. Most of the rest of the time is spinning up the instance and installing the prerequisites, and waiting in the queue for AppVeyor as you said. I don't even know what we could do to speed the rest of it up :(

gravyboat:
Nothing past paying as far as I can think. I'm pretty sure Travis mirrors all the Python repo stuff so it should be fast, but if we're limited resource wise there's only so fast it can go.

beardypig:
True that. The travis builds only take ~5 minutes real-time, I'm fine with that. Maybe we can down the variants in AppVeyor, remove some of the x64 ones? That would save a few minutes. We'll still get stuck in the queue though, so probably not really worth it :)

gravyboat:
Yeah I don't think it's worth it, the queue dwarfs any improvements so much that they wouldn't even be noticeable in the total run time. 4 minutes isn't really a big deal when we spend 2.5 hours in the queue. :smile:

PR #517, opened 2017-02-02
plugins.huomao: Adding plugin for Huomao

svenanderzen:
**Adding plugin to play video sources from huomao.tv / huomao.com.**

I could not find any documentation for developing plugins for Streamlink so winged it a bit (a lot) and used some of the other plugins as inspiration. Let me know if something needs to be changed. Also; Python isn't my strongest side, so let me know if something seems completely bananas.

PR #566, opened 2017-02-10
Huomao plugin tests

svenanderzen:
Adding tests to the Huomao plugin as mentioned in PR #517.

### DIFF
--- a/src/streamlink/plugins/huomao.py
+++ b/src/streamlink/plugins/huomao.py
@@ -48,44 +48,44 @@ stream_info_pattern = re.compile(r"""
     /playlist.m3u8'
 """, re.VERBOSE)
 
-# Returns the stream_id contained in the HTML.
-def get_stream_id(self, html):
-    stream_id = stream_id_pattern.search(html)
-
-    if not stream_id:
-        self.logger.error("Failed to extract stream_id.")
-
-    return stream_id.group("stream_id")
-
-# Returns a list of each stream_url, stream_quality_url and stream_quality_name
-# occurence in the JS.
-def get_stream_info(self, html):
-    stream_info = stream_info_pattern.findall(html)
-
-    if not stream_info:
-        self.logger.error("Failed to extract stream_info.")
-
-    # Rename the "" quality to "source" by transforming the tuples to a
-    # list and reassigning.
-    stream_info_list = []
-    for info in stream_info:
-        if not info[2]:
-            stream_info_list.append([info[0], info[1], "source"])
-        else:
-            stream_info_list.append(list(info))
-
-    return stream_info_list
-
 class Huomao(Plugin):
     @classmethod
     def can_handle_url(self, url):
         return url_re.match(url)
 
+    # Returns the stream_id contained in the HTML.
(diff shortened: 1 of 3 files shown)


beardypig:
Looks good, well done for adding some tests :)

You should move both the `get_stream_id` and `get_stream_info` methods into the `Huomao` class, I don't see any reason for them to exist at the module level. You can call them as instance methods, for example in your tests:
```python
# this 
self.MockHuomao.stream_id(self.MockHtml)
# instead of this
get_stream_id(self.MockHuomao, self.MockHtml)
```

Note about Python style: methods and instance variable are normally named using the `lowercase_separated_by_underscores` convention. eg. `mock_huomao` and `mock_html` would be more Pythonesque :)

svenanderzen:
Absolutely! I'll change them to instance methods and correct the naming. 👍 

Really appreciate the feedback, thanks :)

svenanderzen:
I'll try to squash the changes a bit :)

src/streamlink/plugins/huomao.py
```diff
@@ -1,23 +1,23 @@
 # NOTE: Since a documented API is nowhere to be found for Huomao; this plugin
```
beardypig: Python has a block comment syntax (also it's a string)
```python
"""
A long,
   multi-line
comment
"""
```
If they are at the top of a module, they because the `docstring` for module/method (see [PEP-257](https://www.python.org/dev/peps/pep-0257/) for more info on the subject).

src/streamlink/plugins/huomao.py
```diff
@@ -48,48 +48,48 @@
...
 
-# Returns a list of each stream_url, stream_quality_url and stream_quality_name
-# occurence in the JS.
-def get_stream_info(self, html):
-    stream_info = stream_info_pattern.findall(html.text)
+        return stream_id.group("stream_id")
 
-    if not stream_info:
-        self.logger.error("Failed to extract stream_info.")
+    # Returns a list of each stream_url, stream_quality_url and stream_quality_name
```
beardypig: This comment could also be a `docstring`.
```python
def get_stream_info(self, html):
    """
    Returns a list of each stream_url, stream_quality_url and stream_quality_name 
    occurrence in the JS.
    """
```

src/streamlink/plugins/huomao.py
```diff
@@ -48,48 +48,48 @@
...
-    stream_id = stream_id_pattern.search(html)
-    
-    if not stream_id:
-        self.logger.error("Failed to extract stream_id.")
+class Huomao(Plugin):
+    @classmethod
+    def can_handle_url(self, url):
+        return url_re.match(url)
+
+    # Returns the stream_id contained in the HTML.
```
beardypig: ditto this one.

beardypig:
Some more feedback, just comment style ;) 

Don't worry too much about squashing them, we'll do a squash merge at the end anyway.

beardypig:
Looking good to me :)

gravyboat:
Thanks for the tests @svenanderzen!

Issue #567, opened 2017-02-10
Localization tests fail


mmetak:
### Checklist

- [x] This is a bug report.

### Environment details (operating system, python version, etc.)

Archlinux, python-3.6, streamlink latest git

### Comments, logs, screenshots, etc.
http://pastebin.com/hEdvRRxJ

beardypig:
Are you running this via some kind of build script? Can you link to it, or post it in a gist? The tests actually require that you have `pycountry` installed as well.

mmetak:
Standard build `makepkg -s` in archlinux. I maintain both streamlink packages in AUR and I've update the latest to `0.3.2` which builds fine and this test fail appears later when testing latest git. I already added `python-pycountry` to `checkdepends` as soon as it was merged. ;-)

[Here is also a build log in my repo.](https://build.opensuse.org/build/home:metakcahura/Arch_Extra/x86_64/streamlink-git/_log)

beardypig:
I take it that means you don't have the `iso639` and `iso3166` packages? They are also required for the tests...

mmetak:
I've added `python-iso639` and `python-iso3166` as dependencies, made new packages for them as they weren't packaged in aur and I've also added `python-pycountry` to `checkdepends` in pkgbuild file. Long story short, they're all in there. ;-)

gravyboat:
Awesome @mmetak, thanks so much!

beardypig:
@mmetak If you'd rather not have those 2 extra packages I can make it skip the tests depending on which packages are available. Would be cleaner for the arch package without the extra deps, I guess?

beardypig:
Does it still not build with the iso modules installed? If not I have updated the tests in a PR so they get skipped if those modules are not there. 

Did you get the right iso639 module? It's actually `iso-639` in PyPi... Confusing I know.

mmetak:
@beardypig Those two extra packages are not a problem at all especially for `-git` package I maintain. You write tests and it would be a waste if I didn't enable them to check for issues and report them. ;-) So, I'll leave them all for both sets of tests.

It appears your merged tests work and there's no error now. If I try to build it  without `python-pycountry` it just skips those tests.

beardypig:
OK cool. The tests before assumed that you always have `iso639/3166` available. Which gives misleading errors if you don't, because the `Localization` falls back to `pycountry`, I've fixed this now in that PR.

beardypig:
@mmetak can you share the link to the latest build log?

[...]

PR #569, opened 2017-02-10
tests: merged the Localization tests back in to one module

beardypig:
Depending on which modules are installed iso639+iso3166 or pycountry, different sets of tests will run. If all 3 packages are available, both sets of tests will be run. 

See #567 :)


beardypig:
This is mainly for our downstream buddies if they want to run the tests with the iso639/3166 modules.

gravyboat:
Thanks @beardypig!

PR #570, opened 2017-02-10
Remove deprecation warning for invalid escape sequences

beardypig:
A couple more not-raw string escape sequence deprecation warnings removed :)


gravyboat:
Good stuff, thanks.

PR #578, opened 2017-02-13
Add a donation page

### LINKED ISSUE #347: Should we set up methods by which people can donate?
I've modified this based on some discussion below. Should we set up either a donation page with individuals and their donation information, or set up something like bountysource as mentioned by @bastimeyer?

Edit: I should note I would like to run team donations, but pretty much every service requires PayPal which I personally am very against using since they're a scummy organization. The other option is to do like @bastimeyer does where the donation links are available on the releases page (or maybe we could just make another donation page with links for people since there are multiple people working on this and it might become cluttered). The main issue I see with this also is how you determine who gets paid what amount, as well as transfers to other countries. The more I think about it the more a donation page makes sense.

gravyboat:
This is an initial draft for what a possible donation page would look like. This fixes #347. I believe that this PR addresses https://github.com/streamlink/streamlink/issues/347#issuecomment-270068127 as it prominently features bountysource in addition to direct donation methods. At this time I've only added myself and @bastimeyer as the donation details were easy to find.

Edit: It would probably good to add a donation page link to the releases as well.

### DIFF
--- a/docs/donate.rst
+++ b/docs/donate.rst
@@ -1,0 +1,30 @@
+Donating
+--------
+
+Thank you for considering a donation to the Streamlink team! Streamlink is an
+organization composed of a geographically diverse group of individuals. As such
+please make your donation directly to the appropriate team member. Note that
+donations are not tax deductible as Streamlink does not operate as a charitable
+organization. Your donation to a team member may be used in any way and does
+not come with expectations of work to be provided or as payment for future
+work. If you would like a specific feature implemented considering creating a
+bounty on `Bountysource <https://www.bountysource.com/teams/streamlink>`_.
+
+---------------
+Streamlink Team
+---------------
+
+`Bastimeyer <https://github.com/bastimeyer>`_
+---------------------------------------------
+
+- `Bitcoin <https://blockchain.info/qr?data=1EZg8eBz4RdPb8pEzYD9JEzr9Fyitzj8j8>`_ (`1EZg8eBz4RdPb8pEzYD9JEzr9Fyitzj8j8`)
+- `Dogecoin <https://blockchain.info/qr?data=DGkVP7AMUHr2TbVnWy6QyZqSMvCkeEo1ab>`_ (`DGkVP7AMUHr2TbVnWy6QyZqSMvCkeEo1ab`)
+- `Flattr <https://flattr.com/thing/3956088>`_
+- `Paypal <https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=YUCGRLVALHS8C&item_name=Streamlink%20Twitch%20GUI>`_
+
+`Gravyboat <https://github.com/gravyboat`_
+------------------------------------------
+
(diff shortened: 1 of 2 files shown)


beardypig:
My bitcoin address: [1Ey3KZ9SwTj6QyASE6vgJVebUiJsW1HuRh](https://blockchain.info/qr?data=1Ey3KZ9SwTj6QyASE6vgJVebUiJsW1HuRh) :)

bastimeyer:
Thanks for including me here.

I'm not sure if this data belongs to the git repo, though. This isn't data that is required like copyright stuff, but instead is something additional and optional. I think that this doesn't belong to the git history of such a project. That's why I'm also not including any of this in the Twitch GUI git repo and read donation data during the build process from TravisCI's env vars and then write it to the release pages.

I'd suggest adding this to the repo of the website instead, where it's fine to have this included. The docs build process doesn't interfere with this, does it? The main target should be the website anyway, I think.

beardypig:
This repo contains the source for the website/docs, the docs repo only contains the built docs. Poor docs repo is a fake...

gravyboat:
@bastimeyer As Beardypig said the docs get built out of this repo which is the only reason I added it here. Do you just have concerns over the bitcoin transaction addresses being included?

bastimeyer:
My only concerns are that it's becoming part of the history of the git repo, but if the docs build script overwrites custom changes made to the website repo, then I guess there's no choice.

gravyboat:
@bastimeyer Is the main concern for you that the transaction ID for bitcoin wallets would exist in the repo? I wasn't too worried about it since it's a unique transaction ID. If you have ideas or suggestions for what I could modify to make it more acceptable in terms of the history I'm open to discussing them. I thought about adding the donate content just to the releases page, but doing that would make the releases huge if we included everyone.

bastimeyer:
Let's keep it simple. It's okay including it here, I guess. I don't have any better ideas organizing this for multiple people.

cdrage:
Does anyone know of a service / intermediary that's able to evenly distribute donations to multiple people as a percentage? Just curious. I'm not against putting up a donation page as @bastimeyer and @gravyboat have done a TON of work, but having those donations go evenly between the two people would be beneficial.

gravyboat:
@cdrage I wouldn't advocate for even splits as I think others have done more work than I have (especially @Beardypig). I only used myself, bastimeyer, and beardypig in the example for what the format could possibly look like and it was details I had on hand. Preferably everyone in the project would have an entry on this page which they would add themselves. Right now the best option in that space that I'm aware of is https://gratipay.com/ as you can pick your split over there. Unfortunately Gratipay still only supports PayPal for their payouts, though they want to support more (https://github.com/gratipay/gratipay.com/issues/3671). I've spoken to @whit537 on this topic multiple times and it's an annoying difficult problem to solve. If people have PayPal accounts and want to set up gratipay I would be in support of that and adding a link to the donations page near the bountysource link with those details.

My intention with this PR was never to help determine a split or anything else, but to just get a donation page going so if people want to give back we make it as easy as possible to do so.

gravyboat:
Does anyone else have other discussion here? If not can we get other people to add their details or merge this? @cdrage I wanted to look at adding this to the release page. Looking at the release script right now it looks like you're manually adding it here: https://github.com/streamlink/streamlink/blob/master/script/release.sh#L86 is that still the case?

beardypig:
I'm fine with this. If anyone wants to add their own details we can do that via another PR :-)

PR #585, opened 2017-02-15
setup: choose pycountry as a dependency using an environment variable

beardypig:
See discussion on #567. Paging @mmetak.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,8 +34,11 @@ if version_info[0] == 2:
     deps.append("backports.shutil_get_terminal_size")
 
 # for localization
-deps.append("iso-639")
-deps.append("iso3166")
+if environ.get("STREAMLINK_USE_PYCOUNTRY"):
+    deps.append("pycountry")
+else:
+    deps.append("iso-639")
+    deps.append("iso3166")
 
 # When we build an egg for the Win32 bootstrap we don't want dependency
 # information built into it.


beardypig:
The question is what to name the environment variable that sets `pycountry` vs `iso639+iso3166`.

gravyboat:
I would prefer we make it unique to Streamlink. In the event there is an override by another service I don't want us to be blamed.

beardypig:
What do you mean 'override by another service' @gravyboat?

gravyboat:
In the event someone else uses `USE_PYCOUNTRY` as an environment variable there could be a conflict. Extremely small chance, but it's an edge case we can avoid.

mmetak:
@beardypig Sorry for late reply. I thought about it and you shouldn't trouble yourself with this too much. Distro maintainers can simply comment those two `iso` lines in setup.py and add `pycountry` instead if they wish when building with setuptools which will add `pcountry` to required dependencies instead of those two `iso`. Maybe just a note in docs or something would suffice to make it clear that it's an option, either `pycountry` or two `iso`.

```
# for localization
#deps.append("iso-639")
#deps.append("iso3166")
deps.append("pycountry")
```

beardypig:
@gravyboat, yeah I guess.. `STREAMLINK_USE_PYCOUNTRY` or `SL_USE_PYCOUNTRY` what do you think? 

@mmetak kind of a pain to patch the `setup.py` though, much easier if you can set a variable and have it just work.

gravyboat:
@beardypig I prefer `STREAMLINK_USE_PYCOUNTRY` just so people explicitly know what it is, but I'm not particularly passionate so pick whatever you like the best.

beardypig:
`STREAMLINK_USE_PYCOUNTRY` is clearly the best option :)

gravyboat:
Alright I'm going to merge this in then. Thanks @beardypig.

Techwolf:
Even when I use STREAMLINK_USE_PYCOUNTRY to build and install, running it results in this:

```
streamlink --twitch-oauth-token <redacted> www.twitch.tv/<redacted> mobile -p mpv -v
  File "/usr/lib/python-exec/python3.4/streamlink", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib64/python3.4/site-packages/pkg_resources/__init__.py", line 3030, in <module>
    @_call_aside
  File "/usr/lib64/python3.4/site-packages/pkg_resources/__init__.py", line 3014, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib64/python3.4/site-packages/pkg_resources/__init__.py", line 3043, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib64/python3.4/site-packages/pkg_resources/__init__.py", line 654, in _build_master
    ws.require(__requires__)
  File "/usr/lib64/python3.4/site-packages/pkg_resources/__init__.py", line 968, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib64/python3.4/site-packages/pkg_resources/__init__.py", line 854, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'iso3166' distribution was not found and is required by streamlink
```
I get the same error using python 2.7. The distro I am using only supplies pycountry.

mmetak:
@Techwolf After this commit you should be able to put `export STREAMLINK_USE_PYCOUNTRY=1` line before the python build in distro pkgbuild/ebuild/spec... etc. file and it will add pycountry to requirements. I tried it and it works on arch.

PR #591, opened 2017-02-16
plugins.tvrby: support for live streams of Belarus national TV

beardypig:
Add support for streaming live TV from the Belarus National TV channels.
Most channel provide one HLS stream and one RTMP stream, both are named "live" as they are equivalent and this respects the user's `stream-priority` setting when requesting "best".

Supported URLs:
- tvr.by/televidenie/belarus-1/
- tvr.by/televidenie/belarus-2/
- tvr.by/televidenie/belarus-3/
- tvr.by/televidenie/belarus-5/
- tvr.by/televidenie/belarus-24/

*Note: VLC3-beta does not currently play video for the `rtmp` streams, only audio.*


karlo2105:
Good it's working fine. ;)
When itvplayer used rtmp protocol, their plugin worked with http proxy which was odd as rtmpdump only supports Socks.

beardypig:
Probably the `itvplayer` rtmp streams are not geo-blocked, only the API to get the `rtmp` stream URLs ;)

beardypig:
I think this is OK to merge, @karlo2105 tested it out :)

gravyboat:
Thanks @beardypig!

PR #597, opened 2017-02-16
cli.utils.http_server: avoid "Address already in use" with --player-external-http

### LINKED ISSUE #595: Address already in use
Is there a way to unbind the port once streamlink is stopped? It takes sometimes up to a minute before the port can be used by some other process.

beardypig:
When using `--player-external-http` if streamlink is closed and restarted a `Address already in use` error can be encountered. This is because the socket is still in the `TIME_WAIT` state (more information can be found here: http://hea-www.harvard.edu/~fine/Tech/addrinuse.html).
The solution is to set the socket option `SO_REUSEADDR` so that the address can be reused before the end of the `TIME_WAIT` state. This does **not** mean that multiple streamlink instance can listen on the same port.

This should fix #595 :)

### DIFF
--- a/src/streamlink_cli/utils/http_server.py
+++ b/src/streamlink_cli/utils/http_server.py
@@ -23,6 +23,7 @@ class HTTPRequest(BaseHTTPRequestHandler):
 class HTTPServer(object):
     def __init__(self):
         self.socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         self.conn = self.host = self.port = None
         self.bound = False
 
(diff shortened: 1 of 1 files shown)


gravyboat:
Good stuff, thanks for the quick fix @beardypig!

digitalcelery:
Where would this SO_REUSEADDR option be set?

bastimeyer:
@digitalcelery 
You don't have to set any options. This change was merged in this PR:
https://github.com/streamlink/streamlink/commit/7525a8fa9722133abc2377bdfdf32730cc87332c#diff-c603ce5c9f6ecc47485374e532b7b870R26

Just try running the latest version from the master branch.

digitalcelery:
Installed from master, do not see the changes unfortunately in src/streamlink_cli/utils/http_server.py

PR #598, opened 2017-02-17
plugins.ovvatv: fix b64decoding bug

beardypig:
Affects Python3.3-3.5 where `b64decode` returns a `bytes`, but `json.loads` expects a `str`.

Mea culpa, I didn't test it with Python 3.5...


karlo2105:
OK ;)

gravyboat:

PR #601, opened 2017-02-17
plugin.mitele: use the default plugin cache

beardypig:
A specific mitele cache file was being generated, which is not required as the `Plugin` class already provides a `Cache` instance to use.


gravyboat:
Good stuff, thank you.

PR #602, opened 2017-02-17
plugins.seetv: add support for seetv.tv live streams

beardypig:
Only the streams that are provided directly from seetv.tv will work, embedded (`<iframe />` or `<object />`) streams are not supported.

Example URLs:
- seetv.tv/vse-tv-online/stb-ch-ua
- seetv.tv/vse-tv-online/tv-inter-ua


karlo2105:
Direct streams are working ok.

gravyboat:
Thanks @beardypig!

PR #623, opened 2017-02-24
Window Titles = Stream Titles

Hubcapp:
This pull request is based on https://github.com/chrippa/livestreamer/pull/1167/

I reworked it to only use the first commit though, so that instead of being restricted to just VLC, it allows the user to specify whatever command line argument they need for their special player.

example usage:

```
streamlink https://www.twitch.tv/coolstreamer source -p 'vlc --meta-title="{title}"'

streamlink https://www.twitch.tv/coolstreamer source -p 'mpv --title="{title}"'
```

In this pull request, I've only implemented Twitch support. Other sources we'll try to give them just the channel name as the title, and if we can't do that, we just give them back the url they entered as the title.

There is a mild security concern in this pull request. Because we are putting the stream title into subprocess.Popen, if this were implemented in the worst way possible, a streamer could name their stream e.g. `";echo "hacked` and get arbitrary code execution. Because we don't have shell=True set in subprocess.Popen anywhere and I've replaced ", it should be fine. I'd like confirmation anyway.

It would also be nice to replace the title by default instead of requiring the `--player` argument, but this pull request doesn't do that.


Hubcapp:
I guess this should probably be documented somewhere if it makes it in.

beardypig:
I propose an alternative solution: 35eaccca8cc69b07bce43e4a18d127c1b29b6f4a 

I had added some tests for posix and windows shell argument parsing too.

beardypig:
For posix the python `shlex_quote` method is suitable as it surrounds everything with single quotes, which takes everything as literal. For Windows, it may be more complex... 

I will try to come up with a varied and complete set of tests...

gravyboat:
@beardypig You and @bastimeyer said everything I would have mentioned after reviewing this. I'm not really sure the extra work is worth it but if you guys want to look into it go for it.

beardypig:
I did take a look... after that I have decided it's more complex. As @bastimeyer has mentioned it's not such a simple thing once you start getting in to it.

Hubcapp:
Here's some screenshots of the title displayed from the latest commit

Actual title on twitch:
https://i.imgur.com/JjTcZsc.png

On linux:
https://i.imgur.com/UKM4NXp.png

On windows:
http://i.imgur.com/c4R8BFO.png

On Windows, the unicode characters are messed up, but this is because Windows does not support every unicode character on the command line all the time. It might work if your language settings have the unicode character needed in them. There are some work arounds for Windows users here: https://stackoverflow.com/questions/388490/unicode-characters-in-windows-command-line-how

thinkpad4:
Will this be able to be disabled in the config file?  I don't want titles shown on VLC

Hubcapp:
The only way to use it right now is like
`streamlink https://www.twitch.tv/coolstreamer source -p mpv -a '--title={title} {filename}'`

If you don't go through the trouble of explicitly naming it, it won't work.

src/streamlink/plugin/plugin.py
```diff
@@ -323,5 +323,10 @@ def get_streams(self, *args, **kwargs):
     def _get_streams(self):
         raise NotImplementedError
 
+    def stream_title(self):
```
beardypig: Dunno why you need two methods, `stream_title` seems sufficient :-)

src/streamlink_cli/compat.py
```diff
@@ -33,9 +33,13 @@ def shlex_quote(s):
     if _find_unsafe(s) is None:
         return s
 
-    # use single quotes, and put single quotes into double quotes
-    # the string $'b is then quoted as '$'"'"'b'
-    return "'" + s.replace("'", "'\"'\"'") + "'"
+    if not is_win32:
```
beardypig: Can revert this as it's not being used any more.

tests/test_cli_shlex.py
```diff
@@ -0,0 +1,14 @@
+import pytest
```
beardypig: Can drop this file...

src/streamlink/plugin/plugin.py
Hubcapp: I think this is to deal with plugins that don't have a _get_title method. I just copied it from @chrippa's commit. Could try it without it and see if it works

Hubcapp: Changing 
```    def stream_title(self):
        return self._get_title()```
to 
```    def _get_title(self):
        return self._get_title()```

Works fine if the plugin has a _get_title() method. Else `RuntimeError: maximum recursion depth exceeded`

**Comment by Hubcapp** [comment]
Done working on this pending further feedback.

**Comment by Hubcapp** [comment]
Dear @beardypig @bastimeyer and anyone else,

Hey guys I'd really like to get this finished and merged in, so let me know if there's any more work to be done on it. 

Thanks, Hubcapp

File: src/streamlink/plugin/plugin.py (hunk 1 above)
Comment (wlerin) [inline src/streamlink/plugin/plugin.py]: The relation between `stream_title` and `_get_title` seems to be basically the same as `streams`|`get_streams` and `_get_streams`: `get_streams` is deprecated in favour of `streams` but all plugins still use `_get_streams`. Personally, I like the way this decouples what plugins do from what the public Plugin API does.

Either way, having `_get_title` call itself is obviously undesirable. You could simply remove `stream_title` and leave `_get_title` as is, and it would have the same effect without the recursion. Or perhaps:
```python
    def stream_title(self):
        return self._get_title()

    def _get_title(self):
        return self.url```
The help description says it already defaults to the url, but I don't see where that happens (in fact, it looks like the title is left as "Untitled").

Comment (Hubcapp) [inline src/streamlink/plugin/plugin.py]: Yeah, that seems like a good idea. I really have no problem with it as is, but you're right that it does get changed from None to "Untitled" right now if the plugin has no _get_title function

**Comment by bastimeyer** [comment]
I've just tested this on Arch, but it looks like the `title` variable can only be used as a single part of the player parameters. `streamlink -p mpv -a "--title=\"{title}\" {filename}" URL best` does not work correctly.

Just an example:
If you run this test (title is still `Test "Title"`):
```python
self._test_args(["streamlink", "-p", "/usr/bin/vlc",
                 "-a", "--input-title-format=\"foo {title} bar\" {filename}",
                 "http://test.se", "test"],
                ["/usr/bin/vlc", "--input-title-format=\"foo Test \\\"Title\\\" bar\"", "-"])
```
it'll split the parameter, uses a different title and fails:
```
AssertionError: Expected call: Popen(['/usr/bin/vlc', '--input-title-format="foo Test \\"Title\\" bar"', '-'], bufsize=<ANY>, stderr=<ANY>, stdin=<ANY>, stdout=<ANY>)
Actual call: Popen(['/usr/bin/vlc', "--input-title-format=foo 'Test Title' bar", '-'], bufsize=0, stderr=<_io.TextIOWrapper name='/dev/null' mode='w' encoding='UTF-8'>, stdin=-1, stdout=<_io.TextIOWrapper name='/dev/null' mode='w' encoding='UTF-8'>)
```

Btw, the mode of `src/streamlink_cli/output.py` has been changed from 100644 to 100755. Why?

A tiny suggestion, but maybe just a preference: since we only have one title variable (instead of multiple for various stream properties), I think it would be better for Twitch streams to use the `channel - status` format instead of `status - channel`, so that the channel name comes before the status and player windows can be identified a bit better. The Twitch GUI uses the "channel - game - status" format by default, which are the most prominent stream properties.

Hubcapp:
Okay. I switched around the channel name and the stream title in {title} (seemed like a good preference), changed the mode back to 644, and stripped the outer quotes so that you can put other words around {title} without getting the single quotes surrounding the {title} part. consequently, you'll have to do `streamlink https://www.twitch.tv/coolstreamer source -p mpv -a '--title="{title}" {filename}'` instead of `streamlink https://www.twitch.tv/coolstreamer source -p mpv -a '--title={title} {filename}'` but it's better this way.

Hubcapp:
Haven't tested on windows, they probably need the quotes stripped off too.

Hubcapp:
Can't figure out how to properly escape a space in the title in windows powershell.

`streamlink -p mpv -a '--title="TEST\ {title}" {filename}' https://www.twitch.tv/coolstreamer source` doesn't work, but 

`streamlink -p mpv -a '--title="TEST_{title}" {filename}' https://www.twitch.tv/coolstreamer source` works fine without any kind of extra quotes in the title

The internet says I should be able to escape the space with a carat mark or a backslash, but I can't get either of those to work. I'm going to forget about it, because there is no issue with subprocess.list2cmdline adding extra quotes as far as I can tell.

src/streamlink/plugins/twitch.py
```diff
@@ -599,5 +601,13 @@ def _get_streams(self):
         else:
             return self._get_hls_streams("live")
 
+    def _get_title(self):
+        if self.title is None:
+            info = self.api.channel_info(self.channel_id)
+            if is_py2:
+                self.title = self.channel + " - " + unicode(info["status"])
```
beardypig:
```python
u"{0} - {1}".format(self.channel, info["status"])
```

Should have the safe result in2 and 3.

Hubcapp: ok

Hubcapp:
I don't know how to write the Windows tests. Someone else can do it, or they can just be removed.

src/streamlink/plugin/plugin.py
beardypig: The default should be handled in one place only, if it's the `Plugin` base class it might be better do something like this: 
```python
def stream_title(self):
    title = self._get_title()
    if title is not None:
        return title
    else:
        return self.url

def _get_title(self):
    return None
```
And the `DEFAULT_PLAYER_ARGUMENTS`/`create_format_args` stuff should be refactored. 

side note: to clarify; the `streams` method is the public method for getting the streams, it normalises the result of `_get_streams`. `Plugin` subclasses _must_ implement `_get_streams`.

src/streamlink_cli/main.py
```diff
@@ -240,7 +244,13 @@ def open_stream(stream):
     return stream_fd, prebuffer
 
 
-def output_stream(stream):
+def create_format_args(stream=None, plugin=None):
+    return {
+        "title": plugin.stream_title()
```
beardypig: You probably want to ensure that plugin is not None before calling this...

wlerin:
Windows doesn't use single quotes on the command line, so those last two Actual Calls are invalid. It's not the tests.

Hubcapp:
They use it in powershell.

wlerin:
Are those tests being run in Powershell though? And does [list2cmdline](https://github.com/python/cpython/blob/6f0eb93183519024cb360162bdd81b9faec97ba6/Lib/subprocess.py#L422) care?

Hubcapp:
I don't know or care what the tests do. I've run it on Windows and described the behavior above. It's a dead feature anyway. Who would use this on Windows when there's a gui...?

wlerin:
When `PlayerOutput._create_arguments` calls:
```python
title = subprocess.list2cmdline([self.format_args["title"]])
```
`list2cmdline` places double quotes around the returned title. *However*, not only is it not quoting the whole title, in the failing examples there are also extra double quotes, so most of the title gets unquoted. For example, if you run it with:
```streamlink -p mpv -a "--title=\"TEST {title}\" {filename}" https://www.twitch.tv/esl_sc2```
the call arguments get mangled to:
```mpv --title="TEST "esl_sc2 - LIVE: GuMiho vs. jjakji [TvT] - Bracket 3 - IEM Katowice 2017 - Day 2" "<video url>"```

**Comment by Hubcapp** [comment]
I'll leave it up to Windows users to understand how to escape that string properly in their broken terminals. list2cmdline is only used to ensure that there isn't a security concern in accepting arbitrary titles from streaming services. Working command line examples above, but couldn't figure out how to put a space character in an augmented title.

**Comment by wlerin** [comment]
Your code as written *cannot be escaped properly*. It's not a matter of Windows users figuring out how.

**Comment by Hubcapp** [comment]
Well then it _cannot support_ arbitrary titles. Someone else can work on it if they think they can figure it out. I've got better things to do than write a dead feature for a dead os.

**Comment by wlerin** [comment]
So I've looked at several alternatives, and the simplest solution is to do exactly what you're already doing for Unix, and strip the outer quotes. In fact since this needs to be done for all cases, you could do it here instead:
```python
title = title.replace("$","$$")[1:-1]
```

Hubcapp:
Turns out the existing implementation for fixing altered titles with {title} was mildly broken even on unix. We can't just get rid of the surrounding quotes and do @bastimeyer's command `streamlink -p mpv -a "--title=\"{title}\" {filename}" URL best` because we replaced single quotes with double quotes. You'd have to glue the strings together like `streamlink -p mpv -a '--title='"'"'{title}'"'"' {filename}' URL best` which is not very desirable.

and as @wlerin points out, it's not very possible to escape the string in Windows at all.

So, this commit takes the entire title and puts it into shlex_quote/list2cmdline instead of just the _get_title() part. Unfortunately, I've had to introduce some unique syntax in order to identify the start and end of the user defined title, since media players have different ways of parsing a user defined title from the command line.

-----

Now titles can be used in two ways:

`streamlink https://www.twitch.tv/coolstreamer source -p mpv -a '--title={title} {filename}'` if you're just passing _get_title() along

`streamlink https://www.twitch.tv/coolstreamer source -p mpv -a '--title=@@foo {title} bar@@ {filename}'` if you'd like to include {title} in some longer custom title.

Hubcapp:
This is all just to get rid of a couple quotes in the title. You can still do 

`streamlink https://www.twitch.tv/coolstreamer source -p mpv -a '--title="foo {title} bar" {filename}'

on posix if you don't mind seeing single quotes or

`streamlink https://www.twitch.tv/coolstreamer source -p mpv -a '--title="foo_{title}_bar" {filename}'

in the windows powershell if you don't mind using underscores instead of spaces.

These are the more natural solutions I think users will find if @@ wrapping isn't documented.

src/streamlink_cli/utils/player.py
```diff
@@ -36,3 +38,20 @@ def find_default_player():
...
+        if is_py2:
+            title = shlex_quote(wholeTitle.encode('utf8'))
+        elif is_py3:
+            title = shlex_quote(wholeTitle)
+    else:
+        if is_py2:
+            title = subprocess.list2cmdline([wholeTitle.encode('utf8')])
+        elif is_py3:
+            title = subprocess.list2cmdline([wholeTitle])
+    title = title.replace("$","$$")
```
beardypig: What's this replacement for?

Hubcapp: Don't know which replacement you're referring to. I factored sanitizeTitle() out of _create_arguments() because we use it twice in _create_arguments(). We use it twice in _create_arguments so that the user isn't forced to use @@ wrapping.

beardypig: This replacement, `title.replace("$","$$")`.

Hubcapp: It just turns out to be necessary on both Windows and Posix. I found out that if you have 2 $ in a row, it replaces it with 1 $, since $ is a special character.

beardypig: Weird. Can you include some tests to verify this behaviour?

Hubcapp: Okay.

Hubcapp:
The test I added for checking $$$$$$$$$$$$$ fails, but testing for real I'm pretty sure you need the replacement.

Hubcapp:
I found out why `title.replace("$","$$")` is necessary. @beardypig 

It turns out, it's not. In the two media players I tested, (vlc and mpv) they both happen to use $ as a character to denote the start of some replacement in the title.

See https://mpv.io/manual/stable/#property-expansion for mpv and https://wiki.videolan.org/Documentation:Format_String/ for vlc. The fact that both of them use "$$" to mean "$" is a coincidence.

For us to handle this, we would have to write exceptions for each player that streamlink is meant to support. Since streamlink is meant to support an indefinite amount of players, it's not possible to handle this.

So there's two options that come to mind if StreamLink is to support this feature:

1. Ignore Property Expansion / Format String problems and just let whatever player specific format codes that happen to be in streamer's titles get passed to the player unhandled.

Pros: 
- Streamlink can claim to be player agnostic still
- Don't have to support an indefinite amount of players
- Users who expect Property Expansions / Format Strings to work will still have it work (we aren't escaping their intentional codes)

Cons:
- The title as displayed by various streaming sites will not precisely match the window title as intended in rare circumstances where the stream title happens to contain special player specific format codes

2. Escape the various player codes in {title} for the most popular media players

Pros:
- {title} will be formatted in such a way that it is passed to the media player to be displayed exactly how it appears on the actual streaming site

Cons:
- Streamlink will only support this feature "properly" for a limited number of media players
- Tests should probably be written for this, which may be difficult since we can't get the real window title properly without actually running the program and doing something like "xprop -title [expected title]" in linux and something with win32gui in windows. I'd need help writing these tests, since I'm not familiar enough with python-mock to do so, and don't want to learn it for this feature.
- If not written properly, it could interfere with what the user wants to do. e.g. if we simply do something like `if player is 'mpv': title = "?>" + title` to disable all Property Expansion, the user will get mad when they actually try to use Property Expansion

Hubcapp:
@bastimeyer This issue of "Property Expansions / Format Strings" is a problem in streamlink-twitch-gui. It should be easier for you to solve it since you have only 4 media players that you set presets for (I think only mpv and vlc have the "Player window title" option...)

Hubcapp:
Decided to just go with option 1. If y'all have an opinion on that, let me know...

Meanwhile, all tests should pass now I think. Travis build passed 1 commit ago, and nothing changed that should affect python 3.5, I think Travis just goofed.

gravyboat:
@Hubcapp Sorry for the delay on getting back to you with this, some of the reviewers who originally looked at this are currently out sick/on vacation. Hang tight!

gravyboat:
@beardypig and @bastimeyer Can you re-review this one when you get a chance?

bastimeyer:
Sorry, I forgot about this PR after I had seen your message 11 days ago. 😨 

To be honest, I still have the same concerns I've mentioned in my review summary.

I'm not really a fan of introducing custom syntax for wrapping custom title strings. This is unexpected and can lead to some errors, eg. if people prefer having titles that already include double @ chars. With this implementation you're basically assuming that this string will never be used and is save to be used as a boundary between the player args and custom title.

Regarding the player specific title variables, I think that it's okay to ignore it. At least MPV can be configured to ignore all vars by prepending `$>`. If someone thinks that dollar signs should be escaped for each player separately, they are welcome to add another PR for this.

I don't want to block any features here, so if the rest of you think that it's okay, just ignore me. 😅

gravyboat:
@bastimeyer I agree with you. My main concern is the number of potential edge cases we might not catch that could negatively impact end users. There have already been several of them brought up in your comment and others.

Hubcapp:
I don't like that the user would have to resort to this custom syntax either, but I don't know how to make it better. The only alternative I saw was "not supporting augmented titles with {title} in them" or users accepting underscores & quotes like in https://github.com/streamlink/streamlink/pull/623#issuecomment-283098544

I guess another option would be only supporting VLC and MPV.

I thought I covered @bastimeyer's review with shlex_quote/list2cmd.

btw, the regex that does the @@ wrapping grabs the last @@ it finds, so if there is "@@" in the title, it will be okay. I think the only assumption is that the user won't need to put "@@" in any other parameters.

gravyboat:
@Hubcapp Yeah I don't think anyone is blaming you for this if it came off that way, it's just one of those problems where the streamer can really screw things up.

Hubcapp:
idk, I've been using this build since I first made the pull request and it works. I did a lot of testing with all kinds of funky titles, and even supported unicode. I'm really not sure what edge cases are left now. My opinion is whatever rare edge case there might be left can be handled as a bug fix. Obviously introducing bugs is not desirable, but unless someone can find a problem with this implementation, I can't fix it.

gravyboat:
Any further feedback @bastimeyer and @beardypig? Do we think there are too many edge cases to make this viable?

bastimeyer:
As I've already said, I really don't like the double @ sign custom syntax.
This simply breaks stuff when this string is included somewhere else in the player args, intentionally or not. It is especially problematic for automatically assigned parameters of third party applications, like the Streamlink Twitch GUI for example.

When these Streamlink parameter are being used
`streamlink -p mpv -a "--title \"@@@@ foo\" {filename}" URL QUALITY`
Streamlink will try to run mpv with these parameters:
`mpv '' foo -`

Hubcapp:
If `@@` is not considered unique enough syntax that we can make the safe assumption that the user would not need to use it elsewhere, how about `@:/@` instead?

I choose this new syntax for 3 reasons
* `:` cannot be used as part of a file name in Windows
* `/` cannot be used as part of a file name in Linux/Mac
* `:/` describes the emotion of the user when forced to use such awkward syntax, so it's easy to remember.

I'm also open to any alternative solution to solve the problem `@@` was introduced to solve.

gravyboat:
@bastimeyer What do you think about the proposed changes regarding the syntax?

bastimeyer:
@gravyboat 
Changing the token to something that's *potentially* being used less often by accident or by automated stuff doesn't change anything. This is an issue that can't be solved by regular expressions and requires a real parser, and as @beardypig has already said, it's a bit more complex.

I'm sorry if a lot of time has been spent on implementing this stuff and on waiting here, but it needs a better solution that is not exploitable. 😟

gravyboat:
@bastimeyer Definitely. @Hubcapp I think we're going to have to reject this one for the time being unless a foolproof method can be determined, sorry about that.

Hubcapp:
That's okay. It's a dumb syntax that I hate too. Although I don't really see a problem because the user gets to decide this input. It's not exploitable at all by any third party. The user will have to purposely break it by doing something like @bastimeyer's example from 11 days ago. There's nothing a streamer can do within {title} to break it as long as the user uses the custom quotes the way they were designed to be used.

But like I said, I hate the custom quotes too.

So, for a rewrite that might be more acceptable, we'll forget any custom syntax, and either

1. Not support the use of {title} with any other user defined title elements ("foo {title} bar" will not be supported)

2. Support it for just a couple video players like the bastimeyer's streamlink twitch gui.

Let me know what would be considered acceptable, thanks.

Hubcapp:
Personally I think it would be best to support this feature for per video player. This would allow us to also handle the format codes ($$ -> $) per player.

There's no way to write this with all the desired features (able to augment the title) without removing ambiguity toward how the video player accepts custom titles. That's what `@:/@` was designed to do, but I understand that it is too ugly.

gravyboat:
@bastimeyer Thoughts on @Hubcapp's proposed solutions?

beardypig:
Just a few questions about supporting certain features for certain players;

1.  Which players to support?
2. How does streamlink know which player? Automatic detection, user config?

Hubcapp:
1. VLC, MPV, and MPC-HC are probably the most popular players. Any other players can be implemented as future pull requests or in this one if I forgot any major players for mac e.g.

2. Haven't looked into this yet, but I think streamlink supports launching from a player by default with user config already, which would then have to take `-p` as a higher priority. It should also be possible to just grab the current player while in the "launch player" function like `if playerCommand.startsWith("vlc ") or player == "vlc": (vlc specific code)`

bastimeyer:
@Hubcapp 

> The user will have to purposely break it by doing something

No, what I meant was that parameters which get generated automatically can break the custom syntax implementation, no matter which spearator string you're using. This has nothing to do with the substitution of the `{title}` variable.

I'm not sure if simply ignoring custom user titles is the right way. If you don't want to support it, you have to focus on a specific syntax which ignores single and double quotation marks, like you've described above (`--title {title}` and `--title={title}`). The issue with this idea is that people will use this incorrectly, because it's not intuitive.

@gravyboat 
The per-player implementation is only meant for escaping player specific variables. VLC and MPV both have their own methods here and variable names need to be escaped differently.

@beardypig 
The most popular players among all OSes are VLC, MPV and MPC-HC (Windows only). MPC-HC doesn't have any parameters for setting custom player/window titles, so it can be ignored.

To automatically detect the player being used, the only thing which can be done is having a list of predefined executable names and comparing the basename of the resolved player path with that list.
But since comparing the player executable name is not 100% reliable, you could instead implement a new parameter for automatically resolving the player path so that the player is not being detected but actively chosen. This wouldn't be backwards compatible though, since the "custom" player path (VLC) is automatically set on Windows in the user's config file.

This is the player config file of the Twitch GUI:
https://github.com/streamlink/streamlink-twitch-gui/blob/v1.2.1/src/config/players.json

Hubcapp:
>The issue with this idea is that people will use this incorrectly, because it's not intuitive.

It's already not intuitive to have to include `{filename}` in the player arguments. People already have to do research to use custom parameters with streamlink. It's okay to have unintuitive command line input.

I highly doubt anything would generate `@:/@` twice by accident. But if there was doubt of that, we could change it to like `@:streamlink/@`. The reason it's not acceptable is because it's an ugly hack.

Hubcapp:
Anyway, sorry, none of that matters since we're going ahead with supporting the feature "per player" it looks like. We'll support just VLC and MPV for now.

gravyboat:
Is this being modified to support the specific players as we discussed?

Hubcapp:
It's been an eventful couple months for me, so I haven't worked on it. If anyone else wants to, they can, otherwise I'll get to it some time soon.

gravyboat:
No rush, thanks for the update @Hubcapp!

gravyboat:
@Hubcapp Are you still planning on coming back to this?

bastimeyer:
Just close this PR, it's impossible to track any further progress.
Once all issues are solved which have been mentioned here, a new PR can be opened.

Hubcapp:
studying for cisco networking exams right now -- unfortunately this feature is complete enough for personal use, so I haven't had much motivation to fix it for inclusion in streamlink. Have been using it with no issues for over a year now. @gravyboat 

it does kind of bother me that it's been left undone for so long, but I honestly wouldn't mind if anyone else finished it for me.

Hubcapp:
I agree that a new pull request referencing this one would be a good idea if it's ever completed.

Hubcapp:
Maybe I'll work on it this week. If I haven't made a new PR this month, you can consider it safe to work on without duplicating work. I think it's a good feature that improves streamlink and should be included with every copy, so if I don't do it, someone should.

iirc, the code as it currently is needs to have the "custom quote syntax" used as a hack to circumnavigate cmd.exe's lack of good quoting needs to be erased to make way for only officially supporting 2 different video players.

I think an alias for `-a '--title={title} {filename}'` would be useful, if that flag isn't already used. And actually, @bastimeyer wanted it to support the arrangement of individual elements like author, game, and title etc. maybe `--ta` and `--at` for author first or title first. That should be a little bit more intuitive.

Would adding these flags be objectionable?
`--gta`
`--gat`
`--tga`
`--tag`
`--atg`
`--agt`
`--ta`  (to omit "game" for platforms without "game")
`--at` (to omit "game", for platforms without "game")

As long as adding so many aliases isn't objectionable, I think this would be an even nicer way to make the cli syntax for this feature intuitive and memorable, as well as allowing for the user to avoid dealing with any quoting at all on their end.

Hubcapp:
The big blocker for this feature was that the user would still need to deal with `-a 'title='"'"'my special title for some reason - {title} - {author} - {game}'"'"' {filename}'` for augmented titles that include the `{title}` replacement, and advanced quoting like that is apparently impossible in Windows even with the powershell.

I think the big mistake was trying to use the `-a` flag. If we're no longer calling it player agnostic, we're allowed to introduce per-player support and use an easy system like `--title 'my special title for some reason - {title} - {author} - {game}'`. This will avoid the need to put a quote inside what's already a literal string.

PR #611, opened 2017-02-20
streamlinkrc : Version check disabled by default

karlo2105:
I added line to disable version checking by default. This should stop streamlink ping on launch. 
[https://github.com/streamlink/streamlink/issues/590](url)

Issue #590, opened 2017-02-16
Should we disable CLI version check by default?


gravyboat:
### Checklist

- [x] This is a feature request.
- [x] I used the search function to find already opened/closed issues or pull requests.

### Description

Prior to upgrading to streamlink `0.3.2` I noticed on Windows when I was launching a stream from the CLI there's a short pause while Streamlink is checking for a new version where this message pops up:
`[cli][info] A new version of Streamlink (0.3.2) is available!`. I think we should discuss removal of this notice, or allow users to decide if they want to enable it for privacy purposes. I know it's not a huge deal but some users may not wish to ping an external service other than the site they are trying to load.

### Reproduction steps / Stream URLs to test

1. Install an older version of Streamlink.
2. Load a stream up via the CLI.

beardypig:
Would you want a flag to enable the automatic version checking? 
Currently there is `--no-version-check`, which disable the automatic checking, and `--version-check` which checks for a new version and exits. Do we add a new flag `--auto-version-check`? and always set `no_version_check=True`?

bastimeyer:
Can't `--no-version-check` be turned into a boolean parameter, so it can be turned off by setting `--no-version-check=false` or `--no-version-check false`, so that a new parameter is not required?
Then I'd suggest setting `--no-version-check` to true on *nix systems (when installed via a package manager) and adding a checkbox to the windows installer and writing it to the config file.

beardypig:
It's a bit tricky to modify `--no-version-check` as it currently expects no arguments, could cause some issues.

fishscene:
Just a thought:

- On embedded systems such as RPI, version-checking on startup isn't necessary unless something goes wrong or there's a security issue (which should probably be manually updated anyway). Overall, it just adds a delay to getting to the requested content every time it is called.
- On actively maintained systems, it's nice to know when there's an update, but if other applications depend on streamlink (such as streamlink-gui), they may not support the newest version yet. When they update, they can then run a version check on streamlink to see if a supported one is installed, if not, they can request the user for an updated version.

With this in mind,
What if we disabled the version check on startup on all platforms and if a plugin fails to tap into a stream or there's another error, it either:
A) Automatically runs a version check to see if there's an update to possibly fix the error. We can also spit out the currently installed version number to make it easier - just in case something goes wrong with the version check. Would need some kind of gated version checking so a quick succession of failures doesn't spam the update servers.
or
[...]

PR #630, opened 2017-02-27
cli: disable automatic version checking by default

beardypig:
Adds a new option `--auto-version-check=(yes|no)` and suppresses the existing option `--no-version-check`. `--auto-version-check` defaults to `no`, and can re-enabled by setting `--auto-version-check=yes`.

See discussion in #590.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -131,6 +131,14 @@ def keyvalue(value):
     return match.group("key", "value")
 
 
+def boolean(value):
+    truths = ["yes", "1", "true", "on"]
+    falses = ["no", "0", "false", "off"]
+    if value.lower() not in truths+falses:
+        raise argparse.ArgumentTypeError("{0} was not one of {{{1}}}".format(value, ', '.join(truths+falses)))
+
+    return value.lower() in truths
+
 parser = ArgumentParser(
     fromfile_prefix_chars="@",
     formatter_class=HelpFormatter,
@@ -262,9 +270,9 @@ general.add_argument(
 )
 general.add_argument(
     "--auto-version-check",
-    type=str,
-    choices=["yes", "no"],
-    default="no",
+    type=boolean,
+    metavar="{yes,true,1,on,no,false,0,off}",
+    default=False,
     help="""
     Enable or disable the automatic check for a new version of Streamlink.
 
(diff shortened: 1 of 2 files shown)


gravyboat:
Awesome thanks @beardypig. I'm going to leave this open in case anyone else wants to weigh in, if not I'll merge it.

karlo2105:
@beardypig 
Does it work like old switch in streamlinkrc config file?

beardypig:
What you mean? Any option can go in the config file. In this case to turn the automatic check back in you need to set 
```
auto-version-check=yes
```

The existing `no-version-check` option is effective ignored.

bastimeyer:
As a tiny improvement, I think the value should get implicitly set to "yes" when the `--auto-version-check` parameter is being used but without any values. This doesn't affect its default value of course. Maybe `true`, `false` and `1`, `0` could also be accepted as possible values.

I'm not sure how I feel about completely ignoring `--no-version-check`. I'm using this parameter in the Twitch GUI and it can't be disabled by the user. Adding the new `--auto-version-check` parameter and disabling the old one means that I need to bump the min required Streamlink version and change the parameters.

IMO, the `--no-version-check` should still disable the version check logic when set, regardless the value of `--auto-version-check`. Just for the sake of backwards compatibility, even though the default value of the new one is set to `no`.

beardypig:
`--no-version-check` could be used as an override and always disable the version check regardless. Having an option with an implicit value is a bit troublesome, they can lead to unexpected errors and mean that some options might now work next to other ones.
It's easy to make it accept 1, 0, true, and false as well.

src/streamlink_cli/argparser.py
```diff
@@ -261,10 +269,14 @@ def keyvalue(value):
     """
 )
 general.add_argument(
-    "--no-version-check",
-    action="store_true",
+    "--auto-version-check",
+    type=boolean,
+    metavar="{yes,true,1,on,no,false,0,off}",
+    default=False,
```
cdrage: By default should this be true? (having streamlink check the version on each usage by default)

beardypig: It's to disable that, see #590 :-)

cdrage: Ah poo. That's why. i agree about speeding it up :+1:

gravyboat:
Merging this in, thanks @beardypig, with this change we can close https://github.com/streamlink/streamlink/pull/611 I believe unless anyone has other concerns.

Issue #627, opened 2017-02-27
Plugin: Add support for showroom-live.com


wlerin:
### Checklist

- [ ] This is a bug report.
- [x] This is a plugin request.
- [ ] This is a feature request.
- [x] I used the search function to find already opened/closed issues or pull requests.

### Description

Showroom is a Japanese streaming site, used by artists, entertainers, and amateur streamers. It does not provide VOD services.

### Expected behaviour:

For this code:

    >>> import streamlink
    >>> streams = streamlink.streams("https://www.showroom-live.com/48_Tomu_Mutou")

To return a list of streams (or None if she's not streaming)

### Actual behaviour

Raises `streamlink.exceptions.NoPluginError`

### [Working Plugin](https://github.com/streamlink/streamlink/compare/master...wlerin:master?diff=unified&name=master)

I have written a Showroom plugin, and tested that it correctly returns the desired streams. Some additional testing is required before I create a pull request, but I thought I'd open an issue first and maybe get general feedback or even help resolving some of the issues below.

### API Urls
More (mostly API) urls besides those used in the proposed plugin:

- [Sample Event Room List](https://www.showroom-live.com/event/all_rooms?event_id=1364)
- [Live broadcasts w/ streaming links](https://www.showroom-live.com/api/live/onlives)
- [Upcoming lives in the Idol genre](https://www.showroom-live.com/api/live/upcoming?genre_id=102)
- [Next live for a room](https://www.showroom-live.com/api/room/next_live?room_id=61576)
- [Video Caption](https://www.showroom-live.com/api/live/telop?room_id=61627)
- [Comment Log](https://www.showroom-live.com/api/live/comment_log?room_id=61627)
- [Ranking Summary](https://www.showroom-live.com/api/live/summary_ranking?room_id=61627)

(anything with "room_id" in the url will probably not be interesting unless the room is live)

#### 1. Caption
Live rooms usually have a "telop" or caption, that displays at the top of the video and usually describes whatever's going on in the room that day. The plugin stores this, but does nothing with it. Possibly it could be muxed with the stream like a subtitle? More investigation required.

edit: It appears that I need to support both text and NoneType values for this. For the moment I'm fudging with validate.transform(validate.text).

#### 2. Stream Title
In relation to Hubcapp's title-related work, there are five options for determing a room's title:
1. The "room_url_key", taken either from _get_stream_info, or the first url pattern. This key is meaningful for official rooms, but random and meaningless for amateur rooms.
[...]

PR #633, opened 2017-02-28
plugins.showroom: add support for showroom-live.com live streams

wlerin:
Add support for streaming live broadcasts from the Japanese streaming service showroom-live.com. Showroom provides both HLS and RTMP sources, but as the HLS sources are highly unstable, this plugin only handles RTMP.

Example URLs:
- https://www.showroom-live.com/kimi_dare
- https://www.showroom-live.com/room/profile?room_id=75207

As mentioned in issue #627 

_Tested with mpv 0.23.0, vlc 3.0.0, ffplay (recent build and 2.8.10-0ubuntu0.16.04.1)_

### DIFF
--- a/src/streamlink/plugins/showroom.py
+++ b/src/streamlink/plugins/showroom.py
@@ -41,14 +41,14 @@ _api_data_schema = validate.Schema(
     }
 )
 _rtmp_quality_lookup = {
-    "オリジナル画質": "original",
-    "original spec": "original",
+    "オリジナル画質": "high",
+    "original spec": "high",
     "低画質": "low",
     "low spec": "low"
 }
 # changes here must also be updated in test_plugin_showroom
 _quality_weights = {
-    "original": 720,
+    "high": 720,
     "other": 360,
     "low": 160
 }
--- a/src/streamlink/plugins/showroom.py
+++ b/src/streamlink/plugins/showroom.py
@@ -9,7 +9,7 @@ _url_re = re.compile(r'''^https?://
         (?:\w*.)?
         showroom-live.com/
         (?:
-            (?P<room_title>\w+$)
+            (?P<room_title>[\w-]+$)
             |
             room/profile\?room_id=(?P<room_id>\d+)$
         )
--- a/src/streamlink/plugins/showroom.py
+++ b/src/streamlink/plugins/showroom.py
@@ -16,6 +16,8 @@ _url_re = re.compile(r'''^https?://
 ''', re.VERBOSE)
 
 _room_id_re = re.compile(r'"roomId":(?P<room_id>\d+),')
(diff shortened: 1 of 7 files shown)


src/streamlink/plugins/showroom.py
```diff
@@ -0,0 +1,166 @@
...
+)
+_rtmp_quality_lookup = {
+    "オリジナル画質": "original",
+    "original spec": "original",
+    "低画質": "low",
+    "low spec": "low"
+}
+# changes here must also be updated in test_plugin_showroom
+_quality_weights = {
+    "original": 720,
```
gravyboat: This is the only thing I can see where I imagine @beardypig will chime in and talk about the naming of the weights.

gravyboat:
Other than my note above this looks good, thanks for the tests!

bastimeyer:
@wlerin 
If you rebase your branch, you can remove the unnecessary commits 76c1e00 and fc6179c instead of reverting 76c1e00, which looks like a merging mistake from the master branch. Just force push it again then.

wlerin:
@bastimeyer I'm not completely certain how to do that. My original attempt at rebasing the project is what created 76c1e00 in the first place.

bastimeyer:
It's fine 😄
The commits of the PR will get squashed anyway. But yeah, this looks much cleaner.

src/streamlink/plugins/showroom.py
beardypig: I'm not sure about the names, it's tricky when they are mixed types. Ideally you want equivalent streams to have the same names so they sort correctly, and the users preferred streaming technology can be picked automatically (see [`--stream-types`](https://streamlink.github.io/cli.html#cmdoption-stream-types).

beardypig:
Solid looking plugin. I don't know if there is a way for you to give the streams names parity between hls and rtmp? 

Also, don't forget to add it to the plugins matrix in the docs :-)

wlerin:
@beardypig Without knowing how to distinguish 360p and 720p "original" quality rtmp, no. If I can find another room streaming in 720p maybe I'll be able to fix that. 

Of course it's not an issue at the moment since the plugin only returns rtmp streams.

And I'll get right on that plugin matrix.

src/streamlink/plugins/showroom.py
beardypig: I think `"high"`, `"mid"` and `"low"` might be more neutral names in this case :)

wlerin: What about `"high"`, `"low"`, and `"other"`? The third case is a catch all for any unexpected quality label the API returns.

beardypig: It's only that `"other"` will get sorted between `"high"` and `"low"` because of the stream weight. But I guess `"other"` is OK.

Issue #612, opened 2017-02-21
Plugin: rtve.es VOD


malvinas2:
### Checklist

- [ ] This is a bug report.
- [x] This is a plugin request.
- [ ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description
Would it be possible to add VOD streams for rtve.es ? 

### Reproduction steps / Stream URLs to test

short-URL: http://www.rtve.es/drmn/embed/video/{id}
short-URL: http://www.rtve.es/drmn/embed/video/3886778
long-URL:  http://www.rtve.es/alacarta/videos/cuentame-como-paso/cuentame-como-paso-t18-capitulo-313-amigo-americano/3886778/

### Comments, logs, screenshots, etc.

I don't know if it helps, but one can get information about the video using the API, described by Ulises Gascon at http://ulisesgascon.github.io/RTVE-API/
For example 
http://www.rtve.es/api/videos/{id}.json
http://www.rtve.es/api/videos/3886778.json

malvinas2:
Thank you very much for your quick response. 
I tried the example given in my first post, video-id 3886778, which according to the JSON-information should be available in three different qualities. But when I call the link within python I receive three links with the identical URL, is this correct ? 

`Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (
AMD64)] on win32`
`Type "help", "copyright", "credits" or "license" for more information.`
`>>> import streamlink`
`>>> streams = streamlink.streams("http://www.rtve.es/drmn/embed/video/3886778")`
`>>> streams['worst']`
`<HLSStream('http://hlsvod.lvlt.rtve.es/resources/TE_GL13/mp4/4/6/**1485475369464.mp4**/1485475369464-audio=64000-video=919000.m3u8')>`
`>>> streams['540p']`
`<HLSStream('http://hlsvod.lvlt.rtve.es/resources/TE_GL13/mp4/4/6/**1485475369464.mp4**/1485475369464-audio=64000-video=919000.m3u8')>`
`>>> streams['best']`
`<HLSStream('http://hlsvod.lvlt.rtve.es/resources/TE_GL13/mp4/4/6/**1485475369464.mp4**/1485475369464-audio=64000-video=919000.m3u8')>`

beardypig:
I have added the missing qualities in another PR. Now you should get all these qualities:
```python
[...]

PR #635, opened 2017-02-28
plugins.rtve: return all the available qualities

beardypig:
As pointed out in https://github.com/streamlink/streamlink/issues/612#issuecomment-282900657, some of the available lower qualities were missing. This PR adds all the available qualities from HLS and HTTP.


gravyboat:
Thanks @beardypig.

Issue #105, opened 2016-10-27
Possibility of simultaneous viewing and recording of stream?


flijloku:
It would be good to implement possibility of simultaneous viewing and recording of stream (to save bandwidth). Or this possibility already exists?

scottbernstein:
Someone already implemented this for the main livestreamer build, but of course since there have been no releases, it's still sitting as a pull request.


Perhaps someone could either integrate into streamlink, or they could contact the original coder to ask him to integrate here?


https://github.com/chrippa/livestreamer/pull/1204


[https://avatars1.githubusercontent.com/u/4093184?v=3&s=400]<https://github.com/chrippa/livestreamer/pull/1204>

Add a record option by grazzolini · Pull Request #1204 · chrippa/livestreamer<https://github.com/chrippa/livestreamer/pull/1204>
github.com
I have been using livestreamer for a while now and it is a excellent piece of software. One thing I missed on it was the ability to play a stream while at the same time recording it to a file. This...




________________________________

bastimeyer:
You can use the `-O` parameter that outputs the stream data to stdout and then pipe it into `tee`, which then writes the data into a file and pipes it further into your video player...

```bash
streamlink -O STREAM best | tee recording.ts | mpv -
```

flijloku:
@scottbernstein Thanks, I'll try!
@bastimeyer This is for linux? Is there an analogue for windows?

PR #639, opened 2017-03-01
Add a record option

grazzolini:
Add a record option for recording a stream while also playing it.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -453,6 +453,14 @@ output.add_argument(
     """
 )
 
+output.add_argument(
+    "-R", "--record",
+    action="store_true",
+    help="""
+    When using -o, enable recording the stream while also playing it.
+    """
+)
+
 stream = parser.add_argument_group("Stream options")
 stream.add_argument(
     "--default-stream",
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -57,12 +57,12 @@ def create_output():
 
     """
 
-    if args.output:
+    if args.output and not args.record:
         if args.output == "-":
             out = FileOutput(fd=stdout)
         else:
             out = check_file_output(args.output, args.force)
-    elif args.stdout:
+    elif args.stdout and not args.record:
         out = FileOutput(fd=stdout)
     else:
         http = namedpipe = None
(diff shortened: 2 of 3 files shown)


bastimeyer:
Is a new parameter really required for viewing and recording at the same time?
We had the same question a few months ago in #105. See my answer of how to record streams while watching at the same time [here](https://github.com/streamlink/streamlink/issues/105#issuecomment-256678800).

It also looks like your implementation breaks the `-O` parameter while using `-R`.

grazzolini:
I used to use tee, but tee adjusts the buffer of stdin (see stdbuf man page or tee source code). Therefore, it introduces a lot of buffering and, while the record works, it makes the player to buffer a lot and the playback cuts when this happens. With the recording happening within streamlink itself, that doesn't happen.

I'm the author of that pull request against old livestreamer and I was able to adapt it to streamlink. As for the -O option, I'll check what is happening. It worked on the livestream PR.

grazzolini:
I confirm that using -O while also using -R breaks -O behaviour. I'm not quite sure what should it do in this case, I guess that outright ignoring the -R option should be enough.

fishscene:
If an option to record and view simultaneously can be built in so that it works on all of our supported platforms, I say let's do it.

grazzolini:
@fishscene 
I haven't tested on any other platform than linux. But I don't see why this option wouldn't work on them. I still haven't had time to fix the bad interaction with -O, but I plan on doing it still this week, so this PR can be merged.

gravyboat:
Any further feedback for @grazzolini @bastimeyer? I would like to finish up discussion so we can come to a conclusion on this one.

grazzolini:
@gravyboat 
I was quite busy for the past month, and I haven't had a chance to finish this. If you guys want, close this PR. I'll resubmit it when I have a chance. I didn't even tested this patch against latest streamlink yet. But I'm still packaging it for Archlinux, don't worry.

gravyboat:
@grazzolini Okay thanks, I'm going to close it for now then just so we keep the queue cleared up, I'll try to remember to link this when you open the new one, thanks for understanding!

grazzolini:
@gravyboat 
No problem at all. I hopefully will have more time pretty soon and I'll dig again on streamlink code. And I'm keeping the package on archlinux up to date as fast as I can.

gravyboat:
Thanks for your continued support @grazzolini. Do you want me to add you to the org?

grazzolini:
@gravyboat
I don't know what you mean by that, hehehehe.

gravyboat:
@grazzolini It's just slightly elevated access and the ability to be part of the organization so you can publicly display it on your GitHub profile if you want. I sent you an invite.

Issue #640, opened 2017-03-01
rtmpdump AMF data


ppsspp213:
How do I apply the same in streamlink???
```
−−conn −C type:data
Append arbitrary AMF data to the Connect message. The type must be
B for Boolean, N for number, S for string, O for object, or Z for
null. For Booleans the data must be either 0 or 1 for FALSE or TRUE,
respectively. Likewise for Objects the data must be 0 or 1 to end or
begin an object, respectively. Data items in subobjects may be named,
by prefixing the type with 'N' and specifying the name before the value,
e.g. NB:myFlag:1. This option may be used multiple times to construct
arbitrary AMF sequences.
E.g. −C B:1 −C S:authMe −C O:1 −C NN:code:1.23 −C NS:flag:ok −C O:0
```

gravyboat:
Please review the docs: https://streamlink.github.io/cli.html#playing-built-in-streaming-protocols-directly the second example in that section shows how to pass extra options to an rtmp stream.

beardypig:
This should work:
```shell
streamlink "rtmp://streaming.server.net/playpath conn=B:1 conn=S:authMe conn=O:1 conn=NN:code:1.23 conn=NS:flag:ok conn=O:0"
```
But it will only take one value for conn, the last value. It's a simple fix, so I'll create PR to fix it... 
Meanwhile, I'll reopen this and use it to track the bug.

beardypig:
For multiple `conn` value the following syntax can be used: 
```shell
streamlink "rtmp://streaming.server.net/playpath  conn=['B:1','S:authMe','O:1','NN:code:1.23','NS:flag:ok','O:0']"
```

ppsspp213:
'conn' is not recognized as an internal or external command
operable program or batch file.

Does it work for m3u8 links too?

"hlsvariant://http://???/playlist.m3u8 conn=['S:??? ','S:??? ','S:??? ']"

beardypig:
`conn` is only applicable to `RTMPStream`. What exactly are you trying to do?

beardypig:
@ppsspp213 can you show us the command you ran where you got the error, so we can help you debug it?

ppsspp213:
No I can't sorry, it contains paid for material.
I need to send the amf data back to the host otherwise it won't stream the content, opening the .m3u8 link in vlc just gives a 30 second error message.

Tells me the stream is 360p
[...]

PR #641, opened 2017-03-01
docs: updated the docs for built-in stream parameters

beardypig:
Added some notes in the docs about settings more complex parameters, lists, dicts, etc.
Made some slight changes to the `StreamURL` class and the associated tests, nothing significant.

See #640; as the documentation is not very clear for setting lists/multiple values.


gravyboat:

Issue #645, opened 2017-03-01
Plugin: animelab.com VOD


starchivore:
### Checklist

- [ ] This is a bug report.
- [x] This is a plugin request.
- [ ] This is a feature request.
- [x] I used the search function to find already opened/closed issues or pull requests.

### Description

Both [animelab.com](https://www.animelab.com) and [https://dcgm6i50yfgtk.cloudfront.net](https://dcgm6i50yfgtk.cloudfront.net) are definitely AU/NZ only so either VPN or proxy should be required, though they're doing stuff like this here:

https://github.com/mczub/because-moe/blob/feature/animelab-fix/generate/au.py#L28
https://github.com/mczub/because-moe/blob/feature/animelab-fix/generate/animesources.py#L227

Not really sure about how their API is supposed to work, maybe these links could help?

https://github.com/edpenz/holo/blob/master/src/services/stream/animelab.py

https://github.com/mczub/because-moe/blob/feature/animelab-fix/generate/au.py#L28
https://github.com/mczub/because-moe/blob/feature/animelab-fix/generate/animesources.py#L227

https://github.com/arkenthera/Chiika/blob/master/src/main_process/stream-services.coffee#L27

https://github.com/khloke/play-to-xbmc-chrome/blob/master/js/modules.js#L27
https://github.com/khloke/play-to-xbmc-chrome/blob/master/js/content_scripts/animelab-content-scripts.js

This is how we're getting the latest and greatest:

https://www.animelab.com/api/shows/all?limit=100&page=0
https://www.animelab.com/api/shows/all?limit=100&page=1
https://www.animelab.com/api/shows/all?limit=100&page=2

As usual I'm picking _[Dragon Ball Z](https://www.animelab.com/shows/dragon-ball-z)_ as an example, English dub is only available for paid members and subbed episodes are free:

https://www.animelab.com/player/dragon-ball-z-episode-1/subtitles
https://www.animelab.com/player/dragon-ball-z-episode-2/subtitles
https://www.animelab.com/player/dragon-ball-z-episode-3/subtitles
https://www.animelab.com/player/dragon-ball-z-episode-4/subtitles
https://www.animelab.com/player/dragon-ball-z-episode-5/subtitles

API didn't show us anything helpful except for `"episodeCount"` etc.

https://www.animelab.com/api/shows/271

This is `forward=true`:

```
https://www.animelab.com/api/videos/show/271/subtitles?videoId=10070&position=0&forward=true&_=1488385408682
[...]

PR #648, opened 2017-03-02
plugins.animelab: added support for AnimeLab.com VOD

beardypig:
As requested in #645 by @starchivore.

Two new options were added to support this plugin:
  - `--animelab-email`: the email address to login to AnimeLab
  - `--animelab-password`: the password for your AnimeLab account

The content is only available with a login.

While the website says that it's restricted to Australia and New Zealand, I was able to access using a UK proxy. I was not able to access using a US or German proxy.


beardypig:
@gravyboat just rebased this :)

gravyboat:
Awesome thanks.

Issue #649, opened 2017-03-02
pycountry does not handle some locales


beardypig:
Originally reported here: https://github.com/streamlink/streamlink-portable/issues/12 by @sysilion

### Checklist

- [x] This is a bug report.
- [ ] This is a plugin request.
- [ ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description
 
Using some locales with `pycountry` will produce an `AttributeError` from `pycountry`. 

### Reproduction steps / Stream URLs to test

`streamlink --locale ko_KR twitch.tv/twitch`

### Environment details (operating system, python version, etc.)

When using the `pycountry` module instead of `iso3116/iso-639` modules.

### Comments, logs, screenshots, etc.

```shell
$ streamlink --locale ko_KR -ldebug "twitch.tv/twitch"
[cli][info] Found matching plugin twitch for URL twitch.tv/twitch
[plugin.twitch][debug] Getting live HLS streams for twitch
[plugin.twitch][info] Attempting to authenticate using OAuth token
[plugin.twitch][info] Successfully logged in as beardypig
Traceback (most recent call last):
  File "/home/beardypig/.virtualenvs/streamlink-python3.6/bin/streamlink", line 11, in <module>
    load_entry_point('streamlink', 'console_scripts', 'streamlink')()
  File "/home/beardypig/projects/streamlink/src/streamlink_cli/main.py", line 978, in main
    handle_url()
  File "/home/beardypig/projects/streamlink/src/streamlink_cli/main.py", line 479, in handle_url
    streams = fetch_streams(plugin)
  File "/home/beardypig/projects/streamlink/src/streamlink_cli/main.py", line 391, in fetch_streams
    sorting_excludes=args.stream_sorting_excludes)
  File "/home/beardypig/projects/streamlink/src/streamlink/plugin/plugin.py", line 321, in get_streams
    return self.streams(*args, **kwargs)
  File "/home/beardypig/projects/streamlink/src/streamlink/plugin/plugin.py", line 230, in streams
    ostreams = self._get_streams()
  File "/home/beardypig/projects/streamlink/src/streamlink/plugins/twitch.py", line 600, in _get_streams
    return self._get_hls_streams("live")
  File "/home/beardypig/projects/streamlink/src/streamlink/plugins/twitch.py", line 564, in _get_hls_streams
    force_restart=not stream_type == "live")
[...]

PR #650, opened 2017-03-02
utils.l10n: fix for some locales without an official name in pycountry

beardypig:
Fix for #649 and https://github.com/streamlink/streamlink-portable/issues/12. 

Greetings to our Korean users!


beardypig:
@gravyboat this one is high priority I think :)

gravyboat:
Good stuff, thanks a bunch @beardypig.

Issue #636, opened 2017-02-28
WWEN plugin?


whitetop:
### Checklist

- [ ] This is a bug report.
- [x] This is a plugin request.
- [ ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description
WWEN works on kodi but i'm asking can we get this on streamlink as a fully active py? i tried using it on livestreamer but it has a few py files one for setting and others for logins since its for kodi. streamlink has a bug so i can't get it to start up but read its a problem a dev was going to fix.

Edited: (missing info) this plugin is a WWE Network plugin for kodi url is ( http://network.wwe.com ) Plugin url ( http://forum.kodi.tv/showthread.php?tid=233467 ), some info can be found on that site with the download links for the plugin.

When using on kodi if temp bans your account after 15min so be careful, atres is used for the lower resolutions but it is hidden for some 4500K files like live PPV you have to wait a week for it to show up.

WWE uses the master m3u8 file with the 500k to 4500k m3u8 links.
### Expected / Actual behavior
AIO py to catch 1280x720p files on wwe and not the 3000k 720p, after 4500k ones (they are there but just hidden for a week)

### Reproduction steps / Stream URLs to test

1. Above

### Environment details (operating system, python version, etc.)

win 7 and server 03, all updated.

### Comments, logs, screenshots, etc.

Any more info say it simple of what you need, not cutting corners and thinking i'm a mind reader like below. (not rude but even a child can see what i'm on about below)

beardypig:
@whitetop github seems to be flagging a lot of users lately, particularly people that create accounts and create and issue right away - seems they have an over zealous spam detection system, I doubt anyone reported for you creating this issue :) 

I'm not familiar with WWEN could you provide some URLs that you would like to see supported.

whitetop:
@beardypig I run a site with others and our spam detectors don't do this but i did email them and they sorted the problem out.

The WWEN is here, it works good on kodi but gets flagged after 15 min and keeps breaking on the live stream. http://forum.kodi.tv/showthread.php?tid=233467

beardypig:
It would be more useful for you to provide us with some link that you expect to work, the more information you give the more likely someone will write a plugin.

whitetop:
@beardypig Like what link? asking for a link but not saying what for does not help.

The link in the above post is to the kodi plugin, anyone with a WWE network account can get the m3u8 file with inspect page (console).

beardypig:
[...]

PR #651, opened 2017-03-02
plugins.wwenetwork: support for WWE Network streams

beardypig:
As requested by @whitetop in #636.

Two new options were added to support this plugin:
 - `--wwenetwork-email`: the email for the WWE Network account
 - `--wwenetwork-password`: the password for the WWE Network account

 Both are required to access content.


gravyboat:
Thanks @beardypig.

Issue #657, opened 2017-03-03
TRT/TRT Spor plugins outdated


karlo2105:
### Checklist

- [x] This is a bug report.
- [ ] This is a plugin request.
- [ ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description

Turkish PBS. 

### Expected / Actual behavior

TRT and TRT Spor plugins don't work anymore.

### Reproduction steps / Stream URLs to test

TRT 1 HD
```
streamlink "http://www.trt.net.tr/anasayfa/canli.aspx?y=tv&k=trt1" best
error: No plugin can handle URL: http://www.trt.net.tr/anasayfa/canli.aspx?y=tv&k=trt1
```

TRT Spor
```
streamlink "http://www.trtspor.com/canli-yayin-izle.html" best
error: No plugin can handle URL: http://www.trtspor.com/canli-yayin-izle.html
```

### Environment details (operating system, python version, etc.)

W7 PRO/Streamlink portable

### Comments, logs, screenshots, etc.

No streams from TRT and TRT Spor are handled by streamlink.

beardypig:
Can you try `http://www.trt.net.tr/Anasayfa/canli.aspx?y=tv&k=trt1` (uppercase `A`)?

karlo2105:
```
streamlink "http://www.trt.net.tr/Anasayfa/canli.aspx?y=tv&k=trt1" best
[cli][info] Found matching plugin trt for URL http://www.trt.net.tr/Anasayfa/can
li.aspx?y=tv&k=trt1
Traceback (most recent call last):
  File "C:\Users\Ddr\Downloads\streamlink\\streamlink-script.py", line 12, in <m
odule>
    main()
  File "C:\Users\Ddr\Downloads\streamlink\packages\streamlink_cli\main.py", line
 969, in main
    handle_url()
  File "C:\Users\Ddr\Downloads\streamlink\packages\streamlink_cli\main.py", line
 479, in handle_url
    streams = fetch_streams(plugin)
  File "C:\Users\Ddr\Downloads\streamlink\packages\streamlink_cli\main.py", line
 391, in fetch_streams
    sorting_excludes=args.stream_sorting_excludes)
  File "C:\Users\Ddr\Downloads\streamlink\packages\streamlink\plugin\plugin.py",
 line 321, in get_streams
    return self.streams(*args, **kwargs)
[...]

PR #658, opened 2017-03-03
plugins.trt: make the url test case insensitive and fix py3 bug

beardypig:
As reported by @karlo2105 in #657.

The URL test was case-sensitive and would only work with Anasayfa, not anasayfa.
The base64 decoded data needed to be decoded to unicode to work in Python 3.


karlo2105:
OK ;)

gravyboat:
Thanks @beardypig!

Issue #660, opened 2017-03-03
ARD Live outdated


karlo2105:
### Checklist

- [x] This is a bug report.
- [ ] This is a plugin request.
- [ ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description

ARD is German PBS. 

### Reproduction steps / Stream URLs to test

```
streamlink "http://www.daserste.de/live/index.html"
error: No plugin can handle URL: http://www.daserste.de/live/index.html
```

### Environment details (operating system, python version, etc.)

As usual ;)

gravyboat:
Closing since #661 is merged.

PR #661, opened 2017-03-03
plugins.ard_live: updated to new site layout

beardypig:
As reported by @karlo2105 in #660 the `ard_live` was no longer working.

Updated the plugin to support the new site layout, also added support for VOD streams.

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -19,7 +19,7 @@ aftonbladet         aftonbladet.se       Yes   Yes
 alieztv             aliez.tv             Yes   Yes
 antenna             antenna.gr           --    Yes
 app17               17app.co             Yes   --
-ard_live            live.daserste.de     Yes   --    Streams may be geo-restricted to Germany.
+ard_live            daserste.de          Yes   Yes   Streams may be geo-restricted to Germany.
 ard_mediathek       ardmediathek.de      Yes   Yes   Streams may be geo-restricted to Germany.
 artetv              arte.tv              Yes   Yes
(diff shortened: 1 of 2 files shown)


karlo2105:
Muy bien. ;)

back-to:
You should **change the regex** for `ard_live`
it should
**only works for** `http://www.daserste.de`
or
**not for** `http://mediathek.daserste.de`

`http://mediathek.daserste.de` is the same as the plugin **ard_mediatheak**

`ardmediathek.de` has more videos but all videos from `mediathek.daserste.de` are there.

They have the _same_ urls, just a bit different regex
```
http://www.ardmediathek.de/tv/extra-3/Automobil-Lobbyist-Maxi-Schafroth-zum-Th/Das-Erste/Video?bcastId=23817212&documentId=41101758

http://mediathek.daserste.de/extra-3/Automobil-Lobbyist-Maxi-Schafroth-zum-Th/Video?bcastId=23817212&documentId=41101758
```
 and here the same video for `www.daserste.de` but this works with `ard_live`
```
http://www.daserste.de/unterhaltung/comedy-satire/extra-3/videosextern/automobil-lobbyist-maxi-schafroth-zum-thema-stickoxide-100.html
```

New regex for `ard_mediatheak` something like this
```py
_url_re = re.compile(r"http(s)?://(?:(\w+\.)?ardmediathek.de/tv|mediathek.daserste.de/)")
```

beardypig:
True that. I've updated the regex :)

gravyboat:
Thanks @beardypig and thanks for weighing in @back-to!

Issue #662, opened 2017-03-03
vidio.com outdated


karlo2105:
### Checklist

- [x] This is a bug report.
- [ ] This is a plugin request.
- [ ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description

Vidio.com is streaming platform from Indonesia. 

### Expected / Actual behavior

```
streamlink "hlsvariant://https://livestream-hybrid.akamaized.net/live/smil:sctv.smil/playlist.m3u8" best
[cli][info] Found matching plugin stream for URL hlsvariant://https://livestream
-hybrid.akamaized.net/live/smil:sctv.smil/playlist.m3u8
[cli][info] Available streams: 144p (worst), 270p, 360p (best)
[cli][info] Opening stream: 360p (hls)
[cli][info] Starting player: C:\Users\Ddr\Downloads\VLCPortable\App\vlc\vlc
[cli][info] Player closed
[cli][info] Stream ended
[cli][info] Closing currently open stream...
```

### Reproduction steps / Stream URLs to test

```
streamlink "https://www.vidio.com/en/live/204-sctv-tv-stream"
error: No plugin can handle URL: https://www.vidio.com/en/live/204-sctv-tv-stream

```

### Environment details (operating system, python version, etc.)

W7 Pro/Streamlink portable

beardypig:
Work around is to remove `en/` from the URL :)

gravyboat:
Closing.

PR #663, opened 2017-03-03
plugins.vidio: fix for regex, if the url is the english version

beardypig:
As reported by @karlo2105 in #662 :)

The URLs will have the language code `en` when viewing the English version of the website.


gravyboat:
What a worthwhile exit code from appveyor `code -1073740940`.

beardypig:
Can rerun it, the change is very small though :)

gravyboat:
No reason to do so, I'm going to merge, just wanted to note it in case there were concerns.

PR #665, opened 2017-03-03
plugins.ard_mediathek added mediathek.daserste.de support

back-to:
- Added `mediathek.daserste.de` support to `ard_mediathek` (like i wrote in #661)
- Removed unused code.


gravyboat:

Issue #652, opened 2017-03-02
Docs show usernames from donation page in navigation


gravyboat:
### Checklist

- [x] This is a bug report.
- [ ] This is a plugin request.
- [ ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description

If you navigate over to https://streamlink.github.io/ you'll see that the titles I have on the donate page are showing up as links. I believe this issue is caused by having a toctree depth of 2 set in the index which is picking up the fact that I'm linking to people's profiles.

### Expected / Actual behavior

Navigation doesn't have usernames.

### Reproduction steps / Stream URLs to test

1. Visit https://streamlink.github.io/

### Environment details (operating system, python version, etc.)

N/A

### Comments, logs, screenshots, etc.

My plan to resolve this is simply to remove the links to people's names, or subset their GitHub link underneath their username as opposed to in the title section. If anyone else has better ideas or a specific preference let me know.

beardypig:
Sounds good to me :-)

bastimeyer:
The user names are still in the main menu.
I think you need to change the structure of the rst document and make it so that user names are no section headers, but simple formatted texts elements (or list items).

beardypig:
Yep... Looks like `~~` is the right underline to use :)

beardypig:
Docs have updated and the usernames no longer appear in the navigation hierarchy :)

PR #672, opened 2017-03-07
Fix up the donate docs

gravyboat:
This fixes https://github.com/streamlink/streamlink/issues/652


beardypig:
Cool thanks @gravyboat

Issue #670, opened 2017-03-06
goodgame.py


loboque:
I want to say goodgame.py Plugin not work for me ,
C: \ Users \ loboque> streamlink.exe goodgame.ru/channel/happy/
error: No plugin can handle URL: goodgame.ru/channel/happy/
but i download his from this and he work fine!
chrippa/livestreamer#1518 (comment)

back-to:
Try it with https before the url
`streamlink https://goodgame.ru/channel/happy/ best`
```
ubuntu@ubuntu:~/Desktop/git/streamlink$ streamlink https://goodgame.ru/channel/happy/ best
[cli][info] Found matching plugin goodgame for URL https://goodgame.ru/channel/happy/
[cli][info] Available streams: 240p (worst), 480p, 720p, 1080p (best)
[cli][info] Opening stream: 1080p (hls)
[cli][info] Starting player: /usr/bin/vlc
[cli][info] Player closed
[cli][info] Stream ended
[cli][info] Closing currently open stream...
ubuntu@ubuntu:~/Desktop/git/streamlink$ 
```

The regex only works with https
https://github.com/streamlink/streamlink/blob/34e68540d83dfbb1796bdda701475b0c3fb897f4/src/streamlink/plugins/goodgame.py#L15

loboque:
ok sorry im total noob((

gravyboat:
@loboque No need to apologize, I'm not sure if the plugin didn't support https in the first place for a specific reason or not but I'm labeling this as a bug until we know for sure.

intact:
Looks like they enabled http protocol again.

beardypig:
@loboque it should work with http URLs now, so your original attempt should work now (with the [latest version](https://streamlink-builds.s3.amazonaws.com/nightly/windows/streamlink-latest.exe)) :)

PR #674, opened 2017-03-07
plugins.goodgame: support for http URLs

beardypig:
As reported in #670, by @loboque. 

goodgame.ru has re-enabled http support and no longer has only https URLs.


gravyboat:

Issue #540, opened 2017-02-05
Is there a command to select specific resolution (i.e. 720p) with best bitrate when there are multiple?


Truanifan:
### Checklist

- [ ] This is a bug report.
- [ ] This is a plugin request.
- [ ] This is a feature request.
- [x] I used the search function to find already opened/closed issues or pull requests.

### Description

Forgive me if I don't make enough sense, but I'll try to explain the best I can... I'm trying to write a batch file to grab a selected resolution at the best bitrate, but not sure what commands to use. The m3u8 has multiple bitrates for the same resolution. Example: 720p at 5996k and 720p at 7795k. From using livestreamer I know I can use name_key=bitrate to select the stream based on the bitrate, but I don't know how to tell the difference between the resolutions based on bitrates and the bitrates vary based on the different videos on the site. When I get a list of the available streams via streamlink I see 720p_alt and 720p, but don't know which has the better bitrate.

Is there a way to make streamlink grab the best bitrate when selecting the 720p stream quality in the command? (example: "hlsvariant://....m3u8" best -o "output.ts")

I know this isn't so much an issue as it is a question, but I hope it's still alright to ask. Thank you in advance for any advice.

scottbernstein:
Yes, @truanifan, the way to format your command line for selecting the best bitrate when there are multiple streams of the same bitrate is as follows:

"hlsvariant://....m3u8 name_key=bitrate" best -o "output.ts")

________________________________

Truanifan:
I think I messed up in my explanation before. I'm not looking for just the best bitrate out of all the available resolutions (which is what I get when I use the command you suggested), but the best bitrate of the 720p resolutions or whatever specific size I choose from the available streams. The resolutions go all the way up to 1080p from 234p, but I'm asking what I can do if I want to grab the best bitrate of the two streams available for a specific resolution.

When I use name_key=bitrate it renders the 720p selection option moot and when I use "best" (as you suggest) it just chooses the best of the best which is 1080p. I'm looking for something that will in essence let me select the best bitrate based on the resolution selection and not specifically the bitrate since the bitrates vary from video to video. 

Is there a way to select the best bitrate of multiple bitrates for a specific resolution?

scottbernstein:
If you're savvy enough to use the CURL command to hit the same .m3u8 URL then you can see the list of all of streams, find the resolution / bit rate combination that you want, pick it manually, and then use hls:// (instead of hlsvariant://) to stream the specific bitrate you want.

________________________________

[...]

PR #675, opened 2017-03-07
stream.hls: format string name input for parse_variant_playlist

beardypig:
This allows developers to use a format string to name the streams, which allows more descriptive names. eg. {pixels}_{bitrate} can be used to produce 720p_3500k, 720p_4500k, which is useful if the streams are available in multiple resolutions at the same bitrate.

This feature can also be used when using the builtin stream plugins, eg.
```shell
streamlink "hlsvariant://test.se/master.m3u8 name_fmt=\"{pixels}_{bitrate}\""
```

The limitation being that the name string has to be enclosed by `"` quotes.

This should go someway to fixing issues like #540.

### DIFF
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -7,6 +7,10 @@ from ..cache import Cache
 from ..exceptions import PluginError, NoStreamsError
 from ..options import Options
 
+# FIXME: This is a crude attempt at making a bitrate's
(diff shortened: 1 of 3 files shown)


beardypig:
I'll add some tests :)

scottbernstein:
Thumbs up for this change @beardypig!

gravyboat:
This is great, thanks @beardypig!

PR #677, opened 2017-03-07
plugins.wwenetwork: use the resolution and bitrate in the stream name

beardypig:
Additional for #636, to expose the bit rate.


gravyboat:
Failure is a non-issue and seems like an appveyor problem. They sure have been having a lot of issues recently.

gravyboat:
Thanks @beardypig.

PR #679, opened 2017-03-08
plugins.pluzz: fix SWF player URL search to bring back HDS stream support

melmorabity:
This PR brings back HDS stream support, since the wway to retrieve the flash player on Pluzz has changed.


src/streamlink/plugins/pluzz.py
```diff
@@ -106,7 +106,7 @@ def _get_streams(self):
             res = http.get(player_url)
             match = self._swf_re.search(res.text)
             if match is not None:
-                swf_url = os.path.dirname(player_url) + match.group('swf')
+                swf_url = 'http:' + match.group(0)
```
beardypig: See `streamlink.utils.update_scheme`, if you wanna use the same scheme as the other URLs also works for sure in Python 2+3 :)

melmorabity: Thanks for the suggestion ;)

Issue #683, opened 2017-03-09
Plugin request: camsoda.com [NSFW]


supermagicpony:
### Checklist

- [ ] This is a bug report.
- [x] This is a plugin request.
- [ ] This is a feature request.
- [x] I used the search function to find already opened/closed issues or pull requests.

### Description

Would it be possible to have a plugin for http://camsoda.com (NSFW)? Here is an example of a streaming page:

https://www.camsoda.com/cherryladys

The associated m3u8 url is:

https://vide1.camsoda.com/cam/mp4:cherryladys-enc11_mjpeg/playlist.m3u8?token=1e5ab84f2ca5764b0366a8b6619de668

I am able to get it to work manually by sniffing out the cookie and the m3u8 url:

```
streamlink --http-cookie "token=1e5ab84f2ca5764b0366a8b6619de668" hls://vide1.camsoda.com/cam/mp4:cherryladys-enc11_mjpeg/playlist.m3u8? best

johnthecracker:
I am able to get it to work manually by RTMPDUMPHELPER + RTMPSUCK

RTMP Proxy Server v2.4 GIT-2015-12-14 (Compiled by KSV)
(c) 2010 Andrej Stepanchuk, Howard Chu; license: GPL

Streaming on rtmp://0.0.0.0:1935
ERROR: Handshake failed
Closing connection... done!

ERROR: Handshake failed
Closing connection... done!

WARNING: Trying different position for client digest!
Processing connect
       app : cam?token=5bae312212363371326e6eeab0808c35/
  flashVer : WIN 25,0,0,127
    swfUrl : https://www.sodacdn.com/lib/video-js/video-js.swf
     tcUrl : rtmp://vide1.camsoda.com:1935/cam?token=5bae312212363371326e6eeab08
08c35/
   pageUrl : https://www.camsoda.com/pinkysin
ERROR: Handshake failed
Closing connection... done!

      live : no
  Playpath : ?mp4:pinkysin-enc2
 Saving as : 2017-03-16_10-19-19_mp4_pinkysin-enc2.flv
INFO: Metadata:
INFO:   keyFrameInterval        10.00
INFO:   x_bw                    0.00
INFO:   profile                 main
INFO:   width                   640.00
INFO:   x_cfps                  28.50
INFO:   height                  480.00
INFO:   codec                   H264Avc
INFO:   x_jsbw                  0.00
INFO:   x_version               v44
INFO:   fps                     0.00
[...]

PR #689, opened 2017-03-10
plugins.camsoda Added support for camsoda.com

back-to:
works only for live streams, not for vods

Fixed streamlink/streamlink#683


gravyboat:
Looks as though no one else took a look at this. Are you planning to write tests? If not I'll merge it.

gravyboat:
Awesome, thanks @back-to!

PR #696, opened 2017-03-11
Aliez plugin now accepts any TLD

### LINKED ISSUE #695: Plugin request: Aliez.tv to Aliez.me
### Checklist

- [x] This is a bug report.
- [x] This is a plugin request.
- [ ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description

Aliez.tv does not exist anymore there is Aliez.me which is actually the same streaming site.

### Reproduction steps / Stream URLs to test

1. http://aliez.me/live/fkpev212/ link im trying to open in player

CallMeJuf:
Also fixed Regex verifying we're on a video/live page. Fixes #695, with wildcard TLD suggestion.

### DIFF
--- a/src/streamlink/plugins/aliez.py
+++ b/src/streamlink/plugins/aliez.py
@@ -8,15 +8,8 @@ from streamlink.plugin.api import http, validate
 from streamlink.stream import HTTPStream, RTMPStream
 
 _url_re = re.compile(r"""
-    http(s)?://(\w+\.)?aliez.\w+/
-    (?:
-        (?:
-            live/[^/]+
-        )|
-        (?:
-            video/\d+/[^/]+
-        )
-    )
+    https?://(\w+\.)?aliez.\w+/
+    (?:live/[^/]+|video/\d+/[^/]+)
 """, re.VERBOSE)
 _file_re = re.compile(r"\"?file\"?:\s+['\"]([^'\"]+)['\"]")
 _swf_url_re = re.compile(r"swfobject.embedSWF\(\"([^\"]+)\",")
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/aliez.py
```diff
@@ -8,13 +8,15 @@
 from streamlink.stream import HTTPStream, RTMPStream
 
 _url_re = re.compile(r"""
-    http(s)?://(\w+\.)?aliez.tv
+    http(s)?://(\w+\.)?aliez.\w+/
```
beardypig: You could do the regex like this:
```python
_url_re = re.compile(r"https?://(?:\w+\.)?aliez(?:\.\w+)+/(?:live|video/\d+)/.+")
```
There's no need for so many groups with the live/video part of the URL. If you make the domain part repeatable it can support `.tv`, `.co.me`, etc. They don't really have to be non-capturing groups either :)

gravyboat:
@CallMeJuf When you have some time could you review those changes suggested by Beardypig to see about incorporating, and if you feel they shouldn't be incorporated notate why? I would like to get this merged.

src/streamlink/plugins/aliez.py
CallMeJuf: Agreed it's a bit too expanded. I didn't want to add a repeating group to the TLD (and consequently not accept .co.uk ect) in case other sites happen to use "aliez" as a subdomain. For example if twitch decided to use the following format:
http://aliez.twitch.tv/videos/
This plugin would match. I don't know of any that use this but I could see sites using profile names as subdomains and wouldn't want conflicts.

beardypig: That's a good point :-)

gravyboat:
Awesome thanks @CallMeJuf!

Issue #145, opened 2016-11-09
Support for "canlitv" Sites


Rotarum:
Hi,

there are some web Sites with similar Live Content on it. Did a google search in order to list them as following:

http://canlitv.mobi/
http://www.ecanlitvizle.net/
http://www.canlitvlive.co/
https://www.canlitv.life/
http://www.canlitvlive.co/
http://www.canlitv.info/
https://www.canlitvizle.io/

I'm pretty sure some of them are just a Copy of the other one :)

PR #699, opened 2017-03-12
plugins.canlitv - Added new plugin canlitv

back-to:
Most of the streams works, sometimes there are two iframes.
Just use the link from the iframe and it will works most of the time.

Works only for HLS streams, won't work if the HLS file is broken.

```
canlitv.com
http://www.canlitv.com/24-tv

canlitv.life
http://www.canlitv.life/trt-haber

canlitvlive.co
http://www.canlitvlive.co/izle/ntv.html

canlitvlive.live
http://www.canlitvlive.live/izle/atv.html

ecanlitvizle.net
http://www.ecanlitvizle.net/trt-1-izle/
```

Fixed streamlink/streamlink#145

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -39,6 +39,11 @@ btv                 btv.bg               Yes   No    Requires login, and geo-res
 canalplus           - canalplus.fr       Yes   Yes   Streams may be geo-restricted to France.
                     - c8.fr
                     - cstar.fr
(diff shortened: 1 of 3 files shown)


gravyboat:
Looks good to me, I'll merge it tomorrow if there are no concerns.

gravyboat:
I dropped the ball on getting this merged in a timely fashion, sorry about that @back-to! Merging now. Thanks!

PR #702, opened 2017-03-13
Removed dead plugins

back-to:
| FILES           | URL                        | STATUS |
| ---             | ---                        | ---    |
| bliptv.py       | http://blip.tv             | **REDIRECT** to different website |
| gaminglive.py   | http://gaminglive.tv       | **DEAD ?? REDIRECT** (401 Authorization Required) |
| letontv.py      | http://leton.tv            | **DEAD** |
| livestation.py  | http://www.livestation.com | **DEAD** |
| meerkat.py      | http://meerkatapp.co       | **REDIRECT** to different website |
| veetle.py       | http://veetle.com          | **DEAD** |
| viagame.py      | http://www.viagame.com     | **REDIRECT** to different website |


beardypig:
Either than or mark them as [broken](https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugin/plugin.py#L164).

back-to:
Didn't know that there was a `broken` method

But what would be the point to have `broken` plugins,
when they will never work again because there websites are down/closed.

The `broken` method would work great,
if the website would be still up but with `DRM` `DASH` or something that streamlink does not support.

gravyboat:
Yeah I agree with @back-to. In this event the sites are definitely down and not broken.

PR #710, opened 2017-03-16
Update archlinux installation instructions and maintainer info

mmetak:


gravyboat:
Awesome thanks @mmetak!

PR #711, opened 2017-03-17
plugins.bbciplayer: add support for HLS streams

beardypig:
Added `hls` stream support in addition to the existing `hds` stream support.


gravyboat:
Cool, thanks @beardypig.

PR #713, opened 2017-03-17
setup: choose pycrypto as a dependency using an environment variable

mmetak:


gravyboat:
Good stuff, thanks @mmetak!

PR #715, opened 2017-03-17
Add info about pycrypto and pycountry variables to install doc

mmetak:


docs/install.rst
```diff
@@ -169,13 +169,26 @@ Name                                 Notes
                                      audio and video streams, eg. YouTube 1080p+
 ==================================== ===========================================
 
+Using pycrypto and pycountry
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
beardypig: I'm surprised this didn't fail the build. Could you make the `^^` underline the same width and the text above?

mmetak: Sorry about that @beardypig , forgot to adjust it after I made a change. Should be fixed now.

beardypig: Thankee kindly sai.

gravyboat:
Merging now that the underscores are fixed, thanks @mmetak!

Issue #718, opened 2017-03-18
LookupError: Invalid language code: en_150


Ackmard:
Can anyone help me with that? Just installed it and i can't figure what's wrong.

![livestream error 2](https://cloud.githubusercontent.com/assets/26508542/24074514/fc6a051e-0c12-11e7-8398-e5e7b8e12f9d.jpg)

beardypig:
I'll have to take a closer look as to why that is happening. You are work around it by doing 
```shell
streamlink --locale en_US twitch.tv/quill18
```
That will set your locale to US English. If you don't mind me asking, what language (English by the looks of it) and Country/Region do you have set up in Windows?

Ackmard:
Country/Region is United States, i'm using 2 input languages US English and Greek.

beardypig:
This should be fixed in the latest night version (as #725 was merged).

PR #725, opened 2017-03-21
utils.l10n: use default locale if the system returns an invalid locale

beardypig:
A fix for #718, if the OS returns a locale that cannot be parsed then the default locale will be used (currently US English).


gravyboat:

Issue #729, opened 2017-03-22
Order of qualities returned by API/CLI


enzanki-ars:
### Checklist

- [x] This is a feature request.
- [x] I used the search function to find already opened/closed issues or pull requests.

### Description

Instead of using a `Dict`, it might be better to use an `OrderedDict` object to store the stream URLs, allowing for easier viewing of the available qualities and control of the selected quality from the API side.  

For example, this change could allow this...

```
Available streams: 540p_alt2, 720p_alt, 1080p_alt, 236p_alt, 360p_alt, 236p_alt2, 236p (worst), 
360p, 540p, 720p, 1080p (best)
```

...to instead be displayed as this...

```
Available streams: 236p (worst), 236p_alt, 236p_alt2, 360p, 360p_alt, 540p, 540p_alt2, 720p, 
720p_alt, 1080p (best), 1080p_alt
```

...or even...

```
Available streams: 
 - 236p (worst)
 - 236p_alt
 - 236p_alt2
 - 360p
 - 360p_alt
 - 540p
 - 540p_alt2
 - 720p
 - 720p_alt
 - 1080p (best)
 - 1080p_alt
```

gravyboat:
I think this is a pretty good idea. It wasn't such a major issue for livestreamer as it originally functioned because there were so few options (usually just 3 when I started using it), but as more options show up this could definitely benefit users. Let's see what kind of discussion this issue generates. Thanks @enzanki-ars.

enzanki-ars:
I was poking through the source code to try and see if I could make a simple PR for it, but there is a lot to read through.  In theory, this should be a one line change, but I just need to find the one line (such as changing a `{}` to an `OrderedDict()`).  Also, in theory, this change should be backwards compatible.

enzanki-ars:
https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugin/plugin.py#L257 should be the line.  Changing it from `{}` to `OrderedDict()` should work.  Testing locally now...

enzanki-ars:
I have made a couple of modifications to get this idea to work, and I might have discovered another bug:

[...]

PR #731, opened 2017-03-23
Return sorted list of streams.

enzanki-ars:
See #729 for more information


gravyboat:
Cool, I'm going to leave this open (along with the original issue) to see if we get some discussion going. One of the primary maintainers is out on vacation right now so we might hold off on a more formal review till he gets back in case there is additional input or some sort of dependent logic I am missing.

beardypig:
I would like to review this :-)

enzanki-ars:
There are still some minor changes I would like to make to properly order the `alt` streams.

I think it would still be best to leave it in the following format:

`qualities = ['alt_streams_(worst->best)', 'primary_streams_(worst->best)', 'worst', 'best']`

The placement of worst still bothers me a little bit.  Thoughts?

enzanki-ars:
Extra note: I can confirm that this change has not had any effect on API usage and should be fully backwards compatible (was in my case).  Have not been able to test effect on CLI output yet.

enzanki-ars:
Tested the CLI - Outputs in the sorted order too:

Available streams: 540p_alt, 540p_alt2, 360p_alt, 720p_alt, 236p_alt2, 1080p_alt, 236p (worst), 360p, 540p, 720p, 1080p (best)

gravyboat:
@enzanki-ars Is there any particular reason you wanted to put the alt streams first? Typically when I think about it I would prefer the main options (primary_streams with worst/best notation) first, then the alternates afterwards just so it's a little cleaner for users to quickly scan and decide on a quality value. So instead of looking through the whole list you only need to scan the primary entries at the beginning of the list instead of the end. I could also see it being a bit problematic if there were a lot of alternate language streams, or alternate framerate streams which have been coming up more and more on Twitch.

enzanki-ars:
Good point. So would it be better to sort from best to worst and reverse the current order? If so, should I place `worst` at the end?

gravyboat:
@enzanki-ars I'm not sure that's why I wanted to discuss it! Other people might have different opinions so I would be interested in hearing those as well. I just wasn't sure if there was a specific reason you were sorting the way you were.

karlo2105:
I think it's better in my opinion to sort qualities from worst to best. It's logical as lowest number begins on the far left and while growing heads to far right.

enzanki-ars:
@gravyboat No problem.  I understand the intent of both of your messages.  Sorry about not being clear with the intent of my reply.  I was just responding to confirm my understanding of your point. I should have also stated that there was no reason to my sorting method.  I just used what was pre-existing.  

Currently, I see 3 main choices in sorting:

1. `['alt_streams_(worst->best)', 'primary_streams_(worst->best)', 'worst', 'best']`
2. `['worst', 'alt_streams_(worst->best)', 'primary_streams_(worst->best)', 'best']`
3. `['best', 'primary_streams_(best->worst)', 'alt_streams_(best->worst)', 'worst']`

My personal preference may lean a little more towards option 2 or 3, but there may be others that I did not think of.

malvinas2:
Sorry, to interrumpt you, but what I never dared to ask: 
What is 'alt' the abbreviation for? In case it means 'alternative', what kind of alternative stream is it?

enzanki-ars:
@malvinas2 I have only really been dealing with HLS streams, so my reply will be specific to that.  It's a bit easier to understand, I think.

You would be correct with the statement that it stands for alternative.

---

A good example is http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/master.m3u8 (From http://www.cbsnews.com/live/):

The m3u8 file contains the following info:

```
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=168000,RESOLUTION=320x180,CODECS="avc1.77.30, mp4a.40.2"
http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/index_120_av-p.m3u8?sd=10&rebase=on
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=168000,RESOLUTION=320x180,CODECS="avc1.77.30, mp4a.40.2"
http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/index_120_av-b.m3u8?sd=10&rebase=on
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=424000,RESOLUTION=640x360,CODECS="avc1.66.30, mp4a.40.2"
http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/index_360_av-p.m3u8?sd=10&rebase=on
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=424000,RESOLUTION=640x360,CODECS="avc1.66.30, mp4a.40.2"
http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/index_360_av-b.m3u8?sd=10&rebase=on
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=764000,RESOLUTION=640x360,CODECS="avc1.77.30, mp4a.40.2"
http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/index_700_av-p.m3u8?sd=10&rebase=on
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=764000,RESOLUTION=640x360,CODECS="avc1.77.30, mp4a.40.2"
http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/index_700_av-b.m3u8?sd=10&rebase=on
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=1264000,RESOLUTION=960x540,CODECS="avc1.77.30, mp4a.40.2"
http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/index_1200_av-p.m3u8?sd=10&rebase=on
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=1264000,RESOLUTION=960x540,CODECS="avc1.77.30, mp4a.40.2"
http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/index_1200_av-b.m3u8?sd=10&rebase=on
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=1864000,RESOLUTION=1280x720,CODECS="avc1.77.30, mp4a.40.2"
http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/index_1800_av-p.m3u8?sd=10&rebase=on
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=1864000,RESOLUTION=1280x720,CODECS="avc1.77.30, mp4a.40.2"
http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/index_1800_av-b.m3u8?sd=10&rebase=on
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=2264000,RESOLUTION=1280x720,CODECS="avc1.77.30, mp4a.40.2"
http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/index_2200_av-p.m3u8?sd=10&rebase=on
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=2264000,RESOLUTION=1280x720,CODECS="avc1.77.30, mp4a.40.2"
http://cbsnewshd-lh.akamaihd.net/i/CBSNHD_7@199302/index_2200_av-b.m3u8?sd=10&rebase=on
```
In this stream, there are links to the individual qualities.  While the `resolution` part of every other line does not really mean much as it does not need to be accurate, or even exist, it is good for this example.

In this example, there are 4 different 720p streams.  Streamlink, by default, only cares about 3 of these.  

```
Available streams: 180p_alt, 360p_alt, 360p_alt2, 540p_alt, 720p_alt, 720p_alt2, 180p (worst), 360p, 540p, 720p (best)
```

The last line should be the `720p (best)` stream.  

The `720p_alt` stream should be the second to last link, and also has a resolution of `1280x720` and a bandwidth of `2264000`, therefore it can be considered a mirror.  

The `720p_alt2` stream should be the second to last link, and while it has a resolution of `1280x720`, it does have a bandwidth of `1864000`, therefore, it would be a slightly lower quality.  

---

A great example of the need for the `alt` streams comes from The Weather Channel stream.  I won't give the URL publicly, as there may be some legal issues with it (even though there is no authentication in front  of the HLS stream.  The is authentication on their website though...  - If you Google well, you might find it...  The correct URL contains` twc_1` and it comes from the Akamai CDN)

The Weather Channel has the following qualities:

```
Available streams: 720p_alt2, 540p_alt, 360p_alt, 360p (worst), 540p, 720p (best)
```

(I don't know why 720_alt is missing, even though there are 4 different 720p streams listed in the file.)

In any video player, even if it is not using streamlink, the `720p (best)` stream does not work properly.  `720p_alt2` works perfectly though.

beardypig:
`alt` is alternative. It's a stream of the same type (`hls`/`hds`/etc.) and the same quality. So if there are two`720p hls` streams then one will be named `720p` and the other `720p_alt`. If you ask `streamlink` for `720p` and it fails, it will automatically go to `720p_alt` as an _alternative_ :)

In my opinion the way they are listed it a bit messy, the alternative streams should not be shown in the list and be completely transparent to the user - same goes for different types, as there is already a way to specify which order you want the types (`hls` vs. `hds`) to be sorted. I think the user should be presented with the minimum list of options and any alternatives be chosen by `streamlink`. Of course, there should be options to change the behaviour of this - but in most cases `streamlink` should make the correct choice.

enzanki-ars:
@beardypig There is a case, as noted above, where the 720p fails in video players, and not with the actual processing of the manifest file.  The Weather Channel seems to be improperly encoding one of their streams, which is the one that streamlink and every other program finds as the best quality.  There would be no way for streamlink to notice this error properly, and therefore manually selecting the `alt` stream is still valuable.

enzanki-ars:
@beardypig Missed the last sentence you wrote.  In the ~150 different streams I have, the Weather Channel one is the only one where I need to use the `alt` streams.  Hiding the `alt` for the rest might be a good idea, as long as there is a way to enable display of them in case there are issues.

beardypig:
@enzanki-ars can you email the stream url? I found one that matches this regex `r"https://\w+-\w\w\.akamaihd\.net/i/twc_1@\d+/master\.m3u8"` which listed `720p`, `720p_alt`, and `720p_alt2` but I had no trouble playing any of the streams.

enzanki-ars:
@beardypig I am away from my computer right now, and there isn't a publicly listed email for your account.  I can confirm that the regex provded matches the URL I mentioned.  

I will look into it later, but MPV and video.js both failed to properly play the `720p` stream.  I will check again later to see if it was temporary, or if VLC will play it correctly.

beardypig:
I only tested it with VLC3 :)

src/streamlink/plugin/plugin.py
```diff
@@ -254,7 +255,7 @@ def streams(self, stream_types=None, sorting_excludes=None):
                                 key=partial(stream_type_priority,
                                             stream_types))
 
-        streams = {}
+        streams = OrderedDict()
```
wlerin: This seems to be unnecessary, no? Since you're creating another OrderedDict below, there's no reason to make this one an OrderedDict as well (and if memory serves the builtin dict is slightly more efficient).

enzanki-ars: I missed that.  I originally relied on a slightly different technique on my original attempt to sort, and this is a remnant of that.

gravyboat:
Is there any other feedback on this? If not I think the only discussion left is the list order for the streams. I would like to get this merged in.

roshavagarga:
@Gravyboat - Can I make a dumb suggestion? Since ALT streams tend to have lower bitrate and are backup streams, and to be very honest, all of the sorting ideas I've seen in this thread are, quite literally confusing and not logical, can we just split them up?

Example:
streamlink twitch.tv/randommadeup best
[cli][info] Found matching plugin twitch for URL twitch.tv/randommadeup
[cli][info] Available MAIN streams: worst, in-between, best
[cli][info] Available BACKUP streams: worst_alt, in-between_alt, best_alt1, best_alt2

gravyboat:
@roshavagarga The problem with that is we would have to determine how to do the split, and account for it on all the plugins depending on how they do naming. Unfortunately it's not realistic to be able to do something like that.

enzanki-ars:
In my experience, the alt streams are of the same quality and bitrate. To be honest, if they were hidden by default, I wouldn't care, as long as I can re-enable them.  Also, the sorting seems to work for all sources.

roshavagarga:
@gravyboat Ah, there's the problem. I thought the stream qualities were being parsed at an above-plugin level. Since that's not feasible, I'd suggest going with an arrangement such as this one:
Available streams: 160p, 160p_alt (worst); 480p, 480p_alt; 720p, 720p_alt (best)

This makes much more sense to me as far as being able to quickly read between the different options and having a hierarchy.

gravyboat:
Okay I would like to get this merged in so we can release 0.6.0. . @roshavagarga I like your idea but would prefer `160p_alt (worst), 160p, 480p, 480p_alt, 720p, 720p_alt (best)` as it keeps it easy to digest as a list with the worst and best book ending. Thoughts? I would rather just get this in than discuss it too much more, we can always change it later but this is starting to get a little anal about a minor formatting change. :)

roshavagarga:
@gravyboat Yeah, I think your suggestion is an improvement. We can go ahead with it, no objections from me.

beardypig:
@gravyboat not `160p_alt (worst), 160p, 480p_alt, 480p, 720p_alt, 720p (best)` with the `alt` streams having lower priority?

gravyboat:
@beardypig You're right. I was filling that comment out really late last night. Alt streams as lower priority is fine.

gravyboat:
@enzanki-ars Opinions on the format proposed by @beardypig? If you're fine with that can you rebase to resolve conflicts, update the format, and then we'll get this merged? Sorry about the delay on this one.

enzanki-ars:
Sorry about the delay on my end; just finished with exams. The format sounds good. I will look into getting that adjusted and added in. In theory, this shouldn't be too difficult to get corrected, if my memory serves me right and my plan of attack works...

gravyboat:
@enzanki-ars No worries, I hope your exams went well! Looking forward to getting this merged in, there are a few conflicts that will have to be updated but it should be quick.

gravyboat:
@enzanki-ars How are you looking for this? There's no pressure to get it done I just want to get 0.6.0 built if you think it's going to be a bit still (we'll get it in for the next release no problem).

enzanki-ars:
My apologies for the long delay in response.  I am still looking into having the sorting algorithm output as discussed above.  I know that the `stream_weight()` method is the only major place that needs changed, but I am confused right now on why the `alt` streams are not being sorted the same way as non `alt` streams.  I know I am missing something, but it seems that the `alt` streams should already be sorting in the way that was discussed.

gravyboat:
@enzanki-ars No need to apologize, your update is greatly appreciated. Please don't feel pressured at all.

zraexy:
I made a pull request that fixes the ordering: https://github.com/enzanki-ars/streamlink/pull/1

zraexy:
@enzanki-ars Is there anything else that needs to be done before things can get merged? Based on my testing it works fine with the changes I made.

enzanki-ars:
@zraexy Sorry about the excessively long delay in merging that.  I saw the PR, but forgot to test it.  Will do that tonight.  At a glance, it looks like it will work well.

gravyboat:
@enzanki-ars Were you able to test this?

enzanki-ars:
@gravyboat I have merged in @zraexy's contributions as they work beautifully.  Removing `[WIP]` from the title as I think this completes the PR.

gravyboat:
@enzanki-ars Sounds great! Once the tests are all good we'll get this merged. Thanks again for your work on this, and thanks to @zraexy as well!

Quaraxkad:
I have two comments regarding this change:

First,
`re.match(r"^(\d+)(k|p)?(\d+)?(\+)?(?:_(\d+)k)?(?:_(alt)(\d)?)?$", stream)`
Should probably be:
`re.match(r"^(\d+)(k|p)?(\d+)?(\+)?(?:_(\d+)k)?(?:_(alt)(\d+)?)?$", stream)`
Just in case there are more than 9 alts. The current code does not allow that to occur, but that's a separate issue.

Second, what is the purpose of lowering the weight of alt streams? If they _are_ true mirrors, then their weights should be equal, no? And ideally that would be handled entirely behind-the-scenes instead of being exposed to the end user as being alts. If they _are not_ mirrors, then they should not be identified as "alt" in the first place, which again is a separate issue.

beardypig:
Currently there can be no more than 2 alts (relating to that other issue), _alts_ are handled automatically as sources to retry and should not be treated differently here.

Quaraxkad:
> and should not be treated differently here

Is that a confirmation that this code should not be as-is?

`if match.group(7): 
    weight -= ALT_WEIGHT_MOD * int(match.group(7)) 
else: 
    weight -= ALT_WEIGHT_MOD`

This applies a lower the weight to alt streams.

beardypig:
The alts should be sorted to have a lower priority, but the weight shouldn't be affected. However, at the moment I don't have the time to properly review the implications of this specific change - my gut reaction is that the weight shouldn't be changed, but it might not be a problem...

PR #739, opened 2017-03-28
Fix wrong vlc path for namedpipes

kaszim:
There seems to have been some changes to the way vlc reads named pipes. On master, currently, using named pipes to view a stream simply does not work. These changes allow named pipes to be created with the correct path, while still allowing vlc to read from them.

### DIFF
--- a/src/streamlink/utils/named_pipe.py
+++ b/src/streamlink/utils/named_pipe.py
@@ -19,6 +19,7 @@ class NamedPipe(object):
     def __init__(self, name):
         self.fifo = None
         self.pipe = None
+        self.name = name
 
         if is_win32:
             self.path = os.path.join("\\\\.\\pipe", name)
--- a/src/streamlink_cli/output.py
+++ b/src/streamlink_cli/output.py
@@ -101,7 +101,7 @@ class PlayerOutput(Output):
 
     def _create_arguments(self):
         if self.namedpipe:
-            filename = self.namedpipe.path
+            filename = "stream://\\\\\\.\\pipe\\" + self.namedpipe.name
         elif self.filename:
             filename = self.filename
         elif self.http:


beardypig:
Which OS are you using? Windows I would guess? Which version of VLC are you using? 

Unfortunately this change breaks VLC support on Linux and macOS, as well as breaking support for `mpv`, `mpc-hc`, etc. Perhaps there is a better fix for the issue you're experiencing.

gravyboat:
@kaszim Any chance you're interested in looking at the issues @beardypig has brought up?

kaszim:
You are correct the problem occured under windows 10 with both VLC 2.2.4 and VLC 3.0.0 (latest rev). I believe there have been some changes to how vlc handles named pipes, as the creation of the pipe works. However vlc cannot read from it if the default path for it is supplied.

 I'll take a look at the issues sometime this week.

gravyboat:
Awesome! Thanks for the update @kaszim.

kaszim:
Which way do you wish to have the fix implemented? I see two ways:

* Add an extra flag while running streamlink_cli that uses the changed path for the pipe.
* Simply check which os is running and change the path accordingly.

gravyboat:
@beardypig thoughts?

beardypig:
Well it already uses a different pipe syntax for Windows. I don't know how easy it will be to reliably detect VLC as the player, I assume the old syntax still works with other players? It might be easier to add a flag, or option that sets the flag as well as output to a fifo, `--vlc-fifo` perhaps?

gravyboat:
@beardypig That could definitely work and it avoids breaking issues as noted earlier in the PR.

kaszim:
I do not currently have time to fix this, even though it is a really simple fix. Consequently I am closing the PR to not clutter your repo. I can open it again when I am free from uni assignments 👍

gravyboat:
Thanks for the heads up @kaszim and good luck with your assignments!

Issue #743, opened 2017-03-30
Plugin: rtve.es - add support for quicktime (*.mov)


malvinas2:
### Checklist

- [x] This is a bug report.
- [x] This is a plugin request.
- [ ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description

For some old videos rtve.es seems to use quicktime (*.mov)-files. 
Is it possible to add this to streamlink ? 

Example: 
http://www.rtve.es/alacarta/videos/cantares/marife-triana-cantares-1978/1692684/
http://www.rtve.es/drmn/embed/video/1692684

**EDIT:** 
There also exist movies within an *.avi-containter not working at the moment, e.g. 
http://www.rtve.es/alacarta/videos/version-europea/vida-facil/3453399/
http://www.rtve.es/drmn/embed/video/3453399

gravyboat:
Hmm, I don't know if we can support .mov due to how licensing works: https://developer.apple.com/softwarelicensing/agreements/quicktime.php. I'll leave this open for others to weigh in though.

malvinas2:
As fas as I understood Streamlink delivers the URL of even AES-encrypted HLS streams, which indeed is a communications protocol implemented by _Apple_. So why do you start do care about licensing right now? 
And why should it be critical to return a URL-path which 'coincidentally' contains the term 'mov' ?

beardypig:
@malvinas2 `quicktime` has a difference license to `hls` even though they are both created by Apple. However, I don't see any problem in supporting those VOD streams as they are just URLs. I think Apple is more concerned with people using their trademarks or proprietary formats.

beardypig:
The list of streams returned by the API is a bit weird - it includes quite a few that don't actually work. I added a `HEAD` call for each of the HTTP URLs which might slow down the plugin a bit, but it won't list garbage `1080p` streams which don't actually exist. Standby for PR.

gravyboat:
Closing this since #745 is merged.

PR #745, opened 2017-03-31
plugins.rtve: support for avi/mov VOD streams

beardypig:
As requested by @malvinas2 in #743.

Supports the `.avi` and `.mov` VOD streams on RTVE :)

There is an extra `HEAD` request performed for each stream URL as the API returns some junk/non-existent stream URLs (`1080p` for example), and the `HEAD` request identifies the junk URLs. 

These should work now:
- rtve.es/alacarta/videos/cantares/marife-triana-cantares-1978/1692684/
- rtve.es/drmn/embed/video/1692684
- rtve.es/alacarta/videos/version-europea/vida-facil/3453399/
- rtve.es/drmn/embed/video/3453399

### DIFF
--- a/src/streamlink/plugins/rtve.py
+++ b/src/streamlink/plugins/rtve.py
@@ -122,7 +122,7 @@ class Rtve(Plugin):
         data = http.json(res, schema=self.video_schema)
         qmap = {}
         for item in data["qualities"]:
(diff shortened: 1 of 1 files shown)


gravyboat:
Nice, thanks @beardypig!

Issue #748, opened 2017-04-01
Periscope web URL has changed


solham:
### Checklist

- [x] This is a bug report.
- [ ] This is a plugin request.
- [ ] This is a feature request.
- [x] I used the search function to find already opened/closed issues or pull requests.

### Description

Periscope web URL has changed from www.periscope.tv to www.pscp.tv
periscope.py needs updating to reflect this change accordingly.

### Expected / Actual behavior

Current periscope.py does not recognize www.pscp.tv as a valid plugin address.

### Reproduction steps / Stream URLs to test

1. www.periscope.tv now redirects to www.pscp.tv
2. ...
3. ...

### Environment details (operating system, python version, etc.)

streamlink 0.4.0.1080

### Comments, logs, screenshots, etc.

Changing periscope.py line
_url_re = re.compile(r"http(s)?://(www\.)?periscope.tv/[^/]+/(?P<broadcast_id>[\w\-\=]+)")
to
_url_re = re.compile(r"http(s)?://(www\.)?pscp.tv/[^/]+/(?P<broadcast_id>[\w\-\=]+)")
temporarily fixes the issue

It appears twitter has done this in response to the Turkey situtation, unsure whether this change will be permanent. https://twitter.com/periscopetv/status/847956641450311680

gravyboat:
Closing this out since https://github.com/streamlink/streamlink/pull/749 is merged.

PR #749, opened 2017-04-01
New Periscope URL #748

CallMeJuf:
Old periscope URL still works / is accessible. Using new one temporarily (?) during legal action in turkey. API URL (STREAM_INFO_URL) not interchangeable yet.


gravyboat:

Issue #756, opened 2017-04-04
Plugin request(s): Deutsche Welle, Al Jazeera


roshavagarga:
### Checklist

- [ ] This is a bug report.
- [x] This is a plugin request.
- [ ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description
If possible, a link for two of the more famous news TV channes - Deutsche Welle's live TV and Al Jazeera.
The tricky part is that the former uses this link (http://www.dw.com/en/media-center/live-tv/s-100825) but you can switch between languages that are on the same link for some reason. Al Jazeera can be found on this link - http://www.aljazeera.com/live/. Both should be available worldwide.

gravyboat:
Closing since #758 and #759 are merged.

PR #758, opened 2017-04-04
plugins.deutschewelle: support for live dw streams

beardypig:
As requested in #756 by @roshavagarga.

Only supports the live streams, the different language streams are supported by including the query string argument `channel` in the URL.

Examples:
- dw.com/en/media-center/live-tv/s-100825 (default language)
- dw.com/en/media-center/live-tv/s-100825?channel=5 (German)
- dw.com/en/media-center/live-tv/s-100825?channel=1 (English)
- dw.com/en/media-center/live-tv/s-100825?channel=3 (Spanish)


gravyboat:

PR #759, opened 2017-04-04
plugins.aljazeeraen: support for live and vod streams from aljazeera.com

beardypig:
Only the English variant of the website is support (aljazeera.com).
Live streams and VOD are both supported.

A support class `BrightcovePlayer` was added to simplify the process of playing streams hosted by Brightcove.

Example URLs:
- aljazeera.com/programmes/witness/
- aljazeera.com/live/
- aljazeera.com/programmes/the-big-picture/2017/02/big-picture-people-america-170216063801575.html

I plan to update the `tv8cat` plugin to use the new `BrightcovePlayer` class when/if it is merged.


gravyboat:
Nice this looks good, thanks!

PR #763, opened 2017-04-04
Add readme note about supporting streamers

gravyboat:
This is similar to what @bastimeyer has over on the gui: https://github.com/streamlink/streamlink-twitch-gui#why and I wanted to make sure we're noting that for the CLI as well. I've slightly modified what he had to make sure we're also linking Twitch's official stance regarding ad blockers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,12 @@ The main purpose of streamlink is to convert CPU-heavy flash plugins to a less C
 
 Streamlink is a fork of the [livestreamer](https://github.com/chrippa/livestreamer) project.
 
+Please note that by using this application you're bypassing any ads run by
+Twitch (adblock users do as well and
+[Twitch has taken an agnostic stance regarding ad blockers](https://blog.twitch.tv/introducing-surestream-for-a-better-video-ad-experience-on-twitch-3ca5ce3287c).
+If you want to support Twitch or a single broadcaster, please consider buying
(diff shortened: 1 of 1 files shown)


bastimeyer:
Wouldn't it be better to add a text that targets all plugins and not just Twitch? Despite that, I agree that something like this should be included in the readme here. Maybe even on the website?!

Btw, if you want to still link to the Twitch blog, you can use the short URL format which is a bit better to read:
https://blog.twitch.tv/-3ca5ce3287c

README.md
```diff
@@ -9,6 +9,12 @@ The main purpose of streamlink is to convert CPU-heavy flash plugins to a less C
 
 Streamlink is a fork of the [livestreamer](https://github.com/chrippa/livestreamer) project.
 
+Please note that by using this application you're bypassing any ads run by
+Twitch (adblock users do as well and
```
cdrage: Perhaps saying something along the lines of `...bypassing any ads ran by plugins such as Twitch`. So it still shows that we remain plugin-agnostic (no favouritism!) xD

gravyboat:
Good suggestions, I'll update when I have time.

gravyboat:
I've made some changes, let me know what you think @cdrage and @bastimeyer.

cdrage:
LGTM! Although for some reason travis failed?

cdrage:
Nvm; false alarm, going to restart the job

PR #770, opened 2017-04-06
stream.hds: ensure the live edge does not go past the latest fragment

beardypig:
Fixes a bug in the `HDSStream`, in some cases the end fragment can be past the end of available fragments which results in 404 errors and stutters in the stream.


gravyboat:
Good catch, thanks @beardypig! If you got a notice that Travis failed it was just the run failing, restarted it and that took care of it.

Issue #766, opened 2017-04-05
Dailymotion bug on specific channel


karlo2105:
### Checklist

- [x] This is a bug report.
- [ ] This is a plugin request.
- [ ] This is a feature request.
- [x] I used the search function to find already opened/closed issues or pull requests.

### Description

While checking channels on DM, I just discovered this channel Cnews which is aired officially, doesn't work with streamlink. However it works on browser. 

### Expected / Actual behavior

Three addresses have been used for testing : 

`http://dai.ly/x3b68jn`
`http://www.dailymotion.com/video/x3b68jn_live-cnews_news`
`http://www.dailymotion.com/CNEWS`

### Reproduction steps / Stream URLs to test

```
streamlink "http://dai.ly/x3b68jn" best
[cli][info] Found matching plugin dailymotion for URL http://dai.ly/x3b68jn
error: No playable streams found on this URL: http://dai.ly/x3b68jn
```

```
streamlink "http://www.dailymotion.com/video/x3b68jn_live-cnews_news" best
[cli][info] Found matching plugin dailymotion for URL http://www.dailymotion.com
/video/x3b68jn_live-cnews_news
error: No playable streams found on this URL: http://www.dailymotion.com/video/x
3b68jn_live-cnews_news
```

```
streamlink "http://www.dailymotion.com/CNEWS" best
[cli][info] Found matching plugin dailymotion for URL http://www.dailymotion.com
/CNEWS
error: No playable streams found on this URL: http://www.dailymotion.com/CNEWS
```

### Environment details (operating system, python version, etc.)

W7 Pro/Streamlink Portable

### Comments, logs, screenshots, etc.

I wonder whether they use some DRM but it would be odd for a channel news which is freely aired on DM.

beardypig:
Seems the usual HDS stream URL failed. We can add a Chromecast HLS stream extractor as a fallback :)

beardypig:
See if this updated [dailymotion.py](https://raw.githubusercontent.com/beardypig/streamlink/b3c4c976835cc81ea7888a844bd9dde70da7dbeb/src/streamlink/plugins/dailymotion.py) plugin fixes your issue.

karlo2105:
It works for two addresses : shortcut and regular address. However on third nominative address I get same error message as above.

beardypig:
[...]

PR #773, opened 2017-04-07
Dailymotion feature video and backup stream fallback

beardypig:
Updated the method used to find the featured video for a channel so that it is more reliable. Also added a backup stream by way of the Chromecast stream URL that is embedded in the page, when all other stream URLs fail (as was the case in #766).


karlo2105:
Much appreciated. ;)

beardypig:
Forgot to make the PR yesterday :)

gravyboat:

PR #785, opened 2017-04-11
Will exit with exit code 1 if stream cannot be opened.

### LINKED ISSUE #781: [cli][error] Could not open stream doesn't exit with code 1, but with 0
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [x] I used the search function to find already opened/closed issues or pull requests.

### Description

When streamlink finds a stream, but can not (for whatever reason) open it, the exit code is 0 (success) and not 1 (error).

streamlink should exit with exit code 1 (error).

### Reproduction steps / Stream URLs to test

1. streamlink streamurl best
2. wait for it to open the stream
3. wait for error that it can not open the stream
4. wait for it to exit
5. echo $?
0

### Environment details (operating system, python version, etc.)

Linux, Python 3.6

### Comments, logs, screenshots, etc.

```bash
# streamlink hlsvariant://example/*.m3u8 best 
[cli][info] Found matching plugin stream for URL hlsvariant://example/*.m3u8
[cli][info] Available streams: 240p (worst), 480p (best)
[cli][info] Opening stream: 480p (hls)
[cli][error] Could not open stream: Unable to open URL: http://example/*.m3u8 (404 Client Error: Not Found for url: http://example/*.m3u8)
```
```bash
# echo $?
0
```

Germandrummer92:
Fixes #781 as reported by @strudelkuchen. 

There is a couple other cases where possible a console.exit is better than to just log and exit with success. Namely [L310](https://github.com/Germandrummer92/streamlink/blob/48fb299c556c80a1dd6ac7727abb6f97b4355f9a/src/streamlink_cli/main.py#L310) and/or [L314](https://github.com/Germandrummer92/streamlink/blob/48fb299c556c80a1dd6ac7727abb6f97b4355f9a/src/streamlink_cli/main.py#L314). not sure though if those cases can occur with a "normal" close of the stream (e.g. the streamer stops it) in which case they should stay like they are, or if it's indicative of any error and it should exit with exit code 1.

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -314,9 +314,9 @@ def read_stream(stream, output, prebuffer, chunk_size=8192):
                 break
     except IOError as err:
         console.exit("Error when reading from stream: {0}, exiting", err)
-
-    stream.close()
-    console.logger.info("Stream ended")
+    finally:
+        stream.close()
+        console.logger.info("Stream ended")
 
 
 def handle_stream(plugin, streams, stream_name):
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -249,6 +249,7 @@ def output_stream(stream):
         try:
             stream_fd, prebuffer = open_stream(stream)
             success_open = True
+            break
         except StreamError as err:
             console.logger.error("Try {0}/{1}: Could not open stream {2} ({3})", i+1, args.retry_open, stream, err)
 
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -244,14 +244,16 @@ def output_stream(stream):
     """Open stream, create output and finally write the stream to output."""
     global output
 
+    success_open = False
     for i in range(args.retry_open):
         try:
             stream_fd, prebuffer = open_stream(stream)
-            break
+            success_open = True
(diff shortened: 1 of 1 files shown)


beardypig:
Doesn't that break `--retry-open`, looks like it would exit on the first error instead of retrying? I think the exit should happen on L254 where the `return` is currently, and you'll have to store a value for `err` - the latest one would probably be fine. 

I agree with your comments for L310/L314; I would put exits for those, but also put L316/L317 in a `finally` block.

src/streamlink_cli/main.py
```diff
@@ -249,7 +249,7 @@ def output_stream(stream):
             stream_fd, prebuffer = open_stream(stream)
             break
         except StreamError as err:
-            console.logger.error("{0}", err)
+            console.exit("Could not open stream {0} ({1})", stream, err)
     else:
         return
```
beardypig: Exit should be here to preserve the function of `--retry-open`.

Germandrummer92:
Should be fixed now.  I removed the for else for a more verbose/readable version, and made the output a little more verbose as well.

gravyboat:
Looks good to me, and since @beardypig has approved it as well I'm going to squash and merge it. Thanks @Germandrummer92!

PR #787, opened 2017-04-12
Add support for pc-yourfreetv.com

melmorabity:
This plugin add support for the German TV platform http://pc-yourfreetv.com/. It provides many TV channel streams from Germany, Switzerland, UK, France and Italy.
Although it's free, it requires a valid username and password (the plugin provides the corresponding options).
Once connected, all streams on the following page should be supported:
http://pc-yourfreetv.com/index_livetv.php?page_id=1


gravyboat:
Looks good to me, thanks @melmorabity!

Issue #805, opened 2017-04-17
Plugin request - Telefe VOD


mcaldas:
### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description
Would be very grateful if someone can help making a plugin for http://telefe.com/capitulos-completos/

they use akamai streaming, so should be straightforward. The only issue is that it's geoblocked to argentina only (but i can help debug if needed)
The stream url looks something like this:

http://telefe-vh.akamaihd.net/i/TOK/Amar_Despues_de_Amar/201702/AR/TLF6367307-AMAR_DESPUES_DE_AMAR-ARGENTINA_,1000k,200k,.mp4.csmil/master.m3u8?hdnea=st=1492466426~exp=1492473626~acl=/i/TOK/Amar_Despues_de_Amar/201702/AR/TLF6367307-AMAR_DESPUES_DE_AMAR-ARGENTINA_,1000k,200k,.mp4.csmil/*~hmac=f10ff8c2da31e40a014112edc9c569327a67cfb48f75f753eeb5379931fd3ec3

Thanks in advance!

RosadinTV:
Yo también soy de Argentina, Python no es mi fuerte pero probablemente pueda lograr crear un Plugin 😄

mcaldas:
@RosadinTV si queres te ayudo, tampoco es mi fuerte, pero algo capaz sale

RosadinTV:
@mcaldas No lo probe a fondo pero parece funcionar bien:

**telefe.py**

```
import re

from streamlink.plugin import Plugin
from streamlink.plugin.api import http, useragents, validate
from streamlink.stream import HLSStream


class Telefe(Plugin):
    _url_re = re.compile(r'https?://telefe.com/.+')

    @classmethod
    def can_handle_url(cls, url):
        return Telefe._url_re.match(url)

    def _get_streams(self):
        TELEFE_UA = "Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.133 Safari/537.36 OPR/44.0.2510.1218"
        res = http.get(self.url, headers={'User-Agent': TELEFE_UA})
        busqueda_videoID = res.content.decode()
        busqueda_videoID = busqueda_videoID[busqueda_videoID.index('"PlayerContainer","model":{"id":') + 32 :]
[...]

PR #806, opened 2017-04-18
Add support for Telefe (VOD) and ElTreceTV (VOD and Live)

RosadinTV:
As requested in https://github.com/streamlink/streamlink/issues/805

### DIFF
--- a/src/streamlink/plugins/telefe.py
+++ b/src/streamlink/plugins/telefe.py
@@ -1,0 +1,34 @@
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import http, useragents, validate
+from streamlink.stream import HLSStream
+
+
+class Telefe(Plugin):
+    _url_re = re.compile(r'https?://telefe.com/.+')
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return Telefe._url_re.match(url)
+
+    def _get_streams(self):
+        TELEFE_UA = "Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.133 Safari/537.36 OPR/44.0.2510.1218"
+        res = http.get(self.url, headers={'User-Agent': TELEFE_UA})
+        busqueda_videoID = res.content.decode()
+        busqueda_videoID = busqueda_videoID[busqueda_videoID.index('"PlayerContainer","model":{"id":') + 32 :]
+        busqueda_videoID = busqueda_videoID[: busqueda_videoID.index(',')]
+        #print('ID encontrado: ' + busqueda_videoID)
+
+        hls_streams = HLSStream.parse_variant_playlist(
+            self.session,
+            "http://telefe.com/Api/Videos/GetSourceUrl/" + busqueda_videoID + "/0/HLS",
+            headers={'Referer': self.url,
+            'User-Agent': TELEFE_UA,
+            'X-Requested-With': 'ShockwaveFlash/25.0.0.148'}
+        )
+        for s in hls_streams.items():
+            yield s       
+
+
+__plugin__ = Telefe
--- a/src/streamlink/plugins/telefe.py
+++ b/src/streamlink/plugins/telefe.py
@@ -2,7 +2,8 @@ import re
 
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import http, useragents, validate
-from streamlink.stream import HLSStream
+from streamlink.stream import HLSStream, HTTPStream
+from streamlink.utils import parse_json
 
 
 class Telefe(Plugin):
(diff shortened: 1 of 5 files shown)


gravyboat:
Looks okay to me, do you want to write some tests?

RosadinTV:
@gravyboat Sure, the plugin makes a very basic use of Regex but I will add a test.

src/streamlink/plugins/telefe.py
```diff
@@ -0,0 +1,34 @@
...
+    def can_handle_url(cls, url):
+        return Telefe._url_re.match(url)
+
+    def _get_streams(self):
+        TELEFE_UA = "Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.133 Safari/537.36 OPR/44.0.2510.1218"
+        res = http.get(self.url, headers={'User-Agent': TELEFE_UA})
+        busqueda_videoID = res.content.decode()
+        busqueda_videoID = busqueda_videoID[busqueda_videoID.index('"PlayerContainer","model":{"id":') + 32 :]
+        busqueda_videoID = busqueda_videoID[: busqueda_videoID.index(',')]
+        #print('ID encontrado: ' + busqueda_videoID)
```
beardypig: You could make this line `self.logger.debug('ID encontrado: {0}', busqueda_videoID)` so that if there is a problem later we can get some debugging information :)

src/streamlink/plugins/telefe.py
```diff
@@ -0,0 +1,34 @@
...
+
+class Telefe(Plugin):
+    _url_re = re.compile(r'https?://telefe.com/.+')
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return Telefe._url_re.match(url)
+
+    def _get_streams(self):
+        TELEFE_UA = "Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.133 Safari/537.36 OPR/44.0.2510.1218"
```
beardypig: There are a number of predefined user agent string that could be used here instead, eg. `useragents.CHROME` (you already have the `useragents` module imported 😄)

src/streamlink/plugins/telefe.py
```diff
@@ -0,0 +1,34 @@
...
+        res = http.get(self.url, headers={'User-Agent': TELEFE_UA})
```
beardypig: You can use `{'User-Agent': useragents.CHROME}` instead, same goes for the other `http.get`.

src/streamlink/plugins/telefe.py
```diff
@@ -0,0 +1,34 @@
...
+        busqueda_videoID = res.content.decode()
+        busqueda_videoID = busqueda_videoID[busqueda_videoID.index('"PlayerContainer","model":{"id":') + 32 :]
+        busqueda_videoID = busqueda_videoID[: busqueda_videoID.index(',')]
+        #print('ID encontrado: ' + busqueda_videoID)
+
+        hls_streams = HLSStream.parse_variant_playlist(
+            self.session,
+            "http://telefe.com/Api/Videos/GetSourceUrl/" + busqueda_videoID + "/0/HLS",
```
beardypig: An HTTP version of the stream is also available, just replace `HLS` with `HTTP` - I'm not sure if it works with all streams. You could instead parse the json object `T3.content.cache` (ignoring the empty one) and extract the source URLs from that.

src/streamlink/plugins/telefe.py
```diff
@@ -0,0 +1,34 @@
...
+    _url_re = re.compile(r'https?://telefe.com/.+')
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return Telefe._url_re.match(url)
+
+    def _get_streams(self):
+        TELEFE_UA = "Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.133 Safari/537.36 OPR/44.0.2510.1218"
+        res = http.get(self.url, headers={'User-Agent': TELEFE_UA})
+        busqueda_videoID = res.content.decode()
```
beardypig: `res.content.decode()` is equivalent to `res.text`.

beardypig: I know it's a small thing, but Python variables should be lowercase with `_` separating words  (as is the case throughout streamlink, or it should be 😢) eg. `busqueda_videoID` becomes `busqueda_video_id`.

src/streamlink/plugins/telefe.py
```diff
@@ -0,0 +1,34 @@
...
+from streamlink.plugin.api import http, useragents, validate
+from streamlink.stream import HLSStream
+
+
+class Telefe(Plugin):
+    _url_re = re.compile(r'https?://telefe.com/.+')
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return Telefe._url_re.match(url)
```
beardypig: While this class is unlikely to be subclassed it is better practise to use `cls` instead of the current class, eg. `return cls._url_re.match(url)` that way, if the class is subclassed `_url_re` could be override. It doesn't matter so much in this case, but it does in other parts of the code and is _generally_ a better idea.

tests/test_plugin_telefe.py
```diff
@@ -0,0 +1,15 @@
...
+class TestPluginTelefe(unittest.TestCase):
+    def test_can_handle_url(self):
+        # should match
+        self.assertTrue(Telefe.can_handle_url("http://telefe.com/pone-a-francella/temporada-1/programa-01/"))
+        self.assertTrue(Telefe.can_handle_url("http://telefe.com/los-simuladores/temporada-1/capitulo-01/"))
+        self.assertTrue(Telefe.can_handle_url("http://telefe.com/dulce-amor/capitulos/capitulo-01/"))
+
+        # shouldn't match
+        self.assertFalse(Telefe.can_handle_url("http://local.local/"))
+        self.assertFalse(Telefe.can_handle_url("http://localhost.localhost/"))
```
beardypig: You could test here that it does not match `http://telefe.com/`, as per the regex:
```python
self.assertFalse(Telefe.can_handle_url("http://telefe.com/"))
```

RosadinTV:
@beardypig Thanks for the review, i'm really newbie programming in Python, so these tips are very useful.

mcaldas:
ok, i want to give this a try, how can i do? need to build from source?

src/streamlink/plugins/telefe.py
mcaldas: `res.text` should be used to avoid encoding issues, otherwise we get errors like this:
`error: 'ascii' codec can't decode byte 0xc3 in position 4316: ordinal not in range(128)`

gravyboat:
@RosadinTV As was mentioned in the issue related to this request, can you please make sure all the variables are in English? Thanks!

beardypig:
And the log messages :-)

RosadinTV:
Done, originally left everything in Spanish because Telefe is only available in Argentina (and all its content and website only in Spanish)
Then I thought that those interested in modifying or updating the Plugin would also be Spanish-speaking users.
But definitely the language of the project is English and I must stick to that.
In addition, users of other languages like @beardypig collaborate despite the limitations of geographical block and the best way everyone can handle it is by using the English language.
Thanks for the advices, they always help to progress :)

beardypig:
@RosadinTV you should open a separate PR to add el trece, each plugin needs its own PR.

RosadinTV:
Ok, it's also a good opportunity to make the PR cleaner. Tomorrow i will open a different PR for each Plugin (and separate the fork in branches, i forget to do it, this is the reason why I put two Plugins in the same PR)
I am a living fault :P

bastimeyer:
You can always rebase your PR branch and then force-push it. This will update the PR automatically without the need of opening a new one.

RosadinTV:
@bastimeyer 
This time I went through the easy path and re-created a fork with separate branches for each Plugin.
Sorry for the inconvenience 😢

PR #808, opened 2017-04-18
tests: cmdline tests can fail if there are some config options set

beardypig:
If `stream-types` is set to anything that doesn't include `test` as a stream type, then the tests for the cmd line will fail.

Added the option to assert a specific exit code as well.


gravyboat:
Good catch.

PR #809, opened 2017-04-18
cli: add --url option that can be used in config files to set a URL

### LINKED ISSUE #777: Feature request: Personal "channels"
- [ ] This is a bug report.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description
I've been browsing the information on the configuration pages, but I can't find a way to achieve something I want to do. I would like to have the ability to save streams I frequently watch and reload them with a simpler command, like "streamlink one" for instance. It would save users a lot of time, since longer/weirder links can't be memorized easily and have to be sought out or kept as a bookmark. On lower-spec PCs running Linux, opening a browser and waiting for a bookmark to load can require a lot of time and CPU usage. 

Excuse me if there's already a way of doing this, I was unable to find it.

beardypig:
Fixes #777 by allowing a URL to be set in a config file that can be loaded on demand using the @ syntax.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -467,6 +467,15 @@ output.add_argument(
 
 stream = parser.add_argument_group("Stream options")
 stream.add_argument(
+    "--url",
+    dest="url_param",
+    metavar="URL",
+    help="""
+    Alternative parameter to specify a URL to attempt to
+    extract streams from.
+    """
+)
+stream.add_argument(
     "--default-stream",
     type=comma_list,
     metavar="STREAM",
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -166,6 +166,7 @@ positional.add_argument(
     A URL to attempt to extract streams from.
 
     If it's a HTTP URL then "http://" can be omitted.
+    The URL can also be specified using the --url option.
     """
 )
 positional.add_argument(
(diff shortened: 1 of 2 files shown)


src/streamlink_cli/argparser.py
```diff
@@ -467,6 +467,15 @@ def boolean(value):
 
 stream = parser.add_argument_group("Stream options")
 stream.add_argument(
+    "--url",
+    dest="url_param",
+    metavar="URL",
+    help="""
+    Alternative parameter to specify a URL to attempt to
```
gravyboat: Is there a way we could make this more user friendly in terms of explanation? I feel like it is a bit confusing, not sure if there is a better way though.

gravyboat:
With the addition of this we should probably do some doc updates, more people have been asking about this kind of thing, and even `--default-stream` isn't explained very well, there aren't even any examples.

src/streamlink_cli/argparser.py
beardypig: We could suppress the help and then explain it as a special case in the docs.
I'll have a go at writing something a bit more friendly though :-)

gravyboat: I think it's good to have the help section there, just wondering if there's anything we can do to make it more clear (though I don't have a clarity suggestion so I can't really talk). I'm thinking we should probably start a doc page that specifically explains things like the configuration file. Maybe a more in-depth getting started guide or something.

beardypig: How about this version? :)

I agree about the config guides/etc. there are a lot of options, and the specific behaviour of those options is not exactly well documented :)

gravyboat:
Good changes, thanks!

PR #812, opened 2017-04-19
cli: add wildcard to --stream-types option

beardypig:
When changing the stream priority using `--stream-types` any stream types that are not explicitly included are ignored. For example if a user wants HTTP streams to be preferred and then HLS streams, with the other stream types in any order they might expect that `--stream-types "http,hls"` would be sufficient. However, any stream type not included in this list will be omitted. The default behavior for the `--stream-types` option means that other types are included.

This commit introduces a wildcard character (`*`), where any stream type not listed can be matched. For example, to order the streams with `HTTP`, then `HLS`, then others, the user could specify the `--stream-types` option as `"http,hls,*"`.
The documentation has been clarified to highlight the fact that other stream types not listed will be omitted and that a wildcard character can be used to match the other stream types. The addition of the wildcard does not change the existing behaviour of `--stream-types`.

### DIFF
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -1,6 +1,5 @@
 import operator
 import re
-
 from functools import partial
 
 from ..cache import Cache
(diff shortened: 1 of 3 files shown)


gravyboat:
This is a really good idea, very cool. Thanks @beardypig!

PR #813, opened 2017-04-19
plugins.rtve: stop IOError bubbling up on 404 errors

beardypig:
On some platforms catching OSError is not sufficient.


gravyboat:
Nice, thanks @beardypig!

PR #815, opened 2017-04-19
Add support for Telefe (VOD)

RosadinTV:
As issued in https://github.com/streamlink/streamlink/issues/805


gravyboat:
Looks good, thanks @RosadinTV!

PR #816, opened 2017-04-19
Add support for ElTreceTV (VOD & Live)

RosadinTV:
As suggested in https://github.com/streamlink/streamlink/issues/805

### DIFF
--- a/src/streamlink/plugins/eltrecetv.py
+++ b/src/streamlink/plugins/eltrecetv.py
@@ -24,7 +24,7 @@ class ElTreceTV(Plugin):
             http.headers = {'Referer': self.url,
             'User-Agent': useragents.CHROME}
             res = http.get(self.url)
-            _player_re = re.compile(r'''data\-kaltura="([^"]+)"''')
+            _player_re = re.compile(r'''data-kaltura="([^"]+)"''')
             match = _player_re.search(res.text)
             if not match:
                 return
--- a/src/streamlink/plugins/eltrecetv.py
+++ b/src/streamlink/plugins/eltrecetv.py
@@ -1,0 +1,41 @@
+import re, html
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import http, useragents, validate
+from streamlink.stream import HLSStream, HTTPStream
+from streamlink.utils import parse_json
+
+
+class ElTreceTV(Plugin):
+    _url_re = re.compile(r'https?://(?:www\.)?eltrecetv.com.ar/.+')
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls._url_re.match(url)
+
+    def _get_streams(self):
+
+        video_url_found_hls = ""
+
+        if self.url[self.url.index('eltrecetv.com.ar/') +17:].startswith("vivo"):
+            http.headers = {'Referer': self.url,
+            'User-Agent': "ArtearPlayer/3.2.42 (Linux;Android 4.4.2) ExoPlayerLib/1.5.9"}
+            video_url_found_hls = "http://stream.eltrecetv.com.ar/live13/13tv/13tv1/playlist.m3u8"
+        else:
+            http.headers = {'Referer': self.url,
+            'User-Agent': useragents.CHROME}
+            res = http.get(self.url)
+            video_search = res.text
+            video_search = video_search[video_search.index('data-kaltura="') +14:]
+            video_search = video_search[: video_search.index('"')]
+            video_search = html.unescape(video_search)            
+            json_video_search = parse_json(video_search)
+            video_url_found_hls = "https://vodgc.com/p/111/sp/11100/playManifest/entryId/" + json_video_search["entryId"] + "/format/applehttp/protocol/https/a.m3u8"
+         
+        if video_url_found_hls:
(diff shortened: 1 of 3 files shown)


src/streamlink/plugins/eltrecetv.py
```diff
@@ -0,0 +1,41 @@
...
+            video_url_found_hls = "http://stream.eltrecetv.com.ar/live13/13tv/13tv1/playlist.m3u8"
+        else:
+            http.headers = {'Referer': self.url,
+            'User-Agent': useragents.CHROME}
+            res = http.get(self.url)
+            video_search = res.text
+            video_search = video_search[video_search.index('data-kaltura="') +14:]
+            video_search = video_search[: video_search.index('"')]
+            video_search = html.unescape(video_search)            
+            json_video_search = parse_json(video_search)
```
back-to: you could replace `video_search` and `json_video_search` with this,
to fix the travis error and you need to remove the `import html`
```py
            _player_re = re.compile(r'''data\-kaltura="([^"]+)"''')
            match = _player_re.search(res.text)
            if not match:
                return
            json_video_search = parse_json(match.group(1).replace("&quot;", '"'))
```

src/streamlink/plugins/eltrecetv.py
```diff
@@ -0,0 +1,41 @@
...
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls._url_re.match(url)
+
+    def _get_streams(self):
+
+        video_url_found_hls = ""
+
+        if self.url[self.url.index('eltrecetv.com.ar/') +17:].startswith("vivo"):
```
back-to: you could do it like this, its a bit shorter.
```py
if "eltrecetv.com.ar/vivo" in self.url:
```

src/streamlink/plugins/eltrecetv.py
```diff
@@ -0,0 +1,41 @@
...
+            video_search = video_search[: video_search.index('"')]
+            video_search = html.unescape(video_search)            
+            json_video_search = parse_json(video_search)
+            video_url_found_hls = "https://vodgc.com/p/111/sp/11100/playManifest/entryId/" + json_video_search["entryId"] + "/format/applehttp/protocol/https/a.m3u8"
+         
+        if video_url_found_hls:
+            hls_streams = HLSStream.parse_variant_playlist(self.session, video_url_found_hls)
+            for s in hls_streams.items():
+                yield s  
+
```
back-to: There is **whitespace**, it doesn't really affect the code.
But it looks better without it 👓 

github1

RosadinTV:
@back-to Thanks for the review 🎁
I will apply your suggestions, except for the regex in the first one, i prefer to stay in the more user-readable way.

beardypig:
Personally I agree with @back-to regarding the [regex](https://xkcd.com/208/), I find it easier to read, more understandable and easier to maintain. That being said; _when you're a hammer everything looks like a nail._

RosadinTV:
@beardypig In my case I see Substring easier to edit, because it is very easy to read for someone with basic knowledge, when I see a regex I can not understand it at first sight (unless one knows regular expressions)
So taking this into account and that both methods work, I preferred to use Substring, but if you consider that I should use Regex inevitably no problem, let me know and I will change it 😉 
PS: That meme made my day :rofl:

gravyboat:
I'm fine with either and would like to get this merged, @beardypig you've got the final say on this one, I'm not particularly passionate about one way over another.

beardypig:
OK. Merge it like this, if they change the site at all we can make it more robust with a regex ;)

gravyboat:
@beardypig I mean I would rather just fix it now if that's a concern as opposed to potentially having issues down the road.

RosadinTV:
Regex vs Substring: The final battle 😆
regex vs substring - simspons

gravyboat:
Can we just swap this to regex so the PR can get merged as the regex desire outweighs the substring desire please? At this point I just want to release 0.6.0 and this is blocking.

RosadinTV:
@gravyboat Sure, no problem :wink:

src/streamlink/plugins/eltrecetv.py
```diff
@@ -24,16 +24,16 @@ def _get_streams(self):
             http.headers = {'Referer': self.url,
             'User-Agent': useragents.CHROME}
             res = http.get(self.url)
-            video_search = res.text
-            video_search = video_search[video_search.index('data-kaltura="') +14:]
-            video_search = video_search[: video_search.index('"')]
-            video_search = video_search.replace("&quot;", '"')
-            json_video_search = parse_json(video_search)
+            _player_re = re.compile(r'''data\-kaltura="([^"]+)"''')
```
beardypig: I don't think you need to escape the `-` as it's not in a character range `[]`.

RosadinTV: You are right, i just copy and paste the @back-to regex suggestion :stuck_out_tongue:

RosadinTV:
I think everything is in order now.
Thank you all ( @back-to @gravyboat @beardypig ) :cake:

gravyboat:
@RosadinTV Thanks for all these fixes, I'm going to merge this now. The appveyor fail is due to something fixed elsewhere.

PR #830, opened 2017-04-23
plugins.showroom: Mimic browser request headers

wlerin:
Recently, presumably to fight a wave of view/follow bots, Showroom has started blocking unknown User-Agents from their webpages and api. This PR sends the necessary information so streamlink doesn't get caught in the block.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ install:
 
 script:
   - python -m pytest tests/
-  - coverage run setup.py test
+  - coverage run -m pytest tests/
   # test building the docs
   - if [[ $BUILD_DOCS == 'yes' ]]; then make --directory=docs html; fi
   - if [[ $BUILD_INSTALLER == 'yes' ]]; then ./script/makeinstaller.sh; fi
--- a/src/streamlink/plugins/showroom.py
+++ b/src/streamlink/plugins/showroom.py
@@ -74,7 +74,7 @@ _info_pages = set((
 
 
 class Showroom(Plugin):
-    @staticmethod
+    @classmethod
     def can_handle_url(cls, url):
         match = _url_re.match(url)
         if not match or match.group("room_title") in _info_pages:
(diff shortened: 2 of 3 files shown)


wlerin:
It doesn't look like the travis build failure is related to my changes.

Also, I have two questions unrelated to this PR but related to the Showroom plugin:

If I play the RTMP stream directly through MPV it's a few seconds behind, but if I play it through streamlink it's like 10-15 seconds behind. Is there a way to disable or reduce this delay? I assume it's some kind of caching. Perhaps sending --passthrough or whatever that option was would help.

Is there a way to retry automatically when a stream dies? I think I saw the dailymotion plugin retrying a stream earlier, but I don't see anything about that in the plugin.

gravyboat:
@wlerin If you rebase tests should pass again thanks to #832.

gravyboat:
I think we need a rebase on this so the travis.yml changes don't get pulled in twice. Unless you can think of an easier method @beardypig.

wlerin:
If necessary I can delete the branch and make another PR, the changes weren't very complex.

beardypig:
You should be able to fix it with a rebase. 

I'd do it like this (I have two remotes, my fork on origin, and upstream as the streamlink repo).
```
git checkout master # update master first
git pull upstream master
git checkout showroom 
git reset --hard 50eecaa # go back to your last change commit 
git rebase master # rebase master :)
```

wlerin:
I already attempted something like that, which brought us to the current state. The problem is when I try to push, this happens:

```
error: failed to push some refs to 'https://github.com/wlerin/streamlink.git'
hint: Updates were rejected because the tip of your current branch is behind```

**Comment by bastimeyer** [comment]
Whenever you're rebasing a branch and new commit IDs get generated, you need to force-push to the remote repo, because branches that differ (new commit or ancestor IDs) can't be updated regularly:
`git push --force REMOTE-NAME BRANCH-NAME`

```
$ man git-push
...
       -f, --force
           Usually, the command refuses to update a remote ref that is not an ancestor of the local ref used to overwrite
           it. Also, when --force-with-lease option is used, the command refuses to update a remote ref whose current value
           does not match what is expected.

           This flag disables these checks, and can cause the remote repository to lose commits; use it with care.
...
```

PR #833, opened 2017-04-24
plugins.pcyourfreetv: add token parameters

melmorabity:
This PR fixes the pcyourfreetv plugin which doesn't work anymore:
```
streamlink --pcyourfreetv-username "@@@@@" --pcyourfreetv-password "@@@@@" "http://pc-yourfreetv.com/index_player.php?channel=cartoon%20network&page_id=41"
[cli][info] Found matching plugin pcyourfreetv for URL http://pc-yourfreetv.com/index_player.php?channel=cartoon%20network&page_id=41
[plugin.pcyourfreetv][info] Successfully logged in as @@@@@
error: No playable streams found on this URL: http://pc-yourfreetv.com/index_player.php?channel=cartoon%20network&page_id=41
```
Pc-YourFreeTV.com now requires a single token per connection in the URL parameters.
This PR will fetch and add such a token to the passed URL automatically.

PR #834, opened 2017-04-24
 plugins.pcyourfreetv: add token parameters

melmorabity:
See https://github.com/streamlink/streamlink/pull/833


gravyboat:
Thanks @melmorabity!

PR #841, opened 2017-04-25
[tv1channel.org] New plugin and some changes on [oldlivestream]

back-to:
Some changes on **oldlivestream** so it will work for **tv1channel**

**oldlivestream**
- Support for `cdn` subdomains
- Support for `embed/` streams

**tv1channel**
- New Plugin for embeded streams on tv1channel.org _(youtube.com and livestream.com)_


gravyboat:
Nice this looks good, thanks for adding those tests as well!

Issue #721, opened 2017-03-20
Play video from this url ? 


devwhatsapp:
Hi

How can we play videos from this url in vlc ?

http://cyro.se/watch/homeland/s6/e9

The page loads iframe of google video player.

The videos are stored in google drive or google docs.

Please if anyone can make a plugin to play video on vlc from such url which uploads video on google video servers.

Regards

beardypig:
This is a [work-in-progress plugin](https://github.com/beardypig/streamlink/tree/googledocs) for playing videos hosted on Google Docs - needs more testing :)

back-to:
the regex needs an update
```
https://docs.google.com/file/d/--ID--/edit
https://docs.google.com/file/d/--ID--/view
```
you can watch videos with this aswell.

----

and maybe a check for `fmt_list` in `data`
to avoid `KeyError: 'fmt_list'` for files without a video or dead links

```python
data = dict(parse_qsl(res.text))
if "fmt_list" in data:
    fmts = dict([s.split('/')[:2] for s in data["fmt_list"].split(",")])
```

beardypig:
Thanks for the extra info @back-to and for testing it out :)

PR #842, opened 2017-04-25
[cyro] New plugin for embeded streams from cyro.se

back-to:
Fixed streamlink/streamlink#721


beardypig:
Nice... Forgot about that bit :)

gravyboat:
Looks good to me, thanks @beardypig and @back-to!

PR #852, opened 2017-04-26
[ArteTV] new regex, removed rtmp and better result for available streams

back-to:
- New regex to match **new urls**
- New regex to match **old urls** (some of the old urls still work, some are 404)
- New regex to match **ignore playlists without video ids**
- **Removed rtmp** streams, I don't think they support rtmp anymore
- **Better result for streams**, it will now only show the streams with the same language as the url.

Before (all available languages, only two languages here, but there are streams with five)
```
[cli][info] Found matching plugin artetv for URL http://www.arte.tv/guide/de/074633-001-A/das-testament-des-dr-mabuse
Available streams:
406p_http_alt, 720p_alt, 406p_http, 360p_http_alt, 216p_http, 720p_http,
720p_http_alt, 216p_alt, 360p_http, 61k_alt, 360p_alt, 406p_alt, 216p_http_alt,
61k (worst), 216p, 360p, 406p, 720p (best)
```
After (only language from url, here **de**)
```
[cli][info] Found matching plugin artetv for URL http://www.arte.tv/guide/de/074633-001-A/das-testament-des-dr-mabuse
Available streams:
406p_http, 216p_http, 720p_http, 360p_http,
61k (worst), 216p, 360p, 406p, 720p (best)
```

### DIFF
--- a/src/streamlink/plugins/artetv.py
+++ b/src/streamlink/plugins/artetv.py
@@ -6,18 +6,20 @@ from itertools import chain
 
 from streamlink.compat import urlparse
 from streamlink.plugin import Plugin
-from streamlink.plugin.api import http, validate
-from streamlink.stream import HDSStream, HLSStream, HTTPStream, RTMPStream
+from streamlink.plugin.api import http
+from streamlink.plugin.api import validate
+from streamlink.stream import HDSStream
+from streamlink.stream import HLSStream
+from streamlink.stream import HTTPStream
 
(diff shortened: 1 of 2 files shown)


jalba69:
Hello, Not sure if i should post here or create a new issues report but Arte live is apparently not working with the build 5.0
There is 2 live stream URL's ( this TV channel is present in France and Germany )  : 

http://www.arte.tv/fr/direct/     In French language
http://www.arte.tv/de/live/       In German language 

Thank you very much and sorry if it's the wrong place to post this.

gravyboat:
Thanks @back-to!

PR #853, opened 2017-04-26
[Facebook] New plugin for 360p HLS streams

### LINKED ISSUE #226: Support FacebookLive streaming
My own research shows that Facebook Live streams are MPEG-DASH format with the audio and video in separate streams that need to be remuxed.  Once this project supports DASH video, we should also be able to support Facebook Live streams.

### LINKED ISSUE #845: Add already-created facebook plugin
### Checklist

- [ ] This is a bug report.

### Description
Add an already-created plugin for facebook videos (non-Dash, 360p quality). A link can be found in #226, the original creator left if there and allowed anybody to work on it and use it. Since we have no way of viewing facebook videos, I think this would be a very good addition for the upcoming 0.6 version.

back-to:
- Added @ZP [(gist)](https://gist.github.com/zp/c461761565dba764c90548758ee5ae9f) with a patch file / you can see in the [filehistory ](https://github.com/back-to/streamlink/blame/604806374d36f0e539d8fb42cdd9a5475244aa84/src/streamlink/plugins/facebook.py) that @ZP wrote the plugin 
- Added unittests
- Small regex change

since streamlink does not support DASH we can close the Issues

Fixed #226
Fixed #845

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -95,6 +95,7 @@ earthcam            earthcam.com         Yes   Yes   Only works for the cams hos
 eurocom             eurocom.bg           Yes   No
 euronews            euronews.com         Yes   No
 expressen           expressen.se         Yes   Yes
+facebook            facebook.com         Yes   No    Only 360p HLS streams.
 filmon              filmon.com           Yes   Yes   Only SD quality streams.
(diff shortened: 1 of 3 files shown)


gravyboat:
Looks good, thanks @back-to.

PR #850, opened 2017-04-26
plugin.tvplayer: update to support new site layout

beardypig:
As pointed out in #849 @nicr4wks the tvplayer.com website has been updated. This PR updates the plugin to support the changes.

Issue #849, opened 2017-04-26
tvplayer.com site update


nicr4wks:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description

The website tvplayer.com has recently undergone an update and now the tvplayer plugin is not able to stream video.

streamlink http://tvplayer.com/watch/clubland best
[cli][info] Found matching plugin tvplayer for URL http://tvplayer.com/watch/clubland
error: No playable streams found on this URL: http://tvplayer.com/watch/clubland

### Expected / Actual behavior

Prior to the website update streamlink found the video stream and played without issue.

### Reproduction steps / Explicit stream URLs to test

1. streamlink http://tvplayer.com/watch/clubland

### Environment details

Operating system and version: ...
Debian Jessie

Streamlink and Python version: ...
streamlink 0.5.0
Python 2.7.9

...

### Comments, logs, screenshots, etc.

...

gravyboat:
Closing since #850 is merged.

nicr4wks:
Thankyou @beardypig !

I've updated to latest git and built but now getting `error: Unable to validate JSON: Unable to validate key 'tvplayer': Unable to validate result: Unable to validate key 'response': Unable to validate result: Unable to validate key 'stream': Unable to validate URL attribute 'scheme': u'https' does not equal 'http'`

Is there something extra required for the new plugin?

beardypig:
Looks like they upgraded to https since the PR was created. I have created a new PR to resolve this.

ppsspp213:
How do I use the login?
streamlink https://tvplayer.com/watch/4seven best --tvplayer-email/ bugmenot@mailinator.com --tvplayer-
password bugmenot123

usage: streamlink-script.py [OPTIONS] <URL> [STREAM]
streamlink-script.py: error: unrecognized arguments: --tvplayer-email/ bugmenot@
mailinator.com

nicr4wks:
Try: streamlink --tvplayer-email bugmenot@mailinator.com --tvplayer-password bugmenot123 https://tvplayer.com/watch/4seven

ppsspp213:
error: Unable to open URL: http://api.tvplayer.com/api/v2/stream/live (401 Clien
t Error: Unauthorized for url: http://api.tvplayer.com/api/v2/stream/live)

I know it works with login but it uses 
[...]

PR #867, opened 2017-05-02
plugins.tvplayer: allow https stream URLs

beardypig:
Additional fix for #849, changed since the original PR #850.


nicr4wks:
Thanks @beardypig ! Confirmed working.

beardypig:
No problem, I think they must have changed it since my original PR :(

gravyboat:
If this service continues changing I'm going to add it to the list of services not worth supporting. Thanks @beardypig.

beardypig:
To be fair if hasn't changed for a couple of years and they just had a new design, I doubt it'll change again anytime soon :-)

gravyboat:
Hopefully so!

Issue #859, opened 2017-04-28
NRK + DR plug-ins in need of a brush-up.


Ordval:
- [x]  This is a plugin (improvement) request.
### Description
The NRK + DR plug-ins seem to be in need of a brush-up.
### Reproduction steps / Explicit stream URLs to test
C:\Users\Ordval>streamlink.exe "https://radio.nrk.no/serie/ren-60/DMPO45004516/10-11-2016" best
[cli][info] Found matching plugin nrk for URL https://radio.nrk.no/serie/ren-60/DMPO45004516/10-11-2016
error: Unable to open URL: mediaelement/DMPO45004516 (Invalid URL 'mediaelement/DMPO45004516': No schema supplied. Perhaps you meant http://mediaelement/DMPO45004516?)
... 
C:\Program Files (x86)\Streamlink Nightly build\bin>streamlink.exe "https://radio.nrk.no/serie/ren-60/DMPO45004516/10-11-2016" best
[cli][info] Found matching plugin nrk for URL https://radio.nrk.no/serie/ren-60/DMPO45004516/10-11-2016
error: Unable to open URL: mediaelement/DMPO45004516 (Invalid URL 'mediaelement/DMPO45004516': No schema supplied. Perhaps you meant http://mediaelement/DMPO45004516?)
  
Valid until  11/05 -17

C:\Users\Ordval>streamlink.exe "https://tv.nrk.no/serie/martin-og-mikkelsen/MUHH49000517/sesong-1/episode-5#" best
[cli][info] Found matching plugin nrk for URL https://tv.nrk.no/serie/martin-og-mikkelsen/MUHH49000517/sesong-1/episode-5#
error: Unable to open URL: mediaelement/MUHH49000517 (Invalid URL 'mediaelement/MUHH49000517': No schema supplied. Perhaps you meant http://mediaelement/MUHH49000517?)

No limitation in time.

### Actual behavior: Can't extract a streams MRL with X-Forwarded-For.
C:\Program Files (x86)\Streamlink Nightly build\bin>streamlink.exe --http-header X-Forwarded-For=159.20.6.38 "https://www.dr.dk/tv/se/kongehuset-2/-/kongehuset-4-4" best
[cli][info] Found matching plugin drdk for URL https://www.dr.dk/tv/se/kongehuset-2/-/kongehuset-4-4
error: No playable streams found on this URL: https://www.dr.dk/tv/se/kongehuset-2/-/kongehuset-4-4
[...]

PR #881, opened 2017-05-04
[NRK.NO] fixed regex for _api_baseurl_re

back-to:
The website uses `'` instead of `"` now.

streamlink/streamlink#859

### DIFF
--- a/src/streamlink/plugins/nrk.py
+++ b/src/streamlink/plugins/nrk.py
@@ -13,7 +13,7 @@ COOKIE_PARAMS = (
 
 _id_re = re.compile(r"/(?:program|direkte|serie/[^/]+)/([^/]+)")
 _url_re = re.compile(r"https?://(tv|radio).nrk.no/")
-_api_baseurl_re = re.compile(r'apiBaseUrl:\s*"(?P<baseurl>[^"]+)"')
+_api_baseurl_re = re.compile(r'''apiBaseUrl:\s*["'](?P<baseurl>[^"']+)["']''')
 
 _schema = validate.Schema(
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/nrk.py
```diff
@@ -13,7 +13,7 @@
 
 _id_re = re.compile(r"/(?:program|direkte|serie/[^/]+)/([^/]+)")
 _url_re = re.compile(r"https?://(tv|radio).nrk.no/")
-_api_baseurl_re = re.compile(r'apiBaseUrl:\s*"(?P<baseurl>[^"]+)"')
+_api_baseurl_re = re.compile(r'''apiBaseUrl:\s*["'](?P<baseurl>[^"']+)["']''')
```
beardypig: I like you use capture groups for the quotes then do a negative lookahead to continue until the next quote ;)
```python
r'''apiBaseUrl:\s*(['"])(?P<baseurl>(?:(?!\1).)*)\1'''
```

Probably overkill in this case though ;)

back-to: i think this is an overkill with **\1**, it's maybe useful for longer regex but this one is already short.

beardypig: :) What would be best is to have a JS object parser to handle those, like `dirtyjson` but that hasn't been updated for a while.

gravyboat:
@beardypig Are we good here or do you have further concerns?

beardypig:
Nah, it's cool.

gravyboat:
Thanks @back-to.

Issue #869, opened 2017-05-02
Plugin request: streannlive-lh.akamaihd.net


skilleek:
### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

A simple plugin to open the stream available at http://www.ellobo106.com/index.php/vivo/solo-3audio

gravyboat:
Closing since #887 is merged.

PR #887, opened 2017-05-05
plugins.streann: support for ott.streann.com + ellobo.com

beardypig:
Also supports ellobo.com, fixing #869. Not so straightforward :)


gravyboat:
Thanks @beardypig!

skilleek:
Thanks a lot @beardypig you are awesome!

PR #892, opened 2017-05-07
plugins.pcyourfreetv: fix for new new new layout

melmorabity:
It looks like a game of cat and mouse... At the next layout change, I will probably give up and ask for deleting this plugin.


karlo2105:
Is there any chance to add variables which shall adapt automatically changes?

gravyboat:
@melmorabity Yeah I'm all for not supporting it going forward, too much work for very little gain. We might need to add an unsupported note so people don't ask.

Issue #895, opened 2017-05-07
How do I use a different video player for youtube streams??


Dompyre:
By default it opens up VLC, how do I make it open up MPC-HC instead?

gravyboat:
Please considering reviewing the documentation as this is covered thoroughly over there: https://streamlink.github.io/cli.html#player-options I'm going to close this.

Dompyre:
I read it and didn't really understand anything, not a coder all I see is text

gravyboat:
@Dompyre I've proposed a PR to add some additional notes to the documentation to ensure things are clear.

PR #896, opened 2017-05-08
Add a few more examples to the player option

gravyboat:
This adds a few more examples to the player docs to hopefully address confusion like in https://github.com/streamlink/streamlink/issues/895

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -318,7 +318,7 @@ player.add_argument(
     Paths can also be passed via this option in the event the
     player's path has not been set in the configuration file:
 
-    "streamlink <link> --player='path\to\mpc-hc64.exe'"
+    "streamlink <link> --player=path\to\mpc-hc64.exe"
 
     To use a player that is located in a path with spaces you must
     quote the path:
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -318,7 +318,7 @@ player.add_argument(
     Paths can also be passed via this option in the event the
     player's path has not been set in the configuration file:
 
-      "streamlink --player='path\to\mpc-hc64.exe' <url> <quality>"
+      "streamlink --player="path\to\mpc-hc64.exe" <url> <quality>"
 
     To use a player that is located in a path with spaces you must
     quote the path:
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -307,9 +307,18 @@ player.add_argument(
     default=find_default_player(),
     help="""
     Player to feed stream data to. This is a shell-like syntax to
-    support passing options to the player. For example:
+    support using a specific player, for example:
+
+      "streamlink <url> --player=mpc-hc"
+    
+    Options may also be passed to the player. For example:
 
       "vlc --file-caching=5000"
+
+    Paths can also be passed via this option in the event the
+    player's path has not been set in the configuration file:
+
+    "streamlink <link> --player='path\to\mpc-hc64.exe'"
 
     To use a player that is located in a path with spaces you must
     quote the path:
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -306,27 +306,33 @@ player.add_argument(
     metavar="COMMAND",
     default=find_default_player(),
     help="""
-    Player to feed stream data to. This is a shell-like syntax to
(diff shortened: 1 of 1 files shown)


src/streamlink_cli/argparser.py
```diff
@@ -307,10 +307,19 @@ def boolean(value):
...
+      "streamlink <url> --player=mpc-hc"
+    
+    Options may also be passed to the player. For example:
 
       "vlc --file-caching=5000"
 
+    Paths can also be passed via this option in the event the
+    player's path has not been set in the configuration file:
+
+      "streamlink <link> --player=path\to\mpc-hc64.exe"
```
beardypig: Should this be `<url>` for consistency?

bastimeyer: `streamlink --player "<player>" <url> <quality>`
This would be even better, because the program files path on Windows usually has a space in between the folder names and requires quotations, so the `--player=path` format could be confusing.

gravyboat: Good notes, I'll update it when I get home. @bastimeyer The quote requirement for spaces is already notated further down the page, it could be included in both though, thoughts?

gravyboat: Okay this is now updated. I left the top example calling the URL and then the player (just to show it can be done) then modified the second example to have a full quoted path and the additional items, plus there's the other example below that one which goes into detail regarding the spaces. I think this covers it.

bastimeyer:
> I think there is a problem using ' quotes on Windows in the standard cmd shell, that might cause confusion?...

Yep, that's what I meant with my earlier comment.

gravyboat:
Hmm, I only ever use powershell so I've never encountered that, does it have to be `"`?.

gravyboat:
@bastimeyer Alright fixed the double quotes on there, sorry about the delay, I kept forgetting.

bastimeyer:
Sorry @gravyboat, I opened the PR yesterday, but closed my browser a while later and forgot about it...

I think having a text like this would be better:
```
    Player to feed stream data to. By default, VLC will be used
    if it can be found in its default location.
    This is a shell-like syntax to support using a specific player:
      streamlink --player=vlc <url> <quality>
    Absolute or relative paths can also be passed via this option
    in the event the player's executable can not be resolved:
      streamlink --player=/path/to/vlc <url> <quality>
      streamlink --player=./vlc-player/vlc <url> <quality>
    To use a player that is located in a path with spaces you must
    quote the parameter or its value:
      streamlink "--player=/path/with spaces/vlc" <url> <quality>
      streamlink --player "/path/with spaces/vlc" <url> <quality>
    Options may also be passed to the player:
      streamlink --player "vlc --file-caching=5000" <url> <quality>
    As an alternative to this, see the --player-args parameter,
    which doesn't log the custom player arguments.
```

This is more structured and has strict, unambiguous command lines. First the default value, then a simple player executable, then full paths and paths with spaces with different parameter formats, then player arguments and a reference to `--player-args`. I also removed the unnecessary and confusing wrapping quotation marks.

beardypig:
@bastimeyer that looks like a good suggestion to me 👍

gravyboat:
@bastimeyer @beardypig Okay those changes are in, the only I changed is the second example of passing a player with spaces so it uses mpc instead of VLC so people recognize you CAN use other players there since that's where the initial confusion regarding these options came in.

bastimeyer:
It's okay. If you want, you could change the path to a Windows format. Otherwise, I'm going to merge this.

gravyboat:
@bastimeyer I updated it so the mpc executable was a Windows style path, would have been lame and confusing to leave it otherwise, good catch.

bastimeyer:
The build failed after the merge:
https://travis-ci.org/streamlink/streamlink/builds/235115941
I'm restarting it now...

bastimeyer:
And the rendered page appears to be broken:
https://streamlink.github.io/cli.html#cmdoption-p
What is happening here?

bastimeyer:
Is there a way to disable the option reference added by Sphinx in code blocks? We can't add any markup in the description here, because it also gets used in the streamlink man files.

Also, I have totally missed the unnecessary wrapping quotation of the first code block and the missing closing angle bracket and shouldn't have merged yet.

@gravyboat 
Do you want me to revert this commit?

gravyboat:
@bastimeyer Sorry was sleeping, I'm not sure exactly why that happened considering it had a similar format previously with no issue, but your fix in #943 is fine with me, don't revert.

Issue #902, opened 2017-05-09
Pluzz platform is no more, plugin needs some changes


jalba69:
### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description
Hello,
They removed the French "pluzz" platform and now they put everything on a same more uniform website, here a list of all **live** channels URL's present on the website :

https://www.france.tv/france-2/direct.html
https://www.france.tv/france-3/direct.html
https://www.france.tv/france-4/direct.html
https://www.france.tv/france-5/direct.html
https://www.france.tv/france-o/direct.html  < it's a "o" not a zero

Those are only the live streams for the French channels of this group, there is also a different info channel live stream that use a different URL, but it's from the same group so i don't know what is possible or not, might all work the same way more or less, this is the live stream URL for the information channel : 
http://www.francetvinfo.fr/en-direct/tv.html

Maybe a new plugin " francetv" should be created instead of "pluzz" as this thing doesn't exist anymore now and redirects to france.tv website.

The ludo.fr and zouzou.fr sites that use the pluzz plugin are still alive and not changed i believe, only pluzz itself is no more, i don't think there is live streams there tho.

As for viewing replays, non live videos streams, i leave here a few URL's examples from the new website:

https://www.france.tv/france-2/journal-13h00/140985-edition-du-lundi-8-mai-2017.html
https://www.france.tv/documentaires/histoire/136517-argentine-les-500-bebes-voles-de-la-dictature.html
https://www.france.tv/documentaires/histoire/136611-la-face-cachee-de-naples.html
https://www.france.tv/jeux-et-divertissements/divertissements/133965-le-web-contre-attaque.html
https://www.france.tv/france-2/infrarouge/112981-ils-sont-gendarmes.html
https://www.france.tv/france-o/investigations/119007-investigations.html
https://www.france.tv/france-o/foudre/saison-5/131845-un-nouveau-depart.html
https://www.france.tv/france-5/entree-libre/saison-6/137033-entree-libre.html
https://www.france.tv/france-4/une-saison-au-zoo/saison-7/134929-une-saison-au-zoo.html
https://www.france.tv/france-3/nils-holgersson/139027-le-propulseur.html

Not sure if this can be added again or not, might be a lot of work...

...

### Expected / Actual behavior

...

### Reproduction steps / Explicit stream URLs to test

1. ...
2. ...
3. ...

### Environment details

[...]

PR #903, opened 2017-05-09
plugins.pluzz: rework for new france.tv platform replacing pluzz.francetv.fr

melmorabity:
See https://github.com/streamlink/streamlink/issues/902


gravyboat:
Thanks @melmorabity!

jalba69:
Thanks a lot for the nice work guys  it seems to work now after last modifications  ( using portable version ) .
If i may i would suggest to rename the "pluzz" plugin to "francetv" as now pluzz doesn't exist anymore, it will make more sense, anyways as long as it works it's good :p 

Thanks again for the wonderful tool :)

PR #905, opened 2017-05-10
plugins.streamable: support for videos on streamable.com

beardypig:
Supports VOD streams from streamable.com

- streamable.com/s/5uk0a/ulyjaq
- streamable.com/sf37q


gravyboat:

PR #906, opened 2017-05-10
plugins.euronews: update for live API change

melmorabity:
The euronews plugin is broken for live streams, due to API changes (VOD streams are still OK).
```
$ streamlink "http://fr.euronews.com/live"
[cli][info] Found matching plugin euronews for URL http://fr.euronews.com/live
error: Unable to validate JSON: Unable to validate key 'primary': Type of 'http://euronews-fr-p7-cdn.hexaglobe.net/1c52769160a4c3fc93b932305f4e37dc/59136712/euronews/euronews-euronews-website-web-responsive-2/ewnsabrfrpri_fre.smil/playlist.m3u8' should be 'dict' but is 'str'
```
This PR fixes the plugin to use the new API scheme for live streams. Unit tests are also added.


gravyboat:
Thanks @melmorabity!

PR #907, opened 2017-05-10
Drop pcyourfreetv plugin

melmorabity:
As mentionned in https://github.com/streamlink/streamlink/pull/892, the maintainers of the pc-yourfreetv.com website deliberately and regularly change the layout of their platform to discourage people from watching their streams elsewhere than on their site. And it happened again this week... As a result, I recommend dropping support for the pcyourfreetv plugin... Unless someone wants to patch it every week ^^.

Fortunately, I introduced this plugin after the latest release... Anyone who uses streamlink stable releases won't miss this plugin.


gravyboat:
I'm in favor of this, let's let other people weigh in as well, we might need to add a note that they are unsupported to the docs.

karlo2105:
It would be interesting to find out how long this plugin lasts without being available to wide public.

back-to:
did they change anything since this pull request?

and

```py
_iframe_re = re.compile(r"<iframe .*?\bsrc='(?P<iframe>.+?)'.*?</iframe>")
_player_re = re.compile(r"<script language=JavaScript>m='(?P<player>.+?)'", re.DOTALL)
_video_url_re = re.compile(r'new [a-zA-Z0-9]+.Player\({source: "(?P<video_url>[^"]+?)".+?}.*\);', re.DOTALL)
```

is there no **easier working regex** without `new [a-zA-Z0-9]+.Player\({source: `

something like
`["'](?P<url>(?:[^"']+)?(?:\.m3u8)(?:[^"']+)?)["']` where you only search for `.m3u8`

_i don't have an account, so i don't know how the website code looks like._

gravyboat:
@melmorabity Ugh I dropped the ball, can you rebase?

gravyboat:
Thanks @melmorabity, merging in.

PR #927, opened 2017-05-16
plugins.bbciplayer: add login support

beardypig:
This PR sits on-top of the earlier PR (#925) and adds login support. Two new options were added to facilitate this:
 * `--bbciplayer-username` the email/username used to register with bbc.co.uk
 * `--bbciplayer-password` the password corresponding to that bbc account

The login is currently optional, but will be enforced at some point in the future ("weeks" according to the BBC). 

I have add a info level log message that states that the user should have a TV License.


gravyboat:
Good idea about the TV license as ridiculous as that is. Thanks @beardypig, this looks good.

Issue #934, opened 2017-05-21
can't play any twtich stream


craftwar:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description
Try to watch any twitch stream on my win7 64bit. No one works.
...

### Expected / Actual behavior
Error as below, streamlink stops. 

### Reproduction steps / Explicit stream URLs to test

1. use pip install streamlink in python35-32
2. D:\Python\Python35-32\Scripts\streamlink.exe -p "C:\Program Files\VideoLAN\VLC\vlc.exe" -player-continuous-http http://twitch.tv/divinity_123 best
3. other streams tested
    http://twitch.tv/voyboy, http://twitch.tv/froggen

### Environment details

Operating system and version: win7 64 bit ultimate
Streamlink and Python version: python3.5.2, Streamlink 0.6.0(install from pip)

### Comments, logs, screenshots, etc.
```
D:\Python\Python35-32\Scripts\streamlink.exe -p "C:\Program Files\VideoLAN\VLC\vlc.exe" -player-continuous-http http://twitch.tv/divinity_123 best
[cli][info] Found matching plugin twitch for URL twitch.tv/divinity_123
[cli][info] Available streams: audio_only, 160p (worst), 360p, 480p, 720p (best)

[cli][info] Opening stream: 720p (hls)
[cli][info] Starting player: layer-continuous-http
[cli][info] Closing currently open stream...
Traceback (most recent call last):
  File "d:\programs\python\python35-32\lib\site-packages\streamlink_cli\main.py", line 262, in output_stream
    output.open()
  File "d:\programs\python\python35-32\lib\site-packages\streamlink_cli\output.py", line 23, in open
    self._open()
  File "d:\programs\python\python35-32\lib\site-packages\streamlink_cli\output.py", line 124, in _open
    self._open_subprocess()
  File "d:\programs\python\python35-32\lib\site-packages\streamlink_cli\output.py", line 142, in _open_subprocess
    stderr=self.stderr)
  File "d:\programs\python\python35-32\lib\subprocess.py", line 947, in __init__

    restore_signals, start_new_session)
  File "d:\programs\python\python35-32\lib\subprocess.py", line 1224, in _execute_child
    startupinfo)
FileNotFoundError: [WinError 2] 系統找不到指定的\u6a94案。

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #940, opened 2017-05-22
docs: update Sphinx config to fix the rendering of --

beardypig:
Fixes the documentation error in #934. 

For some reason `html_use_smartypants` has started working and it's broken the `--` in the command line options. This PR updates the `conf.py` and adds `docutils.conf` to disable `smart_quotes`.

### DIFF
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,1 +1,2 @@
-sphinx
+sphinx>=1.6,<1.7
+
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -135,10 +135,6 @@ html_static_path = ['_static']
 # using the given strftime format.
 #html_last_updated_fmt = '%b %d, %Y'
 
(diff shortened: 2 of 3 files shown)


bastimeyer:
> For some reason html_use_smartypants has started working

Wouldn't it make sense limiting the sphinx dependency to a specific version so that those things don't happen in the future? This was introduced by sphinx `1.6.0` as it seems...

This is the commit that turned the double dashes into an en-dash:
https://github.com/streamlink/streamlink.github.io/commit/7b6d144d8fd9910085eb82e654f888ff4008797b (triggered by 9409fd741ce26c227ceadbfb69fcb8d281031b77, unrelated, though)

beardypig:
I wouldn't be against pinning Sphinx to `>=1.6,<1.7`, personally I tend to pin things to specific versions as a matter of course.

gravyboat:
Awesome thanks @beardypig!

Issue #935, opened 2017-05-21
New smashcast.tv (formally hitbox.tv) VODs inaccessible


Incoming5643:
When hitbox.tv became smashcast.tv a fix was added to streamlink to allow you to view smashcast VODs by replacing the domain with hitbox.tv. ( #909 )

However new VODs since the change do not have a hitbox.tv url to access, and thus those VODs cannot be played in streamlink.

Example inaccessible VOD:  www.smashcast.tv/the4chancup/videos/1321506

----

C:\Users\student>streamlink https://www.smashcast.tv/the4chancup/videos/1321506 source --player-passthrough=hls
[cli][info] Found matching plugin hitbox for URL https://www.smashcast.tv/the4chancup/videos/1321506
error: No playable streams found on this URL: https://www.smashcast.tv/the4chancup/videos/1321506

C:\Users\student>streamlink https://www.hitbox.tv/the4chancup/videos/1321506 source --player-passthrough=hls
[cli][info] Found matching plugin hitbox for URL https://www.hitbox.tv/the4chancup/videos/1321506
error: No playable streams found on this URL: https://www.hitbox.tv/the4chancup/videos/1321506

beardypig:
As a work around you can do use www.smashcast.tv/video/1321506 as the URL. I have created a PR with a fix for the issue, as well as supporting hosted channels.

Incoming5643:
Cheers 👍

PR #942, opened 2017-05-22
plugins.hitbox: fix new VOD urls, and add support for hosted streams

beardypig:
Fixes the issue in #935 and adds support for hosted channels.


gravyboat:

PR #943, opened 2017-05-23
docs: fix parameters being linked in code blocks

bastimeyer:
See https://github.com/streamlink/streamlink/pull/896#issuecomment-303317718...
I should have built the docs myself before merging, sorry.

Parameters will now only be linked if the text is not indented. I couldn't find a better way of fixing this. The order of regexp substitutions needed to be changed, so that text inside normal and note blocks can still link to parameters.

I've also fixed the missing closing angle bracket and the unnecessary quotation in the first --player example.

The docs are now looking good to me, but there's a chance I have missing something, so please check it first before merging.

### DIFF
--- a/docs/ext_argparse.py
+++ b/docs/ext_argparse.py
@@ -29,7 +29,7 @@ _Argument = namedtuple("Argument", ["args", "options"])
 _block_re = re.compile(r":\n{2}\s{2}")
 _default_re = re.compile(r"Default is (.+)\.\n")
 _note_re = re.compile(r"Note: (.*)\n\n", re.DOTALL)
-_option_re = re.compile(r"(--[\w-]+)")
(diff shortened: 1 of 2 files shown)


gravyboat:
Nah that was my fault, I should have built them locally as well but wasn't concerned about it as the format was nearly identical to the previous configuration. I'll have time to build them tonight if someone else doesn't sign off first. Thanks @bastimeyer.

gravyboat:
Looks good to me, if no one else double checks by tomorrow morning I'll merge. Thanks @bastimeyer.

beardypig:
I'll double check it now :-)

bastimeyer:
Are we good to go?

gravyboat:
I'm assuming we are so I'm merging.

beardypig:
Oops got distracted, then forgot what I was going to do... Looks like it works though :-)

PR #947, opened 2017-05-24
plugins.tvplayer: fix bug with some channels when not authenticated

beardypig:


gravyboat:

PR #950, opened 2017-05-25
plugins.app17: fix hls url and support UID page

steven7851:
e.g. http://17app.co/share/profile/c87cbac6-7716-40f8-b36d-98919786cb97


gravyboat:
Looks good to me, thanks @steven7851!

PR #958, opened 2017-05-27
plugins.pluzz: add support for regional France 3 stations

melmorabity:
This PR adds support for regional France 3 stations, available on france3-regions.francetvinfo.fr, to the pluzz plugin.


gravyboat:
Good stuff, thanks @melmorabity!

PR #962, opened 2017-05-28
docs: fix typo: specifiying, neverthless

amurzeau:
Lintian was reporting a typo about "specifiying" and I found neverthless in cli.rst, so let's fix them.


gravyboat:
Good catches, thanks!

Issue #963, opened 2017-05-28
InsecureRequestWarning error on connecting to Twitch channels


v1kn:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

error on connecting to Twitch channels
(formatted to increase readability):

    <user-pip-folder>/lib/python3.5/site-packages/urllib3/connectionpool.py:852
    InsecureRequestWarning: Unverified HTTPS request is being made.
    Adding certificate verification is strongly advised.
    See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings

### Environment details

Xubuntu 16.04 LTS
Streamlink 0.6.0
Python 3.5.2

### Other notes

My streamlink invocation (which I've been using since livestreamer days):

    streamlink \
    --http-header (Twitch ID) \
    --retry-open 8000 \
    --retry-streams 10 \
    --player-continuous-http \
    --player-passthrough hls \
    --hls-segment-threads 3 \
    --player "mpv" \
    URL

beardypig:
That's weird, certificate verification should be enabled on  any https request. Can you try curling the URL and see if it also gives a warning about certificates?

v1kn:
Sure, here is the output

    curl https://twitch.tv/<channel-name>

    <html>
    <head><title>302 Found</title></head>
    <body bgcolor="white">
    <center><h1>302 Found</h1></center>
    <hr><center>nginx</center>
    </body>
    </html>

globau:
@beardypig the twitch plugin explicitly disables certificate verification:

https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugins/twitch.py#L207
```
# The certificate used by Twitch cannot be verified on some OpenSSL versions.
res = http.get(url, params=params, verify=False, headers=headers)
```

changing to `verify=True` removes the warnings.

what's interesting is streamlink should be silencing these InsecureRequestWarnings:

https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugin/api/http_session.py#L20
```
# We tell urllib3 to disable warnings about unverified HTTPS requests,
# because in some plugins we have to do unverified requests intentionally.
urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```

so the crux of this issue is that call to disable_warnings is no longer working.  if i remove the `try`/`except` block it works, so it isn't throwing an exception.

[...]

PR #965, opened 2017-05-30
setup: exclude requests version 2.16 through 2.17.1

beardypig:
This is a change to the required version of `requests` to exclude some versions that give an unwanted warning about insecure requests for twitch.tv, fixing #963.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ if version_info[0] == 2:
 if version_info[0] == 2 or (version_info[0] == 3 and version_info[1] < 4):
     deps.append("singledispatch")
 
-deps.append("requests>=2.2,!=2.12.0,!=2.12.1,<3.0")
+deps.append("requests>=2.2,!=2.12.0,!=2.12.1,!=2.16.0,!=2.16.1,!=2.16.2,!=2.16.3,!=2.16.4,!=2.16.5,!=2.17.1,<3.0")
 
 # for encrypted streams
 if environ.get("STREAMLINK_USE_PYCRYPTO"):


bastimeyer:
How likely is it for them to release another version on the 2.16 branch that will contain this bug? Or is there a 2.16.6 version with a fix?

beardypig:
I think it was actually a packaging error, but i'm not 100% sure - I haven't been able to track down the specific change that fixed the issue. However, it is possible that they will release a `2.16.6` version - unfortunately I don't think there is a way to exclude `2.16.?` without explicitly listing them, I could be wrong and would welcome an alternative :)

thinkpad4:
Would defaulting this to True fix this?  The certificate used by Twitch cannot be verified on some OpenSSL versions.
res = http.get(url, params=params, verify=False, headers=headers)  
verify=False to True.  Can that be made true in the twitch.py for the nightlies? That seems to fix the problem seen in https://github.com/streamlink/streamlink/issues/963

beardypig:
It does seem to be broken again. There is already an issue to track re-enabling the SSL verification #966.

gravyboat:
@beardypig Are we good to merge this?

beardypig:
I'm not sure, I think it could break again - but it seems to be more an installer builder issue...

gravyboat:
In that case do we want to merge this or no? I don't see any harm in doing so especially with another release coming up, might save some annoyance for people on edge cases.

beardypig:
Yeah, OK. Merge away.

gravyboat:
Haha, I don't like it either, but what can we do?

PR #967, opened 2017-05-30
Improve CLI documentation

### LINKED ISSUE #751: Default Stream quality to best? (documentation improvement)
### Checklist

- [ ] This is a bug report.
- [x] I used the search function to find already opened/closed issues or pull requests.

### Description
I personally use streamlink for all my stream needs (mostly twitch), so I type a lot of streamlink commands into the terminal so any less typing would be appreciated. 

This would only require a minimal change (see [my fork](https://github.com/streamlink/streamlink/compare/master...Germandrummer92:master)) and would as far as I can see it not break anything, as even if someone just wanted to print the available qualities they are still printed and you can just exit the source stream immediately as you will most likely type into the same terminal window again.

If someone actually wants the behavior of just printing the stream qualities I could implement a separate option for that although I'm not sure anyone actually uses that.

`streamlink twitch.tv/pgl_dota`

defaults to
`streamlink twitch.tv/pgl_dota best` 
instead of just printing the stream qualities

Let me know if this makes sense and I would open a pull request with my one-line change!

bastimeyer:
- Reword several parameters.
- Add missing examples.
- Fix format of the default value of some parameters.
- Fix missing punctuation.
- Fix invalid text formats (notes).
- Wrap text lines which exceed 80 chars per line.

Resolves #751

### DIFF
--- a/docs/ext_argparse.py
+++ b/docs/ext_argparse.py
@@ -28,7 +28,7 @@ _Argument = namedtuple("Argument", ["args", "options"])
 
 _block_re = re.compile(r":\n{2}\s{2}")
 _default_re = re.compile(r"Default is (.+)\.\n")
-_note_re = re.compile(r"Note: (.*)\n\n", re.DOTALL)
+_note_re = re.compile(r"Note: (.*)(?:\n\n|\n*$)", re.DOTALL)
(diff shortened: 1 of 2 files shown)


beardypig:
All the changes seem reasonable to me 👍

gravyboat:
These all look good to me as well, merging.

PR #977, opened 2017-06-02
Add support for TV5Monde

melmorabity:
This plugin adds support for the French-speaking channel network [TV5 Monde](http://www.tv5monde.com/).
It supports both [live](http://live.tv5monde.com/fbs.html) and VOD streams from:

* http://www.tv5monde.com/
* http://www.tv5mondeplus.com/
* http://www.tv5mondeplusafrique.com/

Notice that the live stream is geo-restricted to France, Belgium and Switzerland.


gravyboat:
Good stuff, thanks @melmorabity!

Issue #978, opened 2017-06-02
Handle different YouTube live links


ahtcx:
- [x] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description

Currently handles `/c/<channel>/live` but YouTube also supports that link without the `/c` part. Too small for a pull request.

### Expected / Actual behavior

Now able to match YouTube links without the `/c` prefix.

### Reproduction steps / Explicit stream URLs to test

1. `youtube.com/c/<channel>/live`
2. `youtube.com/<channel>/live`
3. `gaming.youtube.com/<channel>/live`

ahtcx:
Here's a patch

```
diff --git a/src/streamlink/plugins/youtube.py b/src/streamlink/plugins/youtube.py
index 7907aca..23a15c8 100644
--- a/src/streamlink/plugins/youtube.py
+++ b/src/streamlink/plugins/youtube.py
@@ -107,7 +107,7 @@ _url_re = re.compile(r"""
         )
         |
         (?:
-            /c/(?P<liveChannel>[^/?]+)/live
+            /(c/)?(?P<liveChannel>[^/?]+)/live
         )
     )
 """, re.VERBOSE)
```

beardypig:
Looks good, do you want to open a PR for it?

beardypig:
No fix is too small for a PR! Well done for fixing your own issue @ahtcx ;)

PR #980, opened 2017-06-03
Allow live YouTube channel links without '/c/' prefix

ahtcx:
Solution for #978, tiny 3 character commit to allow removing the `/c/` prefix from YouTube live channel links.


gravyboat:
Looks good to me, thanks!

PR #982, opened 2017-06-03
Add support for Deutsche Welle

melmorabity:
This plugin adds support for [Deutsche Welle](http://www.dw.com/), the German public international broadcaster. It supports:

* [live TV streams](http://www.dw.com/en/media-center/live-tv/s-100825)
* [VOD](www.dw.com/en/media-center/all-media-content/s-100826?type=18)
* [Audio streams](http://www.dw.com/en/media-center/all-media-content/s-100826?type=17)

For some video streams, both bitrate and video width are provided by DW. To select the best stream when some videos have the same bitrate, I'm "cheating" by summing bitrate and width (see line 107 in dw.py). Is there a better way to manage stream quality based on these two parameters?

### DIFF
--- a/src/streamlink/plugins/dw.py
+++ b/src/streamlink/plugins/dw.py
@@ -1,0 +1,129 @@
+import re
+
+from streamlink.compat import urlparse
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import http, validate
+from streamlink.stream import HLSStream, HTTPStream, RTMPStream
+from streamlink.utils import parse_qsd
+
+
+class DW(Plugin):
+    _SMIL_API_URL = 'http://www.dw.com/smil/{}'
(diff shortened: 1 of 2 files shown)


back-to:
did not test any of the plugins but there is already one here,

https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugins/deutschewelle.py

maybe you could just update it there, if it needs an update

melmorabity:
@back-to Good catch, I haven't seen a plugin already exist.

beardypig:
It only has support for live streams; VOD and audio would be a welcome addition :)

gravyboat:
@beardypig There's a PR for it here: https://github.com/streamlink/streamlink/pull/983

beardypig:
Yeah... Saw it 2 seconds after posting that reply :-)

PR #983, opened 2017-06-04
plugins.deutschewelle: add support for VOD/audio streams

melmorabity:
This PR extends support for:
* [VOD](www.dw.com/en/media-center/all-media-content/s-100826?type=18)
* [Audio streams](http://www.dw.com/en/media-center/all-media-content/s-100826?type=17)

For some video streams, both bitrate and video width are provided by DW. To select the best stream when some videos have the same bitrate, I'm "cheating" by summing bitrate and width (see line 117 in deutschewelle.py). Is there a better way to manage stream quality based on these two parameters?


gravyboat:
Good stuff thank you!

PR #986, opened 2017-06-04
Add support for radio.net

melmorabity:
This PR provides support for [radio.net](http://radio.net) (and its locale versions radio.(at|dk|de|fr|it|pt|se|es|pl). radio.net offers free and legal access to radio stations from all over the world.


gravyboat:
Not sure how I let this one slip through the cracks, sorry @melmorabity and thanks for adding this!

PR #987, opened 2017-06-05
bbciplayer: fix for api changes to iplayer

beardypig:
Update to the API call URL and the method for determining the VPID of the programme. There was an issue with the ident card stream being selecting instead of the actual programme, this PR makes the plugin more robust.


beardypig:
@gravyboat didn't merge this? :)

gravyboat:
@beardypig I thought I hit the button but apparently not.

PR #970, opened 2017-05-30
Create arconai.py

cirrusUK:
author is user @intact who originally wrote the plugin for livestreamer, i only did the easy part by editing 4 lines and creating the PR.

PR #993, opened 2017-06-09
Update plugin_matrix.rst for arconai.tv

cirrusUK:
just adding arconai.tv plugin


gravyboat:
#970 is the associated plugin.

gravyboat:
These are both now merged, thanks @cirrusUK and @intact!

PR #1001, opened 2017-06-14
plugins.webtv: decrypt the stream URL when applicable

beardypig:
Some (all?) of the stream URLs are now encrypted using AES encryption - the key and IV are prepended to the cipher text.

Added some additional util functions to support the decryption.
 - An unpad method which performs "standard" pkcs5 unpadding.
 - A new validation rule for text; `validate.contains` which validates that the string contains a given substring.


gravyboat:
Nice this looks good! Thanks @beardypig!

PR #1007, opened 2017-06-16
plugins.dogan: small api change for teve2.com.tr

beardypig:
teve2.com.tr updated to use a different api URL (same as cnnturk.com)


gravyboat:
Thanks @beardypig!

bastimeyer:
The builds are failing since this merge:
https://travis-ci.org/streamlink/streamlink/builds

beardypig:
Seems to be unrelated to the actual changes and more that something external changed at the same time. I suspect the `requests` dependencies. I'm going to submit a PR to fix the version to a known working one so the Windows installer works...

PR #1008, opened 2017-06-16
plugins.kanal7: fix for nested iframes

beardypig:
The iframes on the kanal7 site are now nested.


gravyboat:
Dumb, but thanks!

beardypig:
Sure is dumb.

gravyboat:
Iframes all the way down!

beardypig:
That would have made for the best commit message... darn

PR #1018, opened 2017-06-19
plugins.canlitv: simplified and fixed the m3u8 regex

beardypig:
On some pages there is a space before the `:`, after `file`. Added a few tests for the regex too.


gravyboat:
Thanks @beardypig!

PR #969, opened 2017-05-30
win32: fix missing modules when using windows installer

beardypig:
`pkg_resources` (`setuptool`) and `requests` have changed the way their packages are structured so that they no longer vendor packages, this means those previously vendored packages have to be included by the installer builder.

This should fix #945, plus an unreported issue with `requests`.

PR #1020, opened 2017-06-19
win32: update the dependencies for the windows installer

beardypig:
Should fix the broken builds... 

Seems to be related to #969, I think it might be a travis thing ...


gravyboat:
Cool let's see if this does it.

Issue #1006, opened 2017-06-16
picarto video -> mpv does not work?


dahjin:
### Checklist

- [x] This is a bug report.
- [x] This is a feature request.
- [x] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description
Using the command streamlink https://picarto.tv/videopopout/Pixiepowderpuff_2017.06.14.07.27.57.flv -o 1.mp4 gets me nothing except that it thinks the channel is offline.

### Expected / Actual behavior
use above command and it should output the .flv video to a mp4 video on my desktop.

### Reproduction steps / Explicit stream URLs to test

1. open command promt
2. type in streamlink https://picarto.tv/videopopout/Pixiepowderpuff_2017.06.14.07.27.57.flv -o 1.mp4
3. Wonder what is life and why it has wronged you.

### Environment details

Operating system and version: Windows 10 64 bit
Streamlink Version: 0.6.0
Python Version : 3.5.2

### Comments, logs, screenshots, etc.

CLI: https://pastebin.com/raw/sVHeUasj

dahjin:
Using Youtube-dl i have actually gotten around this. But i'm not sure if this should be supported in Streamlink or not due to this being sort of like twitch vods but for Picarto.

beardypig:
How do you get to the VOD page? I cannot find a link on the site... Are there are formats for the VOD links? ie. not videopopout?

skulblakka:
>How do you get to the VOD page? I cannot find a link on the site...

As far as I can tell only Premium streams have VODs. You can find a "Videos" tab under their stream. This one has some for example https://picarto.tv/LiLaiRa (I picked a random one from the popular tab, so I don't know if their content is NSFW)

dahjin:
Pretty much what skulblakka said the only way to access the vods is if that steamer has premium and it's below in a videos tab.

beardypig:
@dahjin hopefully the PR (#1023) will help you to understand life, and come to see that you we not wronged, it just took a little while to balance out ... or something.

gravyboat:
I think we can close this since #1023 is merged yeah?

back-to:
solved in #1023

PR #1023, opened 2017-06-20
plugins.picarto: support for VOD

beardypig:
Adds support for VOD to the picarto plugin as requested in #1006 .


gravyboat:
Nice addition, thanks!

PR #1032, opened 2017-06-23
plugins.app17: change ROOM_URL

steven7851:
17app.co maybe no use.


gravyboat:
Are there any notes or details confirming this? If we can get confirmation from them that this was an intended change and that is the new URL I have no issue switching it over.

steven7851:
I believe that 17.media is the permanent url, and 17app.co is a backup which they used before.

gravyboat:
Okay since no one else weighed in I'm going to merge this. Thanks!

PR #1040, opened 2017-06-26
Add ignore filter to warnings in http_session

azatoth:
Somehow urllib3 isn't reacting to the disable_warnings function. As a
workaround, we'll set a ignore filter to warnings in the http_session
module.

Closes: #963

### DIFF
--- a/src/streamlink/plugin/api/http_session.py
+++ b/src/streamlink/plugin/api/http_session.py
@@ -24,6 +24,9 @@ except (ImportError, AttributeError):
 from ...exceptions import PluginError
 from ...utils import parse_json, parse_xml
 
+import warnings
+warnings.simplefilter("ignore")
+
 __all__ = ["HTTPSession"]
 
 


gravyboat:
Thanks for the PR. I know @beardypig has been working on a PR over here: https://github.com/streamlink/streamlink/pull/965 that excludes some problematic requests versions, let's see what their opinion is.

gravyboat:
@beardypig Sorry to ping you again, but I would like a second opinion on this since #965 was merged. I'm not sure we want to support warnings, or if this is even needed now that the requests changes are in.

beardypig:
I think this is directly related to a specific version(s) of requests and/or urllib3. Gimme until next week to work out the exact cause, ideally we shouldn't suppress these warnings except in very specific cases.

gravyboat:
@beardypig Yep that was my reasoning as well, as generally I want to see errors so we know something is up. No rush.

gravyboat:
I think we are good to close this now due to https://github.com/streamlink/streamlink/pull/1119 so we no longer need to ignore them, thanks for the PR however!

Issue #1045, opened 2017-06-28
artetv plugin can not find streams in any arte.tv url


medialove:
The artetv plugin unfortunately can not find any source in any url on the actual arte.tv website. This is really unfortunate - this tv station produces very interesting documentaries.
Example:
```
$ streamlink http://www.arte.tv/en/videos/072258-000-A/meeting-snowden
[cli][info] Found matching plugin artetv for URL http://www.arte.tv/en/videos/072258-000-A/meeting-snowden
error: No playable streams found on this URL: http://www.arte.tv/en/videos/072258-000-A/meeting-snowden

```

bastimeyer:
It seems like the [VOD API calls](https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugins/artetv.py#L15) are missing the `?platform=ARTE_NEXT` query string:
`https://api.arte.tv/api/player/v1/config/en/072258-000-A`
`https://api.arte.tv/api/player/v1/config/en/072258-000-A?platform=ARTE_NEXT`

I'm not the creator of the plugin, I just wanted to see what's going on here and found this in the code of the arte website, but maybe @back-to who committed the last changes in 7e9ec46517fe76b2b7cc95838d2d411c184600c2 can help here. Adding the query string to the VOD API URL makes the plugin work for this specific URL, but I don't know if it can be added for all VODs.

back-to:
sadly they remove old videos, so you can't compare they with new videos.

But yes it looks like this works for all videos. 

---

At opening some videos i found a different missing language name

`versionShortLibelle = "OmU"`
> **O**riginalfassung **m**it **U**ntertiteln
_original version with subtitles_

Not sure if they only use it on the german version of the API,
but i think the best way is to add this _language name_ to all languages

I will open some other urls tomorrow and take a look if i can find it somewhere again 

http://www.arte.tv/de/videos/062822-002-A/street-art-a-la-francaise-2-10
https://api.arte.tv/api/player/v1/config/de/062822-002-A?platform=ARTE_NEXT

PR #1047, opened 2017-06-29
[arte] vod api url update & add new/missing languages

back-to:
Fixed streamlink/streamlink#1045


gravyboat:
Looks good, thanks @back-to.

Issue #1031, opened 2017-06-23
bigo.tv plugin - cannot seem to play saved file


bigotvuser:
- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description
Hey  guys, the people at bigo.tv have made changes again and now the streams are not playing and saved files open but do not show any content. Could require a similar fix as in issue 234. . 
...

### Expected / Actual behavior
..

### Reproduction steps / Explicit stream URLs to test

streamlink http://bigoweb.co/show/2210868181 best -o kabbi1nes.flv
I tried the latest streamlink version 0.6.0 and see the same results. 
[kabbi1nes.zip](https://github.com/streamlink/streamlink/files/1098918/kabbi1nes.zip)

### Environment details

Operating system and version: ...Win 10
Streamlink and Python version: ..streamlink - .0.6.0

...

### Comments, logs, screenshots, etc.

...

unnutricious:
I can confirm that I see the same thing on any Bigo stream. I haven't figured out what has changed in the protocol yet, though (I wasn't very familiar with the original one). It still seems to be streaming data with FLV headers, but it's not valid.

The flash parameters all look the same as before, but I am guessing we need to send a different set of parameters after opening the socket now.

unnutricious:
I confirmed that the 'login' protocol has not changed at all. The server is just now sending back different data. There are an extra 16 bytes pre-pending before the FLV stream starts, which causes the player to not recognize it. If you strip these bytes out, you can play the first frames of the video. However, there still appear to be errors in the stream, so I suspect it is inserting extra non-FLV data elsewhere as well. I am not that familiar with FLV, though, so it's hard to tell.

After stripping the first 16 bytes from the example capture, flvmeta reports these errors:
```
0x0000014b: warning W70044: 3 bytes missing from tag body after metadata end
0x0000014f: error E12051: previous tag size should be 317, 81160 found instead
0x0000014f: error E30013: unknown tag type 0
0x00000157: error E20025: tag stream id must be zero, 175 found instead
0x0000055e: error E12051: previous tag size should be 1035, 1694485086 found instead
0x0000055e: error E30013: unknown tag type 212
0x0000055e: fatal F30084: consecutive tags with unknown type found, aborting
6 error(s), 1 warning(s)
```

back-to:
the easy way would be to remove the flash stream and use **HLS** instead

[...]

PR #1053, opened 2017-07-01
Add support for hls streams in bigo

unnutricious:
This is a workaround for issue #1031 since FLV streaming is broken

### DIFF
--- a/src/streamlink/plugins/bigo.py
+++ b/src/streamlink/plugins/bigo.py
@@ -47,7 +47,7 @@ class BigoStream(Stream):
 
 
 class Bigo(Plugin):
-    _url_re = re.compile(r"https?://(?:www\.)?(bigo\.tv/\d+|bigoweb\.co/show/\d+)")
+    _url_re = re.compile(r"^https?://(www\.)?(bigo\.tv|bigoweb\.co/show)/[\w\d]+$")
     _flashvars_re = flashvars = re.compile(
         r'''^\s*(?<!<!--)<param.*value="tmp=(\d+)&channel=(\d+)&srv=(\d+\.\d+\.\d+\.\d+)&port=(\d+)"''',
         re.M)
--- a/src/streamlink/plugins/bigo.py
+++ b/src/streamlink/plugins/bigo.py
@@ -4,8 +4,8 @@ import struct
 
 from streamlink import PluginError
 from streamlink.plugin import Plugin
-from streamlink.plugin.api import http
-from streamlink.stream import Stream
+from streamlink.plugin.api import http, useragents
+from streamlink.stream import Stream, HLSStream
 
 
 class BigoStream(Stream):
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/bigo.py
```diff
@@ -67,5 +70,15 @@ def _get_streams(self):
         sid, uid, ip, port = flashvars.groups()
         yield "live", BigoStream(self.session, sid, uid, ip, port)
 
+        page = http.get(self.url,
+                        allow_redirects=True,
+                        headers={"User-Agent": "Mozilla/5.0 (iPhone; CPU iPhone OS 8_0_2 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) Version/8.0 Mobile/12A366 Safari/600.1.4"})
```
beardypig: There is an iPhone user agent const that you can use `streamlink.plugin.api.useragents`.

src/streamlink/plugins/bigo.py
```diff
@@ -67,5 +70,15 @@ def _get_streams(self):
...
+        videomatch = self._video_re.search(page.text)
+        if not videomatch:
+            return
+
+        videourl = videomatch.group(1)
+        yield "hls", HLSStream(self.session, videourl)
```
beardypig: If you name the stream `live` won't it be `best` by default?

src/streamlink/plugins/bigo.py
beardypig: You could update the other http call too as it was written before that module was added :-)

unnutricious:
Addressed comments and also improved the url regex

src/streamlink/plugins/bigo.py
```diff
@@ -59,13 +62,23 @@ def can_handle_url(cls, url):
...
                         allow_redirects=True,
-                        headers={"User-Agent": "Mozilla/5.0 (MSIE 10.0; Windows NT 6.1; Trident/5.0)"})
+                        headers={"User-Agent": useragents.CHROME})
         flashvars = self._flashvars_re.search(page.text)
         if not flashvars:
             return
 
         sid, uid, ip, port = flashvars.groups()
-        yield "live", BigoStream(self.session, sid, uid, ip, port)
+        yield "flv", BigoStream(self.session, sid, uid, ip, port)
```
beardypig: Doesn't it work if both streams are called `live`? HLS shout have a higher priority.

src/streamlink/plugins/bigo.py
```diff
@@ -47,10 +47,13 @@ def close(self):
 
 
 class Bigo(Plugin):
-    _url_re = re.compile(r"https?://(?:www\.)?(bigo\.tv/\d+|bigoweb\.co/show/\d+)")
+    _url_re = re.compile(r"http://(www.bigo.tv/[\w\d]+|bigoweb.co/show/\d+)")
```
beardypig: This won't match www.bigweb.co or https anymore, did you intend to do that?

bigotvuser:
Although fully agree that improving the URL regex is being done at an opportune moment, could just the HLS changes be released for now so that it can be tested. It appears the change is regex would apply to only a minority of bigo tv URLs.

beardypig:
Currently the tests are not passing, so this cannot be merged in its current state.

bastimeyer:
Since @unnutricious seems to be unavailable right now, someone could pull their changes, squash all the commits, add the needed fix and open a new PR with these two commits. This way, everyone will be credited for their work and the plugin gets fixed.

unnutricious:
Sorry, the pr should be in better shape now. I removed flv support because it probably won't work again.

beardypig:
@unnutricious, nice looks good to me :)

gravyboat:
Thanks @unnutricious! Thanks for the review @beardypig!

PR #1060, opened 2017-07-04
plugins.douyutv: update post data

steven7851:
No change, daily update.


gravyboat:
Is this going to be a regular occurrence for this site? If so we may need to discuss removing it.

steven7851:
Even if ```ver``` is not up to date, the plugin still can work normally. I just want to update when I noticed it.

gravyboat:
@steven7851 Gotcha, thanks for the heads up.

Issue #949, opened 2017-05-25
vidio(dot)com plugin not working


lokmiin:
### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description

Hello Admin, Please checking vidio(dot)com plugin no longer working.
Could you fix it please..
Thank You.

### Expected / Actual behavior

...

### Reproduction steps / Explicit stream URLs to test

1. ...
2. ...
3. ...

### Environment details

Operating system and version: ...
Streamlink and Python version: ...

...

### Comments, logs, screenshots, etc.

...

gravyboat:
We're going to need more output from you. What do you get from Streamlink when you try to use the plugin? Can you run with the `-v` and `-l debug` options please?

lokmiin:
@gravyboat 
Thanks for replied..
Here is the error output run both on versions 0.5.0 and 0.6.0
C:\Program Files\Streamlink\bin>streamlink -v https://www.vidio.com/live/777-met
ro-tv-stream best
[cli][info] Found matching plugin vidio for URL https://www.vidio.com/live/777-m
etro-tv-stream
error: No playable streams found on this URL: https://www.vidio.com/live/777-met
ro-tv-stream

Thank You So Much

All The Best

lokmiin:
C:\Program Files\Streamlink\bin>streamlink -l debug https://www.vidio.com/live/7
77-metro-tv-stream best
[cli][info] Found matching plugin vidio for URL https://www.vidio.com/live/777-m
etro-tv-stream
error: No playable streams found on this URL: https://www.vidio.com/live/777-met
ro-tv-stream

gravyboat:
Closed since #1066 is merged.

PR #1066, opened 2017-07-05
[vidio] Fixed Plugin, new Regex for HLS URL

back-to:
Fixed streamlink/streamlink#949


gravyboat:
Wow you were able to chop a lot here, good stuff! Thanks @back-to.

Issue #1073, opened 2017-07-07
Arconai.py  plugin broken ?


cirrusUK:
Well my plugin did not last long, can anyone confirm it is indeed broken and not just PEBKAC, cheers.
[edit] i see their site is going through a transition, and they have changed URL syntax so no doubt plugin is borken.

unlocKing:
It was posted yesterday  #1067 by @back-to

Below is the fix:

import re

from streamlink.plugin import Plugin
from streamlink.plugin.api import http
from streamlink.plugin.api import useragents
from streamlink.stream import HLSStream

_url_re = re.compile(r'''https?://(www\.)?arconaitv\.me/stream\.php\?id=\d+''')
_playlist_re = re.compile(r'''source\ssrc=["'](?P<url>[^"']+)["']''')


class ArconaiTv(Plugin):
    @classmethod
    def can_handle_url(cls, url):
        return _url_re.match(url)

    def _get_streams(self):
        headers = {
            'User-Agent': useragents.CHROME,
            'Referer': self.url
        }

        res = http.get(self.url, headers=headers)

        match = _playlist_re.search(res.text)
        if match is None:
            return

        url = match.group('url')

        if url:
            self.logger.debug('HLS URL: {0}'.format(url))
            yield 'live', HLSStream(self.session, url, headers=headers)

__plugin__ = ArconaiTv

bastimeyer:
You should properly format code you're pasting in any comment section on Github before submitting it.
https://guides.github.com/features/mastering-markdown/
Let me fix this for you...

Btw, a better solution for posting code is using Github gists:
https://gist.github.com/

unlocKing:
@bastimeyer that looks much better, thanks.

cirrusUK:
cheers lads

JourneyOver:
Honestly I could never get arconia to work all that great with livestreamer/streamlink anyways, streams would work for a while and then all of a sudden the stream would freeze because of how arconia has his site/stream setup. Does the same crap even if you are just on his site watching the stream instead of using livestreamer/streamlink.

cirrusUK:
@JourneyOver  Works as expected here. if its that bad you could try using firefox and when playing chosen stream, hit  F12, then hit network tab and filter .m3u8 and/or .ts for source stream URL.

JourneyOver:
@cirrusUK it doesn't seem to matter if it's done through the source stream url or not. 

[...]

Issue #1067, opened 2017-07-05
arconaitv.me plugin broken


saki4u:
anconai made changes to there website streams not working....tnx
http://arconaitv.me/stream.php?id=69

saki4u:
PLEASE FIX ARCONAI plugin too....tnx

back-to:
here you can test it

```py
import re

from streamlink.plugin import Plugin
from streamlink.plugin.api import http
from streamlink.plugin.api import useragents
from streamlink.stream import HLSStream

_url_re = re.compile(r'''https?://(www\.)?arconaitv\.me/stream\.php\?id=\d+''')
_playlist_re = re.compile(r'''source\ssrc=["'](?P<url>[^"']+)["']''')


class ArconaiTv(Plugin):
    @classmethod
    def can_handle_url(cls, url):
        return _url_re.match(url)

    def _get_streams(self):
        headers = {
            'User-Agent': useragents.CHROME,
            'Referer': self.url
        }

        res = http.get(self.url, headers=headers)

        match = _playlist_re.search(res.text)
        if match is None:
            return

        url = match.group('url')

        if url:
            self.logger.debug('HLS URL: {0}'.format(url))
            yield 'live', HLSStream(self.session, url, headers=headers)

__plugin__ = ArconaiTv
```

unlocKing:
C:\Users\j>streamlink http://arconaitv.me/stream.php?id=30 best
[cli][info] Found matching plugin arconai for URL http://arconaitv.me/stream.php?id=30
[cli][info] Available streams: live (worst, best)
[cli][info] Opening stream: live (hls)
[cli][error] Try 1/1: Could not open stream <HLSStream('http://149.56.27.145/live/YXH3hmVh3Up6vr4DJ-_n0g/1499376775/a8107d2e6831bddc1fe8c05a804ca0c1.m3u8')> (Could not open stream: Unable to open URL: http://149.56.27.145/live/YXH3hmVh3Up6vr4DJ-_n0g/1499376775/a8107d2e6831bddc1fe8c05a804ca0c1.m3u8 (404 Client Error: Not Found for url: http://149.56.27.145/live/YXH3hmVh3Up6vr4DJ-_n0g/1499376775/a8107d2e6831bddc1fe8c05a804ca0c1.m3u8))
[...]

PR #1075, opened 2017-07-07
[arconai] Fixed plugin for new website

back-to:
Fixed streamlink/streamlink#1067
Fixed streamlink/streamlink#1073

### DIFF
--- a/src/streamlink/plugins/arconai.py
+++ b/src/streamlink/plugins/arconai.py
@@ -1,14 +1,13 @@
 import re
 
 from streamlink.plugin import Plugin
-from streamlink.plugin.api import http, validate
-from streamlink.plugin.api.utils import parse_json
+from streamlink.plugin.api import http
+from streamlink.plugin.api import useragents
 from streamlink.stream import HLSStream
 
-_url_re = re.compile(r"https://www.arconaitv.me/([^/]+)/")
+_url_re = re.compile(r'''https?://(www\.)?arconaitv\.me/stream\.php\?id=\d+''')
+_playlist_re = re.compile(r'''source\ssrc=["'](?P<url>[^"']+)["']''')
 
-SOURCES_RE = re.compile(r" data-item='([^']+)' ")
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/arconai.py
```diff
@@ -1,34 +1,35 @@
...
+            'User-Agent': useragents.CHROME,
+            'Referer': self.url
+        }
 
-        sources = parse_json(match.group(1))
-        if "sources" not in sources or not isinstance(sources["sources"], list):
+        res = http.get(self.url, headers=headers)
+
+        match = _playlist_re.search(res.text)
+        if match is None:
```
beardypig: I like to do `url = match and match.group('url')`, more concise pretty pythonic too.

back-to: well this might work here,

but since i use a template for plugins,
it would make more work for me if i had to do this always,
when there is more than just `url` from the regex

it's easier to copy and paste with less changes 🦅 

```py
'''
source: -

Plugin for
- https://example.com
'''
import re

from streamlink.plugin.api import http

_url_re = re.compile(r'''https?://(?:\w+\.)?example\.com''')
_playlist_re = re.compile(r'''{\s+file:\s["'](?P<url>[^"']+)["']''')


class PLUGINNAME(Plugin):
    @classmethod
    def can_handle_url(cls, url):
        return _url_re.match(url)

    def _get_streams(self):
        res = http.get(self.url)

        match = _playlist_re.search(res.text)
        if match is None:
            return

        url = match.group('url')
        if url.startswith('//'):
            url = 'https:{0}'.format(url)

        if url:
            self.logger.debug('HLS URL: {0}'.format(url))
            yield 'live', HLSStream(self.session, url)

__plugin__ = PLUGINNAME

```

beardypig: Cool, looks like my template too :-) by the way, there is a method `streamlink.utils.update_scheme` for updating URLs that start with `//` to match the `self.url`.

gravyboat:

back-to:
yes i'm done

gravyboat:
Alright then merging, thanks @back-to!

PR #1079, opened 2017-07-10
plugins.app17: fix HLS url

steven7851:


gravyboat:
Good stuff, thanks @steven7851!

Issue #1062, opened 2017-07-05
npo.nl live streams stopped working


sophof:
----

### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

npo.nl/live streams stopped working. Looks like npo changed its player.

### Expected / Actual behavior

streamlink shows a 'no streams found' error since about last night.

### Reproduction steps / Explicit stream URLs to test

npo.nl/live/npo-1

### Environment details

Operating system and version: Windows 10 and Ubuntu linux
Streamlink and Python version: streamlink: 0.7.0, python 2.7 and 3.6

### Comments, logs, screenshots, etc.

Same issue for youtube-dl, sadly without a fix this time ;-)
https://github.com/rg3/youtube-dl/issues/13568

karlo2105:
It looks like they enabled Widewine. I suppose decryption key is here : https://npo-drm-gateway.samgcloud.nepworldwide.nl/widevine_player_cert.bin

back-to:
# LIVE

- **DASH**
not supported by streamlink
- **HDS**
I don't know if they use HDS, i only found a HLS and a DASH link
- **HLS**
`METHOD=SAMPLE-AES` and this is **not supported by streamlink** #354 #836

```
#EXTM3U
#EXT-X-VERSION:4
## Created with Unified Streaming Platform(version=1.7.25)
#EXT-X-SESSION-KEY:METHOD=SAMPLE-AES,URI="skd://123",KEYFORMAT="com.apple.streamingkeydelivery",KEYFORMATVERSIONS="1"

# AUDIO groups
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="audio-aacl-128",NAME="Dutch/Flemish",LANGUAGE="nl",AUTOSELECT=YES,DEFAULT=YES,CHANNELS="2"

# SUBTITLES groups
#EXT-X-MEDIA:TYPE=SUBTITLES,GROUP-ID="textstream",NAME="Dutch/Flemish",LANGUAGE="nl",AUTOSELECT=YES,DEFAULT=YES,URI="npo-1-tt888_nld=64000.m3u8"

# variants
[...]

PR #1084, opened 2017-07-10
[npo] Update for new website layout, Added HTTPStream support

back-to:
Fixed streamlink/streamlink#1062

If they remove the **old API** _(currently used by streamlink)_,
the livestreams won't work with the **new API**.


gravyboat:
Thanks for the fix @back-to! This is pretty lame of them but not much we can do, people will just have to use the website for the best quality video.

Issue #1029, opened 2017-06-22
Plugin Request for UFC Fightpass


SeanDunford:
### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

Not sure if this is possible but would be cool to have a plugin for ufc fightpass. Happy to contribute to developing the plugin but python isn't my primary language so would need some help.

### Description
I would like to be able to stream ufc fight pass. This is a "paid" service but they do have some free videos. Some that require a login and some that don't. It appears that there are existing "paid" services implemented already (crunchyroll)

Here is an example video that requires no login. 
https://www.ufc.tv/video/fight-night-oklahoma-city-michael-chiesa-vs-kevin-lee-preview

Here is an example of video that requires a paid account (30 day free trial available)
https://www.ufc.tv/video/ufc-fortaleza-2017

### Details
There's a similar project that exists for Kodi. I was actually looking at this to attempt to build a similar project before I found this repo. 

https://github.com/portse/plugin.video.ufcfightpass/blob/master/default.py

If you parse out the xbmc parts of the code you can very easily get stream urls. I was having issues pairing them with cookies in browser though. 

ufc.tv is using neulion and swfobject player to handle all the auth management. The code passed down from the server isn't minified but it's a nightmare to comprehend.

### Source URL 
From the above code I was able to parse out this source url. 
```
http://nlds277.cdnak.neulion.com/nlds_vod/ufc/vod/2017/06/17/28527/2_28527_cam1_cam0_2017_cam0_whole_1_ced.mp4.m3u8?hdnea=expires%3D1497944837%7Eaccess%3D%2Fnlds_vod%2Fufc%2Fvod%2F2017%2F06%2F17%2F28527%2F*%7Emd5%3D80b467247583fc329c2ba2767f552276&nltid=ufc&nltdt=8&nltnt=1&uid=2906116|User-Agent=Mozilla/5.0%20(Linux;%20Android%206.0.1;%20D6603%20Build/23.5.A.0.570;%20wv)%20AppleWebKit/537.36%20(KHTML,%20like%20Gecko)%20Version/4.0%20Chrome/56.0.2924.87%20Mobile%20Safari/537.36%20android%20mobile%20ufc%207.0310
```
The userAgent doesn't seem to be completely necessary as the `.m0_whole_1_ced.mp4.m3u8` downloads regardless. 

[...]

PR #1086, opened 2017-07-11
[ufctv] updated to capture 4500k streams

jgilf:
Fixed issues discussed in streamlink/streamlink#1029

### DIFF
--- a/src/streamlink/plugins/ufctv.py
+++ b/src/streamlink/plugins/ufctv.py
@@ -81,6 +81,7 @@ class UFCTV(Plugin):
         if video:
             self.logger.debug("Found {type}: {name}", **video)
             surl = self._get_stream_url(video['id'], video.get('type', "video"))
+            surl = surl.replace("_iphone.mp4", ".mp4")
             if surl:
                 return HLSStream.parse_variant_playlist(self.session, surl)
             else:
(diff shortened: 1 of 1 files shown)


beardypig:
Cool, are there any live streams on UFC? - I don't remember :)

jgilf:
Yes there are, they often stream their prelims and smaller shows live.

beardypig:
I'd want to make sure the live streams still work with this change before merging, or if higher qualities are now (or could be made) available for live streams too.

jgilf:
I will check it with their 24/7 stream then :)

jgilf:
You were right to check the live stream as it was still only pulling the 1600k stream.

This was due to the the live stream having a slightly different url:

> http://nlds67.cdnak.neulion.com/nlds/ufc/ufcepg1/as/live/ufcepg1_hd_iphone.m3u8

To access the 4500k stream, this needs to be changed to:

> http://nlds67.cdnak.neulion.com/nlds/ufc/ufcepg1/as/live/ufcepg1_hd.m3u8

My new commit should cover both VOD and live streams now.

beardypig:
Cool 👍

gravyboat:
Thanks for the fix @jgilf and thanks for the review @beardypig!

DanStealth:
Anyway to get the 4k stream? it's only available on sony tv's via the ufc.tv app but id imaigne they share similarities.

PR #1089, opened 2017-07-12
Update Windows portable version documentation

RosadinTV:


gravyboat:
Looks good to me!

RosadinTV:
@gravyboat Kinda strange but the font size in "Package maintainers > Windows port. version" is bigger than the others.
package maintainers font bug

gravyboat:
@RosadinTV I think it has to do with the way the table has been broken up due to multiple entries, maybe try removing line 108? I wonder if the blank line is breaking it somehow.

RosadinTV:
@gravyboat I don't know, it seems like the format changes when converted to html in the repository "streamlink.github.io", so the results may be unexpected.
I'm a bit afraid of doing several trial-and-error commits, but let me know if you want me to change anything.

gravyboat:
You can build the docs locally if you want to test it first, I'm not too concerned though, I would just go with the comma if you're worried about making too many. We need to fix it anyways.

RosadinTV:
Now works as expected, thanks @gravyboat 💝

gravyboat:
@RosadinTV Awesome! If we have one of these again let's make sure to PR it instead of pushing directly to master, just makes it easier to track and roll back if we need to! Thanks for fixing that up.

RosadinTV:
@gravyboat Sorry, I'll keep that in mind for the next one 🚀

gravyboat:
No worries it happens! Thanks again for the quick fix.

PR #1091, opened 2017-07-12
plugins.douyutv: use douyu open API

steven7851:
Use douyu open API which no need sign token.


gravyboat:
Huh this is pretty sneaky, nice changes @steven7851! Thanks!

PR #1097, opened 2017-07-14
plugins.app17: new layout

steven7851:
17app webpage has a big change, new url is https://17.live.


gravyboat:
Thanks @steven7851!

PR #1105, opened 2017-07-19
Update donation docs to note open collective

gravyboat:
This goes along with #1122, if anyone has feedback on the wording please let me know. I considered adding sections for tips based on the payment method, but figured that was too complicated. Open Collective is still using PayPal for payouts so I know some of us can't benefit from it (myself included) but at least we can get the ball rolling so people can make credit card donations easily.


gravyboat:
Thanks @bastimeyer, I fixed all of those up minus adding please to the bounty note, but I did fix the grammar problems.

Issue #1114, opened 2017-07-25
Dailymotion parsing error


BjkGkh:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

Since Dailymotion updated their website, player... streamlink seems to not working anymore and throws a parsing error.

### Expected / Actual behavior

Not working anymore

### Reproduction steps / Explicit stream URLs to test

1. Open a dailymotion live stream.

### Environment details

Operating system and version: Windows 10
Streamlink and Python version: Last version for both.

### Comments, logs, screenshots, etc.

https://puu.sh/wSHA9/99230ea948.png

back-to:
update your streamlink

https://github.com/streamlink/streamlink/releases/tag/0.7.0

if was fixed here #974

BjkGkh:
Thank you it's working! Really appreciated, i was sure i had the last version...

PR #1120, opened 2017-07-27
Resolve InsecurePlatformWarnings for older Python2.7 versions

beardypig:
This should resolve the issues some people are having with HTTPS websites when using older versions of Python 2.7. As best I can tell the any version before 2.7.10 is affected.

See discussion in #1114.


gravyboat:
Thanks @beardypig!

PR #1143, opened 2017-08-04
cli: support for SOCKS proxies

### LINKED ISSUE #1138: SOCKS Support?
can we get socks support? way simpler than http, also requests supports it
https://github.com/requests/requests/issues/3172

beardypig:
Any proxy type that is supported by `requests` can be used.
  
For example, SOCKS proxies are support by `requests`, and can be used:
    - `--http-proxy "socks5://[user:password@]host:port"`
    - `--https-proxy "socks5://[user:password@]host:port"`
to set streamlink to use socks proxies for http and https.

Support for SOCKS proxies provided by an extra dependency: `PySocks` via `requests[socks]`.
Fixes #1138.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -329,6 +329,22 @@ Progressive HTTP, HTTPS, etc   httpstream:// [1]_
 .. [1] supports local files using the file:// protocol
 .. _cli-options:
 
+Proxy Support
+-------------
+
+You can use the :option:`--http-proxy` and :option:`--https-proxy` options to
+change the proxy server that Streamlink will use for HTTP and HTTPS requests respectively.
+As HTTP and HTTPS requests can be handled by separate proxies, you may need to specify both
+options if the plugin you use makes HTTP and HTTPS requests.
+
+Both HTTP and SOCKS5 proxies are supported, authentication is supported for both types.
+
+For example:
+.. code-block:: console
+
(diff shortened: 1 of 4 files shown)


gravyboat:
Hmm, do we really need to support this? I am concerned with the potential of bloating due to people requesting a ton of proxy options/one offs that one person uses and going off scope. I know this was requested, but what use case do we resolve here and does it keep us from doing more work or adding more in the future? If we take the stance of only supporting one or two I am fine, but I don't want to support every single request that comes in until the installer is a gig (it already feels heavy to me at 38 megs).

nxtreaming:
Do not need, because http_proxy/https_proxy is enough.
Forget it.

beardypig:
SOCKS is pretty useful and the standard other than http(s) proxies. I use SOCKS a lot by way of SSH using the built-in SOCKS proxy. The code change is really just to tidy it up all the support for SOCKS is already in requests, and the additional package is pretty small. But, it could be an optional package and SOCKS proxies only works if the package is installed - which is what it should be like already, but the proxy URLs are limited to http(s). 

I like SOCKS ;-)

gravyboat:
@beardypig I'm fine with it then as long as we take these requests with a grain of salt so things don't get bloated.

nwgat:
@beardypig socks is awesome
1. `ssh -D 7070 -N user@ip -i "key"`
2. `streamlink --http-proxy "socks5://localhost:7070"`

gravyboat:
Thanks @beardypig!

PR #1144, opened 2017-08-04
plugins: add priority ordering to plugins

beardypig:
Plugins can return a priority level for the URLs that they support.
If two plugins match the same URLs then the plugin with the higher priority will be used. This means that the built-in HLS and HDS plugins can support `.m3u8` and `.f4m` URLs with a lower priority without causing issues with overlapping URL patterns.

By default plugins return NORMAL priority for any URL. It is expected that the `priority` method will only be called with URLs that are actually supported by the plugin.

Thoughts?

### DIFF
--- a/src/streamlink/plugins/hds.py
+++ b/src/streamlink/plugins/hds.py
@@ -1,22 +1,44 @@
 import re
 
 from streamlink.plugin import Plugin
-from streamlink.plugin.plugin import parse_url_params
+from streamlink.plugin.plugin import parse_url_params, LOW_PRIORITY, NORMAL_PRIORITY, NO_PRIORITY
 from streamlink.stream import HDSStream
 from streamlink.utils import update_scheme
+from streamlink.compat import urlparse
 
 
 class HDSPlugin(Plugin):
-    _url_re = re.compile(r"hds://(.+)")
+    _url_re = re.compile(r"(hds://)?(.+(?:\.f4m)?.*)")
+
+    @classmethod
+    def priority(cls, url):
+        """
+        Returns LOW priority if the URL is not prefixed with hds:// but ends with
+        .f4m and return NORMAL priority if the URL is prefixed.
+        :param url: the URL to find the plugin priority for
+        :return: plugin priority for the given URL
+        """
+        m = cls._url_re.match(url)
+        if m:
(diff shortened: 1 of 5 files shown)


src/streamlink/plugins/hds.py
```diff
@@ -1,22 +1,44 @@
...
 
 class HDSPlugin(Plugin):
-    _url_re = re.compile(r"hds://(.+)")
+    _url_re = re.compile(r"(hds://)?(.+(?:\.f4m)?.*)")
+
+    @classmethod
+    def priority(cls, url):
+        """
+        Returns LOW priority if the URL is not prefixed with hls:// but ends with
+        .m3u8 and return NORMAL priority if the URL is prefixed.
```
Koushien: hds://, .f4m

beardypig: True that. Took me a while to notice what you were referring to, a more verbose comment would be more helpful ;-)

src/streamlink/plugins/akamaihd.py
```diff
@@ -1,7 +1,7 @@
 import re
 
 from streamlink.plugin import Plugin
-from streamlink.plugin.plugin import parse_url_params
+from streamlink.plugin.plugin import parse_url_params, LOW_PRIORITY
```
back-to: you don't even use `LOW_PRIORITY` in this file, is it necessary to `import` it?

beardypig: Not strictly, no.

beardypig: Just left overs from some refactoring...

bastimeyer:
This sounds like a reasonable addition, but you should also add some tests for such a change.

beardypig:
Yeah, you're right - I'll add some.

beardypig:
Added a test that tests the correct plugin is returned when multiple plugins support the same URL. The other tests cover a lot of cases already.

gravyboat:
Okay, one last fix here, can you resolve that merge conflict? Sorry @beardypig.

beardypig:
OK, rebased it on master and fixed a little test discovery warning :)

gravyboat:
Thanks @beardypig!

PR #1154, opened 2017-08-08
plugins.tvplayer: extract the channel id when logged in as a subscriber

beardypig:
A small tweak to handle the slightly different layout when the logged in user has a subscription. I also added some more verbose logging for easier debugging.


beardypig:
Paging @gravyboat, this one had failing tests but I fixed them the other day, should be OK to merge now.

gravyboat:
Sweet, thanks @beardypig!

PR #1159, opened 2017-08-11
plugins.bbciplayer: fix for page formatting changes and login

beardypig:
The beeb now enforces logins for watching iPlayer, logins are now
enforced by the plugin.


gravyboat:

PR #1196, opened 2017-09-01
Fix player URL extraction in bloomberg plugin

sqrt2:
The bloomberg plugin currently cannot find any sources for the live streams because a regular expression matches an unintended part of the page. This PR fixes this issue.


gravyboat:
Thanks @sqrt2!

PR #1200, opened 2017-09-04
plugins.arconaitv: fix arconaitv to new url

JourneyOver:
Just updates ArconaiTv to support the new url.


JourneyOver:
Dunno why travis failed...I tested this change myself before even making a pull request and things are working just fine..

for some reason travis is still trying to use a .me url, when new url is a .co

gravyboat:
Thanks!

Issue #1074, opened 2017-07-07
Bongacams is now prefers hls


horacio9a:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description
The last few days the bongacams plugin does not work.
We need new plugin, which should be somewhat shorter and simpler than the previous one that was based on the rtmp protocol.
Now it seems to prefer the hls protocol..
An example of a new link is:

https://ded6573-edge1.bongacams.com/hls/stream_Polya8Nikita/chunks.m3u8

Working also:

https://ded6573-edge1.bongacams.com/hls/stream_Polya8Nikita/playlist.m3u8.m3u8

So it's enough to find the ip address and know the exact name of the performer.

The above link can of course be reproduced in a variety of ways.
I currently put it in Chrome browser and I watch it using the Play HLS M3u8 plugin ;)

As for the websocket, the situation is like this:

Request URL:wss://chatjs.bongacams.com:8443/websocket
Request Method:GET
Status Code:101 Switching Protocols

Request URL:wss://notice.bongacams.com:9998/190/x2322it4/websocket
Request Method:GET
Status Code:101 Switching Protocols

gravyboat:
There's more discussion on this here: https://github.com/streamlink/streamlink/issues/1072

roboscripts:
Any news, its been a while now with no answers what so ever. I clearly remember you guys coming onto the Livestreamer project saying: hey use streamlink its well maintained and regularly updated bla bla bla not like Livestreamer....

Well i started using u guys for Bonga and i disagree with those statements. If you cant do anything about it just say so. At least on the other project users come in and propose solutions...

Doudlidou:
Any News?

Its been a while now with no answers what so ever. I clearly remember you guys coming onto the Livestreamer project saying: hey use streamlink its well maintained and regularly updated bla bla bla not like Livestreamer....

Well i started using u guys for Bonga and i disagree with those statements. If you cant do anything about it just say so. At least on the other project users come in and propose solutions...

gravyboat:
[...]

PR #1201, opened 2017-09-06
[bongacams] replace RTMP with HLS

sdfwv:
Fixed streamlink/streamlink#1074


gravyboat:
Looks good to me, thanks @sdfwv!

PR #1212, opened 2017-09-12
plugins.bbciplayer: support for older VOD streams

beardypig:
When old episodes are made available again (because they were on TV again) they sometimes have rtmp streams available as well, this accounts for the slight difference in the data returned by the API in such cases.


gravyboat:
Thanks @beardypig.

PR #1218, opened 2017-09-13
Update ElTreceTV

RosadinTV:
Fix live content support.

### DIFF
--- a/src/streamlink/plugins/eltrecetv.py
+++ b/src/streamlink/plugins/eltrecetv.py
@@ -2,6 +2,7 @@ import re
 
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import http, useragents
+from streamlink.stream import HLSStream
 from streamlink.utils import parse_json
 
 class ElTreceTV(Plugin):
@@ -15,13 +16,11 @@ class ElTreceTV(Plugin):
         if "eltrecetv.com.ar/vivo" in self.url.lower():
             try:
                 http.headers = {'Referer': self.url,
-                'User-Agent': "ArtearPlayer/3.2.42 (Linux;Android 4.4.2) ExoPlayerLib/1.5.9"}
+                'User-Agent': useragents.ANDROID}
                 res = http.get('https://api.iamat.com/metadata/atcodes/eltrece')
-                live_search = res.text
-                live_search = live_search[live_search.index('"vivo":{"provider":"youtube","youtubeVideo":')+45 :]
-                live_search = live_search[: live_search.index('"')]
-                LIVE_URL_FOUND = 'https://www.youtube.com/watch?v=' + live_search
-                return self.session.streams(LIVE_URL_FOUND)
+                yt_id = parse_json(res.text)["atcodes"][0]["context"]["ahora"]["vivo"]["youtubeVideo"]
+                yt_url = "https://www.youtube.com/watch?v={0}".format(yt_id)
+                return self.session.streams(yt_url)
             except:
                 self.logger.info("Live content is temporarily unavailable. Please try again later.")
         else:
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/eltrecetv.py
```diff
@@ -13,27 +12,31 @@ def can_handle_url(cls, url):
...
+                http.headers = {'Referer': self.url,
+                'User-Agent': useragents.CHROME}
+                res = http.get(self.url)
+                _player_re = re.compile(r'''data-kaltura="([^"]+)"''')
+                match = _player_re.search(res.text)
+                if not match:
+                    return
+                json_video_search = parse_json(match.group(1).replace("&quot;", '"'))
+                VOD_URL_FOUND = "https://vodgc.com/p/111/sp/11100/playManifest/entryId/" + json_video_search["entryId"] + "/format/applehttp/protocol/https/a.m3u8"
+                return self.session.streams(VOD_URL_FOUND)
```
beardypig: This should really just use `HLSStream` or `HLSStream.parse_variant_playlist`, which ever applies.

src/streamlink/plugins/eltrecetv.py
```diff
@@ -13,27 +12,31 @@ def can_handle_url(cls, url):
...
-
         if "eltrecetv.com.ar/vivo" in self.url.lower():
-            http.headers = {'Referer': self.url,
-            'User-Agent': "ArtearPlayer/3.2.42 (Linux;Android 4.4.2) ExoPlayerLib/1.5.9"}
-            video_url_found_hls = "http://stream.eltrecetv.com.ar/live13/13tv/13tv1/playlist.m3u8"
+            try:
+                http.headers = {'Referer': self.url,
+                'User-Agent': "ArtearPlayer/3.2.42 (Linux;Android 4.4.2) ExoPlayerLib/1.5.9"}
+                res = http.get('https://api.iamat.com/metadata/atcodes/eltrece')
+                live_search = res.text
```
beardypig: I think using `http.json` would be better for parsing the API response rather than using substring - it will be more reliable in the event of small changes to the response. It will also be easier for other people to understand, which is important for future maintenance :)

src/streamlink/plugins/eltrecetv.py
```diff
@@ -13,27 +12,31 @@ def can_handle_url(cls, url):
...
-
-        video_url_found_hls = ""
-
         if "eltrecetv.com.ar/vivo" in self.url.lower():
-            http.headers = {'Referer': self.url,
-            'User-Agent': "ArtearPlayer/3.2.42 (Linux;Android 4.4.2) ExoPlayerLib/1.5.9"}
-            video_url_found_hls = "http://stream.eltrecetv.com.ar/live13/13tv/13tv1/playlist.m3u8"
+            try:
+                http.headers = {'Referer': self.url,
+                'User-Agent': "ArtearPlayer/3.2.42 (Linux;Android 4.4.2) ExoPlayerLib/1.5.9"}
```
beardypig: I'd extract the User Agent to a class or module level constant.

src/streamlink/plugins/eltrecetv.py
```diff
@@ -13,27 +12,31 @@ def can_handle_url(cls, url):
...
+                res = http.get('https://api.iamat.com/metadata/atcodes/eltrece')
+                live_search = res.text
+                live_search = live_search[live_search.index('"vivo":{"provider":"youtube","youtubeVideo":')+45 :]
+                live_search = live_search[: live_search.index('"')]
+                LIVE_URL_FOUND = 'https://www.youtube.com/watch?v=' + live_search
```
beardypig: Variables in caps should be avoided, the convention is that caps are reserved for constants. I would make the `https://www.youtube.com/watch?v=` string a constant and put it at the class or module level, eg.
```python
class ElTreceTV(Plugin):
    LIVE_URL = "https://www.youtube.com/watch?v="
```
I'd also use `string.format` as that is generally considered superior to string concatenation, (`"https://www.youtube.com/watch?v={0}".format(yt_id)`)

src/streamlink/plugins/eltrecetv.py
```diff
@@ -13,27 +12,31 @@ def can_handle_url(cls, url):
...
+            try:
+                http.headers = {'Referer': self.url,
+                'User-Agent': useragents.CHROME}
+                res = http.get(self.url)
+                _player_re = re.compile(r'''data-kaltura="([^"]+)"''')
+                match = _player_re.search(res.text)
+                if not match:
+                    return
+                json_video_search = parse_json(match.group(1).replace("&quot;", '"'))
+                VOD_URL_FOUND = "https://vodgc.com/p/111/sp/11100/playManifest/entryId/" + json_video_search["entryId"] + "/format/applehttp/protocol/https/a.m3u8"
```
beardypig: See my comments on `LIVE_URL_FOUND`.

RosadinTV:
@beardypig Thanks for the review, the code is now much cleaner.

beardypig:
@RosadinTV no problem, looks good now :)

gravyboat:
Thanks @RosadinTV and thanks for the review @beardypig!

PR #1220, opened 2017-09-14
installer: include the socks proxy modules

beardypig:
Noticed that the PySocks module isn't actually included in the Windows installer version :) The `win-inet-pton` module isn't required as we're using Python 3 for the installer version.


gravyboat:

Issue #1222, opened 2017-09-15
Plugin request for http://webcast.gov.in/


rusty148:
Hi,

This is a plugin request for http://webcast.gov.in/.

There are following channels inside. 
http://webcast.gov.in/ddpunjabi/
http://webcast.gov.in/ddkisan/
http://www.webcast.gov.in/vyaslive
http://webcast.gov.in/

I am using streamlink portable downloaded from here:  https://github.com/streamlink/streamlink-portable/releases

I am looking for something like this : set url=http://d1hya96e2cm7qi.cloudfront.net/Live/_definst_/amlst:sweetbcha1novD251L720P/playlist.m3u8
streamlink.exe --http-header "Referer=http://www.zengatv.com/live/f132cc60-7635-4b74-b4e9-ce8c43727635.html" hlsvariant://%url% best

source : https://github.com/streamlink/streamlink/issues/890

Thanks
----

### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

...

### Expected / Actual behavior

...

### Reproduction steps / Explicit stream URLs to test

1. ...
2. ...
3. ...

### Environment details

Operating system and version: ...
Streamlink and Python version: ...

...

### Comments, logs, screenshots, etc.

...

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

rusty148:
--http-header "Referer=http://webcast.gov.in/ddpunjabi/" --player "C:\Program Files (x86)\Windows Media Player\wmplayer.exe" -o "C:\ddb.mp4" hlsvariant://http://http://nicls3-lh.akamaihd.net/i/ddpunjabi_1@122402/master.m3u8 144p


I am using this command, please let me know what's missing

sdfwv:
- you are using `http://` twice
- you don't need a Referer
- **144p** does not exists, it's **576p** for me or **best / worst**

```
streamlink hls://http://nicls3-lh.akamaihd.net/i/ddpunjabi_1@122402/master.m3u8 best
# or
streamlink hls://http://nicls3-lh.akamaihd.net/i/ddpunjabi_1@122402/master.m3u8 best -o 123.mp4
```
should work.

rusty148:
Hi, I tried different combinations and finally this command worked for me:
[...]

PR #1224, opened 2017-09-15
Add support for webcast.gov.in

RosadinTV:
As issued in https://github.com/streamlink/streamlink/issues/1222


gravyboat:
Looks good to me, thanks @RosadinTV.

PR #1230, opened 2017-09-17
docs: add new line before codeblock to fix them

amurzeau:
Hi,

A codeblock without an empty new line before cause bad docs generation in man and html formats.

This causes a `man` warning: `warning: macro `.' not defined`.
The manpage contains `.. code\-block:: console` which it does not recognize.

In html format, the  generated line is `For example: .. code-block:: console` and the following codeblock is not correctly formatted.

When adding a empty new line before (as this PR does), the docs are correctly generated.
I've not investigated much if this is a sphinx issue or not (as I don't know much reStructuredText language) but this fix docs generation.


gravyboat:
Good catch, thanks!

PR #1247, opened 2017-09-29
exit with code 130 after a KeyboardInterrupt

### LINKED ISSUE #1234: keyboard interrupt should not exit with code 0
currently streamlink exit with 0 when hit by keyboard interrupt, 
it's hard to handle within automation script,

https://unix.stackexchange.com/questions/223189/what-does-exit-code-130-mean-for-postgres-command
 
 - if the process exited with exit(n) or return n from main(): the lower 8 bits of n (n & 0xFF).
 - if the process was killed by signal n: n + 128.

that do have some convention so people can handle them graceful with script

xgi:
Adds sys.exit(130) where applicable after a KeyboardInterrupt occurs. Notably there are two consecutive KeyboardInterrupt checks; this commit causes a code 130 even if only the first occurs and not the additional "panic" one.

130 is the correct exit code in this case because KeyboardInterrupt causes a SIGINT which has value n=2 and the appropriate exit code is 128+n (see: http://tldp.org/LDP/abs/html/exitcodes.html)

Fixes #1234

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -1043,18 +1043,21 @@ def main():
             setup_options()
             setup_plugin_options()
             handle_url()
-        except KeyboardInterrupt:
+        except KeyboardInterrupt as keyboard_interrupt:
             # Close output
             if output:
                 output.close()
             console.msg("Interrupted! Exiting...")
+            interrupt = keyboard_interrupt
         finally:
             if stream_fd:
                 try:
                     console.logger.info("Closing currently open stream...")
                     stream_fd.close()
+                    if 'interrupt' in locals():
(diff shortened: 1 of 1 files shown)


gravyboat:
Looks good to me, @beardypig any concerns?

beardypig:
Maybe it's better to set an `exit_code` variable at the start of `main()` to `0`, then set it to `130` during the `KeyboardInterrupt`?  That way we can more easily change the exit code for other reasons.

beardypig:
@xgi that looks nicer :)

gravyboat:
Thanks @xgi and thanks for reviewing @beardypig!

PR #1265, opened 2017-10-10
Handle repeated keys in parse_params

worstperson:
The conn option for rtmpdump my need to be added multiple depending on the target server, but parse_params overwrites repeated keys with the last value processed. This patch creates and adds values into a list that the function bake in streamprocess.py handles.

### DIFF
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -149,7 +149,13 @@ def parse_params(params):
         except Exception:
             pass
 
-        rval[key] = value
+        if key in rval:
+            if not isinstance(rval[key], list):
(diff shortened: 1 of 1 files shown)


gravyboat:
This looks fine to me. @beardypig opinions?

beardypig:
You can already pass multiple value for conn by doing `conn=["one", "two"]`. This might be a nicer way to do it, and I guess it wouldn't break the current functionality - just need to test that.

gravyboat:
@beardypig Are we awaiting tests of this nature on the PR or did you have something else in mind?

worstperson:
Thank you beardypig! Closing this request since the functionality already exists.

PR #1274, opened 2017-10-16
plugins.turkuvaz: fix some turkuvaz sites and add support for anews

beardypig:
Fix support for `minikacocuk.com.tr` and `atvavrupa.tv` live streams, and add support for `anews.com.tr`.


gravyboat:
Nice, thanks @beardypig!

PR #1276, opened 2017-10-16
plugins.dogus/startv: fix dogus sites

beardypig:
HDS streams don't appear to be supported any more, and star tv is now an embedded YouTube live stream.


gravyboat:
Thanks @beardypig!

Issue #1273, opened 2017-10-15
[Zattoo] Streamlink crashes on Arch Linux


MaLomm:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

`streamlink -p mpv "https://zattoo.com/watch/daserste" best --zattoo-email email --zattoo-password password` results in: 

```
[cli][info] Found matching plugin zattoo for URL https://zattoo.com/watch/daserste
Traceback (most recent call last):
  File "/usr/bin/streamlink", line 11, in <module>
    load_entry_point('streamlink==0.8.1', 'console_scripts', 'streamlink')()
  File "/usr/lib/python3.6/site-packages/streamlink_cli/main.py", line 1043, in main
    handle_url()
  File "/usr/lib/python3.6/site-packages/streamlink_cli/main.py", line 482, in handle_url
    streams = fetch_streams(plugin)
  File "/usr/lib/python3.6/site-packages/streamlink_cli/main.py", line 394, in fetch_streams
    sorting_excludes=args.stream_sorting_excludes)
  File "/usr/lib/python3.6/site-packages/streamlink/plugin/plugin.py", line 385, in get_streams
    return self.streams(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/streamlink/plugin/plugin.py", line 288, in streams
    ostreams = self._get_streams()
  File "/usr/lib/python3.6/site-packages/streamlink/plugins/zattoo.py", line 224, in _get_streams
    if self._expires < time.time():
TypeError: '<' not supported between instances of 'NoneType' and 'float'
```

### Reproduction steps / Explicit stream URLs to test

1. https://zattoo.com/watch/daserste

### Environment details

Operating system and version: Arch Linux, current
Streamlink version: streamlink 0.8.1 (git)
Python version: Python 3.6.2

PR #1293, opened 2017-10-27
[zattoo] It won't work with None in Python 3.6, set always a default …

back-to:
…date instead of None.

Fixed streamlink/streamlink#1273


gravyboat:
Thanks @back-to!

Issue #1295, opened 2017-10-27
live.me plugin fails


JohnDoh666:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description
The liveme.com plugin fails to load streams

### Reproduction steps / Explicit stream URLs to test

```
streamlink http://liveme.com/live.html?videoid=1509098076123280858 best
[cli][info] Found matching plugin liveme for URL http://liveme.com/live.html?videoid=15090980761232808581&countryCode=undefined
error: Unable to validate JSON: Unable to validate key 'status': u'400' does not equal '200'
```
### Environment details

Operating system and version: ...
MacOS

Streamlink and Python version: ...
Streamlink v0.81 on python 2.7

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

PR #1298, opened 2017-10-30
[liveme] API update

back-to:
closes streamlink/streamlink#1295

### DIFF
--- a/src/streamlink/plugins/liveme.py
+++ b/src/streamlink/plugins/liveme.py
@@ -1,4 +1,5 @@
-from __future__ import print_function
+import math
+import random
 import re
 
 from streamlink.plugin import Plugin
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/liveme.py
```diff
@@ -26,6 +27,16 @@ class LiveMe(Plugin):
     def can_handle_url(cls, url):
         return cls.url_re.match(url) is not None
 
+    def _random_t(self, t):
+        n = "ABCDEFGHJKMNPQRSTWXYZabcdefhijkmnprstwxyz2345678"
+        e = len(n)
+        a = ""
+        r = 0
+        while (r < t):
+            a += n[math.floor(random.randrange(e))]
```
beardypig: `random.choice` might be a better method to use, eg. `random.choice(n)`. 
Personally, I'd just use a list comprehension/generator to do it :) 
```python
def _random_t(self, t):
    return "".join(random.choice("ABCDEFGHJKMNPQRSTWXYZabcdefhijkmnprstwxyz2345678") for _ in range(t))
```

beardypig:
Nice to see you back to streamlink @back-to ... :)

src/streamlink/plugins/liveme.py
back-to: i just did the same, as it is on their website.

```js
function t(t) {
    t = t || 32;
    for (var n = "ABCDEFGHJKMNPQRSTWXYZabcdefhijkmnprstwxyz2345678", e = n.length, A = "", r = 0; r < t; r++) A += n.charAt(Math.floor(Math.random() * e));
    return A
}
```

but `random.choice` looks easier.

gravyboat:
Looks good, thanks @back-to and thanks @beardypig for the review.

Issue #1236, opened 2017-09-21
Error when writing to output: [WinError 10053]


Carrybag1:
### Checklist

- [x] This is a bug report.
- [x] I have read the contribution guidelines.

### Description

When trying to open streams sometimes it will not open the stream and give the below error. Ever since I decided to update my streamlink from 0.5.0

"error: Error when writing to output: [WinError 10053] An established connection was aborted by the software in your host machine, exiting"

This completely stops happening when I uncheck "reconnect to stream" in streamlink settings

Unsure if this means it's something in the streamlink gui or streamlink itself

skulblakka:
This has been reported before but sadly no progress was made. See #916 and #975

Issue #975, opened 2017-06-01
Streamlink sometimes closing stream after closing HTTP connection


x-traxion:
### Checklist

- [x] This is a bug report.
- [x] I have read the contribution guidelines.

### Description
Hi,
Im using Streamlink to get Twitch Streams in my Bouquet of my Linux Reciever (VU+ Solo4k).
Long time i dont use Streamlink or Livestreamer.
Yesterday i started Streamlink with a batch File on my PC as usual. It works fine till i zapped to another Channel (Sat) and zapped back to the Stream but it dont work.
On my PC i got an Error in the cmd Window and i have to restart the batch file.
I get this in 9 out of 10 tries.

Operating system and version: Windows 7 x64 pro, Marusys VU+ Solo4K
Streamlink version: 0.6.0

### Comments, logs, screenshots, etc.
```
C:\Streamlink\bin\streamlink.exe" --player-external-http --player-external-http-port 9999 --player-continuous-http twitch.tv/rocketbeanstv best
[cli][info] Found matching plugin twitch for URL twitch.tv/rocketbeanstv
[cli][info] Available streams: audio_only, 160p (worst), 360p, 480p, 720p, 1080p (best)
[cli][info] Starting server, access with one of:
[cli][info]  http://127.0.0.1:9999/
[cli][info]  http://192.168.124.200:9999/
[cli][info] Got HTTP request from Vu+ STB HbbTV/1.1.1 (+PVR+RTSP+DL;VTi;;;)
[cli][info] Opening stream: 1080p (hls)
error: Error when writing to output: [WinError 10053] Eine bestehende Verbindung wurde softwaregesteuert durch den Hostcomputer abgebrochen, exiting
[cli][info] Stream ended
[cli][info] Closing currently open stream...
```

horacio9a:
I don't know how much it will help my experience.
Operating system and version: Windows 7 x32 Ultimate
Streamlink version: 0.6.0

Started this link in last Firefox: http://twitch.tv/rocketbeanstv

With last IDM received https://video-edge-c2b008.fra02.hls.ttvnw.net ... ffplay has no problems with it
Same is with streamlink!
![2250](https://cloud.githubusercontent.com/assets/17076459/26734952/4527965c-47c0-11e7-93bf-3013d264bf80.jpg)
![2251](https://cloud.githubusercontent.com/assets/17076459/26734965/4e028d40-47c0-11e7-8eac-cd547c0ceb69.jpg)

[...]

Issue #916, opened 2017-05-13
Streamlink 0.6.0 fails to play streams on PotPlayer


Tikounet:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

PotPlayer has been causing an error in streamlink for quite some time.
Until version 0.6 streamlink recovered from it and it was fine.
With version 0.6 streamlink exits after the error.

### Reproduction steps / Explicit stream URLs to test
Here is the command line i use (works with any stream from twitch at least)
```
streamlink --player "D:\PotPlayer64\PotPlayerMini64.exe" --player-continuous-http --player-passthrough rtmp --retry-streams 1 --retry-open 99 --hds-segment-threads 2 --hls-segment-threads 2 --twitch-oauth-token xxxxxxxxxxxxxxx https://www.twitch.tv/aimbotcalvin best

Output with version 0.5
[cli][info] Found matching plugin twitch for URL https://www.twitch.tv/aimbotcalvin
[plugin.twitch][info] Attempting to authenticate using OAuth token
[plugin.twitch][info] Successfully logged in as xxxxxx
[cli][info] Available streams: 720p60_alt, audio_only, 160p (worst), 360p, 480p, 720p, 720p60 (best)
[cli][info] Starting player: D:\PotPlayer64\PotPlayerMini64.exe
[cli][info] Got HTTP request from PotPlayer
[cli][info] Opening stream: 720p60 (hls)
[cli][error] Error when writing to output: [WinError 10053] An established connection was aborted by the software in your host machine
[cli][info] Stream ended
[cli][info] Got HTTP request from WINAMP
[cli][info] Opening stream: 720p60 (hls)

Output with version 0.6
[cli][info] Found matching plugin twitch for URL https://www.twitch.tv/aimbotcalvin
[plugin.twitch][info] Attempting to authenticate using OAuth token
[plugin.twitch][info] Successfully logged in as xxxx
[cli][info] Available streams: 720p60_alt, audio_only, 160p (worst), 360p, 480p, 720p, 720p60 (best)
[cli][info] Starting player: D:\PotPlayer64\PotPlayerMini64.exe
[cli][info] Got HTTP request from PotPlayer
[cli][info] Opening stream: 720p60 (hls)
error: Error when writing to output: [WinError 10053] An established connection was aborted by the software in your host machine, exiting
[cli][info] Stream ended
[...]

PR #1315, opened 2017-11-07
Ignore WinError 10053 / WSAECONNABORTED

back-to:
A nice solution for this error was found in gevent/gevent#1000

Fixed streamlink/streamlink#916
Fixed streamlink/streamlink#975
Fixed streamlink/streamlink#1236

Maybe this will help streamlink/streamlink-twitch-gui#467

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -25,6 +25,10 @@ from .output import FileOutput, PlayerOutput
 from .utils import NamedPipe, HTTPServer, ignored, progress, stream_to_url
 
 ACCEPTABLE_ERRNO = (errno.EPIPE, errno.EINVAL, errno.ECONNRESET)
+try:
+    ACCEPTABLE_ERRNO += (errno.WSAECONNABORTED,)
+except AttributeError:
+    pass  # Not windows
 QUIET_OPTIONS = ("json", "stream_url", "subprocess_cmdline", "quiet")
 
 args = console = streamlink = plugin = stream_fd = output = None


beardypig:
Nice. Seems like it would work :) 
Would be nice to get some verification from a couple of users before closing those issues though.

back-to:
This PR can be easy tested **on windows**

```
streamlink --player-external-http --player-external-http-port 9999 --player-continuous-http twitch.tv/rocketbeanstv best
```

open **http://127.0.0.1:9999/** in the browser(firefox) and click on abort(download)

**without** this PR it will **end streamlink**,
**with** this PR it should **just close** the stream and **not end streamlink**.

beardypig:
@back-to yeah. I just wanted to get some verification that it fixed the issues that would be closed by merging this PR. I tested PotPlayer just now and it does fix it. Good job 🤗

skulblakka:
Just tested it with mpv. Issue seems to be fixed with this PR.

bastimeyer:
Looks like this fixes the issue.
Is this ready for merging? We have 4 hours to merge this (until midnight UTC+0) so that it gets included in the next nightly build, which should help us getting a few more verifications.

gravyboat:
Thanks for the reviews everyone. @bastimeyer apologies on the delay. @back-to Thanks for fixing this up.

bastimeyer:
> apologies on the delay

?
It's all fine... I thought the nightly builds were scheduled by TravisCI at midnight UTC+0, but it's not the case. Would be quite stupid of them triggering builds of all registered projects at the same time. Nightlies are being built at 18:19:49 here (probably because of the time the daily cron job was created).

Thanks for merging though, @gravyboat, and thanks @back-to for the fix.

Tikounet:
I confirm this fixes the issue with PotPlayer (916).
Thanks !

Issue #1271, opened 2017-10-14
camsoda plugin 


ls117:
### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

Unable to validate JSON: Key 'app' not found in {'message': 'Access Denied', 'status': 0}
as of 5PM the plugin no longer works as it once did. i can only record in free chat. Seems like camsoda changed something. I saw someone else had a similar issue with twitch back in february.

edit
error: Unable to parse JSON: Expecting value: line 1 column 1 (char 0) ()

now i get this when im logged in, so it seems they made a change forcing you to actually be logged in, which isnt the problem at all, its just an interesting change. but im hoping someone more experienced with this stuff can help figure it out. i think thats the issue with streamlink now, it doesnt log in as an actual user so you cant view/record anything in a gold show or private.

...

### Reproduction steps / Explicit stream URLs to test

https://www.camsoda.com/prettyangeleyez

### Environment details

Operating system and version:  windows 8.1
Streamlink and Python version:  0.8.1

...

### Comments, logs, screenshots, etc.

...

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

ls117:
it may just be the site, other models in private or ticket shows are also not loading at all, on pc or my phone, and thats if i turn off wifi too. So i know its not an ip ban. i may just be panicking for nothing.

ls117:
error: Unable to parse JSON: Expecting value: line 1 column 1 (char 0) ()

now i get this when im logged in, so it seems they made a change forcing you to actually be logged in, which isnt the problem at all, its just an interesting change. but im hoping someone more experienced with this stuff can help figure it out.

ls117:
url snooper still works, so its technically possible to build a working link that i can save on my own, with rtmpdump. using streamlink/xstreamloader is far more convenient however, and when you're paying for something that you can't quite save, its best to be able to record with one click, as opposed to typing the path.

ls117:
*since you wanted the issue isolated to one thread i deleted this comment from the most recent thread, and im posting it here*

[...]

PR #1334, opened 2017-11-18
[camsoda] Fixed broken plugin

back-to:
- only free public streams are supported
- use "privacy.resistFingerprinting" useragent for FireFox
- removed html tests, regex only tests are enough

Fixed streamlink/streamlink#1271

### DIFF
--- a/src/streamlink/plugin/api/useragents.py
+++ b/src/streamlink/plugin/api/useragents.py
@@ -19,6 +19,6 @@ SAFARI_8 = ("Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_4) "
             "AppleWebKit/600.7.12 (KHTML, like Gecko) Version/8.0.7 Safari/600.7.12")
 SAFARI_7 = ("Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_3) "
             "AppleWebKit/537.75.14 (KHTML, like Gecko) Version/7.0.3 Safari/7046A194A")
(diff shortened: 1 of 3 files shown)


src/streamlink/plugin/api/useragents.py
```diff
@@ -19,6 +19,6 @@
             "AppleWebKit/600.7.12 (KHTML, like Gecko) Version/8.0.7 Safari/600.7.12")
 SAFARI_7 = ("Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_3) "
             "AppleWebKit/537.75.14 (KHTML, like Gecko) Version/7.0.3 Safari/7046A194A")
-FIREFOX = "Mozilla/5.0 (Windows NT 10.0; WOW64; rv:50.0) Gecko/20100101 Firefox/50.0"
+FIREFOX = "Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:52.0) Gecko/20100101 Firefox/52.0"
```
gravyboat: Are we going to have any issues swapping this over? I can't think of anything and all the tests pass, but it would be worth a sanity check.

back-to: it shouldn't be an issue,

it's just an update for a still used User-Agent with Firefox
`privacy.resistFingerprinting` in Firefox57 will change the User-Agent to this

beardypig: Looks perfectly reasonable to me 😎

gravyboat:
Thanks @back-to!

PR #1349, opened 2017-12-01
EOL Python 3.3

beardypig:
Remove Python 3.3 from the automated testing on travis and appveyor.

fixes #1348 and the build issues with #1345

PR #1353, opened 2017-12-02
Add plugin for olympicchannel.com

xela722:
This plugin works for both non-premium live streams, and vods on the site. I'm kinda new to this so if the naming conventions or other things like that are off let me know and I'll do my best to fix them,

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -144,6 +144,7 @@ nrk                 - tv.nrk.no          Yes   Yes   Streams may be geo-restrict
                     - radio.nrk.no
 oldlivestream       - original.li.. [3]_ Yes   No    Only mobile streams are supported.
                     - cdn.livestream.com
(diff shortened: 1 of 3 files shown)


gravyboat:
Can you please rebase? #1349 is merged. Also would you mind adding some tests for this plugin if possible?

xela722:
Sorry, I'm a little new at git and everything. Also, I'm not quite sure how the test plugins work so I'm not sure I could apply one reliably.

gravyboat:
@xela722 No worries, if you feel like doing tests there are a couple ways, you can do something more thorough like https://github.com/streamlink/streamlink/blob/master/tests/test_plugin_adultswim.py, or something simple like https://github.com/streamlink/streamlink/blob/master/tests/test_plugin_bloomberg.py where you just feed urls to the functions and assert that they are true/false.

gravyboat:
Looks good to me! Thanks @xela722! I appreciate you adding tests as well.

Issue #957, opened 2017-05-27
Mixer.com / Beam.pro


LufiaSelan:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description
Greetings. It seems beam.pro changed to mixer.com over the past day and mixer.com links are not currently recognized by streamlink as a valid plugin. Additionally, streamlink requests for beam.pro seem to currently have three behaviors upon attempting to load a stream:
1. stream loads as expected
2. streamlink request returns no data from stream
3. error produced (below)

Error:
[Streamlink for Windows v0.6.0 - Git f286765]
[cli][info] Found matching plugin beam for URL beam.pro/channelone
[cli][info] Available streams: 720p_hls, 480p_hls, 480p (worst), 720p (best)
[cli][info] Opening stream: 720p (rtmp)
[cli][error] Try 1/1: Could not open stream <RTMPStream({'flv': '-', 'rtmp': 'rt
mp://dist4-rtmp.mixer.com:1935/beam//58717_source.mp4'}, redirect=False> (No dat
a returned from stream)
error: Could not open stream <RTMPStream({'flv': '-', 'rtmp': 'rtmp://dist4-rtmp
.mixer.com:1935/beam//58717_source.mp4'}, redirect=False>, tried 1 times, exitin
g
[cli][info] Closing currently open stream...
[End of Streamlink for Windows]

[Streamlink for Windows v0.6.0 - Git f286765]
[cli][info] Found matching plugin beam for URL beam.pro/cliffordakamiku
[cli][info] Available streams: 720p_hls, 480p_hls, 480p (worst), 720p (best)
[cli][info] Opening stream: 720p (rtmp)
[cli][error] Try 1/1: Could not open stream <RTMPStream({'flv': '-', 'rtmp': 'rt
mp://dist5-rtmp.mixer.com:1935/beam//201348_source.mp4'}, redirect=False> (No da
ta returned from stream)
error: Could not open stream <RTMPStream({'flv': '-', 'rtmp': 'rtmp://dist5-rtmp
.mixer.com:1935/beam//201348_source.mp4'}, redirect=False>, tried 1 times, exiti
ng
[cli][info] Closing currently open stream...
[End of Streamlink for Windows]

I picked some random streams that were on the front page for this report.

### Environment details

Operating system and version: Windows 7 Pro x64
[...]

PR #1356, opened 2017-12-04
 [mixer] replaced beam.pro with mixer.com

back-to:
requires two commits, for a proper commit history _(github blame)_

- New domain mixer.com
- Use Mixer API 1.0
- Removed RTMPStream
- Removed HTTP mp4 VODs (they won't load for me)
- Fixed HLS Livestreams
- Added tests for mixer.com

Fixed streamlink/streamlink#957


gravyboat:
Nice @back-to, looks good. Thanks!

Issue #1163, opened 2017-08-13
Bug report: Problems logging in with btv


roshavagarga:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description
Using any combination of the correct password and email/username/nickname will end with a failed login attempt.

### Reproduction steps / Explicit stream URLs to test

1. streamlink http://www.btv.bg/live/ --btv-username XXX --btv-password XXX
2. [cli][info] Found matching plugin btv for URL http://www.btv.bg/live/
[plugin.btv][error] Login failed, a valid username and password is required
error: No playable streams found on this URL: http://www.btv.bg/live/

### Comments, logs, screenshots, etc.
Whomever wishes to test it, I will immediately send you a username and password, as well as the correct email, since the website only allows logging in via email+password at the moment.
BTV is geo-restricted to Bulgaria, so use one of these proxies if needed:

1. http://bgproxy.org/
2. https://nqma.net/
3. http://www.bgproxy.net/bg/index.php
4. http://www.php-proxy.net/

bastimeyer:
I have labeled this as a plugin issue, but it would be good to know if you're using the correct username and password, so that the person who wants to fix this is not wasting their time if there is no issue. Depending on your command line shell, certain characters need to be escaped or the whole parameter quoted, which is a common mistake when using plugins with a required authentication.

PR #1363, opened 2017-12-07
[BTV] Fixed login return message

back-to:
Fixed https://github.com/streamlink/streamlink/issues/1163


gravyboat:
Thanks @back-to!

PR #1368, opened 2017-12-09
Streamlink, specified stream, …, no !

comradekingu:


gravyboat:
Whatever character you used for line 1052 and 1057 isn't working correctly on the Python 2.7 tests, can you revert that or use a standardized character Python 2.7 can handle? The other changes I like, I'm not sure what we gain there over using 3 periods.

comradekingu:
@gravyboat. It does better in text-to-speech. Is typhographically correct. And makes it a little bit easier to search through code. I tried escaping it.

gravyboat:
I see where you are coming from for sure, seems like it is still failing unfortunately:

```
SyntaxError: Non-ASCII character '\xe2' in file c:\projects\streamlink-4m1xi\src\streamlink_cli\main.py on line 1052, but no encoding declared; see http://python.org/dev/peps/pep-0263/
````

**Comment by bastimeyer** [comment]
> SyntaxError: Non-ASCII character '\xe2'

`…` is [U+2026 (horizontal ellipsis)](http://www.fileformat.info/info/unicode/char/2026/index.htm) and gets interpreted incorrectly here, because the utf-8 sequence of that character is `â¦` (\xe2\x80\xa6).

Is utf-8 even supported on Windows? What do we gain here from using a special unicode character instead of three simple dots? Text2speech is not really a convincing argument when changing stdout/stderr output of streamlink...

Btw, we have to be careful when changing log output of any kind. 3rd party applications using streamlink could depend on certain strings.

**Comment by gravyboat** [comment]
@bastimeyer These are all good points. I'm okay with the changes with that reverted as long as we make sure to mention that in the release notes.

File: src/streamlink_cli/main.py [hunk 1]
```diff
@@ -349,7 +349,7 @@ def handle_stream(plugin, streams, stream_name):
 
             console.msg("{0}", cmdline)
         else:
-            console.exit("The stream specified cannot be translated to a command")
+            console.exit("The specified stream cannot be translated to a command")
```
bastimeyer: `The specified stream cannot be translated into a command.`
(translated into + full stop)

src/streamlink_cli/main.py
```diff
@@ -359,7 +359,7 @@ def handle_stream(plugin, streams, stream_name):
         try:
             console.msg("{0}", stream.to_url())
         except TypeError:
-            console.exit("The stream specified cannot be translated to a URL")
+            console.exit("The specified stream cannot be translated to a URL")
```
bastimeyer: `The specified stream cannot be translated into a URL.`
(translated into + full stop)

src/streamlink_cli/main.py
```diff
@@ -981,7 +981,7 @@ def setup_plugin_options():
 def check_root():
     if hasattr(os, "getuid"):
         if os.geteuid() == 0:
-            console.logger.info("streamlink is running as root! Be careful!")
+            console.logger.info("Streamlink is running as root! Be careful!")
```
bastimeyer: I think this was meant to not be capitalized here, because it's referring to the `streamlink` entry script on *nix systems. This was the previous message:
https://github.com/streamlink/streamlink/commit/5f9045aed303de87a2049084a0a397877710fd90#diff-35aaec566eb720d1ce55460c85bf4ce9L841

src/streamlink_cli/main.py
```diff
@@ -1006,7 +1006,7 @@ def check_version(force=False):
                             "available!".format(latest_version))
         cache.set("version_info_printed", True, (60 * 60 * 6))
     elif force:
-        console.logger.info("Your Streamlink version ({0}) is up to date!",
+        console.logger.info("Your Streamlink version ({0}) is up to date.",
```
bastimeyer: Why? This could potentially break 3rd party applications. And since log messages are all over the place and don't follow a certain pattern, having an exclamation mark here instead of a full stop doesn't improve the log messages as a whole.

Having a globally unified log format would be nice, but I guess that would require a major version bump, wouldn't it?

gravyboat:
There hasn't been any movement on these changes so I'm going to close this for now.

comradekingu:
Will look at this once I am home again. :)

Issue #1190, opened 2017-08-28
Plugin request:live.qq.com


kesdoputr:
### Checklist

- [x] This is a feature request.
- [x] I have read the contribution guidelines.

### Description
I use streamlink to view china streams like douyu and huomao.
and live.qq.com is another china streams site for sports.
Could someone write a plugin for the site **live.qq.com**?

ex:
http://live.qq.com/10005384

gravyboat:
Any additional details you can provide regarding what network traffic, etc. looks like would be appreciated.

kesdoputr:
live.qq is like douyutv,and when stream is about sports will on it.
So maybe is similar with douyu.

fozzysec:
live.qq.com uses the same structure like douyu.com there are similar APIs and actually douyu.com and live.qq.com has some co-operations. Some modify on the current Douyutv plugin can make it be able to stream live.qq.com. The only problem existing is that we need to reverse engineering the swf player of live.qq.com to get the API key.

gravyboat:
@fozzysec Is the streaming on this site legitimate streams as opposed to stolen content? I just want to make sure we aren't adding more restream sites.

fozzysec:
@gravyboat There are no legal problem with live.qq.com, I think all of their content are permitted such as some sports channel. But those license may only apply to Mainland China, but it is not what we concern about, watching streams on it out of Mainland China is ok.

PR #1371, opened 2017-12-11
[qq] New Plugin for live.qq.com

back-to:
Fixed https://github.com/streamlink/streamlink/issues/1190

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -157,6 +157,7 @@ pluzz               - france.tv          Yes   Yes   Streams may be geo-restrict
                     - zouzous.fr
                     - france3-reg.. [8]_
 powerapp            powerapp.com.tr      Yes   No
+qq                  live.qq.com          Yes   No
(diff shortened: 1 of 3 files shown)


gravyboat:
@back-to What is the content on this site like? I just want to be sure we aren't merging in a new plugin that's used to restream stolen content.

back-to:
i don't know, i can't read anything in chinese

i only checked that **Tencent** is the owner of `qq.com` and
they have like all livestreaming rights in china.

https://www.theverge.com/2017/8/22/16182754/nfl-tencent-china-streaming-deal
https://www.si.com/nba/2015/01/29/nba-deal-china-tencent

so they must tolerate or permit it at least.

or maybe china has just some other / no rules.

**I don't care if it gets merged or not.**

karlo2105:
Qq.com is just a video hosting as youtube, so I don't see any issues with merging. On the other hand if it's available freely there is not any big deal.

gravyboat:
Thanks @back-to!

Issue #1362, opened 2017-12-07
mlg.tv plugin no longer working


a0bie:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description

streamlink can no longer find mlg.tv (majorleaguegaming.com) player streams. 

### Expected / Actual behavior

Expected streamlink to retrieve stream successfully and forward it to the player, but it failed to do so and returned an error.

### Reproduction steps / Explicit stream URLs to test

```
$ streamlink http://eve.majorleaguegaming.com/overwatchleague

[cli][info] Found matching plugin mlgtv for URL http://eve.majorleaguegaming.com/overwatchleague

error: No playable streams found on this URL: http://eve.majorleaguegaming.com/overwatchleague
```

### Environment details

Operating system and version: macOS Sierra 10.12.6
Streamlink and Python version: streamlink 0.9.0, Python 2.7.10

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

stealthplayer:
Hello, I had the same problem too. From what i could find the strings on line 15 and 16 no longer appear on the page.
the function _get_streams returns on line 76 (after "if not channel_id:")

back-to:
@a0bie @stealthplayer 

this works for cod, it should work for overwatch too.

https://gist.github.com/back-to/3dd388d4cb5282ea5cc1175884e4533f

a0bie:
@back-to 
Recompiling with modified plugin seemed to work great! Thanks.

PR #1372, opened 2017-12-11
[mlgtv] Fixed broken Plugin

back-to:
Fixed streamlink/streamlink#1362


gravyboat:
Good stuff, thanks @back-to!

PR #1373, opened 2017-12-12
Updates and improvements to tigerdile.com plugin

zootboy:
I noticed that this plugin would fail if you missed the trailing slash. Since it works with or without it, I just make the final slash optional in the regex.

### DIFF
--- a/src/streamlink/plugins/tigerdile.py
+++ b/src/streamlink/plugins/tigerdile.py
@@ -10,7 +10,7 @@ STREAM_TYPES = ["rtmp"]
 
 _url_re = re.compile(r"""
     https?://(?:www|sfw)\.tigerdile\.com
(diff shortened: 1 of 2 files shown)


zootboy:
Example to test:

`streamlink https://www.tigerdile.com/stream/overlordfox`
> error: No plugin can handle URL: https://www.tigerdile.com/stream/overlordfox

`streamlink https://www.tigerdile.com/stream/overlordfox/`
> [cli][info] Found matching plugin tigerdile for URL https://www.tigerdile.com/stream/overlordfox/
> Available streams: live (worst, best)

zootboy:
So I ended up doing a little more to this plugin. Now this pull request adds HLS support and proper API polling to determine if the stream is online. I also added a few URL tests to the build.

gravyboat:
Thanks @zootboy!

Issue #1323, opened 2017-11-08
Add python and possibly OS version to --version argument


Porkepix:
### Checklist

- [x] This is a feature request.
- [x] I have read the contribution guidelines.

### Description

Currently, `streamlink --version` only shows streamlink's version.
As for other programs, to make debugging/bug report easier and quicker, the command could also show the python's version it's running on, and possibly if not too complicated to figure out, the OS and OS version it's running on.

### Expected / Actual behavior

Have `streamlink --version` printing python and OS versions

### Reproduction steps / Explicit stream URLs to test

Just by using `streamlink --version`

### Environment details

Operating system and version: macOS 10.9
Streamlink and Python version: streamlink 0.8.1, probably running through system or homebrew's python (streamlink installed with homebrew).

...

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

beardypig:
Nice idea, I like it 👍

beardypig:
How about something like this
```
streamlink 0.8.1 (Python 3.6.1; Linux-4.4.0-97-generic-x86_64-with-debian-stretch-sid)
streamlink 0.8.1 (Python 3.6.0; macOS 10.12.6)
streamlink-script.py 0.8.1 (Python 3.5.2; Windows 10)
```

Porkepix:
Looks good to me for Windows and macOS.
For Linux I think that not all the distributions provide their name in the kernel, and some users could use custom kernels. So, maybe is there some other way to access the information through Python libs?
Some distributions also have one file like `/etc/lsb_release`, `/etc/debian_version` and so on.

I found https://docs.python.org/3/library/platform.html#unix-platforms that seems to do the job with `platform.linux_distribution()`.

It seems you already came up with a patch so I'll let you with this :)

Porkepix:
Oh, and I think that issue template could be updated after this.
Plus if there's common contributors between here and twitch GUI, adding something copyable and pastable there could be nice too. Otherwise I can open an issue there if you want.

bastimeyer:
> How about something like this
> 
> streamlink 0.8.1 (Python 3.6.1; Linux-4.4.0-97-generic-x86_64-with-debian-stretch-sid)
> streamlink 0.8.1 (Python 3.6.0; macOS 10.12.6)
> streamlink-script.py 0.8.1 (Python 3.5.2; Windows 10)

[...]

PR #1374, opened 2017-12-13
[cli-debug] Show current installed versions with -l debug

back-to:
Maybe not exactly what was requested,
but this won't break any third party applications

_Example:_

```
[cli][debug] OS:         Linux-4.14.4-1-ARCH-x86_64-with-arch
[cli][debug] Python:     3.6.3
[cli][debug] Streamlink: 0.9.0
[cli][debug] Requests(2.18.4), Socks(1.6.7), Websocket(0.44.0)
```
Fixed https://github.com/streamlink/streamlink/issues/1323

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -1,5 +1,6 @@
 import errno
 import os
+import platform
 import requests
 import sys
 import signal
@@ -9,8 +10,11 @@ from contextlib import closing
 from distutils.version import StrictVersion
 from functools import partial
 from itertools import chain
+from socks import __version__ as socks_version
 from time import sleep
+from websocket import __version__ as websocket_version
 
+from streamlink import __version__ as streamlink_version
 from streamlink import (Streamlink, StreamError, PluginError,
                         NoPluginError)
 from streamlink.cache import Cache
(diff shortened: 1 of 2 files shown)


gravyboat:
@bastimeyer, @beardypig opinions?

bastimeyer:
Just thought about the version string again. I'm not sure if this belongs into this PR, but I also didn't want to open a new issue...

The Streamlink version that's being returned by `--loglevel debug` and `--version` is always the version string of the latest release. It'll never show the commit id if Streamlink has been built from git or if it's a nightly build on Windows. I know, this has always been the case, but since we're trying to improve the issue reports here, I think it would be nice to use a better version string.

Btw, the version constant is still being exported (and [imported](https://github.com/streamlink/streamlink/blob/0.9.0/src/streamlink_cli/argparser.py#L202)) as [`LIVESTREAMER_VERSION`](https://github.com/streamlink/streamlink/blob/0.9.0/src/streamlink_cli/constants.py#L3) from the [`__version__`](https://github.com/streamlink/streamlink/blob/0.9.0/src/streamlink/__init__.py#L25) constant that gets updated on every release.

back-to:
[versioneer](https://github.com/warner/python-versioneer) could be used for this.

For a new release:
`0.9.0`

For a new commit:
`0.9.1+1.geef06b3`

`TAG[+DISTANCE.gHEX[.dirty]]`
https://github.com/warner/python-versioneer#styles

but it's maybe an overkill, if there is some other easy solution.

bastimeyer:
This looks like it would solve the issue and would also make the release process a bit easier by not having to edit the version string manually (the releases are already built automatically on each git tag push to the repo, see e4fc612a3fe7f1021d0b6c494b1aa7b0140ad28b).

> it's maybe an overkill

Probably, but how would you solve this issue otherwise? The version string has to be retrieved from the git repo or from a static string generated by the official builds that don't include git information.

back-to:
i tested the _**versioneer**_ and it doesn't work with the windows installer,
because there is no **.git** folder or files from **sdist** shipped with the `.exe` file

so it's not a solution

beardypig:
I was trying the same thing @back-to, I managed to get it to work with some extra jiggery pokery.

gravyboat:
Are we done discussing this?

src/streamlink_cli/main.py
```diff
@@ -984,6 +988,26 @@ def check_root():
             console.logger.info("streamlink is running as root! Be careful!")
 
 
+def check_current_versions():
```
beardypig: maybe this method should have a different name, like log current versions? might be confusing otherwise...

bastimeyer:
Why is the build on TravisCI running since 12h?
It looks like someone tried to cancel it (all build jobs are cancelled), but it somehow didn't work.
https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts

back-to:
travis-ci had some issues.

https://www.traviscistatus.com/incidents/c1s4dlyxd4lf

gravyboat:
@back-to I restarted the builds associated with that job.

cdrage:
Code LGTM and Tests pass :+1:

PR #1375, opened 2017-12-14
New option to limit the number of fetch retries

zootboy:
Sometimes I want to launch streamlink before a stream has gone live. The --retry-streams argument lets me loop until the stream starts, but there was no way to tell it to give up after a while. I've added a new argument, --retry-fetch-max, to make streamlink only try N fetch retries before giving up.

I also reworded the documentation for --retry-streams and --retry-open, since I had gotten confused and thought that --retry-open did what I've now done with --retry-fetch-max.

Suggestions / criticisms welcome. I was thinking about implementing an exponential backoff, but I figured I'd keep it simple for now.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -541,8 +541,23 @@ stream.add_argument(
     metavar="DELAY",
     type=num(float, min=0),
     help="""
-    Will retry fetching streams until streams are found while
-    waiting DELAY (seconds) between each attempt.
+    Retry fetching the list of available streams until streams are found
+    while waiting DELAY second(s) between each attempt. If unset, only one
+    attempt will be made to fetch the list of streams available.
+
+    The number of fetch retry attempts can be capped with --retry-max.
+    """
+)
+stream.add_argument(
+    "--retry-max",
+    metavar="COUNT",
+    type=num(int, min=-1),
+    help="""
+    When using --retry-streams, stop retrying the fetch after COUNT retry
+    attempt(s). Fetch will retry infinitely if COUNT is zero or unset.
+
+    If --retry-max is set without setting --retry-streams, the delay between
+    retries will default to 1 second.
     """
 )
 stream.add_argument(
@@ -551,7 +566,8 @@ stream.add_argument(
     type=num(int, min=0),
     default=1,
     help="""
-    Will try ATTEMPTS times to open the stream until giving up.
+    After a successful fetch, try ATTEMPTS time(s)
+    to open the stream until giving up.
 
     Default is 1.
     """
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -402,8 +402,9 @@ def fetch_streams(plugin):
                               sorting_excludes=args.stream_sorting_excludes)
 
 
-def fetch_streams_infinite(plugin, interval):
-    """Attempts to fetch streams until some are returned."""
+def fetch_streams_with_retry(plugin, interval, count):
+    """Attempts to fetch streams repeatedly
+       until some are returned or limit hit."""
 
     try:
         streams = fetch_streams(plugin)
@@ -414,6 +415,8 @@ def fetch_streams_infinite(plugin, interval):
     if not streams:
         console.logger.info("Waiting for streams, retrying every {0} "
                             "second(s)", interval)
+    attempts = 0
+
     while not streams:
         sleep(interval)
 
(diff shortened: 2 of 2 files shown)


gravyboat:
@beardypig or @bastimeyer any concerns? I think this is a good addition to have and I'm fine with the implementation. My only concern is removing the infinite command in the event people are relying on that to work.

bastimeyer:
> My only concern is removing the infinite command in the event people are relying on that to work.

Yeah, the default value should be set to infinite, in order to not break anything.

Another concern from my side is the name of the parameter. Since the other related params don't have the `-max` suffix in their names, I think this one shouldn't have it either.

Apart from that I think that's a good addition / bugfix / docs improvement. Thanks!

zootboy:
I'll change the default behavior if unset back to infinite, that makes sense.

As for the names, I was trying to come up with a more intuitive name to prevent the same confusion I had with retry-streams & retry-open. Any suggestions for a better name?

bastimeyer:
> to prevent the same confusion

True. The `-max` suffix makes it perfectly clear what the param does. Unfortunately, this makes the naming scheme inconsistent.

> Any suggestions for a better name?

Would it make sense adding the `-max` suffix to the other params as an alternative, so we can have a unified naming scheme?

zootboy:
I could change --retry-open to --retry-open-max and make its behavior consistent with --retry-fetch-max (and keep the old --retry-open option available but undocumented?). --retry-streams is both an enable switch and a delay input, so it wouldn't make sense to add -max to it. It would be nice if its name telegraphed that behavior better, but I'm struggling to come up with something concise.

gravyboat:
What do we have left to discuss on this? Any further input @bastimeyer?

gravyboat:
I'd like to get this one either merged or closed if we can.

zootboy:
Let me know what (if any) changes you'd like made, I'm happy to do them.

gravyboat:
@zootboy I think at this point we're just waiting for @bastimeyer to weigh in regarding your questions and notes from this comment: https://github.com/streamlink/streamlink/pull/1375#issuecomment-354372108

bastimeyer:
I didn't have time this week and couldn't take a look at this and respond sooner. I also somehow missed the earlier comments. Sorry about that...

So the default behavior is indeed correct and it loops infinitely if the parameter is unset. Regarding the naming scheme, I think the parameter should just be called `retry-max` instead of `retry-fetch-max`, because it is shorter and only contains one dash. Renaming the other parameters shouldn't be done, as it is unnecessary and would also require a major version bump which we can't do now.

Then we would have `retry-streams`, which is responsible for fetching the streams multiple times as a "boolean" param with `delay` seconds in between, `retry-max` to limit the `number` of fetches and `retry-open` for setting the `number` of attempts for the selected stream to be opened.

Additionally, I think that `retry-max` should set `retry-streams` to 1, if it is not set. Does that make sense?

It would also be cool to have tests for these three params. They don't necessarily have to be added in this PR, though.

zootboy:
I altered my commit to use `retry-max` as the option name. 

> Additionally, I think that retry-max should set retry-streams to 1, if it is not set. Does that make sense?

It already does that. If unset, the `retry-streams` behavior defaults to the old single attempt behavior.

As for tests, I'd be happy to write them, but I'm not very familiar with the testing system of streamlink. If anyone could point out a similar test that I could crib from, that would be fantastic.

bastimeyer:
Don't fotget to change this line as well:
https://github.com/streamlink/streamlink/pull/1375/commits/8d53f8b2ef6644300a7a4753591a9bbd3b357d85#diff-35aaec566eb720d1ce55460c85bf4ce9R493

> If unset, the retry-streams behavior defaults to the old single attempt behavior

I meant that when `retry-max` is set, but `retry-streams` is not, `retry-streams` should be set to 1, because otherwise, `retry-max` doesn't do anything, as `fetch_streams_with_retry` doesn't get called. This way, you don't need to set both parameters at the same time, if you're okay with a one second delay.

> I'm not very familiar with the testing system of streamlink

Me neither, unfortunately :disappointed:. Maybe this is a little bit of help:
https://github.com/streamlink/streamlink/blob/master/tests/test_cli_main.py

- **fetch_streams_with_retry** should be relatively simple to mock (see `fetch_streams`, `console.logger.error`, `sleep`).
- **handle_url** is a bit more complex...

But as I've said, you don't necessarily have to add this in this PR if you don't feel comfortable enough with this. The tests should have been there beforehand, which is a little bit of a shame.

zootboy:
OK, I expanded the logic to correctly handle all four permutations of setting/not setting `retry-max` and `retry-streams`. I also reworded the docs for both to explain the behavior more clearly.

I'll try to set aside some time to dig into the testing framework in the future, but I have manually tested this code for the moment.

gravyboat:
I'm good with this.

zootboy:
OK, rebased to current HEAD. Thanks for your help with this, @bastimeyer and @gravyboat

gravyboat:
It was pretty much all @bastimeyer. Thanks for taking the time to get this fixed up @zootboy!

Issue #1376, opened 2017-12-15
TV3Play has changed domain to Viafree


eublem:
- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

TV3Play has changed their domain and probably API, so the plugin doesn't work.

(VIASAT Plugin)

karlo2105:
If you want developers fix you should provide further details in template.

eublem:
Viafree has quit adding the ID into the URL, making the plugin unable to find the correct video.

So a temporary fix is to check the thumbnail picture URL and copy its ID (6 integers), which is the same as the video. And add the video ID at the end.

EXAMPLE
**Original URL:**
https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/episode-1
**Thumbnail:**
https://playapi.mtgx.tv/imagecache/240x135-retina,lowquality/cloud/content-images/seasons/16372/**822763**/5cda7a5119f8919363666fc3dd097c64-llSqK5R.jpg
**Working Streamlink URL's** (doesn't actually work in the browser)**:**
https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/episode-1/822763
or
https://www.viafree.no/programmer/822763

back-to:
duplicat of https://github.com/streamlink/streamlink/issues/727

https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/822763
is the working url in streamlink and browser

Issue #727, opened 2017-03-21
Some links on viafree.no (viasat plugin) not working (updated with correct error message)


kmyhre:
### Checklist

- [x] This is a bug report.
- [ ] This is a plugin request.
- [ ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description

Some links on Viafree does work, other gives an error.

### Expected / Actual behavior

Expected behavior:
```
user$ streamlink https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/episode-5 best
[cli][info] Found matching plugin viasat for URL https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/episode-5
[cli][info] Available streams: 48k (worst), 180p, 270p, 432p, 576p, 720p (best)
[cli][info] Opening stream: 720p (hls)
[cli][info] Starting player: /Applications/VLC.app/Contents/MacOS/VLC
```

Actual behavior:
```
user$ streamlink https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/episode-5 best
error: No plugin can handle URL: https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/episode-5
```

### Stream URLs to test

URLs:
1. https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/episode-5 (not working)
2. https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/817236 (working)

gravyboat:
Can you please run the request with the `-v` and `-l  debug` options? Then we'll get more output to see if there's something missing.

kmyhre:
Sure, but it doesn't seem to give any more information:

```
user$ streamlink -v https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/episode-1 best
error: No plugin can handle URL: https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/episode-1

```
```
user$ streamlink -l debug https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/episode-1 best
error: No plugin can handle URL: https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/episode-1

```

kmyhre:
While the working links looks like this:
```
user$ streamlink -l debug https://www.viafree.no/programmer/underholdning/paradise-hotel/sesong-9/816551 best
[...]

PR #1383, opened 2017-12-19
[viasat] Added support for urls without a stream_id

back-to:
- removed dead domains from _url_re
- added a error message for geo blocking
- new regex for stream_id from image url
- Removed old embed plugin
- try to find an iframe if no stream_id was found.
- added tests

Fixed https://github.com/streamlink/streamlink/issues/727
Fixed https://github.com/streamlink/streamlink/issues/1376


gravyboat:
Looks good, thanks @back-to.

PR #1395, opened 2017-12-28
docs: remove flattr-badge.png image

amurzeau:
It was used in README.rst before, but is no longer used since commit
b4470f3e7cc087994a1b574b040e31d577d7f15f.

Also remove _static directory and its reference from docs/conf.py as it
only contains that image.


gravyboat:
Good catch thanks @amurzeau!

amurzeau:
Don't thank me, it was found by Paul Wise when reviewing the package :)
That image had an unclear license and so was to be removed from source .tar.gz before entering Debian.

Have a nice day

PR #1403, opened 2018-01-02
build: use pynsist<2 to build the windows installer

### LINKED ISSUE #1340: Install directory for Windows setup have changed between 0.8.1 and 0.9
### Description

On Windows, default installation path have changed between 0.8..1 and 0.9.
This may be wished, but I searched into changelog/release note and commits and couldn't find anything related.
Version 0.8.1 setup tryto install into `C:\Program Files (x86)\Streamlink` by default.
Version 0.9 setup tryto install into `C:\Users\%USER%\AppData\Local\Streamlink` by default

Installation path should be the same or change should have been visible somewhere.

Just run both setups and take a look at the installation path

### Environment details

Tested on Windows 10, might be able to test W7 in the coming days if needed

This might also come from choices at the setup build step, but maybe this is not in the repository.

beardypig:
There is currently no way to change the install location from the
per-user location to the global location when using pynist>=2. pynist
1.x uses the global install location.

fixes #1340

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,5 +4,5 @@ pytest-cov
 codecov
 coverage
 mock
-pynsist
+pynsist<2
 unittest2; python_version < '2.7'


bastimeyer:
I think I've found a better solution... Please don't merge this yet. I'll post it in the other thread once I've tested it

beardypig:
Superseded by #1407

Issue #1264, opened 2017-10-09
Cryptic error with youtube (related to audio?)


vadosnaprimer:
For [this channel](https://www.youtube.com/channel/UCfVFSjHQ57zyxajhhRc7i0g) streamlink fails to fetch anything for me. Other channels work fine. This is on latest release.

```
>streamlink https://www.youtube.com/watch?v=S1pnO942Sdk
[cli][info] Found matching plugin youtube for URL https://www.youtube.com/watch?v=S1pnO942Sdk
Traceback (most recent call last):
  File "C:\Program Files (x86)\Streamlink\bin\streamlink-script.py", line 15, in <module>
    main()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 1045, in main
    handle_url()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 482, in handle_url
    streams = fetch_streams(plugin)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 394, in fetch_streams
    sorting_excludes=args.stream_sorting_excludes)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\plugin\plugin.py", line 385, in get_streams
    return self.streams(*args, **kwargs)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\plugin\plugin.py", line 288, in streams
    ostreams = self._get_streams()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\plugins\youtube.py", line 259, in _get_streams
    if best_audio_itag is None or self.adp_audio[itag] > self.adp_audio[best_audio_itag]:
KeyError: 256
```

PR #1414, opened 2018-01-05
[youtube] added Audio m4a itag 256 and 258

back-to:
Fixed streamlink/streamlink#1264

Test url

https://www.youtube.com/watch?v=aqz-KE-bpKQ


gravyboat:
Good fix, thanks.

beardypig:
Could probably add some warnings in here. Like test to see if the itag is in the lookup dict, skip if not and print a warning :)

PR #1415, opened 2018-01-05
Implement PKCS#7 padding decoding with AES-128 HLS

amurzeau:
According to the the HLS RFC [0], AES-128 streams use PKCS#7 padding.
This commit implement PKCS#7 padding decoding.

Although streamlink wasn't implementing this padding, MPEG TS streams
are working because they use a synchronization byte so the non-removed
padding was ignored by video players.

This require to have the full segment buffer instead of iterating the
result in chunks, but requests was still downloading the full segment in
memory as stream=True is not set. See [1].

----
I found that by using this test:
Execute this gist: https://gist.github.com/amurzeau/78e9f9ce958a562e86bba46b18ea8a7f
This script does:
- Create a fake stream
- Encrypt it
- Create HLS playlists for that stream
- Use streamlink on it and output the result in a file
- Compare the streamlink output with concat unencrypted stream files and expect them to be equal

Tested with the encrypted HLS AES-128 stream at address [2].

[0] https://tools.ietf.org/html/rfc8216#section-4.3.2.4
[1] http://docs.python-requests.org/en/master/api/#requests.Response.iter_content
[2] hls://https://cdn.theoplayer.com/video/big_buck_bunny_encrypted/stream-800/index.m3u8

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -17,6 +17,17 @@ Sequence = namedtuple("Sequence", "num segment")
 def num_to_iv(n):
     return struct.pack(">8xq", n)
 
+def pkcs7_decode(paddedData, keySize = 16):
+        '''
+        Remove the PKCS#7 padding
+        '''
+        # Use ord + [-1:] to support both python 2 and 3
+        val = ord(paddedData[-1:])
+        if val > keySize:
+            raise StreamError("Input is not padded or padding is corrupt, got padding size of {0}".format(val))
+
+        return paddedData[:-val]
+
 
 class HLSStreamWriter(SegmentedStreamWriter):
     def __init__(self, reader, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


amurzeau:
Currently, the integration test is not included in the code coverage report as streamlink  is run as a black box command line tool.
I'm trying to fix that so the code coverage report can include this test.

amurzeau:
Done on coverage fix, I made this:
- Detect whether the test run under coverage
- If yes, run streamlink using load_entry_point
- If no, run streamlink using subprocess.open (so this also check streamlink as a black box command line tool)

Coverage include now this test but codecov.io ignore it (maybe it doesn't like amended commits).
The true coverage change can be seen here (+4.73%): https://codecov.io/gh/streamlink/streamlink/compare/e7d849a919e8764ad2f2bf769b880b3083fff555...574bbddb4e309cbc66eb918c32eaed3358f53707/list/src

amurzeau:
This kind of test need streamlink to be installed beforehand which is not always the case.
I will change that test to a simpler one focused on PKCS#7.

beardypig:
Nice addition, thanks. When you sort out the tests we can merge it in :)

amurzeau:
Thanks :)

I'm going to add requests-mock as tests dependency.
Is it sufficient to add it in dev-requirements.txt ?

amurzeau:
Ready to merge now 👍 , +3.74% of test coverage while testing hls and segmented modules :-)

https://codecov.io/gh/streamlink/streamlink/compare/7f1492595613442869ea1d2a5e519b5a34f687f1...943539797ad11237e53b67c9b368c2d9512768d0/list/src

gravyboat:
Looks great to me, thanks @amurzeau! As soon as @beardypig gives final sign off we'll merge it in.

src/streamlink/stream/hls.py
```diff
@@ -17,6 +17,17 @@
 def num_to_iv(n):
     return struct.pack(">8xq", n)
 
+def pkcs7_decode(paddedData, keySize = 16):
+        '''
+        Remove the PKCS#7 padding
+        '''
+        # Use ord + [-1:] to support both python 2 and 3
+        val = ord(paddedData[-1:])
```
beardypig: 👍

Issue #1337, opened 2017-11-20
Mediaset (mitele.es) not work propperly


vk496:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description

http://www.mitele.es not get the stream propperly

### Expected / Actual behavior

Expected: Get the stream link
...
Actual: `error: Unable to open URL: https://mdslivegeo.mediaset.es/out/u/3vkkbgnvsm2r5/104951/NmbWM5YjE6LAH8o-VL3vr3hw3JqSTnRU/cn/fb3d6852c1ac44c982aa727fe9e791cd.m3u8 (403 Client Error: Forbidden for url: https://mdslivegeo.mediaset.es/out/u/3vkkbgnvsm2r5/104951/NmbWM5YjE6LAH8o-VL3vr3hw3JqSTnRU/cn/fb3d6852c1ac44c982aa727fe9e791cd.m3u8)`
### Reproduction steps / Explicit stream URLs to test

`streamlink http://www.mitele.es/directo/boing best -l debug`

### Environment details

Operating system and version: Manjaro XFCE 17.1
Streamlink and Python version: streamlink 0.9.0 / Python 3.6.3

### Comments, logs, screenshots, etc.

```bash
$ streamlink http://www.mitele.es/directo/boing worst -l debug
[cli][info] Found matching plugin mitele for URL http://www.mitele.es/directo/boing
[plugin.mitele][debug] Using cached sessionKey
[plugin.mitele][debug] Got api key: tsbmYyOjRH6dK3rmGxeTP2uleUj_.VG0Ka
[plugin.mitele][debug] Got channel ID ZvOXRpODE6c__EQ05epWGBX6530x-EhS for channel boing
error: Unable to open URL: https://mdslivegeo.mediaset.es/out/u/3vkkbgnvsm2r5/104951/ZvOXRpODE6c__EQ05epWGBX6530x-EhS/cn/ddf752ff80af43709eabdf84e1d65366.m3u8 (403 Client Error: Forbidden for url: https://mdslivegeo.mediaset.es/out/u/3vkkbgnvsm2r5/104951/ZvOXRpODE6c__EQ05epWGBX6530x-EhS/cn/ddf752ff80af43709eabdf84e1d65366.m3u8)
```

beardypig:
Are you streaming from Spain? Have you tried any other channels other than Boing?

[...]

PR #1416, opened 2018-01-06
[mitele] Update for different api response

back-to:
- fallback if not hls_url was found, just the suffix.
- added url tests
- added all channels

Fixed streamlink/streamlink#1337


gravyboat:
Thanks @back-to!

PR #1420, opened 2018-01-08
[orf_tvthek] Work around broken HTTP connection persistence

sqrt2:
The server hosting the HLS playlists for tvthek.orf.at does not respect HTTP connection persistence. It closes the connection after a second request is made instead of responding to the request, making streamlink crash with an error while stream sources are collected.

This pull request adds a function to swap out an existing HTTP session for a new one to the Streamlink session object, and calls this function in the appropriate place to have stream enumeration succeed in the orf_tvthek plugin.

### DIFF
--- a/src/streamlink/plugins/orf_tvthek.py
+++ b/src/streamlink/plugins/orf_tvthek.py
@@ -47,7 +47,8 @@ class ORFTVThek(Plugin):
             except KeyError:
                 continue
             stream = HLSStream.parse_variant_playlist(self.session, url)
-            self.session.new_http_session()
+            # work around broken HTTP connection persistence by acquiring a new connection
+            http.close()
             streams.update(stream)
 
         return streams
--- a/src/streamlink/plugins/orf_tvthek.py
+++ b/src/streamlink/plugins/orf_tvthek.py
@@ -47,6 +47,7 @@ class ORFTVThek(Plugin):
             except KeyError:
                 continue
             stream = HLSStream.parse_variant_playlist(self.session, url)
+            self.session.new_http_session()
             streams.update(stream)
 
         return streams
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -454,9 +454,6 @@ class Streamlink(object):
         if file:
             file.close()
 
-    def new_http_session(self):
-        self.http.close()
-
     @property
     def version(self):
         return __version__
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -454,6 +454,9 @@ class Streamlink(object):
         if file:
             file.close()
 
+    def new_http_session(self):
+        self.http.close()
+
     @property
     def version(self):
         return __version__


src/streamlink/plugins/orf_tvthek.py
```diff
@@ -47,6 +47,7 @@ def _get_streams(self):
             except KeyError:
                 continue
             stream = HLSStream.parse_variant_playlist(self.session, url)
+            self.session.new_http_session()
```
back-to: why not just `self.session.http.close()` or `http.close()`

---

do you have an example url where the video will only work with this?
or are this the geo-blocked videos

src/streamlink/session.py
```diff
@@ -454,6 +454,9 @@ def load_plugin(self, name, file, pathname, desc):
         if file:
             file.close()
 
+    def new_http_session(self):
```
back-to: if needed it should be in `src/streamlink/plugins/orf_tvthek.py`

beardypig:
Thanks for your contribution, this project is kept alive by contributions by people like yourself. 
I agree with @back-to, it seems like a special case that can better handled in the plugin. I don't think there needs to be an additional method, just use `http.close()`, it's cleaner and easier for people to understand and follow :)

sqrt2:
The effect can be observed with any currently live stream linked from http://tvthek.orf.at/live (featuring a red dot). However, as frequently content is streamed that I doubt is licensed internationally, there might very well be geoblocking in place.

Introducing a new function to reset HTTP session state seemed like a good idea to me for several reasons:

1. The internal state of one of the Streamlink object's member variables is modified. In the spirit of data encapsulation, it should be the Streamlink object that does this (managing the HTTP session state ought to be the Streamlink object's task; its module name is even streamlink.session).
2. streamlink uses a custom subclass of requests.Session. Should in the future some state be added to this class that needs to be reinitialised when the HTTP session is, the new_http_session() function is easier to find than an equivalent instruction in a plugin.
3. Resetting HTTP session state may be useful to other plugins, too -- its use is not restricted to misbehaving HTTP servers.

If you insist, I can of course directly call close() in the plugin. Would a new pull request be the right way to go about this?

karlo2105:
@sqrt2 
I always wondered is there any page on ORF TVTHEK available where you can catch all channels which don't depend on tv broadcast aired currently on TV ? 
Thanks for your contribution.

back-to:
it should be in `src/streamlink/plugins/orf_tvthek.py`
because it won't be used that often for other plugins and it's just one line.

you can just add the changes to this pr, it can get squashed to one commit at the end.

---

_some debug_

**ORF 2** _geo-blocked_
not sure if this changes without geo-block

```
error: Unable to open URL: http://apasfiisl.apa.at/ipad/orf2_qxb/orf.sdp/playlist.m3u8
(('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',)))
```

**ORF 2** _geo-blocked_ with `http.close()`
no data to play
```
[cli][info] Found matching plugin orf_tvthek for URL http://tvthek.orf.at/live/Sturm-der-Liebe-Teil-2843/13946156
[cli][info] Available streams: 360p (worst, best)
[cli][info] Opening stream: 360p (hls)
[stream.hls][debug] First Sequence: 0; Last Sequence: 0
[stream.hls][debug] Start offset: 0; Duration: None; Start Sequence: 0; End Sequence: 0
[stream.hls][debug] Download of segment 0 complete
^C[stream.hls][debug] Closing worker thread
[stream.hls][debug] Closing writer thread
[cli][info] Stream ended
Interrupted! Exiting...
```


**ORF 3** _not geo-blocked_
```
$ streamlink http://tvthek.orf.at/live/Expeditionen-Der-Mehlspeis-Adel/13946158 best
[cli][info] Found matching plugin orf_tvthek for URL http://tvthek.orf.at/live/Expeditionen-Der-Mehlspeis-Adel/13946158
[cli][info] Available streams: 180p (worst), 360p, 540p, 720p (best)
[cli][info] Player closed
[cli][info] Stream ended
```

sqrt2:
I have updated the PR.

@back-to The first output is the error that is fixed by this PR, not necessarily indicative of geoblocking.

@karlo2105 Do you mean VODs? You can find these at http://tvthek.orf.at/schedule and http://tvthek.orf.at/profiles. streamlink can play these, but unfortunately each segment of each program is its own VOD, so it is not the most comfortable way of viewing.

karlo2105:
@sqrt2 
No I meant unique Id for live channels. I guess it doesn't exist for Orf.

PR #1423, opened 2018-01-09
[youtube] New params for get_video_info

back-to:
- removed double import
- use https
- **detailpage** for normal videos
- **embedded** as a fallback for age-restricted videos

Fixed streamlink/streamlink#1421


src/streamlink/plugins/youtube.py
```diff
@@ -205,11 +204,16 @@ def _get_stream_info(self, url):
...
+        for el in ("detailpage", "embedded"):
+            params = {
+                "video_id": video_id,
+                "el": el
+            }
+            res = http.get(API_VIDEO_INFO, params=params, headers=HLS_HEADERS)
+            if "status=fail" in res.text:
+                continue
+            break
+
```
beardypig: I know there is no error checking at the moment for the fail status, maybe now is a good time to add a logging message if there is a fail on both? The current feedback for the youtube plugin is pretty minimal :)

src/streamlink/plugins/youtube.py
```diff
@@ -205,11 +204,21 @@ def _get_stream_info(self, url):
...
-            "el": "player_embedded"
-        }
-        res = http.get(API_VIDEO_INFO, params=params, headers=HLS_HEADERS)
+        for el in ("detailpage", "embedded"):
+            params = {
+                "video_id": video_id,
+                "el": el
+            }
+            res = http.get(API_VIDEO_INFO, params=params, headers=HLS_HEADERS)
+            info_parsed = parse_query(res.text)
```
back-to: `parse_query` is now called twice, here and

`return parse_query(res.text, name="config", schema=_config_schema)`

i don't know if i can pass `info_parsed` with `name="config", schema=_config_schema` somehow

back-to:
ok, `-l debug` will now print error messages for `_get_stream_info`

Kagami:
Any reasons not to merge this?

Twilight0:
I actually merged this PR in my kodi module, imho it performs really well: https://github.com/Twilight0/script.module.streamlink.base/blob/7c0a2e47ccfb9359d41b615f5f8e5bed708a58ca/resources/lib/streamlink/plugins/youtube.py

gravyboat:
@kagami because it hasn't been reviewed again after the recent changes? Look at the discussion so far. @beardypig any other feedback?

TheSHEEEP:
The problem I see here is that in case of age restriction, "embedded" will fail, too, if the author did not explicitly allow embedding (at least in live streams, as I've tested). 
If there was a way to "fake the age", that would be optimal, but I kind of doubt that...

back-to:
@TheSHEEEP can you give me an example url with age restriction and embedded disallowed

TheSHEEEP:
@back-to Not really, but I'd suggest getting OBS Studio, start streaming to YouTube and setting your stream to be age restricted (embedded disallowed is default). That's how I usually test stuff like that.

back-to:
no interest in installing obs for an almost never used combination.

```
streamlink https://www.youtube.com/watch?v=-8hKiyG7RZ0 best -l debug
```

this will get the data, so it should work for other aswell.

Issue #1419, opened 2018-01-08
Weird issue, youtube live stream starts playing 4 hours backwards?


mmetak:
### Checklist

- [x] This is a bug report.
- [x] I have read the contribution guidelines.

### Description
Youtube live stream starts playing from 4 hours backwards into the live stream.

### Expected / Actual behavior
Play stream as it should, close as possible to real time.

### Reproduction steps / Explicit stream URLs to test

1. `streamlink -p mpv https://www.youtube.com/watch?v=2TrGSO2ZQ_Q 480p`
2. `streamlink -p mpv https://www.youtube.com/watch?v=rGNNXg8YJoU 480p`

### Environment details

Operating system and version: Archlinux, up to date
Streamlink and Python version: 0.9.0.r55.gaa45712e and python-3.6.4-1

Edit: downgraded to stable version 0.9.0 and the issue was gone.

### Comments, logs, screenshots, etc.
Weird part is that livestreamer plays these same streams just fine. ;-)
Below is a screenshot with live stream playing in the background in Opera and in front of it is mpv window opened with streamlink. Look at the time stamp. Both were started second or two apart.

![screenshot_20180108_162259a](https://user-images.githubusercontent.com/6218030/34678329-1ea4c646-f493-11e7-92ab-0fefbfbaaffd.png)

mmetak:
@beardypig this appers after this PR: Merge pull request #1303 from beardypig/hls-skip

beardypig:
I'll take a look :) (unless @back-to beats me to it ;))

back-to:
@beardypig 
it's because of
https://github.com/streamlink/streamlink/commit/3ff1fcd0fe50085e0f3b9c39f8fedb8ac319af56

but you need this for every normal vod.

---

https://github.com/streamlink/streamlink/blob/master/src/streamlink/stream/hls.py#L132-L137

`self.duration_offset_start` is **0** there for the livestream

changing it to **-1** when it's **0**, could be a workaround.

mmetak:
@back-to That fixes it. 😅

mmetak:
@gravyboat I think you merged #1424 too soon?

@back-to said this in that PR:
> streamlink/streamlink#1423 required for youtube.

back-to:
@mmetak this issue here was not youtube related,
livestreams with a playback are fixed now.

---

But **youtube** itself is broken. #1421

so it's fine.

PR #1424, opened 2018-01-09
[hls] Don't try to skip a stream if the offset is 0

back-to:
and don't raise KeyKerror if the m3u8 file is empty this allows the file to reload.

---

Fixed streamlink/streamlink#1419

streamlink/streamlink#1423 required for youtube.

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -136,15 +136,17 @@ class HLSStreamWorker(SegmentedStreamWorker):
             # live playlist, force offset durations back to None
             self.duration_offset_start = -self.duration_offset_start
 
-        self.playlist_sequence = self.duration_to_sequence(self.duration_offset_start, self.playlist_sequences)
-        if self.duration_limit and self.playlist_end:
-            self.playlist_end = self.duration_to_sequence(self.duration_offset_start + self.duration_limit,
-                                                          self.playlist_sequences)
-        self.logger.debug("First Sequence: {0}; Last Sequence: {1}",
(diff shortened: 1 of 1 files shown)


src/streamlink/stream/hls.py
```diff
@@ -136,15 +136,17 @@ def __init__(self, *args, **kwargs):
...
-        self.playlist_sequence = self.duration_to_sequence(self.duration_offset_start, self.playlist_sequences)
-        if self.duration_limit and self.playlist_end:
-            self.playlist_end = self.duration_to_sequence(self.duration_offset_start + self.duration_limit,
-                                                          self.playlist_sequences)
-        self.logger.debug("First Sequence: {0}; Last Sequence: {1}",
-                          self.playlist_sequences[0].num, self.playlist_sequences[-1].num)
-        self.logger.debug("Start offset: {0}; Duration: {1}; Start Sequence: {2}; End Sequence: {3}",
-                          self.duration_offset_start, self.duration_limit, self.playlist_sequence,
-                          self.playlist_end)
+        if self.duration_offset_start is not 0:
```
beardypig: While `0` is a singleton in CPython, I think that `!=` should be used in this case :)

beardypig:
I was just working on both these things ;)

src/streamlink/stream/hls.py
back-to: i changed it 👍

gravyboat:
Thanks @back-to and thanks for reviewing @beardypig.

PR #1427, opened 2018-01-12
PEP8

hicrop:


back-to:
looks good,
but it might make other PR's incompatible or this PR will be incompatible

so some rebase will be required, depends on what get merged first.

beardypig:
All sensible PEP8 fixes :)

gravyboat:
Le's merge some of the other PRs first since they have more critical changes and this should be easier to rebase instead of getting a bunch of different users to do so.

gravyboat:
@hicrop Let me try to get what we can merged tonight then you shouldn't have to rebase multiple times.

Issue #687, opened 2017-03-09
Sinclair Broadcast Channels (Charge!, Comet, and TBD).


RedPenguin2:
- [x] This is a bug report.
### Description
Sinclair Broadcast Group has 3 separate channels that are free to live stream online.

They use an m3u8 file and show the stream as expected but where the problem seems to lie is in commercial breaks.

They will show the TV shows as usual online but then jump to a screen saying "We're Be Right Back!" which some users of devices like the Fire TV said it causes their players to really use CPU.

### Expected / Actual behavior

I expect to have an mp4 outputted that I can watch or edit in any software just like any other stream I've ever recorded with streamlink.

### Reproduction steps / Stream URLs to test

1. ... streamlink -f --hls-segment-threads 10 --hls-segment-attempts 25 "hlsvariant://http://content.uplynk.com/channel/c1c931fb16184c9391bd0b0c91f30354.m3u8?tc=1&exp=1614631202&rn=1318676852&ct=c&cid=c1c931fb16184c9391bd0b0c91f30354&ad=charge&ad.adUnit=%2FCHARGE%2FLivestream_midroll&ad._debug=CHARGE_ssai_livestream_midroll&ad.ad_rule=1&ad.pmad=12&ad.output=xml_vmap1&sig=28e5dcf3f704effb0e8c5dea5f718bb00d79abc2d2c7b8d2552b1fd94eddaecd" best

### Environment details (operating system, python version, etc.)

Windows 10 Home

### Comments, logs, screenshots, etc.

streamlink appears normal when recording, counting the size and time.

Yet when you look at the file afterwards, it appears to be the correct size based on 720P resolution but Kodi, ffmpeg, VLC, and other software have an extremely tough time seeking it and today's file the players think is only a few seconds long even though it's at least 4 hours long.

A repair tool I have just completely crashed while trying to repair the mp4.

RedPenguin2:
It seems while ffmpeg failed to fix the output file it appears that avconv is able to fix it.

Though I wonder if there is a way to have it record a file that isn't already a convoluted mess that plays right off the bat.

I know it has to be something on Sinclair or Uplynk's end as no other stream I've used through streamlink has ever done this.

beardypig:
What's the site with the actual player on it? Might be easier to make a plugin or atleast understand the context.

beardypig:
[...]

PR #1432, opened 2018-01-13
[hls] New option --hls-segment-ignore-names

back-to:
You can **ignore segment** names that might **corrupt** or **lag** your output

- this will freeze the player until there is a valid segment name
- this will not add the segment to an output file,
  so it won't be damaged and can be edited.

Note: The `--hls-timeout` must be increased, to a time that is longer
than the ignored break.

Example:
`streamlink URL best --hls-segment-ignore-names 000,001,002
--hls-timeout 1200`

This will ignore every segment that ends with 000.ts, 001.ts and 002.ts

---

some names/text can be changed, if someone has a better idea.

---

Hey @RedPenguin2

it would be nice if you could test this,
if you don't know how to, i can help you.

`streamlink http://www.tbd.com best -l debug --hls-segment-ignore-names 0000,0001,0002 -o test.ts --hls-timeout 2400`
change the url with your hls link

this PR would fix streamlink/streamlink#687

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -1,3 +1,4 @@
+import re
 import struct
 from collections import defaultdict, namedtuple
 
@@ -35,11 +36,19 @@ class HLSStreamWriter(SegmentedStreamWriter):
         kwargs["retries"] = options.get("hls-segment-attempts")
         kwargs["threads"] = options.get("hls-segment-threads")
         kwargs["timeout"] = options.get("hls-segment-timeout")
+        kwargs["ignore_names"] = options.get("hls-segment-ignore-names")
         SegmentedStreamWriter.__init__(self, reader, *args, **kwargs)
 
         self.byterange_offsets = defaultdict(int)
         self.key_data = None
         self.key_uri = None
+        if self.ignore_names:
+            # creates a regex from a list of segment names,
+            # this will be used to ignore segments.
+            self.ignore_names = list(set(self.ignore_names))
+            self.ignore_names = "|".join(list(map(re.escape, self.ignore_names)))
+            self.ignore_names_re = re.compile(r"(?:{blacklist})\.ts".format(
+                    blacklist=self.ignore_names),  re.IGNORECASE)
 
     def create_decryptor(self, key, sequence):
         if key.method != "AES-128":
@@ -86,6 +95,11 @@ class HLSStreamWriter(SegmentedStreamWriter):
 
         try:
             request_params = self.create_request_params(sequence)
+            # skip ignored segment names
+            if self.ignore_names and self.ignore_names_re.search(sequence.segment.uri):
+                self.logger.debug("Skipping segment {0}".format(sequence.num))
+                return
+
             return self.session.http.get(sequence.segment.uri,
                                          timeout=self.timeout,
                                          exception=StreamError,
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -709,6 +709,22 @@ transport.add_argument(
     Default is 10.0.
     """)
 transport.add_argument(
+    "--hls-segment-ignore-names",
+    metavar="NAMES",
+    type=comma_list,
+    help="""
+    A comma-delimited list of segment names that will not be fetched.
+
+    Example: --hls-segment-ignore-names 000,001,002
+
+    This will ignore every segment that ends with 000.ts, 001.ts and 002.ts
+
+    Default is None.
+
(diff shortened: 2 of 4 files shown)


RedPenguin2:
@back-to I got streamlink setup with the pull request but now I seem to have found a completely different issue now.

[stream.hls][debug] Segments in this playlist are encrypted

I don't remember getting that before at all.

back-to:
the
`[stream.hls][debug] Segments in this playlist are encrypted`
is
`#EXT-X-KEY:METHOD=AES-128,URI=...`
it's supported by streamlink

---

the `skipping segments` are the corrupt mp4 files that **won't get added**,
so you can edit the output file with a video editor, what did not work before.

there are maybe dead minutes in the output file, but you can just cut it.

---

you need to run the command longer than the Pause break,
so you will get some data to the output file.

i only tested tbd, the other channels might have different filenames

RedPenguin2:
Oh thanks for the info, going to try it again.

I really wish they really had a better commercial break method.

Watching a screen just saying "We'll Be Right Back" for 3 minutes straight with no music is even more boring than actually watching a normal commercial break. LoL

RedPenguin2:
So far it appears to work for TBD.

Originally I had an outdated URL, that just gave me a commerical break the entire time.

I noticed on the website I guess it supplements with ads.

The only thing I noticed is, I still had to do a quick ffmpeg copy audio/video to fix the header, because even though the file looks great being just the program with no commercials, it oddly wouldn't rewind correctly in ffplay at first.

amurzeau:
VLC 3.0 handle TBD stream good enough. VLC 3.0 is at RC stage as of now (downloadable in nightly builds) (see also my comment just added on the linked issue)

gravyboat:
@bastimeyer, @beardypig Any concerns over this new option?

beardypig:
Is it a work around for discontinuity issues? Seems generally useful though :)

gravyboat:
Is there any other feedback here from anyone? If not I want to get this merged since it's a good addition.

src/streamlink/stream/hls.py
```diff
@@ -35,11 +36,19 @@ def __init__(self, reader, *args, **kwargs):
...
 
         self.byterange_offsets = defaultdict(int)
         self.key_data = None
         self.key_uri = None
+        if self.ignore_names:
+            # creates a regex from a list of segment names,
+            # this will be used to ignore segments.
+            self.ignore_names = list(set(self.ignore_names))
+            self.ignore_names = "|".join(list(map(re.escape, self.ignore_names)))
+            self.ignore_names_re = re.compile(r"(?:{blacklist})\.ts".format(
```
bastimeyer: > This will ignore every segment that ends with ...

Just a minor suggestion: you should add the dollar anchor to the end of the regex, just to be sure that a segment doesn't get matched by accident if a URL somehow contains another `.ts`, you'll never know 😃

src/streamlink_cli/argparser.py
```diff
@@ -708,6 +708,22 @@ def hours_minutes_seconds(value):
...
+    help="""
+    A comma-delimited list of segment names that will not be fetched.
+
+    Example: --hls-segment-ignore-names 000,001,002
+
+    This will ignore every segment that ends with 000.ts, 001.ts and 002.ts
+
+    Default is None.
+
+    Note: The --hls-timeout must be increased, to a time that is longer than the ignored break.
```
bastimeyer: I guess this can't be automated, right? Segment lengths are dependent on the streaming service (may even be variable) and there could also be multiple segments in a row which get removed.

src/streamlink/stream/hls.py
back-to: > ends with

i search for the **end of the filename** not for the uri end,
if there are params after `00.ts?foo=bar` it's not the end of the filename

`$` won't work for this

---

if i would parse the uri and just work with the _path_, then sure
but using `urlparse` might be to advanced.

bastimeyer: True... I had static segment URLs in mind without query data.
`\.ts(?:?.*)$` should work, but we could also just leave it as it is, it's good enough. :)

src/streamlink_cli/argparser.py
back-to: i could change `timeout=self.timeout` to a bigger value
if this command is used

but that's not really automated

gravyboat:
@bastimeyer Any other feedback? If not I would like to get this merged.

bastimeyer:
No, let's merge this... Further improvements, if needed, can be added in another PR

PR #1442, opened 2018-01-20
ecanlitvizle.net changed url to ecanlitvizle.tv

Rotarum:


Rotarum:
ecanlitvizle.net is not working!

Please change .net to .tv

Why has the check failed??

bastimeyer:
> Why has the check failed??

https://travis-ci.org/streamlink/streamlink/jobs/331219757#L1642

You also have to change the tests (which try to match the plugin's URLs):
https://github.com/streamlink/streamlink/blob/0.10.0/tests/test_plugin_canlitv.py#L21

dreamboxco:
on the page e canli no picture is reproduced I think that is because of the side

beardypig:
Closing this PR as we are removing the ecanlitvizle plugin, most of the channels are covered by other plugins - if there are channels that are not covered, please open an issue to support that channel's  *official* site.

Issue #1404, opened 2018-01-02
Support for ok.ru?


Bluebuster116:
*Thanks for reporting an issue!*
*Please read the contribution guidelines first! (see the link above)*
*Also check the list of known issues before reporting an issue!*

*Feel free to use the following template. Be as detailed as possible.*
*Please see the text preview to avoid unnecessary formatting errors.*
*Don't forget to remove this text before submitting.*

----

### Checklist

- [ ] This is a bug report.
- [x] This is a feature request.
- [x] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

Is Ok.ru supported by streamlink?. I have looked in different sources but there is no information which I could find on this matter. There is this stream http://ok.ru/videoembed/412346031626 which I would like to use with streamlink but I am unable to get it work.

### Expected / Actual behavior

I have only been able to catch this "source" from inspecting the player https://vsd0.mycdn.me/dash/stream_333533874698/stream.manifest/sig/La5bYbT_yFY/expires/1515003608246/srcIp/201.240.21.221/clientType/0/srcAg/CHROME/frag-medium-9937.dash but the fragments make it not possible to use it with the existing parameters on streamlink. Any suggestions or possibly support of ok.ru on streamlink?

### Reproduction steps / Explicit stream URLs to test

1. None

### Environment details

Operating system and version:  Windows 7
Streamlink and Python version: current

### Comments, logs, screenshots, etc.

None

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

bobvargas:
Hopefully they can support urls for tv tokyo on ok.ru and the url https://www.ok.ru/live/412362939914

gravyboat:
What kind of content do they stream on this site? Based on what @bobvargas is saying it looks like the site is kind of a Russian Facebook, but people are using their streaming function to restream illegal content, and the link from @Bluebuster116 is also restreaming a television station illegally.

atykhyy:
ok.ru is indeed a sort of Russian Facebook (the other one being vk.com). Russians are using it for camshows and what-not (look at their [top page](https://ok.ru/live)) so it's no more inherently illegal than vaughnlive.tv or any other streaming network. Here is a plugin that does the job for live videos, feel free to add it or not:
```
import urllib3
import re

from streamlink.plugin import Plugin
from streamlink.plugin.api import http, validate
from streamlink.plugin.api import useragents
[...]

PR #1451, opened 2018-01-24
Add ok.ru/live plugin

atykhyy:
This PR adds a plugin for ok.ru live videos (#1404)

### DIFF
--- a/src/streamlink/plugins/ok_live.py
+++ b/src/streamlink/plugins/ok_live.py
@@ -1,0 +1,41 @@
+import urllib3
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import http, validate
+from streamlink.plugin.api import useragents
+from streamlink.stream import HLSStream
+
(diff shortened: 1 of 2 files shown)


gravyboat:
I'm not okay with adding an additional dependency (bs4) to the project for a single site, do they not have some kind of API or anything that can be used? We can check with @beardypig and @bastimeyer as well but I highly doubt they will want to add such a dependency either.

atykhyy:
Not in their [public API](https://apiok.ru/dev/methods/rest/video/), and their mobile app is obfuscated. It's possible, I suppose, to fish the HLS playlist URL out of the raw HTML with a regular expression, as many other plugins do, but HTML is structured data, why not use it as such? Navigating HTML with XPath or equivalent results in more readable code, and is less brittle to minor textual changes in the HTML than regular expressions, where e.g. a comment or swapped tag attributes can easily ruin one's day.

gravyboat:
Good stuff, thanks for working around that @atykhyy.

PR #1454, opened 2018-01-26
Remove Debian directory

amurzeau:
This directory is not used for downstream Debian package. The package is
maintained in a separate git repository.
So remove it to avoid maintenance burden here (like updating years).


gravyboat:
Good stuff, thanks for cleaning this up!

PR #1455, opened 2018-01-27
docs/install: add Debian instructions

amurzeau:
Add instructions for installation on Debian in docs.

Note: the package is only available in sid/unstable as of now and in testing in around 10 days.
It will reach stable distribution once the next one (Buster) is released.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -22,6 +22,10 @@ Distribution                         Installing
                                         # pacaur -S streamlink-git
 
                                      `Installing AUR packages`_
+`Debian (sid, testing)`_             .. code-block:: console
+
+                                        # apt update
+                                        # apt install streamlink
 `Fedora`_                            .. code-block:: console
 
                                         # dnf install streamlink
@@ -48,6 +52,7 @@ Distribution                         Installing
 
 .. _Arch Linux: https://www.archlinux.org/packages/community/any/streamlink/
 .. _Arch Linux (aur, git): https://aur.archlinux.org/packages/streamlink-git/
+.. _Debian (sid, testing): https://packages.debian.org/unstable/streamlink
 .. _Fedora: https://apps.fedoraproject.org/packages/python-streamlink
 .. _Gentoo Linux: https://packages.gentoo.org/package/net-misc/streamlink
 .. _NetBSD (pkgsrc): http://pkgsrc.se/multimedia/streamlink
@@ -96,6 +101,7 @@ Distribution/Platform                Maintainer
 Arch                                 Giancarlo Razzolini <grazzolini at archlinux.org>
 Arch (aur, git)                      Josip Ponjavic <josipponjavic at gmail.com>
 Chocolatey                           Scott Walters <me at scowalt.com>
+Debian (sid, testing)                Alexis Murzeau <amubtdx at gmail.com>
 Fedora                               Mohamed El Morabity <melmorabity at fedoraproject.org>
 Gentoo                               soredake <fdsfgs at krutt.org>
 NetBSD                               Maya Rashish <maya at netbsd.org>
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -26,6 +26,12 @@ Distribution                         Installing
 
                                         # apt update
                                         # apt install streamlink
+`Debian (stable)`_                   .. code-block:: console
+
+                                        # wget -qO- "https://bintray.com/user/downloadSubjectPublicKey?username=amurzeau" | apt-key add -
+                                        # echo "deb https://dl.bintray.com/amurzeau/streamlink-debian stretch-backports main" | tee "/etc/apt/sources.list.d/streamlink.list"
+                                        # apt update
+                                        # apt install streamlink
 `Fedora`_                            .. code-block:: console
 
                                         # dnf install streamlink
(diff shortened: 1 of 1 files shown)


bastimeyer:
> the package is only available in sid/unstable as of now and in testing in around 10 days

Wouldn't it make sense adding a small note to the docs? Or just call it `Debian (sid, testing)` instead of just `Debian` (similar to the Arch AUR package)...

Also, could you please add yourself to the package maintainer list down below?
Thanks!

amurzeau:
Updated, thanks !

amurzeau:
For debian stable, I'm pushing packages backport in bintray, https://bintray.com/amurzeau/streamlink-debian/streamlink.
I'm adding this information to the install page.

gravyboat:
This looks good to me! Any further feedback @bastimeyer?

docs/install.rst
```diff
@@ -22,6 +22,16 @@ Distribution                         Installing
...
 
                                      `Installing AUR packages`_
+`Debian (sid, testing)`_             .. code-block:: console
+
+                                        # apt update
+                                        # apt install streamlink
+`Debian (stable)`_                   .. code-block:: console
+
+                                        # wget -qO - "https://bintray.com/user/downloadSubjectPublicKey?username=amurzeau" | apt-key
+                                        # deb https://dl.bintray.com/amurzeau/streamlink-debian stretch-backports main
```
bastimeyer: I think we should be explicit here and not just tell the user the contents of the `/etc/apt/sources.list.d/streamlink.list` file which he has to create on his own.
```bash
echo "deb https://dl.bintray.com/amurzeau/streamlink-debian stretch-backports main" | sudo tee "/etc/apt/sources.list.d/streamlink.list"
```
I'm not sure though how it'll look on the rendered page. Need to check this first...

docs/install.rst
```diff
@@ -22,6 +22,16 @@ Distribution                         Installing
                                         # pacaur -S streamlink-git
 
                                      `Installing AUR packages`_
+`Debian (sid, testing)`_             .. code-block:: console
+
+                                        # apt update
+                                        # apt install streamlink
+`Debian (stable)`_                   .. code-block:: console
+
+                                        # wget -qO - "https://bintray.com/user/downloadSubjectPublicKey?username=amurzeau" | apt-key
```
bastimeyer: Isn't `apt-key` missing a few params here? Doesn't it also need root access?
```bash
wget -qO- "https://bintray.com/user/downloadSubjectPublicKey?username=amurzeau" | sudo apt-key add -
```

amurzeau: Yes you are right about params, I used what I had on bintray description but it is wrong here too.
Ok about `sudo`, I was expecting `#` to mean "run as root" as opposed to "$"

docs/install.rst
amurzeau: You are right too

bastimeyer: > I'm not sure though how it'll look on the rendered page.

I knew this was gonna happen. Long lines of text are adding a horizontal scroll bar to the list. I had the same issue recently when I updated the Windows installer docs. That's so stupid... Do we need to find a different solution here, or should we keep this? The other lists also have scroll bars, but those are rather short and the lists themselves are not as important as this one.
screenshot from 2018-02-02 19-31-25

bastimeyer:
> I'm not sure though how it'll look on the rendered page.

I knew this was gonna happen. Long lines of text are adding a horizontal scroll bar to the list. I had the same issue recently when I updated the Windows installer docs. That's so stupid... Do we need to find a different solution here, or should we keep this? The other lists also have scroll bars, but those are rather short and the lists themselves are not as important as this one.
screenshot from 2018-02-02 19-31-25

amurzeau:
I suggest something like this:
image

But to do that, I have added a fixed pixel width on the <pre> tag. I don't know much about CSS that would allow to have that in a clean manner.
Using the official RTD theme doesn't help, it has the same behavior.

bastimeyer:
I'll see if I can fix this later on this evening when I find the time. Let's merge this PR after the scrollbar issue has been fixed.

bastimeyer:
Submitted a PR for fixing the table layout: #1467

I had another thought about the `sudo` command and the root user indicated by the `#` prompt. Not using `sudo` here makes it consistent with the other commands listed in the table. I didn't notice that when I reviewed the commit changes. Not including `sudo` may be confusing though for unexperienced users. However, the `apt update` and `apt install streamlink` commands also don't use `sudo`.

So what should we do here? I guess being consistent and not using `sudo` is the better choice, right? If you want to keep it the way it is now, I'm also fine with that. If you want to change it back again, you could use the opportunity and save one unnecessary character by removing the space in between wget's params, like this: `wget -qO-`
😅

amurzeau:
Maybe not using sudo is better, as commands for other distributions don't use sudo either (even Ubuntu).
In fact, Debian doesn't have sudo by default, so I think it's better to remove it.

On the other side, Ubuntu uses sudo, and it might be better to add it there (and other distributions that use sudo with the default install). That can be done in another PR.

amurzeau:
Removed both `sudo` for Debian stable install instructions and change `wget -q0 -` to `wget -q0-`.

bastimeyer:
@gravyboat Should be good to go, right, or do you have any concerns? I don't want to merge this, because of my PR #1467 for fixing the table layout which I obviously shouldn't merge myself. :)

gravyboat:
I've merged both so we should be good to go @bastimeyer, thanks for the input and contribution @amurzeau.

PR #1457, opened 2018-01-30
[TF1] Fix plugin

BZHDeveloper:
TF1 group renamed some of her channels, NT1 & HD1, as "TFX" & "TF1 Séries Films". So I've updated plugin according to last TF1 changes.

### DIFF
--- a/src/streamlink/plugins/tf1.py
+++ b/src/streamlink/plugins/tf1.py
@@ -8,13 +8,13 @@ from streamlink.stream import HLSStream
 
 
 class TF1(Plugin):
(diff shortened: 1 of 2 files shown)


gravyboat:
What do you think about adding some tests to https://github.com/streamlink/streamlink/blob/master/tests/test_plugin_tf1.py to cover the new domains?

BZHDeveloper:
NOTE : TF1 redirect old channel names to new ones (for now).
so I added tests only for TF1 & TMC (NT1 & HD1 failure tests are commented and disabled)

melmorabity:
Works fine :)

gravyboat:
Good stuff thanks @BZHDeveloper, sorry for the slow response. Thanks for the review @melmorabity.

PR #1467, opened 2018-02-03
docs: fix table layout on the install page

bastimeyer:
- Add a custom class to all tables on the install page, so tables on
  other pages don't get changed
- Ignore the colgroup data, set consistent table cell widths and unset
  the white-space property for being able to have automatic line breaks
- Set explicit line breaks at certain positions
- Fix document white space


amurzeau:
I checked this merged with #1455, works fine with small screen and has no conflict with responsive behavior.

gravyboat:
Thanks @bastimeyer! And thanks everyone else for the reviews.

PR #1468, opened 2018-02-04
docs/install: use sudo for Ubuntu and Solus

amurzeau:
After a check of which distribution uses sudo with their default
installation, I found that only Ubuntu and Solus use sudo by default.


gravyboat:
Good catch, thank you.

Issue #1412, opened 2018-01-05
bigo.tv plugin issue


JohnDoh666:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ x] I have read the contribution guidelines.

### Description

They seem to have changed streaming protocol or something. I'm not able to pick up a streaming url via  the inspect elements/network feature in Chrome.

### Reproduction steps / Explicit stream URLs to test

streamlink https://www.bigo.tv/b190890 best
[cli][info] Found matching plugin bigo for URL https://www.bigo.tv/b190890
error: No playable streams found on this URL: https://www.bigo.tv/b190890

### Environment details

Win10, Python 3.4.4 and Streamlink 0.9.0

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

back-to:
the regex is different now

https://github.com/streamlink/streamlink/blob/498783880bf8ad1dd9c93e66be66dd301056a3b8/src/streamlink/plugins/bigo.py#L54-L56

```py
_video_re = re.compile(r"""videoSrc:\s?["'](?P<url>[^"']+)["']""")
```

---

but the real issue is, that the `.m3u8` will only play for **1 minute**

reloading the `.m3u8` url from the website after 45sec might work,
but it could be laggy.

unnutricious:
I am no longer seeing any issue with the hls stream playing for a limited amount of time, so I think we can just go ahead and update the regex.

JohnDoh666:
I have tested the regex fix and it's working without interruptions for me.
Thanks for you help @back-to

gravyboat:
Closing since associated PR is merged.

PR #1474, opened 2018-02-08
[bigo] update video regex to match current website (#1412)

unnutricious:
Update regex per #1412 so that the bigo plugin will work again.


gravyboat:
Thanks @unnutricious!

Issue #1263, opened 2017-10-09
Request: zattoo plugin should also download PVR files


anschluss-org:
### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

The zattoo plugin only gives access to live video and VOD streams. I am requesting that the plugin also allows user recordings to be downloaded.

### Expected / Actual behavior

Pasting the URL from the PVR/Recordings section should download the actual recording, not the live stream of the channel in the URL:

$  streamlink "https://zattoo.com/ondemand/watch/srf_zwei/110223896-die-schweizermacher/52845783/1455130800000/1455137700000/6900000" 5000k  --zattoo-email x@y.z --zattoo-password xxxxx

results in the output of the live stream of the srf_zwei channel instead of the recorded file

### Reproduction steps / Explicit stream URLs to test

sorry, for this I would have to indicate my zattoo credentials

### Environment details

Operating system and version: ...
MacOS 10.11.6, 10.12.6
(should not matter)
Streamlink and Python version: ...
Streamlink 0.8.1
Python 2.7.10
...

### Comments, logs, screenshots, etc.

...

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

karlo2105:
You should add switch -o and it shall save video from live or vod source.

anschluss-org:
> Am 09.10.2017 um 18:55 schrieb karlo2105 <notifications@github.com>:
> 
> You should add switch -o and it shall save video from live or vod source.
> 
Hi,

this is not the problem - the plugin selects the live video channel instead of the PVR/recording. Even with the -o switch added, the live stream from the channel is recorded instead of the recording.

back-to:
I'm geo-blocked to srf_zwei

can you try to create one highlight for 

```
- https://zattoo.com/watch/24h
- https://zattoo.com/watch/france-24-fr
- https://zattoo.com/watch/france-24-en
- https://zattoo.com/watch/bloomberg-europe
- https://zattoo.com/watch/r1
- https://zattoo.com/watch/rt_doc
- https://zattoo.com/watch/tve
```

I don't know if i can access it without a Premium account,
but i can give it a try.

anschluss-org:
Sure, here is a recording from TVE:

[...]

PR #1480, opened 2018-02-12
[zattoo] Added support for zattoo recordings

back-to:
Ref streamlink/streamlink#1263


gravyboat:
Good stuff, thanks @back-to!

Issue #1493, opened 2018-02-20
Neulion DMCA Takedown Notice


gravyboat:
I received the following email today and in the spirit of transparency am providing it here below:

------------
Hi there,

I'm contacting you on behalf of GitHub Support because we've received a DMCA takedown notice regarding the following content:

https://github.com/streamlink/streamlink

You can view the notice and the details of the requested changes below.

In this case only certain files were identified as allegedly infringing. Since it's not possible to disable individual files within a GitHub repository, we're giving you a 24 hour opportunity to remove the content named in the takedown. Please follow the steps in the following article to remove the content from the history:

https://help.github.com/articles/remove-sensitive-data

Once you've done that, please reply to this message to confirm the change. If we don't hear from you that a change is made within the next 24 hours, we will need to disable the entire repository according to our GitHub DMCA Takedown Policy, which you can find here:

https://help.github.com/articles/dmca-takedown-policy/

If you believe your content on GitHub was mistakenly disabled by a DMCA takedown request, you have the right to contest the takedown by submitting a counter notice, as described on the DMCA Takedown Policy page.

PLEASE NOTE: It is important that you reply within 24 hours to confirm whether you have made the requested changes. If you do not, the repository will be disabled.

Are you the copyright owner or authorized to act on the copyright owner's behalf?

NeuLion is the sole owner of all intellectual property rights in the NeuLion® Digital Platform as well as the NeuLion SDK/API. The sole use of the files that should be taken down is to infringe upon NeuLion’s intellectual property rights in the NeuLion Digital Platform as well as to provide a means for those that use such files to do so.

Please provide a detailed description of the original copyrighted work that has allegedly been infringed. If possible, include a URL to where it is posted online.

https://github.com/portse/plugin.video.ufcfightpass/blob/master/default.py is meant to steal content and violate terms of service for ufc.tv, based on the NeuLion Digital Platform and hosted/managed by NeuLion on behalf of the UFC. This code enables piracy activity that increases our CDN traffic costs.

https://github.com/streamlink/streamlink includes plugins such as Neulion meant to steal content and violate the terms of services on the NeuLion digital Platform and hosted customers including ufc.tv, Fanpass.co.nz, http://elevensports.tw/, http://elevensportsnetwork.sg/, http://elevensportsnetwork.be/, http://elevensports.lu/, http://elevensportsnetwork.pl/, http://elevensportsusa.com/, http://watch.nba.com/ and more.

[...]

PR #1497, opened 2018-02-21
[neulion] Remove plugin. See #1493

bastimeyer:
Opening a PR myself since nobody else did yet...
Please check if I've missed something, thanks!


gravyboat:
Good solution based on the discussion that occurred while I was asleep, thanks @bastimeyer!

Issue #990, opened 2017-06-07
Recent issues with facebook plug-ins.


DawgNewb:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description

Hello there, I'm usally used facebook plug-in to record stuff.
Until 2 day ago it seem like fb changing somethings and it not working anymore.

I also had one more request is to do fb authentication for live in age restrict page.

### Environment details

Operating system and version: Win7 x64
Streamlink version: Portable exe v0.5.0 - Git a6bb875

### Comments, logs, screenshots, etc.

```
Streamlink.exe https://www.facebook.com/*AnyPages*/videos/1893527364233958/ best -o "C:\today.ts"
[Streamlink for Windows v0.5.0 - Git a6bb875]
[cli][info] Found matching plugin facebook for URL  https://www.facebook.com/*AnyPages*/videos/1893527364233958/
[cli][info] Available streams: 642p (worst, best)
[cli][info] Opening stream: 642p (hls)
[stream.hls][error] Failed to open segment 438: Unable to open URL: https://video.fbkk1-3.fna.fbcdn.net/hvideo-ash5/v/rNXCUe1akD_Vc3JznmsI6/live-md/1893527370900624__U6_5_577RuoWr1iYscRfgg-438.ts (400 Client Error: Bad request - Invalid broadcast id for url: https://video.fbkk1-3.fna.fbcdn.net/hvideo-ash5/v/rNXCUe1akD_Vc3JznmsI6/live-md/1893527370900624__U6_5_577RuoWr1iYscRfgg-438.ts)
[stream.hls][error] Failed to open segment 439: Unable to open URL: https://video.fbkk1-3.fna.fbcdn.net/hvideo-ash5/v/rNXCUe1akD_Vc3JznmsI6/live-md/1893527370900624__U6_5_577RuoWr1iYscRfgg-439.ts (400 Client Error: Bad request - Invalid broadcast id for url: https://video.fbkk1-3.fna.fbcdn.net/hvideo-ash5/v/rNXCUe1akD_Vc3JznmsI6/live-md/1893527370900624__U6_5_577RuoWr1iYscRfgg-439.ts)
[...]

Issue #880, opened 2017-05-04
Facebook : Nominative address support


karlo2105:
### Checklist

- [ ] This is a bug report.
- [x] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

Facebook is a social video platform which supports live streaming. 

### Expected / Actual behavior

```
streamlink "https://www.facebook.com/IndexVijesti/videos/1906400442908386/" best
[cli][info] Found matching plugin facebook for URL https://www.facebook.com/Inde
xVijesti/videos/1906400442908386/
[cli][info] Available streams: 360p (best, worst)
[cli][info] Opening stream: 360p (hls)
[cli][info] Starting player: C:\Users\Ddr\Downloads\VLCPortable\App\vlc\vlc
[cli][info] Player closed
[cli][info] Stream ended
[cli][info] Closing currently open stream...
```

```
streamlink "https://www.facebook.com/IndexVijesti/"
error: No plugin can handle URL: https://www.facebook.com/IndexVijesti/

```
### Environment details

Operating system and version: Windows 7
Streamlink and Python version: Streamlink portable

johnthecracker:
C:\>youtube-dl.exe -v -F --no-check-certificate https://www.facebook.com/IndexVijesti/videos/1892803307601433
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', '-F', '--no-check-certificate', 'https://www.facebook.com/IndexVijesti/videos/1892803307601433']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2017.05.01
[debug] Python version 3.4.4 - Windows-7-6.1.7601-SP1
[debug] exe versions: ffmpeg n3.2.1, rtmpdump 2.4
[debug] Proxy map: {}
[facebook] 1892803307601433: Downloading webpage
[info] Available formats for 1892803307601433:
format code               extension  resolution note
247717965692890ad         m4a        audio only [eng] DASH audio   66k , mp4a.40.2 (48000Hz)
1419985091395935vd        mp4        428x240    [eng] DASH video  253k , avc1.4d401e, video only
122107244999658v          mp4        642x360    [eng] DASH video  503k , avc1.4d401e, video only
dash_sd_src               mp4        unknown
dash_sd_src_no_ratelimit  mp4        unknown
dash_hd_src               mp4        unknown
[...]

PR #1501, opened 2018-02-22
[facebook] mark as broken, they use dash now.

back-to:
more information https://github.com/streamlink/streamlink/issues/990#issuecomment-354804542

currently not fixable with streamlink

Closes https://github.com/streamlink/streamlink/issues/990
Closes https://github.com/streamlink/streamlink/issues/880


gravyboat:
Thanks @back-to!

Issue #1438, opened 2018-01-15
Periscope: Could not open stream HLSStream, but HLS variant playlist directly succeeds


prech:
When loading live Periscope streams (archived/saved videos stream worked well without issue), I've recently encountered this error every time:

> streamlink https://www.pscp.tv/w/1MYGNjnkkBZGw?channel=sporting-green
[cli][info] Found matching plugin periscope for URL https://www.pscp.tv/w/1MYGNjnkkBZGw?channel=sporting-green
[cli][info] Available streams: replay (worst, best)
[cli][info] Opening stream: replay (hls)
[cli][error] Try 1/1: Could not open stream <HLSStream('https://prod-video-ap-southeast-2.periscope.tv/NQAp9dKjjTHgPqpT2JHARL75vd8T8nQxsMYlZWwJAbQqM7n3JS945oO1ExWxcUZyJ9UPKXYImGolqfXuD9n6vg/live/ap-southeast-2/master_hlsproducer.m3u8')> (Could not open stream: Attempted to play a variant playlist, use 'hlsvariant://https://prod-video-ap-southeast-2.periscope.tv/NQAp9dKjjTHgPqpT2JHARL75vd8T8nQxsMYlZWwJAbQqM7n3JS945oO1ExWxcUZyJ9UPKXYImGolqfXuD9n6vg/live/ap-southeast-2/master_hlsproducer.m3u8' instead)
error: Could not open stream <HLSStream('https://prod-video-ap-southeast-2.periscope.tv/NQAp9dKjjTHgPqpT2JHARL75vd8T8nQxsMYlZWwJAbQqM7n3JS945oO1ExWxcUZyJ9UPKXYImGolqfXuD9n6vg/live/ap-southeast-2/master_hlsproducer.m3u8')>, tried 1 times, exiting

> streamlink hlsvariant://https://prod-video-ap-southeast-2.periscope.tv/NQAp9dKjjTHgPqpT2JHARL75vd8T8nQxsMYlZWwJAbQqM7n3JS945oO1ExWxcUZyJ9UPKXYImGolqfXuD9n6vg/live/ap-southeast-2/master_hlsproducer.m3u8
[...]

PR #1504, opened 2018-02-22
[periscope] Fix for variant HLS streams

back-to:
They use normal and variant playlists now.

Closes https://github.com/streamlink/streamlink/issues/1438


gravyboat:
Good stuff, thanks!

PR #1413, opened 2018-01-05
build: use versioneer to set the build number

beardypig:
First attempt at getting versioneer to work, mainly for the windows installer.

PR #1509, opened 2018-02-26
[codecov] use pytest and upload all data

back-to:
- use **pytest** not coverage for windows
- fixed not working linux upload, only windows upload worked before this.
- `pip install -e .` or pytest upload won't work
- don't run tests twice in travis
- don't ignore init files
- codecov target to 30
- ignore _version.py for #1413

---

`pip install .` will fix the 3.7 build,
`python setup.py install` is still broken on 3.7

---

codecov will go down from 50% to ~30%
with this it will now show the real %, because it will upload the 0% files as well.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,10 +5,9 @@ source =
 
 [report]
 omit =
-  */python?.?/*
-  *__init__*
   src/streamlink/packages/*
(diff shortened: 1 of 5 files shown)


gravyboat:
Thanks @back-to!

PR #1514, opened 2018-02-27
Add support for IDF1

melmorabity:
This plugin adds support for the French-local TV channel [IDF1](http://www.idf1.fr/).
It supports both [live](http://www.idf1.fr/live) and [VOD](http://www.idf1.fr/videos) streams.


gravyboat:
@melmorabity Is this site only for their own content? I went to it and I can't tell past seeing they seem to rely on ads and be from the early 2000s for site design.

melmorabity:
@geavyboat the website is entirely dedicated to IDF1. It's not a national TV channel (it only broadcasts in Paris metropolean area), that's why their website may look "cheap". The design was recently refreshed by the way ^^

melmorabity:
Thanks :)

PR #1518, opened 2018-02-28
plugins.canlitv: support for ecanlitvizle.live

beardypig:
Replacement PR for #1442.


back-to:
since this is a restreaming website
maybe removing it would make more sense,
most of the streams are already covered from there original source anyways.

gravyboat:
Yeah I agree with @back-to that we should just remove this plugin, we need to take a stance on this at some point and we've been cracking down on sites that are basically just restreams of stolen content.

dreamboxco:
here is the new site http://www.canlitvizle.ca/?i=1

beardypig:
Closing this as it seems to be a maintenance burden, we should open another PR to remove this plugin.

PR #1524, opened 2018-03-02
added cd streamlink

offboard:
navigate to streamlink before python setup.py install


gravyboat:
Good catch, thanks @offboard.

Issue #1484, opened 2018-02-16
baltic live cam plugin support


bobvargas:
*Thanks for reporting an issue!*
*Please read the contribution guidelines first! (see the link above)*
*Also check the list of known issues before reporting an issue!*

*Feel free to use the following template. Be as detailed as possible.*
*Please see the text preview to avoid unnecessary formatting errors.*
*Don't forget to remove this text before submitting.*

----

### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

...

### Expected / Actual behavior

...

### Reproduction steps / Explicit stream URLs to test

1. https://balticlivecam.com/cameras/united-states/new-york/brighton-beach/
2. ...
3. ...

### Logs

_Logs are always required for a bug report, use `-l debug` [(help)](https://streamlink.github.io/cli.html#cmdoption-l)
Make sure to **remove username and password**
You can upload your logs to https://gist.github.com/ or_

```
REPLACE THIS TEXT WITH YOUR LOG
```

### Comments, screenshots, etc.
I hope anyone submit this plugin request...
...

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

gravyboat:
We've decided we're not going to support this site for the reasons stated over on #1525.

PR #1525, opened 2018-03-02
[balticlivecam] New Plugin for balticlivecam.com

back-to:
Closes https://github.com/streamlink/streamlink/issues/1484

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -28,6 +28,7 @@ ard_mediathek       - ardmediathek.de    Yes   Yes   Streams may be geo-restrict
                     - mediathek... [5]_
 artetv              arte.tv              Yes   Yes
 atresplayer         atresplayer.com      Yes   No    Streams are geo-restricted to Spain.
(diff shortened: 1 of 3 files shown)


gravyboat:
I have concerns over the impact this plugin would have on the site. It seems like they are a pretty small operation that primarily relies on ad traffic to keep these livestreams going (along with supporting their other enterprise supporting animal shelters via these livestreams), and they even have a small window relating to ads supporting them. I'd like to see what other people think. @beardypig, @bastimeyer, thoughts?

bastimeyer:
Does it even make sense adding plugins for such small websites? I can't imagine anyone watching public livecams regularly with Streamlink.

This plugin request looks like something in the sense of "I just stumbled upon this site by accident and couldn't find a plugin for it in Streamlink"... Your concerns are absolutely justified, @gravyboat.

gravyboat:
@bastimeyer That's a good point to bring up as well. I think I'll close this for now, sorry to waste your time @back-to. Definitely my fault for not investigating the issue when it was originally opened.

Issue #1435, opened 2018-01-14
Plugin tvrplus.ro : No playable streams found


karlo2105:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

tvrplus.ro is Romanian public broadcasters service which offers all state channels online. TVR International is the only channel unlocked abroad. 

### Expected / Actual behavior

```
streamlink "http://www.tvrplus.ro/live-tvr-international" best
[cli][info] Found matching plugin tvrplus for URL http://www.tvrplus.ro/live-tvr
-international
error: No playable streams found on this URL: http://www.tvrplus.ro/live-tvr-int
ernational
```

### Reproduction steps / Explicit stream URLs to test

1. http://www.tvrplus.ro/live-tvr-international

When adding Referer : 

```
streamlink --http-header "Referer=http://www.tvrplus.ro/live-tvr-international" "hlsvariant://http://89.136.209.33:1935/liveorigin/TVRI.smil/playlist.m3u8" best
[cli][info] Found matching plugin hls for URL hlsvariant://http://89.136.209.33:1935/liveorigin/TVRI.smil/playlist.m3u8
[cli][info] Available streams: 360k (worst), 1200k, 2500k (best)
[cli][info] Opening stream: 2500k (hls)
[cli][error] Try 1/1: Could not open stream <HLSStream('http://89.136.209.33:193
5/liveorigin/TVRI.smil/chunklist_w1335124422_b2500000.m3u8')> (Could not open st
ream: Unable to open URL: http://89.136.209.33:1935/liveorigin/TVRI.smil/chunkli
st_w1335124422_b2500000.m3u8 (404 Client Error: Not Found for url: http://89.136
.209.33:1935/liveorigin/TVRI.smil/chunklist_w1335124422_b2500000.m3u8))
error: Could not open stream <HLSStream('http://89.136.209.33:1935/liveorigin/TV
RI.smil/chunklist_w1335124422_b2500000.m3u8')>, tried 1 times, exiting
```

### Environment details

Windows 7 Pro
streamlink 0.9.0/Python 2.9.13

### Comments, logs, screenshots, etc.

They changed broadcasting addresses. 
Currently : 

`http://89.136.209.33:1935/liveorigin/TVRI.smil/playlist.m3u8`

back-to:
regex changed

[...]

PR #1526, opened 2018-03-02
[TVRPlus] Fix for hls_re and use headers for HLSStream

back-to:
`best` and `worst` won't work, the source is **404**

`streamlink "http://www.tvrplus.ro/live-tvr-international" 1200k`

```
360k (worst) - broken 404
1200k - works
2500k (best) - broken 404
```

Closes https://github.com/streamlink/streamlink/issues/1435


gravyboat:
Good stuff, thanks @back-to.

PR #1532, opened 2018-03-04
update plugin for longzhu.com to support new url pattern

fozzysec:
Support new url pattern `http://y.longzhu.com/<channel>?from=challcontent`


gravyboat:
Thanks @fozzysec.

Issue #1542, opened 2018-03-09
Dailymotion plugin is broken


melmorabity:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description
Dailymotion plugin is failing on live and VOD streams since today

### Reproduction steps / Explicit stream URLs to test

On a VOD stream:
```
$ streamlink -l debug "https://www.dailymotion.com/video/x6g0kq0"
[cli][debug] OS:         Linux-4.15.6-300.fc27.x86_64-x86_64-with-fedora-27-Twenty_Seven
[cli][debug] Python:     3.6.4
[cli][debug] Streamlink: 0.11.0
[cli][debug] Requests(2.18.4), Socks(1.6.7), Websocket(0.40.0)
[cli][info] Found matching plugin dailymotion for URL https://www.dailymotion.com/video/x6g0kq0
[plugin.dailymotion][debug] Found media ID: x6g0kq0
error: Unable to open URL: https://www.dailymotion.com/sequence/full/x6g0kq0 (404 Client Error: Not Found for url: https://www.dailymotion.com/sequence/full/x6g0kq0)
```
On a live stream:
```
$ streamlink -l debug https://www.dailymotion.com/video/x4rdeu6
[cli][debug] OS:         Linux-4.15.6-300.fc27.x86_64-x86_64-with-fedora-27-Twenty_Seven
[cli][debug] Python:     3.6.4
[cli][debug] Streamlink: 0.11.0
[cli][debug] Requests(2.18.4), Socks(1.6.7), Websocket(0.40.0)
[cli][info] Found matching plugin dailymotion for URL https://www.dailymotion.com/video/x4rdeu6
[plugin.dailymotion][debug] Found media ID: x4rdeu6
error: Unable to open URL: https://www.dailymotion.com/sequence/full/x4rdeu6 (404 Client Error: Not Found for url: https://www.dailymotion.com/sequence/full/x4rdeu6)
```

### Comments, screenshots, etc.

Looks like the API URL called to retrieve stream data is dead (https://www.dailymotion.com/sequence/full/{video_id}).
By sniffing video playing in my browser, with different combinations of settings (UAs, HTML5 enabled/disabled), it seems also that:
* http://www.dailymotion.com/player/metadata/video/{video_id} can be used instead to retrieve stream data
[...]

PR #1543, opened 2018-03-09
[dailymotion] Fix for new stream data API

melmorabity:
* Fix https://github.com/streamlink/streamlink/issues/1542
* Add unit tests for dailymotion plugin

### DIFF
--- a/src/streamlink/plugins/dailymotion.py
+++ b/src/streamlink/plugins/dailymotion.py
@@ -9,7 +9,7 @@ COOKIES = {
     "family_filter": "off",
     "ff": "off"
 }
-STREAM_INFO_URL = "http://www.dailymotion.com/player/metadata/video/{0}"
+STREAM_INFO_URL = "https://www.dailymotion.com/player/metadata/video/{0}"
 USER_INFO_URL = "https://api.dailymotion.com/user/{0}"
 
 _url_re = re.compile(r"""
--- a/src/streamlink/plugins/dailymotion.py
+++ b/src/streamlink/plugins/dailymotion.py
@@ -1,37 +1,17 @@
-import json
 import re
 
-from functools import reduce
-
-from streamlink.compat import urlparse, range
 from streamlink.exceptions import NoStreamsError
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import http, validate
-from streamlink.stream import HDSStream, HLSStream, HTTPStream, RTMPStream
-from streamlink.stream.playlist import FLVPlaylist
+from streamlink.stream import HLSStream, HTTPStream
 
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/dailymotion.py
```diff
@@ -92,138 +46,24 @@
...
+        for quality, streams in media['qualities'].items():
+            for stream in streams:
+                if stream['type'] == 'application/x-mpegURL':
+                    if quality != 'auto':
+                        # Avoid duplicate HLS streams with bitrate selector in the URL query
+                        continue
+                    for s in HLSStream.parse_variant_playlist(self.session, stream['url']).items():
+                        yield s
+                elif stream['type'] == 'video/mp4':
+                    bitrate = quality + 'p'
```
back-to: _this is not so important but_
**bitrate** is currently the **resolution**,
maybe change it to `resolution` , `name` or something

---

`$ streamlink  https://www.dailymotion.com/video/x5qnozs 1080 -l debug`

currently the names of bitrate are
```
144p
240p
380p
480p
720@60p
1080@60p
```
but @ is invalid so it will be 720 and 1080

```
[cli][info] Available streams: 480p_http, 720, 1080, 112p_alt (worst),
112p, 144p, 184p_alt, 184p, 240p, 288p_alt, 288p, 380p, 480p_alt,
480p, 720p_alt, 720p, 1080p_alt, 1080p (best)
```

---

maybe use `re.sub` and replace @ with p
or
use only the resolution and remove the FPS

src/streamlink/plugins/dailymotion.py
```diff
@@ -1,37 +1,17 @@
...
-    "sd": "360p",
-    "hq": "480p",
-    "hd720": "720p",
-    "hd1080": "1080p",
-    "custom": "live",
-    "auto": "hds",
-    "source": "hds"
-}
-STREAM_INFO_URL = "https://www.dailymotion.com/sequence/full/{0}"
+STREAM_INFO_URL = "http://www.dailymotion.com/player/metadata/video/{0}"
```
back-to: `https://` is available, maybe use it instead.

src/streamlink/plugins/dailymotion.py
melmorabity: Good catch ;).
I suppose @[0-9]+ corresponds to a FPS.

src/streamlink/plugins/dailymotion.py
melmorabity: Totally right, I've fixed it too

gravyboat:
Good stuff, thanks @melmorabity and thanks for the review @back-to.

Issue #1505, opened 2018-02-23
Ott Streams (ellobo) plugin error


skilleek:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

Plugin for ott streams streann.py, returns the error:

`streamlink "ellobo106.com/index.php/vivo/solo-3audio"
[cli][info] Found matching plugin ellobo for URL ellobo106.com/index.php/vivo/solo-3audio
error: Unable to parse JSON: No JSON object could be decoded ()`

url: http://www.ellobo106.com/index.php/vivo/solo-3audio

I would be grateful if you could fix it. Thank you as always for all your support!

back-to:
i think it works again, could you try it @skilleek ?

```
$ streamlink http://www.ellobo106.com/index.php/vivo/solo-3audio
[cli][info] Found matching plugin ellobo for URL http://www.ellobo106.com/index.php/vivo/solo-3audio
Available streams: 96k (worst, best)
[ka@ka-MSI ~]$ streamlink http://www.ellobo106.com/index.php/vivo/solo-3audio best
[cli][info] Found matching plugin ellobo for URL http://www.ellobo106.com/index.php/vivo/solo-3audio
[cli][info] Available streams: 96k (worst, best)
[cli][info] Opening stream: 96k (hls)
[cli][info] Starting player: /usr/bin/mpv
```

PR #1544, opened 2018-03-10
[streann] Fixed broken plugin

back-to:
- **more debug** msg
- set **headers** on every **get request**
- use **new** url for **stream_url**, the old one still works but the website uses this url

Closes https://github.com/streamlink/streamlink/issues/1505


gravyboat:
Thanks @back-to.

beardypig:
FYI, you should be able to do `http.headers = {...}`, to set them for every request.

PR #1560, opened 2018-03-18
plugins.live_russia_tv New Plugin implementation for live.russia.tv

piddubnyi:
Adds support for Russian TV channels from https://live.russia.tv/
E.g.
$ streamlink https://live.russia.tv/index/index/channel_id/4 best

### DIFF
--- a/src/streamlink/plugins/live_russia_tv.py
+++ b/src/streamlink/plugins/live_russia_tv.py
@@ -14,16 +14,20 @@ class LiveRussia(Plugin):
 
     def _get_streams(self):
         res = http.get(self.url)
-        iframe_url = self.check_re(self.iframe_re, res.text)
-        res = http.get(iframe_url)
-        stream_url = self.check_re(self.stream_re, res.text)
-        return HLSStream.parse_variant_playlist(self.session, stream_url)
+        iframe_result = re.search(self.iframe_re, res.text)
 
-    def check_re(self, search, text):
-        re_result = re.search(search, text)
-        if re_result:
-            return re_result.group(1)
-        self.logger.error("The requested content is unavailable.")
+        if not iframe_result:
+            self.logger.error("The requested content is unavailable.")
+            return
+
+        res = http.get(iframe_result.group(1))
+        stream_url_result = re.search(self.stream_re, res.text)
+
+        if not stream_url_result:
+            self.logger.error("The requested content is unavailable.")
+            return
+
+        return HLSStream.parse_variant_playlist(self.session, stream_url_result.group(1))
 
 
 __plugin__ = LiveRussia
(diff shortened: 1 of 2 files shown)


beardypig:
Looks like a site that iframes streams from other sites?

karlo2105:
@beardypig 
You are wrong. Those are channels of Russian public TV which belong to Vgtrk.
Take a look : 
https://en.wikipedia.org/wiki/All-Russia_State_Television_and_Radio_Broadcasting_Company
https://en.wikipedia.org/wiki/Russia-1

It's really annoying asking always to justify plugin for any website even for official websites. If it's free who cares there aren't any reasons not to merge it.
I suggest in future to separate plugins from github and host them on alternative websites outside of usa which will not put in jeopardy streamlink development.

gravyboat:
@karlo2105 It matters because if a site is stealing content then that site with the stolen content is negatively impacted. How would you feel if it was your site and some random people were stealing traffic from you and you couldn't monetize their viewing? It's important that we're still responsible internet citizens when we can be. The real question instead of `What's the purpose of streamlink without plugins?` is `Do we need a plugin for every single site?`. The original purpose was "The main purpose of streamlink is to convert CPU-heavy flash plugins to a less CPU-intensive format." which applied to sites like Twitch and a few others.

For me it's just as annoying to ask for justification for every plugin because people kept trying to get shady plugins added, porn plugins, plugins for sites which no one uses legitimately, etc. We work on this for free so cut us some slack here. It's a lot of thankless unpaid work and we're doing the best we can.

karlo2105:
> How would you feel if it was your site and some random people were stealing traffic from you and you couldn't monetize their viewing? 

> 

If monetisation was my main money mearning I would protect it either with Drm or similar system as streamlive.to, otherwise i wouldn't give a damn for it. 
Yes streamlink is very good for Cpu heavy flash plugins. Believe me html5 is even worse than flash plugins.
I wasn't rude and i gave above evidences of Russia Tv official channels.

I wasn't rude with anybody

beardypig:
@karlo2105 it would have been sufficient to say that russia.tv is the official site for russian state television - I don't read/speak Russian and the live.russia.tv site looks like a re-framing website. 
The reason we don't want to support those websites is for maintenance, it's all well and good having 100s of plugins but without the developers to support them we will end up with non-working plugins all over the place. Obviously we welcome contributions, but we do want to ensure that the quality of streamlink remains high.

karlo2105:
@beardypig 
Much obliged for clarification. ;)

beardypig:
@piddubnyi any chance you can review some of the requested changes?

piddubnyi:
@beardypig Please check.

beardypig:
Looks good, thanks @piddubnyi :)

gravyboat:
Thanks @piddubnyi for the contribution and thanks for the review @beardypig.

PR #1576, opened 2018-03-26
Window Titles = Stream Titles + Other Attributes

Hubcapp:
This pull request is influenced by the code and comments left in https://github.com/streamlink/streamlink/pull/623, but it's been rewritten to address some concerns which prevented this feature from entering Streamlink about a year ago now.

**What does this add?**
* A `--title` flag which allows the user to change the text that appears in the title bar of the video player Streamlink opens based on attributes specific to the stream.
* {title}
* {category} and its synonym {game}
* {author}

**What doesn't this add?**
* Any format codes to the `-a` or `-p` flags, because it is either impossible or impractical to support literal quotes inside literal quotes in the context of `cmd.exe`.
* Any silly custom quotes to get around above limitation.
* Support for `--title` in players other than `vlc` or `mpv`.
* [Twitch support](https://gist.github.com/Hubcapp/65521f9182150c4d84a598838ef901c8) (separate PR)
* [YouTube support](https://gist.github.com/Hubcapp/eb232a5fcc5bca4d1502270b5b8e4ba0) (separate PR)

**What could it support?**
* Using the `--title` flag as the "weird quote" for the `-a` and `-p` flag, but since this feature has been blocked for so long already, I'd prefer that as a separate PR.

**Example usage:**
`streamlink --title 'title: {title} -- author: {author} -- game: {game} -- category (same as game): {category}' https://www.twitch.tv/bobross 720p`
...
`streamlink -p 'mpv' --title 'title: {title} -- author: {author} -- game: {game} -- category (same as game): {category}' https://www.twitch.tv/bobross 720p`


Hubcapp:
Just realized, could add the `{title}` format code back to the `-a` flag for the user to use in the following way:

`streamlink -p 'unknown_player' --title '{title} -- {author} -- {category}' -a '--weird_player_title_flag={title} {filename}' https://www.twitch.tv/bobross 720p`

In that way, the `--title` flag would act as our "weird quote", holding the entire contents of what we want the title to be.

This would allow any arbitrary player that supports passing the window title over command line to be supported in Streamlink.

src/streamlink_cli/main.py
```diff
@@ -123,7 +127,17 @@ def create_http_server(host=None, port=0):
 
     return http
 
-
+def create_title(plugin=None):
```
beardypig: Looks like there is a bug in the Python 2 part, `args.encode('utf8')`. 

I think you could simplify this method by encoded (or not) at the end - didn't test it though. 
```python
def create_title(plugin=None):
	if args.title and plugin:
        title = args.title.format(title=plugin._get_title(),
                                  author=plugin._get_author(),
                                  category=plugin._get_category(),
                                  game=plugin._get_category())
    else:
        title = args.url

	if is_py2:
        return title.encode("utf8")
	else:
        return title
```

src/streamlink/plugin/plugin.py
```diff
@@ -398,6 +398,15 @@ def get_streams(self, *args, **kwargs):
     def _get_streams(self):
         raise NotImplementedError
 
+    def _get_title(self):
```
beardypig: I think these should be public method (ie. not prefixed with `_`) or possibly `@property`s - this is adding something to the API that other plugin devs might have an opinion on (paging dr. @back-to).

beardypig:
@Hubcapp thanks for the work on this, I have left a few comments (before even testing it, which I'm going to do now)... :)

Hubcapp:
Thanks for the feedback. I hadn't tested for Python 2 yet, because I really wanted to get everyone's opinion on the feature before continuing to develop it. I'm not 100% sure it's working after the merge conflict resolution. I haven't tested it after the merge conflict, but have been using 5c8990b with no problems.

Just wanted feedback on whether or not this is starting to look like something that won't be blocked for the same reasons as the last PR

beardypig:
Have you tested it out on Windows? I can help to write some tests if you'd like?

Hubcapp:
I would love it if you would write some tests. I tested briefly on Windows, it worked there.

src/streamlink_cli/main.py
beardypig: Also, what should happen in the case where `_get_author` is not implemented?

Hubcapp: It's supposed to fallback to these values:

     DEFAULT_STREAM_METADATA = {
         "title"    : u"Untitled",
         "author"   : u"Unknown Author",
         "category" : u"No Category",
         "game"     : u"No Game/Category"
     }

beardypig: I can't see where that happens...

Hubcapp: Maybe not implemented yet...

src/streamlink/plugin/plugin.py
Hubcapp: plugin developers who are more familiar with the streamlink API than I am could implement `@property`s if it's not too big of a job. Personally I don't know what a `@property` is. Just removed the leading underscore for now.

src/streamlink_cli/main.py
Hubcapp: I changed plugin.get_author, plugin.get_title, and plugin.get_category to no longer `raise NotImplementedError`. Now it'll return the values in `DEFAULT_STREAM_METADATA`.

Hubcapp:
All feedback from beardypig implemented (except making `get_title` etc into `@property`s, don't know what those are)

Still looking for more feedback & confirmation that `--act --atc --tca --tac` etc don't need to be implemented.

src/streamlink/plugin/plugin.py
beardypig: It was more that for other Plugins to give a Title, Author, etc. they would need to implement these methods, so once they have names that's it :) FYI `@propery`: https://docs.python.org/3/library/functions.html#property

Hubcapp: I've gone into deep meditation on `@property`, and decided that I don't understand the benefit of creating this structure versus just defining methods as they're written. Seems like an appropriate place to use it, but I just don't understand what benefit it has.

Not sure what "once they have names that's it" means, sorry :s

beardypig: For @property I just think it looks nicer to write `plugin.title` vs. `plugin.get_title()` - but probably `get_title` is better ;)

Hubcapp: I mean, I want to make whatever changes are necessary to get the feature into streamlink. But personally I like `get_title()` better because it's clear that it's explicitly calling a function. The reader would have to know that `plugin.title` is this `@property` structure, or else while skimming, they could just think it's a static value that doesn't do anything.

If not using `@property` makes it harder to refactor later and you want it to be a `@property` then I'll make it a property. I just want people to care about the feature and work with me.

beardypig: Yeah, `get_title()` is better as it might have to make extra calls, like in your YouTube Twitch examples.

Hubcapp: Ok, thanks... :-)

src/streamlink_cli/output.py
```diff
@@ -108,12 +109,27 @@ def _create_arguments(self):
...
         args = self.args.format(filename=filename)
         cmd = self.cmd
+        
+        if self.title is not None:
+            #vlc
+            if self.cmd in SUPPORTED_PLAYERS['vlc']:
+                self.title.replace("$","$$").replace("\$$","$") #see https://wiki.videolan.org/Documentation:Format_String/, allow escaping with \$
+                if is_win32:
+                    return cmd + " --input-title-format=\"" + self.title + "\" " + args
+                return shlex.split(cmd) + shlex.split("--input-title-format=" + shlex_quote(self.title)) + shlex.split(args)
```
beardypig: Here is the real meat of the issue... On macOS and Linux it works great to let `list2cmdline` deal with the escaping (as it should). 

```python
return shlex.split(cmd) + ["--input-title-format", self.title] + shlex.split(args)
```

But on Windows we don't want to apply `shlex.split`, as it doesn't work quite right. 

Doing this works for me on Windows, Linux and macOS. 
```python
args = self.args.format(filename=filename)
cmd = self.cmd
extra_args = []

if self.title is not None:
    # vlc
    if self.cmd in SUPPORTED_PLAYERS['vlc']:
        # see https://wiki.videolan.org/Documentation:Format_String/, allow escaping with \$
        self.title.replace("$", "$$").replace("\$$", "$")
        extra_args.extend(["--input-title-format", self.title])

    # mpv
    if self.cmd in SUPPORTED_PLAYERS['mpv']:
        # see https://mpv.io/manual/stable/#property-expansion, allow escaping with \$
        self.title.replace("$", "$$").replace("\$$", "$")
        extra_args.extend(["--title", self.title])

# other player
if is_win32:
    return cmd + " " + subprocess.list2cmdline(extra_args) + " " + args
return shlex.split(cmd) + extra_args + shlex.split(args)
```

Hubcapp: Yeah, that looks better to me. I think I had it in my head that win32 is already protected from title-based attacks somewhere down the line with how the player is launched, but your code looks cleaner and uses list2cmdline, so I'll use it.

src/streamlink_cli/output.py
```diff
@@ -108,13 +109,27 @@ def _create_arguments(self):
...
         else:
             filename = "-"
-
         args = self.args.format(filename=filename)
         cmd = self.cmd
+        extra_args = []
+
+        if self.title is not None:
+            # vlc
+            if self.cmd in SUPPORTED_PLAYERS['vlc']:
```
beardypig: There is a problem with this, it doesn't work if you have some extra args with `-p`, for example `-p "mpv --no-audio"`. However, I have a fix ;)

```diff
diff --git a/src/streamlink_cli/output.py b/src/streamlink_cli/output.py
index 31af016..14fc797 100644
--- a/src/streamlink_cli/output.py
+++ b/src/streamlink_cli/output.py
@@ -82,6 +82,7 @@ class PlayerOutput(Output):
         self.namedpipe = namedpipe
         self.http = http
         self.title = title
+        self.player = self.supported_player(self.cmd)

         if self.namedpipe or self.filename or self.http:
             self.stdin = sys.stdin
@@ -100,6 +101,13 @@ class PlayerOutput(Output):
         sleep(0.5)
         return self.player.poll() is None

+    @staticmethod
+    def supported_player(cmd):
+        """
+        Check if the current player supports adding a title
+
+        :param cmd: cmd to test
+        :return: name of the player|None
+        """
+        for player, possiblecmds in SUPPORTED_PLAYERS.items():
+            for possiblecmd in possiblecmds:
+                if possiblecmd in cmd:
+                    return player
+
     def _create_arguments(self):
         if self.namedpipe:
             filename = self.namedpipe.path
@@ -115,13 +123,13 @@ class PlayerOutput(Output):

         if self.title is not None:
             # vlc
-            if self.cmd in SUPPORTED_PLAYERS['vlc']:
+            if self.player == "vlc":
                 # see https://wiki.videolan.org/Documentation:Format_String/, allow escaping with \$
                 self.title = self.title.replace("$", "$$").replace("\$$", "$")
                 extra_args.extend(["--input-title-format", self.title])

             # mpv
-            if self.cmd in SUPPORTED_PLAYERS['mpv']:
+            if self.player == "mpv":
                 # see https://mpv.io/manual/stable/#property-expansion, allow escaping with \$
                 self.title = self.title.replace("$", "$$").replace("\$$", "$")
                 extra_args.extend(["--title", self.title])    
```

src/streamlink/plugin/plugin.py
```diff
@@ -398,6 +400,15 @@ def get_streams(self, *args, **kwargs):
     def _get_streams(self):
         raise NotImplementedError
 
+    def get_title(self):
```
beardypig: I think these should return `None` instead, and the `create_title` method should handle the `None` response. This is better because if a Plugin cannot find the title or category for some stream, then it doesn't have to look up what the default is and return it - easier for Plugins to not screw it up ;)

src/streamlink_cli/main.py
```diff
@@ -123,7 +127,19 @@ def create_http_server(host=None, port=0):
 
     return http
 
+def create_title(plugin=None):
+    if args.title and plugin:
+        title = args.title.format(title=plugin.get_title(),
```
beardypig: If possible this should only call the methods that are required to build the title, otherwise there could be slowdowns from extra API calls, etc.

src/streamlink/plugin/plugin.py
beardypig: Oe maybe it could be one method that returns a `dict` so that plugins could add their own metadata, like `episode` or `season`... but, maybe that is crazy.

src/streamlink_cli/main.py
beardypig: Might be nice, if they all returned `None` (previous comment), to default back to `args.url` for the title.

beardypig:
FYI, I experimented and added this to the `bbciplayer` plugin - works reeeaaal nice :)

src/streamlink_cli/constants.py
```diff
@@ -3,8 +3,20 @@
...
+DEFAULT_STREAM_METADATA = {
+    "title"    : u"Untitled",
+    "author"   : u"Unknown Author",
+    "category" : u"No Category",
+    "game"     : u"No Game/Category"
+}
+SUPPORTED_PLAYERS = {  #these are the players that streamlink knows how to set the window title for with `--title`. key names are used in help text
+    #name      #possible binary names
+    "vlc"    : ["vlc",find_default_player()],
+    "mpv"    : ["mpv"]
```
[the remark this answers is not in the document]
beardypig: I suggested an alternative method for doing this, https://github.com/streamlink/streamlink/pull/1576#pullrequestreview-126887060

src/streamlink/plugin/plugin.py
```diff
@@ -12,6 +12,8 @@
 from streamlink.exceptions import PluginError, NoStreamsError
 from streamlink.options import Options, Arguments
 
+from streamlink_cli.constants import DEFAULT_STREAM_METADATA
```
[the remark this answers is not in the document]
beardypig: I think this should be handled by the CLI, and the `get_title`, etc. methods should return `None` (see https://github.com/streamlink/streamlink/pull/1576#discussion_r193837320)

src/streamlink/plugins/youtube.py
```diff
@@ -138,6 +172,14 @@ class YouTube(Plugin):
         258: 258,
     }
 
+    def __init__(self, url):
```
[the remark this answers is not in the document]
beardypig: This should be up to the individual plugins to handle, I think.

src/streamlink/plugin/plugin.py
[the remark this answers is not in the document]
back-to: my comment was just meant for the `import` not the `function`

but `None` is probably better for the function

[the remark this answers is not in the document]
beardypig: Ah yeah, I just meant that if they return `None` then the import is no longer required :)

beardypig:
Hey @Hubcapp, are you still planning to work on this? :)

Hubcapp:
Yeah, I'll put the changes you both have made into my branch and resolve the new conflict. I think since 3 out of 3 people agree on not needing the --at --ta --cta --cat --act --atc --tca and --tac flags, I'll remove references to them as well.

src/streamlink/plugin/plugin.py
Hubcapp: I kind of like the `dict` idea, but it does create the problem of "secret format codes" for users. If there are only 3 (4 counting the category/game synonym) format codes, then it's easier for users to know all the possible format codes without reading documentation for each plugin. Plus I think it's easier to write for plugin developers...

beardypig: Yeah, probably. 

My only concern is that `{game}` is very specific and `{category}`/`{author}` don't apply to everything. If we include `{game}`, why not include more? if we don't include more options, why include `{game}`?

Hubcapp: To me it seems like a lot of online streaming is either gaming oriented like twitch, or more like youtube. So that's the perspective I had adding those format codes. `category` doesn't have to be implemented, and I think every stream should have a human/corporate_entity that starts it (`author`)

I'm open to including more format codes, but I don't know what they would be. I'd need the perspective of someone more involved with streamlink. You said maybe `episode` and `season`, but I think that could acceptably be covered by `category`, or really, that particular information would usually be in the `title` of the stream.

Hubcapp: yeah, if I were implementing `{title} {category} {author}` for a plugin where the `season` and `episode` are available, I'd put it in `{title}` like `S01E01 It all begins`

beardypig: There are a few gaming plugins, but the majority are live TV - which I guess doesn't really have `{game}` or `{category}` :) You are right about `{author}`, that should be applicable in most cases.

src/streamlink_cli/main.py
```diff
@@ -144,17 +170,21 @@ def create_title(plugin=None):
         if _game is None:
             _game = DEFAULT_STREAM_METADATA["game"]
         
+        #python 2 unicode handling
```
beardypig: What was the problem here? If `args.title` is unicode it should be fine right?

Hubcapp: python's string.format doesn't want to format unless I put that there.

src/streamlink_cli/output.py
```diff
@@ -166,7 +169,18 @@ def _open_call(self):
     def _open_subprocess(self):
         # Force bufsize=0 on all Python versions to avoid writing the
         # unflushed buffer when closing a broken input pipe
-        self.player = subprocess.Popen(self._create_arguments(),
+        if is_py2:
```
beardypig: This might be useful as a general util method in `streamlink_cli.utils` :)

src/streamlink_cli/output.py
```diff
@@ -143,8 +143,11 @@ def _create_arguments(self):
 
         # player command
         if is_win32:
-            return cmd + " " + subprocess.list2cmdline(extra_args) + " " + args
-        return shlex.split(cmd) + extra_args + shlex.split(args)
+            if is_py2:
+                return unicode.join( u'', [ cmd, " ", subprocess.list2cmdline(extra_args).decode('utf-8'), " ", args ] )
```
beardypig: Is this equivalent to 

```python
  u''.join([cmd, " ", subprocess.list2cmdline(extra_args).decode('utf-8'), " ", args])
```

Hubcapp: I'm not sure. I was having problems concatting the string and found `unicode.join` as a suggestion. I'll use ` u''.join([cmd, " ", subprocess.list2cmdline(extra_args).decode('utf-8'), " ", args])` if it works and is preferrable

beardypig: We should add a util method for this, something like
```python
def maybe_decode(text, encoding="utf8"):
    if is_py2 and isinstance(text, str):
        return text.decode(encoding)
    else:
        return text
```

Then you could just do 

```python    
return u' '.join([cmd, maybe_decode(subprocess.list2cmdline(extra_args), "utf8"), args])
```

beardypig: And the reverse, `maybe_encode` :)

src/streamlink_cli/main.py
```diff
@@ -130,11 +132,35 @@ def create_http_server(host=None, port=0):
...
 def create_title(plugin=None):
     if args.title and plugin:
-        _title = plugin.get_title()
-        _author = plugin.get_author()
-        _category = plugin.get_category()
-        _game = _category
+        _title = _author = _category = _game = None
+        
+        #"did the user include a format code?" logic
+        tokens = ["t17L3_XXQZ","4uTh0r_qZ","c4Teg0ry_zQ","G4m3_z4"]
```
beardypig: What's this trying to do, looks very cryptic to me :)

Hubcapp: basically just "create a substring not in original string" so that we can use the same logic of string.format to see if we will alter the `--title` string given by the user without creating key errors. 

we can't just do `if args.title != args.title.format(title="different")` because if `{author}` is in args.title then we get a key error. So I create these tokens that are defined in a way that they cannot possibly be substrings of `args.title`, and we're still able to track which format code was replaced.

If those substring tokens appear in arg_title_formatted, then I know that args.title.format put them there. and I can compare what token is in args_title_formatted to determine which format code was in the original string.

a simpler implementation would be something like `if "{author}" in args.title:`, but "what if the user wants to put {{author}} in the title?". I had flashbacks to when I implemented custom quotes in #623 and everyone hated it. This implementation uses no regex's and only uses the same exact string.format logic to detect if string.format will string.format the string.

src/streamlink_cli/main.py
Hubcapp: I didn't, but I think I would need the "maybe_decode" function you mentioned if I want it to be prettier than what's written here.

src/streamlink_cli/main.py
bastimeyer: > So I create these tokens that are defined in a way that they cannot possibly be substrings of args.title

This method was one of the reasons why the previous PR was rejected. I haven't looked into this PR yet (was just briefly reading the comments), but doing string replacement based on (static) tokens which are "not likely to be used" is bad. If this method really has to be used here, at least generate a random prefix with an additional timestamp.

Hubcapp: it is literally impossible now because we loop until we generate string not in args.title.

src/streamlink_cli/main.py
beardypig: Yeah, like that :)

src/streamlink_cli/main.py
beardypig: So the idea is you want to avoid calling the `get_{author,title,game}` method unless the format code is in the title?

Hubcapp: correct, that's what this does. it prevents unnecessary api calls.

beardypig: Here is an alternative, more generic (but no less magical), implementation... 

https://gist.github.com/beardypig/14da8dc260cb249b3c0d7508130b1d24

Hubcapp: This looks better. :) I'll put `class LazyFormatter(object):` into `streamlink_cli/util/lazy_formatter.py` and use it.

beardypig: Needs a couple more tests for non-callable items.

Hubcapp: `create_title` currently looks like this if you're writing tests for `LazyFormatter`. I had to change it so that it could supply a default value.
```
def create_title(plugin=None):
    if args.title and plugin:
        _title = _author = _category = _game = None
        
        title = LazyFormatter.format(maybe_encode(args.title),
                                     title=[plugin.get_title(),DEFAULT_STREAM_METADATA["title"]],
                                     author=[plugin.get_author(),DEFAULT_STREAM_METADATA["author"]],
                                     category=[plugin.get_category(),DEFAULT_STREAM_METADATA["category"]],
                                     game=[plugin.get_category(),DEFAULT_STREAM_METADATA["game"]])
    else:
        title = args.url
    return title
```

src/streamlink_cli/main.py
```diff
@@ -10,6 +10,8 @@
 import signal
 import webbrowser
 
+from random import randint
+
```
Hubcapp: forgot to scrub this

src/streamlink_cli/main.py
beardypig: Unfortunately this defeats the purpose of the `LazyFormatter`. 
What you have now is the same as 
```python
maybe_encode(args.title).format(
  title=plugin.get_title() or DEFAULT_STREAM_METADATA["title"],
  author=plugin.get_author() or DEFAULT_STREAM_METADATA["author"],
  ...
)
```
ie. The methods always get called.

Hubcapp: Hmm, okay. I did notice that just putting each function in an array seemed to call it.

DEFAULT_STREAM_METADATA must be available at the time of formatting the final title, so I'll import it inside LazyFormatter instead of supplying it here.

beardypig: To have a default value use the original `LazyFormatter` I proposed,

```python
title = LazyFormatter.format(maybe_encode(args.title),
                             title=lambda: plugin.get_title() or DEFAULT_STREAM_METADATA["title"],
                             author=lambda: plugin.get_author() or DEFAULT_STREAM_METADATA["author"],
                             category=lambda: plugin.get_category() or DEFAULT_STREAM_METADATA["category"],
                             game=lambda: plugin.get_category() or DEFAULT_STREAM_METADATA["game"])
```

That way the methods are only called when needed.

Hubcapp: ok, thanks. :)

beardypig: I have updated the `LazyFormatter` slightly, it not allows the format string to contain `{fmt}`. 

https://gist.github.com/beardypig/14da8dc260cb249b3c0d7508130b1d24

All the tests are included too, so you can include that file in the `tests` directory and just move `LazyFormatter` out to utils.

tests/test_lazy_formatter.py
```diff
@@ -0,0 +1,81 @@
+import unittest
+from mock import MagicMock
+from streamlink.utils import *
```
beardypig: Don't like to nit-pick, but `import *` is a bad habit and is best avoided unless done for a very specific reason :)

Hubcapp: I used it because I have no knowledge of how to test your tests, and I saw "from streamlink.utils import *" in a different test file.

beardypig: If you have the dev dependencies installed (`pip install -r dev-requirements.txt`) you can just run `pytest tests` or `pytests tests/test_lazy_formatter.py`

src/streamlink_cli/utils/__init__.py
```diff
@@ -7,10 +8,11 @@
 from .progress import progress
 from .player import find_default_player
 from .stream import stream_to_url
+from .lazy_formatter import *
```
beardypig: Should avoid `*` here too. In this case it will also import `string` - which you probably didn't expect...

src/streamlink_cli/utils/lazy_formatter.py
```diff
@@ -0,0 +1,21 @@
+import string
```
beardypig: This is in the wrong module :) It should be in `src/streamlink/utils/lazy_formatter.py`.

tests/test_lazy_formatter.py
Hubcapp: wow, great. :) :) maybe I can try to fix the tests then. Unless you were looking forward to it... I can at least make a start at it now.

I'm not sure why appveyor is complaining for python 3.5. I think it's documentation related but I don't know why it's mad.

src/streamlink_cli/utils/__init__.py
Hubcapp: I did not expect that, no.

src/streamlink_cli/output.py
```diff
@@ -99,6 +101,20 @@ def running(self):
         sleep(0.5)
         return self.player.poll() is None
 
+    @classmethod
+    def supported_player(cls, cmd):
```
beardypig: I think this method still needs a bit of work. Need to write some extensive tests for it :)

beardypig:
@Hubcapp, if you like you could grant me access to your streamlink fork on GitHub and I can push some fixes to the tests, and add some new tests.

Hubcapp:
That would be great... I still can't figure out how to do this simple import inside the test. I'll work on figuring out how to grant you access to my fork instead...

beardypig:
Should be under [Settings -> Collaborators](https://github.com/Hubcapp/streamlink/settings/collaboration).

Hubcapp:
Thanks so much for the help.

beardypig:
I’ll push some fixes for you later, maybe a separate branch that you can pull and merge :)

src/streamlink/plugins/youtube.py
```diff
@@ -364,6 +412,68 @@ def _get_streams(self):
                               "try youtube-dl instead")
 
         return streams
+    
+    def handle_403(self, url, query):
+        res = http.get(url, params=query, raise_for_status=False)
+        if res.status_code == codes.ok:
+            return res
+        else:
+            while res.status_code == codes.forbidden:
```
beardypig: This could loop for quite a while, it might be nice to have some retries here and a graceful fallback.

```python
def handle_403(self, url, query, retries=5):
    for _ in range(retries):
        res = http.get(url, params=query, raise_for_status=False)
        if res.status_code == codes.ok:
            return res
        elif res.status_code == codes.forbidden:
            self.logger.debug("YouTube says \"Forbidden\". Retrying api call to {0} in 1 second", url)
            sleep (1)
        else:
            # other cases? break? retry?
```

Then in `set_title_info` you check to see if it returns `None`, log a warning saying that the title information was not available at this time, and continue on.

src/streamlink_cli/output.py
Hubcapp: I think this review is outdated now. I saw tests for the support_player method in your PR.

src/streamlink/plugins/youtube.py
back-to: I would just set it to **None** on the first fail instead of this

src/streamlink/plugins/youtube.py
```diff
@@ -288,6 +334,8 @@ def _get_stream_info(self, url):
                 query_info = dict(parse_qsl(urlparse(url).query))
                 if "channel" in query_info:
                     video_id = self._get_channel_video(query_info["channel"])
+        
+        self.video_id = video_id
```
back-to: **This was only meant for testing,**

If you call `set_title_info` without `_get_streams()`, you will get a _NameError_

This needs an update when the **main feature is done**.

src/streamlink/plugins/youtube.py
Hubcapp: That might be best, but sometimes the youtube api key is just totally broken. Doesn't matter how long I backed off, (1 second vs 10 seconds, doesn't matter because other users are using the same API key). It just "wouldn't work" until I spammed the call enough times with handle_403. It gets very hard to be lucky enough times in a row to successfully get {title} {author} and {category} all at the same time, since each one is a separate api call.

Hubcapp: Maybe Streamlink needs more api keys to cycle through if we get a 403 error? I don't know much about their API or what it costs.

src/streamlink/plugins/youtube.py
Hubcapp: Hmmm, I'll try to make that not happen. I didn't think anyone would want to do that.

src/streamlink/plugins/youtube.py
beardypig: Maybe @back-to is right. We discussed adding more API keys but decided against it. We did change the YouTube plugin to avoid the main use of the API (searching, which costs 100 api points, I think lookups are cheap 1 or 2 points), but most people are not using that version yet. But, we could change the API key, so that people using the new version are not penalised.

Hubcapp: Changing the API key sounds like a good idea. Was it a ToS issue that prevented adding a rotation of API keys?

beardypig: No one wanted to sign up the keys, you need to give a phone number iirc. Plus, it wasn’t needed after the updated plugin. If people upgrade to the latest version (I think it was merged for 0.13.0) they shouldn’t have any more 403 issues. For these API calls I think they use leas quota so there would need to be 30x as many users to cause the same problem (assuming we change the key).

src/streamlink_cli/output.py
beardypig: Yep it is.

src/streamlink/plugins/youtube.py
beardypig: I don’t think we want to go the API key rotation route, as I said before it shouldn’t be an issue any more. We can open a new issue to discuss changing the API key for YouTube.

Hubcapp: I'll just change handle_403 to not retry as @back-to suggested. Will be renamed to handle other errors as you suggested.

Hubcapp:
I've tested on linux py2 now, and added tests for mpv title with $> in it. Thanks again to @beardypig for getting me started on tests. Wouldn't have caught some bugs in this implementation without them.

Hubcapp:
I think the only issue I know about now is that the Documentation needs to be updated.

src/streamlink_cli/output.py
```diff
@@ -140,8 +140,47 @@ def _create_arguments(self):
 
             # mpv
             if self.player == "mpv":
-                # see https://mpv.io/manual/stable/#property-expansion, allow escaping with \$
-                self.title = self.title.replace("$", "$$").replace("\$$", "$")
+                # see https://mpv.io/manual/stable/#property-expansion, allow escaping with \$, respect mpv's $>
+                ##mpv has a "disable property-expansion" token which must be handled in order to accurately represent $$ in title
+                if '\$>' in self.title:
```
beardypig: Need some tests for this :)

Hubcapp: there are some! if you want to write more, that'd be cool. :) not sure what more I would add.

beardypig: True, I didn't see them hiding there. I would probably separate this in to a separate method as it is quite complex, would make it easier to test standalone and possibly refactor.

Hubcapp: It could be put into a separate function, yeah, as there are a lot of branches and complexity to disabling $ doubling partway through a string depending on how many leading $ are in front of a > & still replacing \$. I tried a different implementation before parsing the string one character at a time, but it failed on some tests and actually wasn't much "better to read"/"more maintainable" than this version... :)

What file should a refactored "mpv format title string" method go into?

beardypig: I would keep it as a method of `Player` for now, `_mpv_title_escape` or something. Later I was thinking we could refactor `Player` so that subclasses are possible, like `PlayerMPV` and `PlayerVLC` to handle the specifics of those players. But that would like require a factory method for `Player` or some `__new__` shenanigans.

beardypig: Are the idea of these transformations to stop Twitch channels from putting formatting commands in the titles on their streams? Like if I set the title of my stream to 'You're using MPV v${mpv-version}', it wouldn't render the MPV version, but the literal string?

Hubcapp: There's really no way to stop channels from putting format commands in the titles of their streams, unless you want to look for them and strip them before LazyFormatter puts them into PlayerOutput's self.title. The format codes are harmless and can't execute naughty code, it could just cause a desync of what is displayed in webbrowser at twitch.tv and in streamlink's window title. Because there's no motivation for evil hackers to insert format codes, and because streamers must put the substring "\$" in their title to activate them, it's more likely a "feature" than a "bug" that streamers could put player specific format codes in their stream title.

The idea of `_mpv_title_escape' is to render multiple `$` in a row correctly, to reduce accidentally triggering format codes, and to enable triggering format codes when desired. VLC can do it in a one-liner, but mpv has that annoying "disable format codes from this point on" token.

beardypig: The only reason I asked it because I couldn't get it to work ... but now it does ;)

beardypig: `--title '{title} (\${{mpv-version}})'` works as expected with mpv :)

Hubcapp:
Currently there are no issues known to me. Waiting for more review (particularly if the YouTube plugin modifications meet @back-to's standard).

If @bastimeyer or @gravyboat would like to review, I'd appreciate it.

Hubcapp:
Thanks for the patch, it's been applied. @back-to

beardypig:
FYI @Hubcapp, you could have downloaded [@back-to's patch](https://github.com/back-to/streamlink/commit/8a5b55a32767a838c8ec5f0c35aaaf84fd4c9c3f.patch) and applied it directly using `git am <patchfile>`. That would have set @back-to as the author etc. :)

Hubcapp:
ah, sorry. I just downloaded it and applied with `git apply <patchfile>`

src/streamlink/plugins/youtube.py
```diff
@@ -365,5 +419,75 @@ def _get_streams(self):
...
+
+    def get_category(self):
+        if self.category is None:
+            self.logger.debug("get_category")
+            if self.category_id is None:
+                self.set_title_info()
+            # translate from category_id to category name
+            query = {
+                "part": "snippet",
+                "id": self.category_id,
```
beardypig: `self.category_id` could still be `None` here (same goes for `self.channel_id` in `get_author`.)

Should check for `None` and `return` if it is still `None` after the request.

Hubcapp: I see what you mean. Will do.

beardypig:
I think this is starting to look pretty good - I thin we should be able to get it merged soon.

Hubcapp:
Yeah, I think it looks really solid, and unlike #623, there's nothing I'm dissatisfied with. Thanks again for your substantial help making this PR possible. @beardypig

back-to:
it is probably easier to remove the changes in
`src/streamlink/plugins/youtube.py`

and use a **different PR** for it when this is already merged
as there are changes in the plugin which breaks this version https://github.com/streamlink/streamlink/pull/1888

and the `self.video_id` must be changed in the other future PR as it looks a bit ugly

Hubcapp:
I can remove YouTube support from this PR if desired, but I can't write `self.video_id` to be "less ugly", since I don't know what's ugly about it.

beardypig:
> it is probably easier to remove the changes in src/streamlink/plugins/youtube.py

I say the same for `twitch`, better to have these changes in separate PRs. 

FYI, the `maybe_{de,en}code` methods already got added :)

Hubcapp:
Okay, only 15 files changed now, instead of 17, and new merge conflicts are solved. :+1:

back-to:
it is currently not implemented for `--player-passthrough`

https://github.com/Hubcapp/streamlink/blob/9b1152eb1754fc22b854df285647e23cb48fcfc8/src/streamlink_cli/main.py#L227-L234

https://streamlink.github.io/cli.html#cmdoption-player-passthrough

```
streamlink https://www.youtube.com/user/france24 --title 'ABC {title} 123' --player-passthrough hls -p vlc
```

maybe it should be added there too

Hubcapp:
Ah, good catch. After grepping the project for other `PlayerOutput(` occurrences, I put `title=title` for `output_stream_http` as well.

src/streamlink_cli/output.py
```diff
@@ -99,6 +103,62 @@ def running(self):
...
+        cmd = os.path.basename(cmd.lower())
+        for player, possiblecmds in SUPPORTED_PLAYERS.items():
+            for possiblecmd in possiblecmds:
+                if cmd.startswith(possiblecmd):
+                    return player
+
+    @classmethod
+    def _mpv_title_escape(cls, title_string):
+        # mpv has a "disable property-expansion" token which must be handled in order to accurately represent $$ in title
+        if '\$>' in title_string:
```
back-to: `W605 invalid escape sequence '\$'`

could you use `r''` or `\\`

src/streamlink_cli/output.py
```diff
@@ -99,6 +103,62 @@ def running(self):
...
+            processed_title = ""
+            double_dollars = True
+            i = dollars = 0
+            while i < len(title_string):
+                if double_dollars:
+                    if title_string[i] == "\\":
+                        if title_string[i+1] == "$":
```
back-to: `E226 missing whitespace around arithmetic operator`

could you use `[i + 1]`

src/streamlink_cli/output.py
```diff
@@ -99,6 +103,62 @@ def running(self):
...
+                            processed_title += "$"
+                            dollars += 1
+                            i += 1
+                            if title_string[i+1] == ">" and dollars % 2 == 1:
```
back-to: `E226 missing whitespace around arithmetic operator`

could you use `[i + 1]`

src/streamlink_cli/output.py
```diff
@@ -99,6 +103,62 @@ def running(self):
...
+                                i += 1
+                        else:
+                            processed_title += "\\"
+                    elif title_string[i] == "$":
+                        processed_title += "$$"
+                    else:
+                        dollars = 0
+                        processed_title += title_string[i]
+                else:
+                    if title_string[i:i+2] == "\\$":
```
back-to: `E226 missing whitespace around arithmetic operator`

could you use `[i:i + 2]`

src/streamlink_cli/output.py
```diff
@@ -99,6 +103,62 @@ def running(self):
...
+                        processed_title += "$"
+                        i += 1
+                    else:
+                        processed_title += title_string[i]
+                i += 1
+            return processed_title
+        else:
+            # not possible for property-expansion to be disabled, happy days
+            return title_string.replace("$", "$$").replace("\$$", "$")
```
back-to: `W605 invalid escape sequence '\$'`

could you use `r''` or `\\`

src/streamlink_cli/output.py
```diff
@@ -108,13 +168,30 @@ def _create_arguments(self):
...
-
         args = self.args.format(filename=filename)
         cmd = self.cmd
+        extra_args = []
+
+        if self.title is not None:
+            # vlc
+            if self.player_name == "vlc":
+                # see https://wiki.videolan.org/Documentation:Format_String/, allow escaping with \$
+                self.title = self.title.replace("$", "$$").replace("\$$", "$")
```
back-to: `W605 invalid escape sequence '\$'`

could you use `r''` or `\\`

tests/test_cmdline.py
```diff
@@ -3,10 +3,14 @@
 from tests.mock import patch, ANY
 
 import os.path
+import unittest
```
back-to: `unittest` is already imported, can be removed

tests/test_cmdline.py
```diff
@@ -3,10 +3,14 @@
...
+
 import streamlink_cli.main
 
+
 from streamlink import Streamlink
 from streamlink_cli.compat import is_win32
+from tests.mock import patch, ANY
```
back-to: `patch, ANY` is already imported, can be removed

tests/test_output.py
```diff
@@ -0,0 +1,121 @@
...
+                         PlayerOutput.supported_player("C:\\mpc\\mpc-hd.exe"))
+        self.assertEqual(None,
+                         PlayerOutput.supported_player("C:\\mplayer\\not-vlc.exe"))
+
+    def test_open_player_with_title_mpv_escape_1(self):
+        self.assertEqual(PlayerOutput._mpv_title_escape("no escape $$ codes $"),
+                         'no escape $$$$ codes $$')
+
+    def test_open_player_with_title_mpv_escape_2(self):
+        self.assertEqual(PlayerOutput._mpv_title_escape("\$> begins with escape code $$"),
```
back-to: same in this file use `r''` or `\\`

```
tests/test_output.py:64:10: W605 invalid escape sequence '\$'
tests/test_output.py:68:35: W605 invalid escape sequence '\$'
tests/test_output.py:72:10: W605 invalid escape sequence '\$'
tests/test_output.py:72:47: W605 invalid escape sequence '\$'
tests/test_output.py:92:22: W605 invalid escape sequence '\$'
tests/test_output.py:92:32: W605 invalid escape sequence '\$'
tests/test_output.py:92:43: W605 invalid escape sequence '\$'
tests/test_output.py:92:53: W605 invalid escape sequence '\$'
tests/test_output.py:92:57: W605 invalid escape sequence '\$'
tests/test_output.py:92:61: W605 invalid escape sequence '\$'
tests/test_output.py:92:64: W605 invalid escape sequence '\$'
tests/test_output.py:96:25: W605 invalid escape sequence '\$'
tests/test_output.py:100:25: W605 invalid escape sequence '\$'
tests/test_output.py:100:27: W605 invalid escape sequence '\$'
tests/test_output.py:112:24: W605 invalid escape sequence '\$'
tests/test_output.py:112:31: W605 invalid escape sequence '\$'
tests/test_output.py:112:33: W605 invalid escape sequence '\$'
tests/test_output.py:116:23: W605 invalid escape sequence '\$'
tests/test_output.py:116:25: W605 invalid escape sequence '\$'
tests/test_output.py:116:33: W605 invalid escape sequence '\$'
```

src/streamlink_cli/argparser.py
```diff
@@ -481,6 +481,74 @@ def build_parser():
         This option will instead let the player decide when to exit.
         """
     )
+    player.add_argument(
+        "-t", "--title",
+        metavar="TITLE",
+        help="""
+        This option allows you to supply a title to be displayed in the 
+        title bar of the window that the video player is launched in.
+
```
back-to: could you remove the not used whitespace

`W293 blank line contains whitespace`

line

490
496
501
504
507
512
516
532
535
542
546

src/streamlink_cli/argparser.py
```diff
@@ -481,6 +481,74 @@ def build_parser():
...
+            A full list of the format codes VLC uses is available here:
+            https://wiki.videolan.org/Documentation:Format_String/
+            
+        mpv specific information:
+            mpv has certain codes you can use inside your title.
+            These are accessible inside --title by using a backslash 
+            before the dollar sign mpv uses to denote a format character.
+            
+            e.g. to put the current version of mpv running inside your
+            mpv window title, the string "\${{{{mpv-version}}}}" could be
```
back-to: could you use `\\$`

src/streamlink_cli/argparser.py
```diff
@@ -481,6 +481,74 @@ def build_parser():
...
+
+        This option is only supported for the following players: {0}.
+        
+        VLC specific information:
+            VLC has certain codes you can use inside your title.
+            These are accessible inside --title by using a backslash 
+            before the dollar sign VLC uses to denote a format character.
+            
+            e.g. to put the current date in your VLC window title,
+            the string "\$A" could be inserted inside your --title string.
```
back-to: could you use `\\$`

src/streamlink_cli/argparser.py
```diff
@@ -481,6 +481,74 @@ def build_parser():
...
+            
+            Otherwise, it is the string "{3}"
+
+        {{game}}
+            This is just a synonym for {{category}} which may make more sense for
+            gaming oriented platforms. "Game being played" is a way to categorize
+            the stream, so it doesn't need its own separate handling.
+            
+       Examples:
+           %(prog)s -p vlc --title '{{title}} -!- {{author}} -!- {{category}} \$A' <url> [stream]
```
back-to: could you use `\\$`

src/streamlink_cli/argparser.py
```diff
@@ -481,6 +481,74 @@ def build_parser():
...
+           %(prog)s -p mpv --title "{{title}} -- {{author}} -- {{category}} -- (\${{{{mpv-version}}}})" <url> [stream]
```
back-to: could you use `\\$`

src/streamlink_cli/argparser.py
```diff
@@ -481,6 +481,74 @@ def build_parser():
...
+            
+            e.g. to put the current date in your VLC window title,
+            the string "\$A" could be inserted inside your --title string.
+            
+            A full list of the format codes VLC uses is available here:
+            https://wiki.videolan.org/Documentation:Format_String/
+            
+        mpv specific information:
+            mpv has certain codes you can use inside your title.
+            These are accessible inside --title by using a backslash
```
back-to: could you remove the whitespace after **backslash**

`W291 trailing whitespace`

src/streamlink_cli/argparser.py
```diff
@@ -481,6 +481,74 @@ def build_parser():
...
+        
+        This value can contain formatting variables surrounded by curly braces,
+        {{ and }}. If you need to include a brace character, it can be escaped
+        by doubling, e.g. {{{{ and }}}}.
+
+        This option is only supported for the following players: {0}.
+        
+        VLC specific information:
+            VLC has certain codes you can use inside your title.
+            These are accessible inside --title by using a backslash
```
back-to: could you remove the whitespace after backslash

`W291 trailing whitespace`

src/streamlink_cli/argparser.py
```diff
@@ -481,6 +481,74 @@ def build_parser():
         This option will instead let the player decide when to exit.
         """
     )
+    player.add_argument(
+        "-t", "--title",
+        metavar="TITLE",
+        help="""
+        This option allows you to supply a title to be displayed in the
```
back-to: could you remove the whitespace after the

`W291 trailing whitespace`

tests/test_cmdline.py
```diff
@@ -51,72 +56,69 @@ def fn(*args):
...
-    @unittest.skipIf(is_win32, "test only applicable in a POSIX OS")
     def test_open_space_path_player(self):
-        self._test_args(["streamlink", "-p", "\"/Applications/Video Player/VLC/vlc\"", "http://test.se", "test"],
-                        ["/Applications/Video Player/VLC/vlc", "-"])
+        self._test_args(["streamlink", "-p", "\"/Applications/Video Player/player\"", "http://test.se", "test"],
+                        ["/Applications/Video Player/player", "-"])
         # escaped
-        self._test_args(["streamlink", "-p", "/Applications/Video\ Player/VLC/vlc", "http://test.se", "test"],
-                        ["/Applications/Video Player/VLC/vlc", "-"])
+        self._test_args(["streamlink", "-p", "/Applications/Video\ Player/player", "http://test.se", "test"],
```
back-to: could you use `Video\\ Player`

src/streamlink_cli/main.py
```diff
@@ -27,12 +27,14 @@
 from streamlink.stream import StreamProcess
 from streamlink.plugins.twitch import TWITCH_CLIENT_ID
 from streamlink.plugin import PluginOptions
+from streamlink.utils import LazyFormatter
 
 import streamlink.logger as logger
 from .argparser import build_parser
-from .compat import stdout, is_win32
+from .compat import stdout, is_win32, is_py2, is_py3
```
back-to: `is_py2, is_py3` is not used, can be removed

src/streamlink_cli/main.py
```diff
@@ -125,6 +129,16 @@ def create_http_server(host=None, port=0):
 
     return http
 
+def create_title(plugin=None):
+    if args.title and plugin:
+        title = LazyFormatter.format(maybe_encode(args.title),
```
back-to: could you format it like

```py
def create_title(plugin=None):
    if args.title and plugin:
        title = LazyFormatter.format(
            maybe_encode(args.title),
            title=lambda: plugin.get_title() or DEFAULT_STREAM_METADATA["title"],
            author=lambda: plugin.get_author() or DEFAULT_STREAM_METADATA["author"],
            category=lambda: plugin.get_category() or DEFAULT_STREAM_METADATA["category"],
            game=lambda: plugin.get_category() or DEFAULT_STREAM_METADATA["game"]
        )
    else:
        title = args.url
    return title
```

`E128 continuation line under-indented for visual indent`

src/streamlink_cli/main.py
```diff
@@ -125,6 +129,18 @@ def create_http_server(host=None, port=0):
...
+    if args.title and plugin:
+        title = LazyFormatter.format(
+            maybe_encode(args.title),
+            title=lambda: plugin.get_title() or DEFAULT_STREAM_METADATA["title"],
+            author=lambda: plugin.get_author() or DEFAULT_STREAM_METADATA["author"],
+            category=lambda: plugin.get_category() or DEFAULT_STREAM_METADATA["category"],
+            game=lambda: plugin.get_category() or DEFAULT_STREAM_METADATA["game"]
+        )
+    else:
+        title = args.url
```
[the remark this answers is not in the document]
Hubcapp: Oh, I see what you're saying.

I thought args.url would only have the url opened when I wrote that, so to conform with that expected behavior, you would like an args.url of `hls://https://nasa-i.akamaihd.net/hls/live/253565/NASA-NTV1-Public/master.m3u8 name_key='bitrate'` to be converted to `hls://https://nasa-i.akamaihd.net/hls/live/253565/NASA-NTV1-Public/master.m3u8`.

Which you say should be just doing `title = args.url.split(' ')[0]`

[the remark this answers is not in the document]
Hubcapp: Doing that might not be acceptable (URLs can have spaces in them if they're not converted to %20, although), so if this is considered a blocker, we can just use the current behavior of changing the title to " - ".

I just thought args.url would be more helpful than " - ".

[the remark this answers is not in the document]
Hubcapp: Possibly the better fix is to rewrite the use of `args.url`to not put those params in it. I would want that done as a separate pull request, by someone who is not me. So for this PR, it should remain `args.url` or be " - ".

src/streamlink_cli/main.py
```diff
@@ -101,11 +103,13 @@ def create_output():
         elif args.player_http:
             http = create_http_server()
 
+        title = create_title(streamlink.resolve_url(args.url))
```
back-to: currently the `first plugin` is used for the **title**,
but if you use embedded plugins it might not be the best title

**example**

https://www.cnews.fr/le-direct

it uses **dailymotion** as 2nd plugin

---

here is a patch which **would use the last plugin**,
removes also the rerun of `resolve_url`

https://gist.github.com/back-to/699722cc180769d11579630c3e256e37

Hubcapp: Respond again if the patch is ready to be applied. A bit confused by the "not quite possible how I wanted it to work" comment.

I'll apply with `git am` when it's ready this time. :)

back-to: the patch just removes the reusage of `streamlink.resolve_url`

as it was already used here
https://github.com/Hubcapp/streamlink/blob/6a61f70c44ab1804705c3693fa5c50f8a1ec37e1/src/streamlink_cli/main.py#L527
so you don't run the same script twice

**this part is ready**

---

what I wanted is to get an updated `plugin`
but the streamlink session won't update it automatically

but that is **not related to this PR**
as it would requiere some changes to the session part of streamlink

src/streamlink_cli/main.py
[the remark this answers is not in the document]
back-to: 1. I rethinked it, the builtin plugin will never be called without `--title`
so it will always be `args.url` not `self.title`
`plugin.url` could be used but it won't do any big impact, 
 because `self.url` is changed very rarely

2. maybe it is better not to use a default title,
   some third party apps might break (already uses some sort of own `--title`)

3. something else could be used instead of `DEFAULT_STREAM_METADATA["title"]`,
   maybe `plugin.url` so it won't show up as Unknown

this is my opinion, maybe there are also some other solutions.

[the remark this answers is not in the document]
Hubcapp: 2. My opinion is that software needs to be free to make changes that make it better, without worrying about hypothetical 3rd party software incompatibilities. If it's a blocker, then I don't care that much, but the URL of the stream is a much more useful default than " - " in my opinion.

3. Why? If the title is unknown, then it's accurate. It indicates that the plugin used does not have a title getting method in it. The title is not the URL. This will just lead to unwanted behavior by users who were pleased by the behavior of getting the URL when using {title} and then being upset that it's changed once the feature is implemented for their plugin. Possibly you want a `{url}` format code, but I don't want to implement anything else on this PR that could block it.

Also, I'm going on vacation now until August. I'll still accept PRs but won't be working on this at least until then.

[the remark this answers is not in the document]
back-to: > If it's a blocker, then I don't care that much,

it is not really a blocker, just some opinion from me it can be used like it is.

> DEFAULT_STREAM_METADATA["title"]

true, I did not think about the `"go away" once someone implements the feature for the plugin.`

gravyboat:
Really glad to finally see this in. Thanks to everyone for their contributions and especially @Hubcapp!

beardypig:
Congratulations @Hubcapp

Hubcapp:
Yayayayayayayay

Still on vacation but yay

The YouTube and twitch diffa are still above, feel free anyone out there to work on them, I can't make prs until later

Issue #1187, opened 2017-08-25
vaughnlive plugin not working..


JourneyOver:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

Vaughnlive streams are not working

### Expected / Actual behavior

...

### Reproduction steps / Explicit stream URLs to test

seems to be a replica of the closed issue https://github.com/streamlink/streamlink/issues/1034

Every stream just comes back with "error: No playable streams found" 

### Environment details

Operating system and version: Windows 10 x64
Streamlink and Python version: 0.7.0 ( and the latest dev buld from https://dl.bintray.com/streamlink/streamlink-nightly/ ) 

### Comments, logs, screenshots, etc.

This is happening even on the latest night/dev build
Doing "-l debug" doesn't seem to produce anything at all either, just throws the same error: no playable streams found even though the streams are working just fine in browser..

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

beardypig:
@JourneyOver which channel? I tested a few and they still work for me ...

JourneyOver:
I tested about 5-10 of them, all in the misc section, but as you say they are working now..so idk what was going on with vaughnlive when I was trying to watch something earlier..

gravyboat:
@beardypig I'm waiting for the 500th change where we decide to stop supporting it (especially considering a lot of it is stolen content).. Closing since we can't duplicate.

mmetak:
I tried a couple of channels and I get this:

```
$ streamlink -l debug https://vaughnlive.tv/atlantis4 worst
[cli][info] Found matching plugin vaughnlive for URL https://vaughnlive.tv/atlantis4
error: No playable streams found on this URL: https://vaughnlive.tv/atlantis4
```

Streams play in browser though. Maybe they changed something upstream again?

fadster:
Seems changes have been made to the site today. The plugin was working this morning but all of a sudden I'm getting "No playable streams found" on all channels. Should I open a new issue?

gravyboat:
@fadster No let's just open this one back up, there are already 25 vaughnlive issues...

mmetak:
I get the "No playable streams found" after successfully opening and playing two or three times.

gravyboat:
[...]

Issue #1034, opened 2017-06-24
vaughnlive plugin is buggy


bobvargas:
### Checklist

- [x] This is a bug report.
- [x] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description

vaughnlive streams may not work all the time as it shows not found

### Expected / Actual behavior

...

### Reproduction steps / Explicit stream URLs to test

1. vaughnlive.tv/sherming66

### Environment details

Operating system and version: Windows 7
Streamlink and Python version: 0.6.0

beardypig:
Could you please include what is printed out on the console when it does not work?

bobvargas:
it prints out connection failed message sometimes when I open a vaughnlive stream.

beardypig:
When it happens again can you copy the full message here, and if possible use the `-l debug` option as that will give us a much better idea of what is going wrong.

bobvargas:
ok thanks. @beardypig

techNeffect:
Vaughnlive recently switched to a HTML player from Flash, so that could be part of the issue. It says "no playable streams"

I got no extra meaningful output by applying the debug flag, but I could have done it incorrectly.

![screenshot001](https://user-images.githubusercontent.com/12178152/27619691-4f6391d0-5b92-11e7-8cda-5b1a282c9f96.jpg)

beardypig:
Ah OK, I didn't know that - I haven't actually been able to check the website. Sounds like we might need a rewrite of the plugin if they have finally stopped using flash!

techNeffect:
Seems they are still broadcasting a flash stream. If one appends /old to any Vaughnlive URL it'll access a flash version of the stream that'll work with Streamlink. Don't know how long it'll work.

RedPenguin2:
I can attest to techNeffect's comment. If you attach /old to the URL, it works immediately but fails with the mentioned above error if you don't.

sync5:
since few moments, /old urls are now disabled on Vaughn

EFEVE:
For me, it stopped working yesterday. This is the message I'm receiving. Curiously, even when I'm not using the '/old' appendix, the website still uses a Flash Player (on Firefox, at least):

```
C:\>streamlink vaughnlive.tv/erobtv/old live
[cli][info] Found matching plugin vaughnlive for URL vaughnlive.tv/erobtv/old
[...]

PR #1583, opened 2018-03-29
plugins.vaughnlive: Updated server map

### LINKED ISSUE #1456: Vaughnlive changed IP's to break Streamlink
This will be a very brief bug report... As of tonight the head vaughnlive.py references IPs which were disconnected by vaughn to thwart streamlinking. I've observed vaughn serving video now from "66.90.93.44","66.90.93.35" and have personally gotten it to work overwriting the IP's in rtmp_server_map with those two alternating. I would submit the commit but I think some more testing is needed as I only use streamlink with one occasional stream and don't know how far those IPs will get more frequent SL users.

 #1187 contains lengthy discussion on the history of the war vaughn has waged against streamlink, this is probably not the last time the IPs will change.

beardypig:
Bit slow in creating this PR... fixes #1456.

### DIFF
--- a/src/streamlink/plugins/vaughnlive.py
+++ b/src/streamlink/plugins/vaughnlive.py
@@ -38,10 +38,10 @@ class VaughnLive(Plugin):
                                                                                                    range(1, 6))]
     origin = "https://vaughnlive.tv"
     rtmp_server_map = {
-        "594140c69edad": "66.90.93.42",
-        "585c4cab1bef1": "66.90.93.34",
-        "5940d648b3929": "66.90.93.42",
-        "5941854b39bc4": "198.255.0.10"
+        "594140c69edad": "192.240.105.171:1935",
+        "585c4cab1bef1": "192.240.105.171:1935",
+        "5940d648b3929": "192.240.105.171:1935",
+        "5941854b39bc4": "192.240.105.171:1935"
     }
     name_remap = {"#vl": "live", "#btv": "btv", "#pt": "pt", "#igb": "instagib", "#vtv": "vtv"}
     domain_map = {"vaughnlive": "#vl", "breakers": "#btv", "instagib": "#igb", "vapers": "#vtv", "pearltime": "#pt"}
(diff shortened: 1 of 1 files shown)


gravyboat:
@beardypig I think vaughnlive should be added to our potential list of chopping block plugins. It's mostly stolen content and maintaining the plugin is a pain in the butt.

fadster:
@gravyboat I must disagree. A vast majority of the content on that site and its secondary sites is legitimate. Any channel broadcasting "stolen" content is promptly shutdown by the site owners as they are very responsive to DMCA complaints. This is a very popular plugin and once the site stabilizes after the transition it's going through, keeping it updated shouldn't be too much of an issue.

gravyboat:
@fadster It would be nice to go onto the site for once when this comes up and to not see half a dozen TV shows, anime, UFC fights, and various news stations being rebroadcast but that really isn't the main issue, it's just a convenient secondary concern. The site maintainer purposefully changes configuration to try and block us (https://github.com/streamlink/streamlink/issues/1034#issuecomment-318326584), and you have also commented previously on the constantly changing configuration (https://github.com/streamlink/streamlink/issues/1187#issuecomment-333080247), how many years of configuration changes are okay? I probably should have eliminated the concern of illegal content from my comment entirely and focused on the annoyance of maintenance this plugin creates for us, but I'm not going to edit my comment for posterity's sake.

fadster:
@gravyboat I'm sure you're aware of the reasons for using Streamlink to access the site. The built-in player has been notoriously buggy. Among other problems that have been plaguing it with every update, it doesn't handle buffering very well and forces a 16:9 aspect ratio on all streams, distorting the video in many cases. An external player such as vlc provides controls that resolve these issues entirely.

Regarding the shifting configurations, I believe they are mostly due to changes in hosting platforms, server software and data center locations, in particular over the past few months to accommodate the migration to HTML5. As such, I doubt they are explicitly targeted at blocking Streamlink. I'm confident the plugin won't require frequent updates once the site stabilizes.

gravyboat:
@fadster Yeah their player is... yeesh. I'm going to merge this, if the configuration keeps being swapped around for yet another year we'll have to reconsider support but we'll push it off till then.

PR #1591, opened 2018-04-04
stream.hls: change --hls-audio-select to take a list and wildcard

mattiasa:
This changes the --hls-audio-select flag to take a comma-separated
list of audio source names or language codes. This will cause the
ffmpeg muxer to include all selected audio tracks in the output
stream. It also changes so the option takes an asterisk character to
select all available audio tracks.

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -289,7 +289,13 @@ class MuxedHLSStream(MuxedStream):
     __shortname__ = "hls-multi"
 
     def __init__(self, session, video, audio, force_restart=False, ffmpeg_options=None, **args):
-        substreams = map(lambda url: HLSStream(session, url, force_restart=force_restart, **args), [video, audio])
+        tracks = [video]
+        if audio:
+            if isinstance(audio, list):
+                tracks.extend(audio)
+            else:
+                tracks.append(audio)
+        substreams = map(lambda url: HLSStream(session, url, force_restart=force_restart, **args), tracks)
         ffmpeg_options = ffmpeg_options or {}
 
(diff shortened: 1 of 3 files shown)


beardypig:
Looks good, I cannot test it right now but I cannot see any obvious problems. Thanks @mattiasa 🗡 
Lemme test it out later today, or tomorrow morning - do you have any publicly access sources that we can test with?

mattiasa:
You can use the MLB.tv free game of the day. You still have to register for an account, but it seems the stream is available without a paid subscription. If you want to do that you can go to http://mlb.com and register for an account in the upper right-hand corner. The stream is encrypted so the easiest way to get the stream is to use https://github.com/tonycpsu/mlbstreamer

Install that in the same virtual env as streamlink, then run:

```
mlbplay  --init-config
```

and enter your mlb.com credentials. You can use /bin/true for player. Then run:

```
mlbplay -s . -d 2018-04-04 -v hou
```

to get last night's free game. That will print streamlink command including the Authorization header needed for the stream. With `--hls-audio-select *` that downloads all the audio streams.

mattiasa:
I forgot to mention that because of a bug in ffmpeg you have to run the latest master branch of ffmpeg with https://patchwork.ffmpeg.org/patch/8333/ applied to actually get the additional audio streams from mlb.com. It should work on streams which don't have the weird duplicated ID3 tags in them.

mattiasa:
It does look like this test stream from https://github.com/mangui/flashls/issues/538 can be used to test this too: http://sample.vodobox.com/planete_interdite/planete_interdite_alternate.m3u8

beardypig:
@mattiasa thanks for the public source without login :) It does appear to be working as expected :)

gravyboat:
Good stuff, thanks for the PR @mattiasa, and thanks for the review @beardypig!

PR #1605, opened 2018-04-15
[build] Fixed AppVeyor build pip10 error

back-to:
```
  File "c:\python35\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\python35\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Python35\Scripts\pip.exe\__main__.py", line 5, in <module>
ImportError: cannot import name 'main'
Command exited with code 1
```

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -36,11 +36,11 @@ install:
 
   # Upgrade to the latest version of pip to avoid it displaying warnings
   # about it being out of date.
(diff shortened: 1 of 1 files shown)


beardypig:
Thanks for the fix @back-to

PR #1610, opened 2018-04-20
Filmon

Billy2011:


src/streamlink/plugins/filmon.py
```diff
@@ -134,7 +134,7 @@ def _get_streams(self):
                 yield stream["quality"], FilmOnHLS(self.session, vod_id=vod_id, quality=stream["quality"])
 
         else:
-            if not channel:
+            if not channel or channel == "sat-1-schweiz":
```
beardypig: I'd create a list of channels that should be handled differently and then check `channel in self._special_case_channels`. In case there are other in the future :)

src/streamlink/plugins/filmon.py
```diff
@@ -95,7 +95,7 @@ class Filmon(Plugin):
         )
     """, re.VERBOSE)
 
-    _channel_id_re = re.compile(r'channel_id\s*?=\s*"(\d+)"')
+    _channel_id_re = re.compile(r"""channel_id\s*?=\s*"?(\d+)["']""")
```
beardypig: You have `"` at the start of the id, but `"` or `'` at the end, should it be `["']` on both sides?

Billy2011: No it's not nessecary, the id itself isn't quoted in that case.

src/streamlink/plugins/filmon.py
Billy2011: OK, I'll do it.

src/streamlink/plugins/filmon.py
bastimeyer: What @beardypig meant was that usually when you're matching (optional) strings, you should try to match the same kind of quotation mark at the end of the string which was used at the beginning of it. The regex should look something like that:
```python
r"""channel_id\s*=\s*(?P<quote>['"]?)(?P<value>\d+)(?P=quote)"""
```
This also replaces the unnecessary `\s*?` with a simple `\s*` and gives each capture group a name, as it increments the position of the value capture group by one. This needs to get fixed in the code below...

Billy2011:
It's annoying, I can spend my time better.

gravyboat:
@Billy2011 Thanks for your work on this! I ended up merging #1620 since it had the changes noted above, thanks for pushing this to get finished.

PR #1624, opened 2018-04-26
build/deploy: do not deploy streamlink-latest, and remove old nightlies

beardypig:
This updates the Travis/BinTray deploy to stop deploying `streamlink-latest.exe` and to delete old nightly versions so we don't run out of quota on BinTray.

### DIFF
--- a/script/bintrayupdate.sh
+++ b/script/bintrayupdate.sh
@@ -1,8 +1,21 @@
 #!/usr/bin/env bash
+set -e
 # Update the Bintray release
 
+old_date=$(date +"%Y.%m.%d" -d "30 days ago")
 build_dir="$(pwd)/build"
-current_api_url="https://api.bintray.com/packages/streamlink/streamlink-nightly/streamlink/versions/$(date +'%Y.%m.%d')/release_notes"
-latest_api_url="https://api.bintray.com/packages/streamlink/streamlink-nightly/streamlink/versions/latest/release_notes"
-curl -u "${BINTRAY_USER}:${BINTRAY_KEY}" -H "Content-Type: application/json" -d @"${build_dir}/bintray-changelog.json" ${current_api_url}
-curl -u "${BINTRAY_USER}:${BINTRAY_KEY}" -H "Content-Type: application/json" -d @"${build_dir}/bintray-changelog.json" ${latest_api_url}
+api_base="https://api.bintray.com/packages/streamlink/streamlink-nightly/streamlink"
+current_api_url="${api_base}/versions/$(date +'%Y.%m.%d')/release_notes"
+files_api_url="${api_base}/files"
+versions_api_url="${api_base}/versions"
+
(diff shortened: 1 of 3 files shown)


script/bintrayupdate.sh
```diff
@@ -1,8 +1,21 @@
...
-curl -u "${BINTRAY_USER}:${BINTRAY_KEY}" -H "Content-Type: application/json" -d @"${build_dir}/bintray-changelog.json" ${latest_api_url}
+api_base="https://api.bintray.com/packages/streamlink/streamlink-nightly/streamlink"
+current_api_url="${api_base}/versions/$(date +'%Y.%m.%d')/release_notes"
+files_api_url="${api_base}/files"
+versions_api_url="${api_base}/versions"
+
+curl -u "${BINTRAY_USER}:${BINTRAY_KEY}" -H "Content-Type: application/json" -d @"${build_dir}/bintray-changelog.json" "${current_api_url}"
+
+echo "Deleting versions older than ${old_date}..."
+versions=$(curl -s -u "${BINTRAY_USER}:${BINTRAY_KEY}" "${files_api_url}" | grep -Po '(?<="version":")[^"]+(?=")')
```
bastimeyer: Does TravisCI have `jq` installed by default? Then you could replace the (unsafe) regex
`grep -Po '(?<="version":")[^"]+(?=")'`
with a simple
`jq -r '.[].version'`
that returns the version property of all objects in the json array response in a raw format (no quotations)

beardypig: I doubt it, but if there is an ubuntu package it could be added... lemme see

bastimeyer: Looks like `jq` needs to be added to the apt package list in the travisci config. Package is whitelisted, just checked it.

beardypig: Cool. I read that it's pre-installed, can test it out.

script/bintrayupdate.sh
```diff
@@ -9,10 +9,10 @@ current_api_url="${api_base}/versions/$(date +'%Y.%m.%d')/release_notes"
 files_api_url="${api_base}/files"
 versions_api_url="${api_base}/versions"
 
-curl -u "${BINTRAY_USER}:${BINTRAY_KEY}" -H "Content-Type: application/json" -d @"${build_dir}/bintray-changelog.json" "${current_api_url}"
+#curl -u "${BINTRAY_USER}:${BINTRAY_KEY}" -H "Content-Type: application/json" -d @"${build_dir}/bintray-changelog.json" "${current_api_url}"
```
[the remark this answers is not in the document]
beardypig: damn... that was for testing :(

script/bintrayupdate.sh
beardypig: `jq` is cool, thanks for the tip

gravyboat:

PR #1635, opened 2018-05-02
Spelling: Offline, URL, Logged in

comradekingu:


src/streamlink/plugins/pixiv.py
```diff
@@ -94,7 +94,7 @@ def _get_streams(self):
         if login_username and login_password:
             self.logger.debug("Attempting login as {0}".format(login_username))
             if self._login(login_username, login_password):
-                self.logger.info("Successfully logged in as {0}".format(login_username))
+                self.logger.info("Logged in as {0}".format(login_username))
```
back-to: it's just a copy and paste https://github.com/streamlink/streamlink/search?q=%22Successfully+logged+in+as+%7B0%7D%22&unscoped_q=%22Successfully+logged+in+as+%7B0%7D%22

src/streamlink/plugins/pixiv.py
```diff
@@ -56,7 +56,7 @@ def find_videopage(self):
 
         m = self._videopage_re.search(res.text)
         if not m:
-            self.logger.debug("No stream path, stream might be offline or invalid url.")
+            self.logger.debug("No stream path. Offline stream or invalid URL?")
```
back-to: because I wrote this plugin I am biased,
but this changes doesn't bring any real value.

gravyboat:
My main issue is that if anyone has any sort of automated tooling that relies on reading the debug output this would break it. In addition I don't really see these as being significantly more "correct" grammatically, nor do address an issue related to someone being confused about the output. As @back-to pointed out we also use some of that specific wording in a variety of plugins. I'll leave it open in case anyone else has feedback.

comradekingu:
@gravyboat This annoys me as a user every single day.
"Successfully" is redundant. "url" is not correct, as per https://en.wikipedia.org/wiki/URL
Overall it is shorter, which is also good.

@back-to Surely you can't be invested in this. I can correct the others too.

gravyboat:
@comradekingu I don't mind it being changed if it bothers you that much, my main concern is potential for reliance on the current log output. @bastimeyer is this going to cause any issues for the GUI if this is modified for multiple plugins?

bastimeyer:
> if this is modified for multiple plugins

Depends on what will get changed. There are a couple of error messages with "redundant" words, so if that's the motivation behind this, then a lot of other messages also have to be rewritten.

I personally think that changing log output is bad if you also don't do a major version bump, as it will simply break dependencies. There is no other way than reading and parsing log output if you're building an application around Streamlink which is not written in Python or does have proper bindings or other IPC methods.

So if you want to change already "established" log messages (which were already included in the latest stable release), the entire log output should at least be tagged with unique codes based on the class/plugin, so that other applications can reliably parse it. Currently, only the log channel, its origin and message is being logged. Changing it would still require a major version bump, but since these changes are so minor and IMO irrelevant, I don't think they should be done now and we should instead wait until Streamlink is ready for a proper 1.0.0. Maybe #1690 will make this a bit easier to change.

PR #1638, opened 2018-05-04
Plugin Arguments API

beardypig:
TLDR; new API for plugin arguments to allow side loaded plugins to add command line arguments, with ease.

I noticed recently that there have been a few plugins that have added extra command line arguments, and it has been more difficult for users to test as you cannot just side load the plugin. This is my attempt at solving that problem.

I have added two claases: `Arguments` (`PluginArguments`) and `Argument` (`PluginArgument`),
`PluginArguments` is just to provide a wrapper around a list og `PluginArgument`s, mainly for possible future use. 
The idea is that the plugin developer defines the arguments needed by the plugin (replacing the existing `PluginOptions` usage), and the front-end deals with collecting them and supplying them back to the plugin.

`PluginArguent` accepts most of the same parameters as `ArgumentParser.add_argument`, except `requires` is a special case as in this case it is only enforced if the plugin is in use. In addition the `name` parameter is the name relative to the plugin eg. `username`, `paasword`, etc. The accepted  parameters are (in addition to the `add_argument` ones.

 - `argument_name` (default=None): a name override for the argument eg. `"ustream-password`, for backwards compatibility (shouldn't be used going forward)
 - `requires=None` (default=None): a list of other arguments that this argument will require if used eg. `["password"]` for the `"username"` argument.
 - `prompt=None` (default=None): for a required argument, the prompt that will be displayed to the user. The default in the cli is `"Enter <plugin_name> <argument_name>"`
 - `sensitive=False` (default=False): if the argument contains sensitive data, and should be omitted from logs etc, and will generate and `askpass` prompt in the cli.

There are plenty of examples, and I plan on expanding the docs next week (before/if this gets merged).

This is significant change set, that touches many files - I have tried my best to test it but I imagine there will be some more bugs found :)


beardypig:
Already found a bug with `required`...

back-to:
```
streamlink "http://zattoo.com/watch/daserste" best -l debug
  --zattoo-email *** --zattoo-password ***
```

it asks for my password even when I use it.

```
[cli][debug] Python:     3.6.5
[cli][debug] Streamlink: 0.11.0
Enter zattoo password:

Interrupted! Exiting...
```

---

some plugins with different names for `_` and `-` must be updated,

example zattoo

old name `purge_credentials`

new name `purge-credentials`

https://github.com/beardypig/streamlink/blob/0c8c5ac3061a7def2778e62475bf31bae947d07d/src/streamlink/plugins/zattoo.py#L254

---

exceptions are big now

**master**

```
error: No plugin can handle URL:
```
**this pr**

```
  File "streamlink/parts/bin/streamlink", line 11, in <module>
  File "streamlink/src/streamlink_cli/main.py", line 968, in main
    setup_plugin_options(streamlink)
  File "streamlink/src/streamlink_cli/main.py", line 853, in setup_plugin_options
    plugin = streamlink.resolve_url(args.url)
  File "streamlink/src/streamlink/utils/__init__.py", line 170, in memoizer
    cache[key] = obj(*args, **kwargs)
  File "streamlink/src/streamlink/session.py", line 390, in resolve_url
    raise NoPluginError
streamlink.exceptions.NoPluginError
```

src/streamlink/plugin/plugin.py
```diff
@@ -189,6 +190,10 @@ def set_option(cls, key, value):
     def get_option(cls, key):
         return cls.options.get(key)
 
+    @classmethod
+    def get_argument(cls, key):
```
back-to: `self.arguments.get` and `self.get_argument` are
`<streamlink.options.Argument object at 0x0000000>`

shouldn't it be like `self.options.get` and `self.get_options`

src/streamlink/plugin/plugin.py
```diff
@@ -189,6 +190,10 @@ def set_option(cls, key, value):
     def get_option(cls, key):
```
back-to: when the argument is called `b-netloc`

`-` won't work in get_option `_` must be used,
well thats **backport compatibility**

but shouldn't it work with `-` now too or should `self.get_argument` be used now



```
self.get_option("b-netloc") == None
self.get_option("b_netloc") == ['test.com', 'test.test']
```

back-to:
And something  that I have noticed but **it's not important**

at https://github.com/streamlink/streamlink/blob/master/src/streamlink_cli/main.py#L615

arguments can't be called in

```
@classmethod
def can_handle_url(cls, url):
```

when `streamlink.resolve_url(args.url)` get's used the first time

`master`

get_option will be the **default** value such as **False**
but you can't change it to True or something else.
 
`this PR`

it's always None on the first call

---

when `can_handle_url` gets called again the 2nd time,
it will work normal as expected

src/streamlink/plugin/plugin.py
beardypig: I did intend it to return the actual `Argument` object, in this case `get_option` should still be used. And as you pointed out in the other comment, the name of the option/argument should be normalised :)

src/streamlink/plugin/plugin.py
beardypig: I guess the only caveat here is that the keys `b_netloc` and `b-netloc` will be equivalent - I don't think this is a problem though.

beardypig: The other option is to use the `dest` keyword argument to override the option name.

back-to:
Plugins with `_` in the name will have commands with `_`

**works** `--ard_mediathek-example`

**unrecognized arguments** `--ard-mediathek-example`

maybe this can be normalised as well, if it's possible

beardypig:
Didn't realise any plugins with `_` in the module name had arguments, but should normalise those too :)

src/streamlink_cli/argparser.py
```diff
@@ -156,1317 +156,1011 @@ def hours_minutes_seconds(value):
...
+        add_help=False,
+        usage="%(prog)s [OPTIONS] <URL> [STREAM]",
+        description=dedent("""
+        Streamlink is command-line utility that extracts streams from
+        various services and pipes them into a video player of choice.
+        """),
+        epilog=dedent("""
+        For more in-depth documentation see:
+          https://streamlink.github.io
+
```
back-to: a lot of flake8 errors `W293 blank line contains whitespace` in argparser.py

back-to:
well it might work,
but it's hard to test so many different command situations

if you are done with it and and doesn't plan to change something,
it's probably easier to just merge it and let more people test it.

beardypig:
I think I'm done with this for now, I don't plan to make any more changes - unless someone finds some more bugs.

gravyboat:
@beardypig Made some time to review everything tonight and this looks great. I'm going to leave it open until tomorrow in case anyone else wants to review it. If no one else weighs in I'll merge it and people can test it out. Explanations and code are clean and thorough and I didn't have any issue following along even though this is a massive change so good job there!

beardypig:
Thanks @back-to and @gravyboat for reviewing.

tests/test_plugin_funimationnow.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def test_arguments(self):
+        from streamlink_cli.main import setup_plugin_args
+        session = Streamlink()
+        parser = MagicMock()
+        plugin_parser = MagicMock()
+        parser.add_argument_group = MagicMock(return_value=plugin_parser)
+
+        session.plugins = {
+            'ustreamtv': FunimationNow
```
back-to: typo

PR #1646, opened 2018-05-09
Test coverage increase

beardypig:
A current work in progress, increasing the test coverage. 

Base increase ~10% with 25dbc76.

- [ ] 1 %
- [ ] 2 %
- [ ] 5 %
- [ ] 10 %


gravyboat:
Nice @beardypig. I was actually thinking about this a few days ago when reviewing your mpeg dash changes!

beardypig:
Again this one -11%, doesn't make sense to me ... looking in to it :/

back-to:
for the **-11%** error

---

in **.coveragerc**, if you change

```
[run]
source =
    streamlink
    streamlink_cli
```

to

```
[run]
source =
    src
```

**master** will change from `33%` to `47%`

**test-coverage** will be 48%

---

```
pytest --cov=src/
```

this is the same but the other might be cleaner

---

all the `0%` plugins will change because `can_handle_url` already run at some point,

but I am not sure how accurate the results are.

tests/__init__.py
```diff
@@ -0,0 +1,8 @@
+import sys
+
+if sys.version_info[0:2] == (2, 6):
```
back-to: > Oct 29 2013: Python 2.6.9 final released (security-only)

you don't need to support **python 2.6**

https://github.com/streamlink/streamlink/pull/544
https://www.python.org/dev/peps/pep-0361/#release-schedule

beardypig: True, it was still working until recently - but not now, so I'm not going to fix it. #544 was less about removing support for 2.6, but rather not going out of our way to maintain it.

beardypig:
Rebased for #1707

back-to:
since this changes the **default locations,**

it's maybe easier to rebase/merge this,
so new plugins won't be added to the old locations
as it would break this pr all the time.

further changes could be done in a different pr

beardypig:
@back-to yeah, gotta rebase it a lot... didn't even make 1% yet though :D

gravyboat:
Sound good, as soon as the tests pass I'll merge it.

gravyboat:
Restarted the Travis build since it was done and successful but not noting as such here. You'll also have to rebase again due to another PR.

gravyboat:
Thanks @beardypig.

PR #1647, opened 2018-05-10
plugins.ok_live: fix <hlsurl> extraction

Billy2011:
Fixes HTTP 400 Error


gravyboat:
Good stuff, thanks @Billy2011!

PR #1651, opened 2018-05-14
plugins.btsports: add plugin bt sports

beardypig:
Basic plugin to support BT Sports, requires a login to use.


gravyboat:

PR #1655, opened 2018-05-15
docs: add a note about socks4 and socks5 vs socks4a and socks5h

### LINKED ISSUE #1654: Need a option for DNS through proxy
### Checklist

- [ ] This is a bug report.

### Description
Steamlink can't resolve hostnames through the proxy server. If you receive a incorrect ip of the host via DNS spoofing, the option --http-proxy and --https-proxy may not work correctly.

Expected: Steamkink connect to youtube server through proxy and work fine.
Actual: Read timed out.

1. Run a socks5 proxy server on port 1080
2. streamlink --hls-live-restart --http-proxy "socks5://127.0.0.1:1080" --https-proxy "socks5://127.0.0.1:1080" https://www.youtube.com/watch?v=fO8x9MZ8m9g

### Logs
[cli][debug] Python:     3.6.5
[cli][debug] Streamlink: 0.12.1
[cli][info] Found matching plugin youtube for URL https://www.youtube.com/watch?v=fO8x9MZ8m9g
error: Unable to open URL: https://youtube.com/get_video_info (SOCKSHTTPSConnectionPool(host='youtube.com', port=443): Read timed out. (read timeout=20.0))

### Comments, screenshots, etc.
This is the log of my socks5 proxy when I run the step 2:
[2018-05-15 21:26:52] connect to 8.7.198.45:443

And this is the log when I visit https://youtube.com on firefox(Use the same socks5 proxy and it works fine)
[2018-05-15 21:28:51] connect to youtube.com:443

And this is the ping log to youtube.com on my pc
Pinging youtube.com [8.7.198.45] with 32 bytes of data:
Request timed out.
Request timed out.

Stramlink resolve the hostname 'youtube.com' direct and receive the incorrect ip '8.7.198.45' via DNS spoofing. Firefox works fine because it send the DNS request through the proxy.

beardypig:
Fixes #1654 by updating the docs to include a note detailing the difference in DNS lookup handling for the different socks scheme.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -339,18 +339,17 @@ options if the plugin you use makes HTTP and HTTPS requests.
 
 Both HTTP and SOCKS proxies are supported, authentication is supported for both types.
 
+.. note::
+    When using a SOCKS proxy the ``socks4`` and ``socks5`` schemes mean that DNS lookups are done
+    locally, rather than on the proxy server. To have the proxy server perform the DNS lookups the
+    ``socks4a`` and ``socks5h`` schemes should be used instead.
+
 For example:
 
 .. code-block:: console
 
     $ streamlink --http-proxy "http://user:pass@10.10.1.10:3128/" --https-proxy "socks5://10.10.1.10:1242"
-
-.. note::
-
-    When using a SOCKS proxy the ``socks4`` and ``socks5`` schemes mean that DNS lookups are done locally, rather than
-    on the proxy server. To have the proxy server perform the DNS lookups the ``socks4a`` and ``socks5h`` schemes should
-    be used instead.
-
+    $ streamlink --http-proxy "socks4a://10.10.1.10:1235" --https-proxy "socks5h://10.10.1.10:1234"
 
 Command-line usage
 ------------------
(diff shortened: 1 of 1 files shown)


gravyboat:
@beardypig This is a good addition, what do you think about adding an example of `socks4a` or `socks5h` just in case there are any language barriers that could result in confusion:

```
.. code-block:: console
 
     $ streamlink --http-proxy "socks5h://10.10.1.10:3128/" --https-proxy "socks5h://10.10.1.10:1242"
```

beardypig:
Could add another example :) Where do you suggest? After the note? Or do we remove the note and just expand the socks explanation?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

gravyboat:
@beardypig I think keeping the note is important for clarity, so I'd say just after the note.

beardypig:
I moved the note to above the examples, and added an example for `socks5h/socks4a`.

socks-docs-preview

bastimeyer:
Is there a requests doc site other than
https://github.com/requests/requests/blob/master/docs/user/advanced.rst#socks
which could be added here as an additional source of information?
Their external docs site is served via regular http and doesn't contain any information about the socks proxy usage :unamused:

gravyboat:
@bastimeyer I took a look and couldn't find anything so I'm going to merge this. If we discover something later we can add it. Thanks @beardypig!

Issue #521, opened 2017-02-03
Sometimes Ctrl-C doesn't end the connection to Twitch


thinkpad4:
### Checklist

- [x] This is a bug report.
- [ ] This is a plugin request.
- [ ] This is a feature request.
- [ ] I used the search function to find already opened/closed issues or pull requests.

### Description
Sometimes when a stream cuts off, for tech reasons or twitch reasons I can't Ctrl-C the connection in command prompt.

### Expected / Actual behavior
To have Ctrl-C end a connection every time

### Reproduction steps / Stream URLs to test
It doesn't happen all the time so reproducing it is hard

### Environment details (operating system, python version, etc.)
Win7 64 bit latest nightly

beardypig:
Does it work if you press Ctrl-C again?

gravyboat:
I noticed this on my Linux box as well. @beardypig Yeah you just have to hit `Ctrl-C` again.

beardypig:
What does it print to the console when it doesn't work? 
It could be some exception handling grabbing `KeyboardInterrupt` when it shouldn't...

gravyboat:
Hmm, I just tested this on `0.3.1` and I'm not encountering the issue, it does hang for a split second because it's shutting everything down, but it does exit correctly with a single `Ctrl-C`. I'll do some more testing in the next day or two to see if I can duplicate. @beardypig when it was happening it just looked like:
```
^C^CInterrupted! Exiting...
[cli][info] Closing currently open stream...
```

beardypig:
Would it eventually quit? It could be the StreamWorker thread still chomping through the segments and not interrupting the download... 

Side note: There are quite a few issues with the Stream downloader classes... I have been redesigning them a bit :)

gravyboat:
I honestly don't know, it was never a big enough deal to me to wait it out so I just hit `Ctrl-C` again. It doesn't even happen all the time.

beardypig:
Fair enough, it's probably the threads just not stopping right away...

thinkpad4:
Like I said, it's not every single time, but it seems to be related to a stream going down quick, like if the streamer's pc rebooted and they didn't have time to click end

thinkpad4:
@beardypig and to answer your first question, no repeated Ctrl-C does not do anything, I have to close the command prompt to end it.  Mind you VLC closed already so it has to be some process in streamlink

gravyboat:
[...]

PR #1658, opened 2018-05-16
Allow streamlink to exit faster when using Ctrl-C

beardypig:
The segmented stream downloader uses a threadpool to allow multiple segments to be downloaded simultaneously. The thread pool installs an atexit hook so that causes them to complete their current tasks, this means that if segments are still downloading streamlink will continue running until the segments are downloading.

By unregistering this atexit hook (because there aren't any consistency issues) the exit time can be sped up in some cases.

See #521.

### DIFF
--- a/src/streamlink/stream/segmented.py
+++ b/src/streamlink/stream/segmented.py
@@ -1,3 +1,5 @@
+import concurrent.futures.thread
+
 from concurrent import futures
 from threading import Thread, Event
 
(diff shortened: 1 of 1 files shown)


beardypig:
Only works for Python 3+, hence the build errors.

beardypig:
@gravyboat if you get time to look at this one as well.

gravyboat:
Looks good to me, nice fix @beardypig.

PR #1665, opened 2018-05-18
Add the README file to the Python package

beardypig:
Now that PyPI is migrating to their new version (warehouse, I think) we have the ability to include the README as Markdown instead of converting it to RST.

This will make the project page on PyPI look much nicer. You just have to check https://pypi.org/project/streamlink/0.12.1/ to see how great it looks right now 😉 

I also updated the classifiers while I was in `setup.py`...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,13 @@
 #!/usr/bin/env python
-
+import codecs
 import os
-import versioneer
 from os import environ
-from os.path import abspath, dirname, join
-from setuptools import setup, find_packages
+from os import path
 from sys import version_info, path as sys_path
+
+from setuptools import setup, find_packages
+
+import versioneer
 
 deps = []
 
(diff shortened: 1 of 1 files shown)


gravyboat:
Nice @beardypig, more good changes here!

back-to:
here are some additional options, that can be used.

```py
python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, <4',
```

https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

and

```py
project_urls={
    'Documentation': 'https://streamlink.github.io/',
    'Tracker': 'https://github.com/streamlink/streamlink/issues',
},
```


https://packaging.python.org/guides/distributing-packages-using-setuptools/#project-urls

beardypig:
@gravyboat I think this is good to merge.

gravyboat:
Cool, thanks @beardypig!

PR #1666, opened 2018-05-18
Add plugin for www.kingkong.com.tw

littlebtc:
This pull request will add a plugin for www.kingkong.com.tw, a game streaming website in Taiwan.

This plugin is mostly modified from the one of Douyutv.

Example of stream URLs can be found in the Kingkong front page.
A VOD URL example is provided FYI:
https://www.kingkong.com.tw/video/2152350G38400MJTU


gravyboat:
@littlebtc This looks pretty good! Do you think you could add some tests for this plugin?

littlebtc:
Added tests for `can_handle_url`.

beardypig:
Looks fine, some negative test cases would be good too. They are useful to make sure the plugin isn’t matching sites that it shouldn’t. Thanks @littlebtc

gravyboat:
Thanks @littlebtc!

PR #1671, opened 2018-05-19
plugins.btsports: add missing plugin matrix entry and tests

beardypig:
Forgot to update the docs plugin matrix.


gravyboat:

Issue #1669, opened 2018-05-19
[plugin issue] plugin.vidio.com / can't handle stream URLs anymore


akiwatti:
- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description
plugin handling of https://www.vidio.com/live  URls not working anymore due to changes at provided stream structure

### Expected / Actual behavior
streamlink -l debug www.vidio.com/live/665-rcti-tv-stream
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.5.2
[cli][debug] Streamlink: 0.12.1
[cli][debug] Requests(2.18.4), Socks(1.6.7), Websocket(0.47.0)
[cli][info] Found matching plugin vidio for URL www.vidio.com/live/665-rcti-tv-stream
[plugin.vidio][debug] HLS URL: https://kmklive-lh.akamaihd.net/i/rcti_ta_regular@94478/master.m3u8
error: Unable to open URL: https://kmklive-lh.akamaihd.net/i/rcti_ta_regular@94478/master.m3u8 (403 Client Error: Forbidden for url: https://kmklive-lh.akamaihd.net/i/rcti_ta_regular@94478/master.m3u8)

beardypig:
Looks like there are some extra steps required to get the actual stream url. Seems pretty straightforward though :)

beardypig:
@akiwatti I have submitted a PR to fix this, hopefully it will be merged soon and available in the nightly version.

akiwatti:
@beardypig ..can confirm for Win10/VLC is working. Many thanks,
But would appreciate if also for enigma2 openatv..but get this...

root@ax51:~# streamlinksrv manualstart
[streamlinksrv][info] Sun May 20 07:45:10 2018 Server started - Host: ax51, Port: 8088
[streamlinksrv][info] Python:     2.7.13
[streamlinksrv][info] Streamlink: 0.12.1 / 07.05.2018
[streamlinksrv][info] Requests(2.18.4), Websocket(0.44.0)
[streamlinksrv][info] ffmpeg is usable.
127.0.0.1 - - [20/May/2018 07:45:14] "GET /https://www.vidio.com/live/204-sctv-tv-stream HTTP/1.1" 200 -
127.0.0.1 - - [20/May/2018 07:45:14] URL: https://www.vidio.com/live/204-sctv-tv-stream Quality: best
[...]

PR #1672, opened 2018-05-20
plugins.vidio: py2 vs. py3 unicode fix

beardypig:
See #1669


gravyboat:

PR #1673, opened 2018-05-20
tests: test to ensure that all plugins are listed in the plugin matrix

beardypig:
I wrote a test that tests to ensure that each plugin exists in the plugin matrix in the docs (excluding the built in plugins). I found these plugins didn't have a row in the docs: live_russia_tv, ok_live, piczel, streann, and teamliquid. I updated it as best I could, please point out an omissions/mistakes.

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -9,268 +9,272 @@ they support. Streamlink's primary focus is live streams, so VOD support
 is limited.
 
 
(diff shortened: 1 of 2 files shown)


beardypig:
Looking in to the -11% coverage...

gravyboat:
Cool, merging. Thanks @beardypig this is a good addition.

Issue #1519, opened 2018-02-28
URL building in plugins


amurzeau:
### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [X] This is a plugin (improvement) request.
- [X] I have read the contribution guidelines.

### Description

Many plugins manipulate URLs to retrieve the underlying stream url.
An example: https://github.com/streamlink/streamlink/blob/b839cfd76b63eec3646a58ca19d7be560c422592/src/streamlink/plugins/dailymotion.py#L235

While uploading the previous version of streamlink to Debian, Paul Wise suggested to improve URL handling using something to manipulate URL safely which would understand, for example, query parameters and encode them.

What do you think about adding a URL utility class that would take care of URL handling ?
I though about something like `string.format` or `%` but urlencoding every parameters. I'm not sure though that this URL handling improvement can be used in every URL manipulations.

As an example, if the first string contains query parameters, the concatenation with `/videos` won't work (the result would be like `https://dailymotion.com/first_path?param1=data/videos`).
In this very case, `USER_INFO_URL` does not contains query parameters, but `username` might contains special characters.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

beardypig:
In this case, `urljoin` is probably the correct method. However, I'm sure we could whip up a nice `urlformat` method that could be used here :)

beardypig:
A review of all the plugins and how they build URLs needs to be done. We should have some guidelines on how to implement this kind of URL building. Most of the plugins are offenders, so we should sort it out. I have it on my todo list :)

back-to:
closing as https://github.com/streamlink/streamlink/pull/1675 is already merged

new plugins or plugin updates can use the new tools,
but updating every old plugin is to much work for a low value.

https://github.com/streamlink/streamlink/blob/cd6f94a5a4ee83cb5b61a295df9e13956bcfcb27/tests/test_utils_url.py#L39-L48

PR #1675, opened 2018-05-21
URL builder utils

beardypig:
As part of the solution for the issues raised in #1519 by @amurzeau, I have started adding some URL utils that can be used in plugins to better manipulate URLs. 

As well as a useful method for iterating through HTML tags, using regex in a brute force kind of way - like most of the plugins do. 

`update_qsd` and `url_concat` are the two URL manipulation methods. `url_concat` will join together URL parts to make a URL, and `update_qsd` can be used to add, remove or update query string parameters in a URL. 

I updated a couple of plugins as examples.

### DIFF
--- a/src/streamlink/plugins/gardenersworld.py
+++ b/src/streamlink/plugins/gardenersworld.py
@@ -4,12 +4,13 @@ import re
 
 from streamlink import NoPluginError
 from streamlink.plugin import Plugin
+from streamlink.plugin.api import http
+from streamlink.plugin.api.utils import itertags
 from streamlink.utils import update_scheme
 
 
 class GardenersWorld(Plugin):
     url_re = re.compile(r"https?://(?:www\.)?gardenersworld\.com/")
-    iframe_re = re.compile('''iframe src=(?P<quote>["'])(?P<url>.*?)(?P=quote)''')
 
     @classmethod
     def can_handle_url(cls, url):
@@ -17,13 +18,13 @@ class GardenersWorld(Plugin):
 
     def _get_streams(self):
         page = self.session.http.get(self.url)
-        iframe = self.iframe_re.search(page.text)
-        if iframe:
-            self.logger.debug("Handing off of {0}".format(iframe.group("url")))
+        for iframe in itertags(page.text, u"iframe"):
+            url = iframe.attributes["src"]
+            self.logger.debug("Handing off of {0}".format(url))
             try:
-                return self.session.streams(update_scheme(self.url, iframe.group("url")))
+                return self.session.streams(update_scheme(self.url, url))
             except NoPluginError:
-                self.logger.error("Handing off of {0} failed".format(iframe.group("url")))
+                self.logger.error("Handing off of {0} failed".format(url))
                 return None
 
 
--- a/src/streamlink/plugins/gardenersworld.py
+++ b/src/streamlink/plugins/gardenersworld.py
@@ -4,7 +4,6 @@ import re
 
 from streamlink import NoPluginError
 from streamlink.plugin import Plugin
-from streamlink.plugin.api import http
 from streamlink.plugin.api.utils import itertags
 from streamlink.utils import update_scheme
 
--- a/src/streamlink/plugin/api/utils.py
+++ b/src/streamlink/plugin/api/utils.py
@@ -17,11 +17,13 @@ def itertags(html, tag):
     """
     Brute force regex based HTML tag parser. This is a rough-and-ready searcher to find HTML tags when
     standards compliance is not required. Will find tags that are commented out, or inside script tag etc.
+
     :param html: HTML page
     :param tag: tag name to find
     :return: generator with Tags
     """
     for match in tag_re.finditer(html):
         if match.group("tag") == tag:
-            attrs = dict((a.group("key").lower(), a.group("value")) for a in attr_re.finditer(match.group("attr")))
+            attrs = {a.group("key").lower(): a.group("value") for a in attr_re.finditer(match.group("attr"))}
             yield Tag(match.group("tag"), attrs, match.group("inner"))
+
--- a/src/streamlink/plugin/api/utils.py
+++ b/src/streamlink/plugin/api/utils.py
@@ -24,6 +24,6 @@ def itertags(html, tag):
     """
     for match in tag_re.finditer(html):
         if match.group("tag") == tag:
-            attrs = {a.group("key").lower(): a.group("value") for a in attr_re.finditer(match.group("attr"))}
+            attrs = dict((a.group("key").lower(), a.group("value")) for a in attr_re.finditer(match.group("attr")))
             yield Tag(match.group("tag"), attrs, match.group("inner"))
 
--- a/src/streamlink/plugins/tf1.py
+++ b/src/streamlink/plugins/tf1.py
@@ -6,6 +6,7 @@ from streamlink.plugin import Plugin
 from streamlink.plugin.api import useragents
 from streamlink.stream import HDSStream
 from streamlink.stream import HLSStream
+from streamlink.utils.url import update_qsd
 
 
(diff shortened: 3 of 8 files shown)


gravyboat:
This fell through and I didn't ever merge it, sorry @beardypig. Do you want to rebase and we can get this in?

beardypig:
I cannot work out why it's failing on 3.7 right now ... When I have some time I will look at it, unless @back-to is able to work it out :)

back-to:
I know why it fails

```
re.finditer
Changed in version 3.7: Non-empty matches can now start just after a previous empty match.
```

https://docs.python.org/3/library/re.html#re.finditer
https://github.com/python/cpython/pull/4471
https://bugs.python.org/issue25054

https://docs.python.org/3/whatsnew/3.7.html#changes-in-the-python-api

---

here is something similar https://bugs.python.org/issue33585
might be the new expected behavior

> If you don't want to find an empty string, change you patter so that it will not match an empty string: ".+".

that has maybe something todo with it,
**but not sure whats the best way to fix it.**

```py
import re
test_html = """
<title>Title</title>
<meta property="og:type" content= "website" />
<meta property="og:url" content="http://test.se/"/>
<meta property="og:site_name" content="Test" />
<script src="https://test.se/test.js"></script>
<link rel="stylesheet" type="text/css" href="https://test.se/test.css">
<script>Tester.ready(function () {
alert("Hello, world!"); });</script>
<a
href="http://test.se/foo">bar</a>
"""
tag_re = re.compile(r'''(?=<(?P<tag>[a-zA-Z]+)(?P<attr>.*?)(?P<end>/)?>(?:(?P<inner>.*?)</\s*(?P=tag)\s*>)?)''', re.MULTILINE | re.DOTALL)
print([m.groups() for m in tag_re.finditer(test_html)])
```

Python 3.7

```py
[('title', '', None, 'Title'),
('meta', ' property="og:type" content= "website" ', '/', 'Title'),
('meta', ' property="og:url" content="http://test.se/"', '/', None),
('meta', ' property="og:site_name" content="Test" ', '/', None),
('script', ' src="https://test.se/test.js"', '/', ''),
('link', ' rel="stylesheet" type="text/css" href="https://test.se/test.css"', None, ''),
('script', '', None, 'Tester.ready(function () {\nalert("Hello, world!"); });'),
('a', '\nhref="http://test.se/foo"', None, 'bar')]
```

Python 3.6

```py
[('title', '', None, 'Title'),
('meta', ' property="og:type" content= "website" ', '/', None),
('meta', ' property="og:url" content="http://test.se/"', '/', None),
('meta', ' property="og:site_name" content="Test" ', '/', None),
('script', ' src="https://test.se/test.js"', None, ''),
('link', ' rel="stylesheet" type="text/css" href="https://test.se/test.css"', None, None),
('script', '', None, 'Tester.ready(function () {\nalert("Hello, world!"); });'),
('a', '\nhref="http://test.se/foo"', None, 'bar')]
```

beardypig:
I must be missing something - I cannot see why the `inner` group would be `"Title"` for the second match. It seems like the `inner` group is not being reset, and gets the previous value for some reason (same for when it is `''`).

back-to:
@beardypig 

with removing the `(?= ... )` part, the tests will pass on 3.7

```diff
diff --git a/src/streamlink/plugin/api/utils.py b/src/streamlink/plugin/api/utils.py
index f6f3f12..00445e2 100644
--- a/src/streamlink/plugin/api/utils.py
+++ b/src/streamlink/plugin/api/utils.py
@@ -7,9 +7,9 @@ from ...utils import parse_qsd as parse_query, parse_json, parse_xml
 __all__ = ["parse_json", "parse_xml", "parse_query"]
 
 
-tag_re = re.compile('''(?=<(?P<tag>[a-zA-Z]+)(?P<attr>.*?)(?P<end>/)?>(?:(?P<inner>.*?)</\s*(?P=tag)\s*>)?)''',
+tag_re = re.compile(r'''<(?P<tag>[a-zA-Z]+)(?P<attr>.*?)(?P<end>/)?>(?:(?P<inner>.*?)</\s*(?P=tag)\s*>)?''',
                     re.MULTILINE | re.DOTALL)
-attr_re = re.compile('''\s*(?P<key>[\w-]+)\s*(?:=\s*(?P<quote>["']?)(?P<value>.*?)(?P=quote)\s*)?''')
+attr_re = re.compile(r'''\s*(?P<key>[\w-]+)\s*(?:=\s*(?P<quote>["']?)(?P<value>.*?)(?P=quote)\s*)?''')
 Tag = namedtuple("Tag", "tag attributes text")
 
 
@@ -26,4 +26,3 @@ def itertags(html, tag):
         if match.group("tag") == tag:
             attrs = dict((a.group("key").lower(), a.group("value")) for a in attr_re.finditer(match.group("attr")))
             yield Tag(match.group("tag"), attrs, match.group("inner"))
-
```

---

```
platform linux -- Python 3.7.0, pytest-3.6.3, py-1.5.4, pluggy-0.6.0
rootdir: /run/media/ka/N/github/back-to/streamlink, inifile:
plugins: requests-mock-1.5.2, cov-2.5.1
collected 5 items                                                                                                                                            

tests/test_plugin_utils.py .....
```

beardypig:
Not sure that the positive look ahead is even required, might have been left over from a previous `iter_tags` implementation (I had a few goes) :)

src/streamlink/plugins/gardenersworld.py
```diff
@@ -4,26 +4,27 @@
 
 from streamlink import NoPluginError
 from streamlink.plugin import Plugin
+from streamlink.plugin.api import http
```
back-to: `http` can be removed, it got added with a rebase.

beardypig: True that, forgot to clean up after I rebased. I saw it (using PyCharm to do the rebase), and i was going to tidy it up but I forgot :)

back-to:
@beardypig 

`itertags` is actually broken now, because the **test html code** is missing the `<html>` tag,
which will work for the tests but not for real websites.

it can be tested with the `gardenersworld` plugin as it is broken with the current changes.

`<>` must be ignored for inner tags or there will be only one tag `html`

here is a posible fix

https://github.com/back-to/streamlink/commit/ab7d6bf25345a1a866e1ef6411b493fc074fc9a2
https://travis-ci.org/back-to/streamlink/builds/410256162

beardypig:
@back-to I updated the test html with your changes, and added a test for inner/outer tags - to maintain the existing behaviour. 

I think this might be a bug in Python. I have created a [bug report](https://bugs.python.org/issue34294), so we can see what they say.

back-to:
well, this might take awhile.

**Maybe** the **tests** should be **ignored on py37 and py38** for now,
so `url_concat` and `update_qsd` can be used.

As the Issue was merged in https://github.com/streamlink/streamlink/pull/1693 and it is not really related to this PR,
it was only discovered because of the additional tests.

---

also `from streamlink.plugin.api import http` got added again with the rebase :elephant: 

---

https://github.com/beardypig/streamlink/blob/e449aa326a1cece9a8dc4755c413caa179d915f5/tests/test_utils.py#L100-L111

`test_url_equal` should be removed,
when you moved it to `test_utils_url.py` it was not removed in `test_utils.py`.

beardypig:
🤦‍♂️ The rebase for this one got complex :)

I'll add a `skipTest` for Python 3.7/3.8.

beardypig:
I updated the tests so those that were failing allow failure on 3.7+. With a note to monitor bpo-34294.

beardypig:
@gravyboat, I think this is OK to merge with the exception that it might be wonky on Python 3.7.

gravyboat:
@beardypig Sounds good, changes look good as well. We can always open another issue depending on what occurs.

beardypig:
@gravyboat I opened an issue to remind us :)

Issue #1679, opened 2018-05-22
Bintray deploy won't work because of an invalid includePattern


back-to:
### Checklist

- [x] This is a bug report.

### Description

Ref https://github.com/streamlink/streamlink/pull/1413#issuecomment-389561062
https://bintray.com/streamlink/streamlink-nightly/streamlink

the `includePattern` won't work with the new version name,

https://github.com/streamlink/streamlink/blob/416c8347b394c9b1cbaff150450d51a75509dd24/script/bintrayconfig.sh#L32-L36

`/home/travis/build/streamlink/streamlink/dist/nsis/streamlink-0.12.1+8.ge2a5546.exe`

it won't match because of the `+`

> includePattern: Pattern in the form of Ruby regular expression

https://docs.travis-ci.com/user/deployment/bintray

beardypig:
I think bbe6674 will fix it.

back-to:
yes the fix works, maybe merge it into master

https://travis-ci.org/streamrepo/streamlink/jobs/382262907#L2872
https://bintray.com/back-to/Test/streamlink

beardypig:
Thanks for testing @back-to, I didn't have time to test it yet :)

gravyboat:
Closing since #1681 is merged.

PR #1681, opened 2018-05-22
build: use _ instead of + in the Windows installer exe filename

beardypig:
Fix for #1679


gravyboat:
Nice catch @beardypig.

PR #1682, opened 2018-05-22
tests: test to ensure each plugin listed in the plugin matrix exists

beardypig:
Ensure that each plugin that is listed in the plugin matrix in the docs actually exists, for example if a plugin was deleted but not removed from the plugin matrix. It's a complimentary test to the one added in #1673.


gravyboat:
Good stuff, thanks @beardypig! You've been cranking out the enhancements lately!

beardypig:
@gravyboat I've got a big TODO list :)

Issue #1664, opened 2018-05-18
Release Improvements


beardypig:
The aim of this issue is to track the changes required to simplify the release process and make @cdrage's life easier, or allow others to perform releases.

I think the process should end up being three steps only, via a script or possibly the GitHub UI. 

1. Create PR with updated change log for the new version.
2. Merge PR.
3. Create tag for the new version. 

The rest of the process should be handled automatically, triggered by creating the tag. 

----
 
 - [x] Automatically add the Windows installer to GitHub release (#328)
 - [x] Update the release information using travis from the change log (#328)
 - [x] Include README in the package long_description (#1665)
 - [x] Automatically deploy the `streamlink` package to PyPI (#1701)
 - [x] Automatically add the sdist package GitHub release (#1701)
 - [x] Deploy a wheel to PyPI (#1701)
 - [ ] stable/dev versions of the docs (#1684 WIP)
 - [x] Make the change log consistently markdown, and include it in the docs using `recommonmark` (#1667)
 - [x] Only include installation/open collective in GitHub release info (#1667) 

1. We can include the README.md file in the package metadata as long_description to improve the appearance of the `streamlink` package on PyPI. See https://pypi.org/project/streamlink/0.12.1/ to see how it looks right now.

1. We _could_ deploy the `streamlink` package to PyPI automatically when creating a release tag, this would streamline the release process and remove some of the manual work and possibility for mistakes.

1. We _could_ add the `sdist` package to the GitHub release automatically with the release tag, as we do with the Windows installer. 

1. We _could_ also deploy a wheel to PyPI either automatically or manually, depending on what we do with the others. This might speed up installations in some cases, and is the preferred binary distribution method - but each platform variant might need its own wheel.

1. Separate the docs in to 2 streams (dev and stable), where stable is the default page. A link to the dev version will also need to be added a long with a note stating that is it the development/nightly version. Possibly we could deploy the dev docs at the same time as the nightlies are built, and deploy the stable version on release tag. 

1. The change log is currently a mix of reStructuredText and Markdown, it should be Markdown only. It should be able to be included in the docs by using `recommonmark` in Sphinx to render the Markdown. 

1. The installation instructions and open collective info don't need to be included for every single version in the release log, they are useful in the GitHub releases though. The script that updates the GitHub releases can be modified to include this template, then it doesn't need to be included in the change log. 

As always, it would be nice to have some feedback from the rest of you! I'm sure there are some other things we can improve 😄

[...]

PR #1684, opened 2018-05-23
Deploy two docs streams: stable and unstable

### LINKED ISSUE #254: Documentation doesn't match stable version
The documentation is build from master whenever there is a commit on master, which is nice for some things like errors or omissions in the docs, but not so good for new features. It is unclear which features the latest stable version supports. For example, the docs currently list bigo as a supported site, however it is only supported in the nightly/HEAD version rather than the latest tagged version (0.1.0). 

There are a couple of ways I can think of resolving this. 
1. We could have a separate repo for the docs and ask people to submit separate PRs to update the docs. This might lead to the docs going unloved as it's an extra bit of work.
2. We could build the docs only on tagged revisions. This would fix the new features only in nightlies, but we'd have to have a tag/release to fix errors and omissions in the docs. We can build multiple versions of the docs, for each tag and for the current HEAD revision (similar to the readthedocs site).
3. We could mark each feature with the version that it was added or if it is currently only supported in the nightly version. This would be more work to update the versions when we release a new version as we'd have to updated all the "this feature is only supported in the nightly version" type messages to "added in x.y.z". It's possible that some automation (sphinx magic) could be used to make the release managers life easier. 

All of these suggestions add extra work, but would ultimately would benefit the users. If anyone has any other suggestions to improve the quality of the docs they would be most welcome. 

I think the 3rd option is pretty good, but the 2nd option would be OK too. 

Thought? Suggestions?

beardypig:
Opening this PR to continue discussion from #1664. 

When there is a tag the stable version of the docs will be built to the root directory (`/`) ~~and to a directory of the tag (`/0.12.1`, etc.)~~. For each commit on `master` the docs will be build to the `/latest` directory. 

There is a version picker for stable and latest statically rendered in to each page.

Will fix #254.


.travis.yml
```diff
@@ -1,5 +1,10 @@
 language: python
 
+env:
+  global:
+    # Doctr deploy key for streamlink/streamlink.github.io
+    - secure: "e9rxQKoEqdJvk8l66qqpBtYN8uhnizwynCzAGTLb6q1wx7PA8V9zGqByMTBKEo+/BnfRTTxahcxHPu21qujLxvL6cbYraDBYSLJj06gQgIFnAueXbtHb/9Yjhdzlvz9jLwYA2S2LgZInor2FfduEIQlSivSlKSLIyO+JRh/1pa005+Gndx0DHMrFMdFO2+YisZksJxwpXgCwIqa4d4rLkLjg9p4a7E713CcDz1OhUNNyA2oGe8SMm96oq9z2O5CJrQ6E2weKflqu+LEITJylC3Pkz5klp9XK1BsQdA8J5aD5FdIxQIdam9Xjg03V5ww71RIHGSvKcOPu1sHpkpVuW6gc35Texk3M2Bl6OSKes2k49RmegTg/qhh16w/en8hm5Knymrkwq0/KPInjvOjg6d2g5781Mkr68cXGxI7PFb6on7LG5O0R6azrhJx5BAnUsfEb5E+CJ/c75os5fe/XDCcxNkCyM5uVlUeBRvKdr+Cq5CfCUm6x1EkxYf6E982fUcXnHWn6ofrW4wz+FRrtdYOh2GNtHq6SNcRpZvSN1am4Ztu3xTtT8kxne7ECqis9SFLIbPN1YI2b2d9nkE9W9xZu7eK9OpIb4f/wJtLJRp8+CWsQ2SN8milSvOYiK9XqvRlr+lW1hlLY27m492lKX+s6hYJ+7HljCJTJ9EXihnI="
```
bastimeyer: In order to not bloat up the travis conf file any further, I'd prefer if the env var could simply be added to the env var list on the web panel.

docs/_themes/sphinx_rtd_theme_violet/static/css/theme.css
```diff
@@ -2425,6 +2425,22 @@ a .fa, a .rst-content .admonition-title, .rst-content a .admonition-title, a .rs
   background: #e7f2fa;
 }
 
+.rst-conent .version-warning {
```
bastimeyer: typo here

beardypig:
I think we should build just the two versions for now, stable and unstable. Doubling the size of the repo isn't too bad, and if we work out a better way to do it at a later date we can squash all the commits down.

beardypig:
@bastimeyer did you have any html/css suggestions?

bastimeyer:
I don't have the time right now to work on this or other projects, sorry. I'll see if I can take at look at this next week or so. My suggestions in the other thread regarding the navigation of the unstable/stable docs were not that important. The box at the bottom right corner should be working fine for now. The changes can be made later on, if you want to merge your doc separation work.

gravyboat:
Closing as per requested by @beardypig.

Issue #1352, opened 2017-12-01
Plugin Request: 1tv.ru


ghost:
*Thanks for reporting an issue!*
*Please read the contribution guidelines first! (see the link above)*
*Also check the list of known issues before reporting an issue!*

*Feel free to use the following template. Be as detailed as possible.*
*Please see the text preview to avoid unnecessary formatting errors.*
*Don't forget to remove this text before submitting.*

----

### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

Russia's main television channel

https://www.1tv.ru/live

gravyboat:
If you can provide any additional details about what a request looks like that would be great. I'm not going to get a Russian proxy to check out this site.

roshavagarga:
I can help out with information about this plugin request if it can be reopened @gravyboat

gravyboat:
@roshavagarga Sure, is the content on this site legal? I don't speak Russian or have familiarity with video there.

roshavagarga:
@gravyboat The content is legal but as far as I've seen, some of the content gets blocked as Russia-only, while other content doesn't. This is basically the main national tv station for the whole country.
They're using tokens as far as I can see for that. I don't really know Russian either, but Bulgarian is close enough as far as languages get.

beardypig:
Looks legit, based on wikipedia anyway :)

gravyboat:
Closing since #1687 was merged.

PR #1687, opened 2018-05-24
plugins.onetv: add support for 1tv.ru and a few other related sites

beardypig:
Support for 1tv.ru as requested in #1352, also supports a couple of other live streams and VOD from 1tv.ru.


gravyboat:
Thanks @beardypig.

Issue #1386, opened 2017-12-22
Getting streamdata directly into Python and opencv


DanielTea:
Hi, 

is it possible to stream twitch streamdata directly into opencv via streamlink? 
I'm looking for something like this (does not work but you'll get the concept):

```
streams = streamlink.streams("https://www.twitch.tv/grimmmz")
stream = streams["best"]
fd = stream.open()
bytes=''

while True:
    bytes+= fd.read(1024)

    a = bytes.find('\xff\xd8')
    b = bytes.find('\xff\xd9')

    if a!=-1 and b!=-1:
        jpg = bytes[a:b+2]
        bytes= bytes[b+2:]
        img = cv2.imdecode(np.fromstring(jpg, dtype=np.uint8),cv2.CV_LOAD_IMAGE_COLOR)
        cv2.imwrite('messigray.png', img)
        cv2.imshow('cam2', img)

    else:
        continue

```

gravyboat:
You could try https://streamlink.github.io/cli.html#cmdoption-stdout but I've never used that command so I'm not sure what the data output looks like. What would your use case be? Most of the opencv examples I've seen have been for modification of video in ways to try and circumvent automated takedown systems by making slight changes to videos so they avoid detection.

DanielTea:
Hi gravyboart!

Thanks for your fast reply!
We try to do a realtime emotion analysis on streamers for a  university project, classifing them raging, laughing etc. That’s why we have to pipe the stream through openCV to do a face recognition.

I didn‘t try the stdout command yet. Will do. But I assume it will output a raw byte stream which we have to interpret or decode somehow. 

What works is creating a .ts file and read that file via openCV. But it is super Laggy and far from realtime.

gravyboat:
No problem, if you don't have any luck with the stdout option, maybe give youtube-dl a try? They had a similar issue on this a while back: https://github.com/rg3/youtube-dl/issues/11291 that has some suggestions that might work, but it might produce the slowness issue you were encountering as well.

beardypig:
Something similar was discussed in #857, might be a worth taking a look at that - if you didn't already :)

AndyZhu1991:
```
streams = streamlink.streams('live.bilibili.com/1')
url = streams['best'].url
cap = cv2.VideoCapture(url)

while True:
    succ, frame = cap.read()
    if succ:
        # Do something here
    else:
        break

cap.release()
```
[...]

Issue #857, opened 2017-04-27
Can't decode piped data


aharris02:
Hi - I'm trying to decode the piped data and seem to be coming up empty.

I've tried piping directly from streamlink-->OpenCV  and I've tried streamlink -> ffmpeg -> OpenCV. Neither works.

streamlink->opencv:
```
pipe = Popen(["streamlink -O --twitch-oauth-token " + self.oauth_token + " -Q twitch.tv/" + self.username + " " + self.quality], stdout = PIPE, bufsize=10**8, shell=True)
```

streamlink->ffmpeg->opencv:
```
pipe = Popen(["streamlink --twitch-oauth-token " + self.oauth_token + " -Q twitch.tv/" + self.username + " " + self.quality + " | ffmpeg -i pipe:0 -c:v rawvideo -r 1 -f image2pipe -"], stdout = PIPE, bufsize=10**8, shell=True)
```

Decoder:
```
raw_image = pipe.stdout.read()
frame = cv2.imdecode(np.frombuffer(raw_image, np.uint8), 1)`
```

Error:
```
File "recorder.py", line 114
    frame = cv2.imdecode(np.frombuffer(raw_image, np.uint8), 1)
cv2.error: /home/user/opencv-3.1.0/modules/imgcodecs/src/loadsave.cpp:490: error: (-215) !buf.empty() && buf.isContinuous() in function imdecode_
```

Really appreciate any help you can provide on this.

aharris02:
Logfile from ffmpeg.

[ffmpeg-20170427-164454.txt](https://github.com/streamlink/streamlink/files/962875/ffmpeg-20170427-164454.txt)

beardypig:
Assuming you're using Linux, this might get you in the right direction.
You can use `cv2.VideoCapture` to read from a named pipe. If you create a named pipe and write the twitch stream to that pipe.
```
mkfifo /tmp/stream
streamlink twitch.tv/legema best -o /tmp/stream
```
Then the data will be streamed to /tmp/stream and you can read it from a separate process.

```python
import cv2
import numpy


def extract(path):
    cap = cv2.VideoCapture(path)
    fnum = 0

    try:
        while True:
            # Capture frame-by-frame
            ret, frame = cap.read()
            if ret:
                average_color_per_row = numpy.average(frame, axis=0)
                average_color = numpy.average(average_color_per_row, axis=0)
                print(average_color)
    finally:
        # When everything done, release the capture
        cap.release()


[...]

PR #1689, opened 2018-05-24
example: added an opencv face detection example

beardypig:
As this has come up a couple of times (#857 and #1386), I decided to make a more expanded example. 

In this example a stream from streamlink is read by `opencv` (`opencv-python` module is required), and the faces in the stream a detected and highlighted in real-time. It might not work for every stream, and it doesn't support any special streamlink options (proxies, etc.) - but it's an example. Plus, it was quite fun to do and it's amusing to watch 😄

### DIFF
--- a/examples/opencv-face.py
+++ b/examples/opencv-face.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+import logging
+import sys
+import streamlink
+import os.path
+
+try:
+    import cv2
+except ImportError:
+    sys.stderr.write("This example requires opencv-python is installed")
+    raise
+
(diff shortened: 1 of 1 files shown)


gravyboat:
Very cool, thanks @beardypig.

beardypig:
@felipegcampos I’m not sure now. I think I tested it on some live TV stream. Lemme check and get back to you next week. I don’t see any particular reason why it wouldn’t work on twitch though.

felipegcampos:
Hey @beardypig . I appreciate your time. I feel it could be some sort of cross-origin restriction.  I'm just guessing though.

PR #1690, opened 2018-05-24
logging: refactor to use python logging module

beardypig:
The logging output remains identical to before. The API has changed, instead of passing a Logger instance around, a module level `log` instance is used. This is the standard way to do logging in most Python applications and module. 
 
 - [x] Replace old crufty logging with standard Python logging
 - [x] Add `trace` log level
 - [x] Add compatibility class for old Logger
 - [x] ~~Update plugins to use new logging~~ Too many plugins!!!

`Session` still has a `logger` attribute, which passes the log messages to the standard logger. A `DeprecationWarning` is issued if this is done though. The `Logger` and `LoggerModule` classes have been reimplemented as a compatibility layer just incase.  

There are a number of advantages to doing this:
  1. more Pythonic
  2. easier to add logging message from utils and other methods without passing session or loggers around (this is my main motivation) 
  3. if required, log messages from other Python modules can be enabled
  4. the log destination could be changed to (or duplicated to) other places, syslog for example
  5. when other applications include streamlink they can get the streamlink logging message

Note: I want to refactor the logging for the plugins too, but there are too many damn plugins - it's optional, developers can use either system as it ends up in the same place anyway. As a developer of plugins you would not notice any difference. 

FYI: I wanted this to add logging to the `MPD` class, with the current system it's not very easy and make very ugly code to log via `Session`.


beardypig:
@Billy2011 how are you using it? Do you have the code on GitHub?

Billy2011:
The code is not on GitGub: [Enigma2 streamlinksrv](https://pastebin.com/75ggE4J6)

beardypig:
@Billy2011 your current usage will continue to work as is :)

src/streamlink_cli/argparser.py
```diff
@@ -275,12 +276,14 @@ def build_parser():
...
         default="info",
         help="""
         Set the log message threshold.
-
-        Valid levels are: none, error, warning, info, debug
-        """
+        Can be repeated to load multiple files, in which case
+        the options are merged on top of each other where the
+        last config has highest priority.
+    """
```
gravyboat: This needs to be indented 4 spaces further.

gravyboat: I also wonder if we could potentially make what exactly this does a little more clear in terms of the documentation? I don't really see in what case you would run multiple log levels so the modified description makes it a bit more confusing, and also removes information regarding what the valid levels are.

beardypig: Seems to have got messed up when I rebased before.

src/streamlink/logger.py
```diff
@@ -1,61 +1,212 @@
...
+        handler.setFormatter(formatter)
+
+        root.addHandler(handler)
+        level = kwargs.get("level")
+        if level is not None:
+            root.setLevel(level)
 
 
-__all__ = ["Logger"]
+__all__ = ["StreamlinkLogHandler", "Logger", "basicConfig", "root", "levels"]
```
back-to: it's not used anywhere

`src/streamlink/logger.py:212:1: F822 undefined name 'StreamlinkLogHandler' in __all__`

src/streamlink/logger.py
```diff
@@ -1,61 +1,212 @@
...
+
+    def formatMessage(self, record):
+        if self.style == "{":
+            return self.fmt.format(**record.__dict__)
+        else:
+            return self.fmt % record.__dict__
+
+    def format(self, record):
+        for rbase in self.remove_base:
+            record.name = record.name.replace(rbase+".", "")
```
back-to: `E226 missing whitespace around arithmetic operator`

just some Flake8 errors

beardypig:
Thanks for the review as always @back-to

gravyboat:
Thanks for the contribution @beardypig, and thanks for the review @back-to!

beardypig:
Cool! Time to add a bunch of logging to the dash stuff :)

src/streamlink_cli/main.py
```diff
@@ -886,12 +885,12 @@ def setup_plugin_options(session, plugin):
...
     if hasattr(os, "getuid"):
         if os.geteuid() == 0:
-            console.logger.info("streamlink is running as root! Be careful!")
+            log.info("streamlink is running as root! Be careful!")
 
 
 def log_current_versions():
     """Show current installed versions"""
-    if args.loglevel == "debug":
+    if logger.root.isEnabledFor(logging.DEBUG):
```
back-to: @beardypig I found a small issue

when `loglevel=debug` is setup in the config file,
it won't print `log_current_versions()` because `logger.root` won't change

```text
$ streamlink
>>> setup_args(parser, ignore_unknown=True)
---- args.loglevel
info
---- logger.root
<Logger streamlink (warning)>
>>> setup_config_args(parser)
---- args.loglevel
debug
---- logger.root
<Logger streamlink (info)>
>>> log_current_versions()
---- args.loglevel
debug
---- logger.root
<Logger streamlink (info)>
```

beardypig: Good catch @back-to, I'm taking a look now.

beardypig: Opened #1715 which resolves it for me, can you take a look?

back-to:
@beardypig 
any stream from https://goodgame.ru/ and `-l debug` will end in a TypeError

```
  File "env/bin/streamlink", line 11, in <module>
  File "src/streamlink_cli/main.py", line 999, in main
    handle_url()
  File "src/streamlink_cli/main.py", line 530, in handle_url
  File "src/streamlink_cli/main.py", line 412, in fetch_streams
  File "src/streamlink/plugin/plugin.py", line 396, in get_streams
    return self.streams(*args, **kwargs)
  File "src/streamlink/plugin/plugin.py", line 298, in streams
  File "src/streamlink/plugins/goodgame.py", line 50, in _get_streams
    **channel_info)
  File "/usr/lib/python3.6/logging/__init__.py", line 1294, in debug
    self._log(DEBUG, msg, args, **kwargs)
TypeError: _log() got an unexpected keyword argument 'id'
```

https://github.com/streamlink/streamlink/blob/20831c500f60a7802500f6251877e187ac786b5a/src/streamlink/plugins/goodgame.py#L49-L50

the easy way is to change the plugin, but it's maybe an issue with the logging change

beardypig:
I think the fix for this is to change the plugin - the easiest way is to just use `str.format`. The logging in livestreamer->streamlink is a bit different to regular python logging, and that’s what this PR tried to fix. Obviously there are a few kinks :)

src/streamlink_cli/main.py
```diff
@@ -41,12 +43,14 @@
 
 args = console = streamlink = plugin = stream_fd = output = None
 
+log = logging.getLogger("streamlink.console")
```
bastimeyer: This changes the main log output from `[cli][foo]bar` to `[console][foo]bar`, which breaks the Twitch GUI's error handling.
I had to fix this in the Twitch GUI, but since there's no new release yet, everyone who's updated Streamlink will run into potential issues with that. Same for other applications relying on the log output, ofc. Unfortunately, I didn't notice this earlier and I should have commented on this sooner.

bastimeyer: @beardypig Is there are reason this was changed from "cli" to "console"?

beardypig: Nah, was a mistake

bastimeyer: Can we revert this and publish a bugfix release, please? :confused:

beardypig: If you’re able to do it now you can, I won’t be able to do it until tomorrow - don’t have access to a PC right now...

Issue #1599, opened 2018-04-12
[PlugIn Request] Support for Gol TV


nickiman:
### Checklist

- [ ] This is a bug report.
- [x] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description

Would it be possible to support Gol TV online:
https://www.goltelevision.com/
**https://www.goltelevision.com/live**

You don´t need a special account to watch it.

I can play the stream with VLC in this way:
**masterManifest**: https://api-cdn.overon.es/SmilLive/getLiveIOS/playlist.m3u8?system=webgol&stream=webgol_gol24h&token=1514d6a55a2193c1d810ae7138627089&expire=5acf5a00
**mediaManifest**: https://webgol-hls-live-overon.secure2.footprint.net/OveronHttpOrigin/_definst_/webgol_gol24h_MB_3496/chunklist.m3u8?secureHash=MTk1LjIzNS4xNS4yMDA=&nva=20180413091844&token=02ffdb5df38255bf5bbc0

Manifest content:
```
#EXTM3U
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=896000
https://webgol-hls-live-overon.secure2.footprint.net/OveronHttpOrigin/_definst_/webgol_gol24h_MB_896/chunklist.m3u8?secureHash=MTk1LjIzNS4xNS4yMDA=&nva=20180413092219&token=064fde37d48a562f1fea1
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=1896000
https://webgol-hls-live-overon.secure2.footprint.net/OveronHttpOrigin/_definst_/webgol_gol24h_MB_1896/chunklist.m3u8?secureHash=MTk1LjIzNS4xNS4yMDA=&nva=20180413092219&token=03e8014dc23609e302906
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=2896000
https://webgol-hls-live-overon.secure2.footprint.net/OveronHttpOrigin/_definst_/webgol_gol24h_MB_2896/chunklist.m3u8?secureHash=MTk1LjIzNS4xNS4yMDA=&nva=20180413092219&token=0122fca24320935986fe5
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=3496000
[...]

PR #1694, opened 2018-05-25
plugins.goltelevision: support for the live stream

beardypig:
As requested in #1599. Geo-blocked to Spain.


gravyboat:
Thanks @beardypig.

PR #1685, opened 2018-05-23
deploy: automatically deploy sdist and wheel builds to PyPI.

beardypig:
Deploys using a new Streamlink user on PyPI, I have emailed the credentials to @cdrage and @gravyboat and already added the credentials to travis.

PR #1701, opened 2018-05-28
build: build and sign sdist in travis using an RSA sign-only key

beardypig:
I added a signing key to the repo, encrypted using travis, to sign the packages. The key is an 4096 byte RSA sign-only key, and has the ID `2E390FA0`. There is no passphrase on the key stored in the repo - we rely on travis to keep it secret. If this PR is merged, I will publish the key to the MIT PGP key server, and distributed copies of the signing key with a passphrase to the normal people. 

The script is fairly simple, it creates a temp directory to use as the gpg keyring, adds the key, then signs the tar ball. gpg won't use a key unless it is stored in a keyring - hence, the temporary keyring. The resulting tarball and signature file are added to the GitHub release. 

Tested on my fork.


back-to:
if you want to do it like this, you can drop https://github.com/streamlink/streamlink/pull/1685

and add **twine** in this script

_example_

```
pip install wheel twine

twine upload --username $PYPI_USER --password $PYPI_PASSWORD ...
```
and
https://github.com/tliron/ronin/blob/f91fc4b0645150de7a62d4caefbdf48faf91f21b/deploy-to-pypi.sh#L16-L20

you would only **build it once** and sign them all

---

the whl could also be upload to releases on github

---

also for MANIFEST.in, something like this

```
prune */__pycache__
global-exclude *.pyc *~ *.bak *.swp *.pyo
```

because if there is no cleanup, the docs will create a `__pycache__` folder

beardypig:
Good catch on the MANIFEST entries. I also think it makes sense to combine the two, streamlines the process a bit more :)

beardypig:
Combined with #1685

gravyboat:
Thanks @beardypig! @cdrage We'll need to make sure things look good for the next release with this being merged.

beardypig:
@cdrage @gravyboat I noticed an issue with the release script. It won't work quite as intended, the CHANELOG.md file won't be formatted correctly. I have been working on an updated release script though :)

script/sdistsign.sh
```diff
@@ -0,0 +1,33 @@
...
+version=$(python setup.py --version)
+dist_dir=${STREAMLINK_DIST_DIR:-dist}
+temp_keyring=$(mktemp -d) && trap "rm -rf ${temp_keyring}" EXIT || exit 255
+
+if [[ -n "${TRAVIS}" ]]; then
+      openssl aes-256-cbc -K ${encrypted_eeb8b970d3a3_key} -iv ${encrypted_eeb8b970d3a3_iv} -in signing.key.enc -out "${SDIST_KEY_FILE}" -d
+fi
+
+echo "build: Installing twine and wheel" >&2
+pip -q install twine wheel
```
back-to: `wheel` can be removed, it won't work with **python 2.7**

```
$ pip list
Package    Version
---------- -------
pip        10.0.1 
setuptools 39.2.0 
wheel      0.31.1 
```
```
$ pip install streamlink
Collecting streamlink
  Using cached .../streamlink-0.13.0-py2.py3-none-any.whl
  ...
Installing collected packages: iso-639, iso3166, six, websocket-client, chardet, certifi, urllib3, idna, requests, pycryptodome, PySocks, isodate, streamlink
Successfully installed PySocks-1.6.8 certifi-2018.4.16 chardet-3.0.4 idna-2.6 iso-639-0.4.5 iso3166-0.8 isodate-0.6.0 pycryptodome-3.6.1 requests-2.18.4 six-1.11.0 streamlink-0.13.0 urllib3-1.22 websocket-client-0.48.0
```

```
$ streamlink 
...
ImportError: No module named backports.shutil_which
```

```
$ pip list
Package          Version  
---------------- ---------
certifi          2018.4.16
chardet          3.0.4    
idna             2.6      
iso-639          0.4.5    
iso3166          0.8      
isodate          0.6.0    
pip              10.0.1   
pycryptodome     3.6.1    
PySocks          1.6.8    
requests         2.18.4   
setuptools       39.2.0   
six              1.11.0   
streamlink       0.13.0   
urllib3          1.22     
websocket-client 0.48.0   
wheel            0.31.1 
```

beardypig: Can add a fix for this :)

beardypig: @back-to would you be able to test out building a whl with #1758 and seeing that installs for you? I assume this is an issue for Enigma2 boxes?

Issue #1179, opened 2017-08-20
Chaturbate: Invalid URL '': No schema supplied.


soredake:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description
```
[cli][info] Found matching plugin chaturbate for URL https://chaturbate.com/$regional blocked streamer/
error: Unable to open URL:  (Invalid URL '': No schema supplied. Perhaps you meant http://?)
```
...

### Expected / Actual behavior
Playing stream.
...

### Reproduction steps / Explicit stream URLs to test

1. streamlink $regional blocked streamer best
2. ...

### Environment details

streamlink 0.7.0
Python 3.4.5
...

### Comments, logs, screenshots, etc.

...

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

gravyboat:
Did you try using `http` as noted in the error?

soredake:
Yes, same error.

beardypig:
Have you tired putting quotes (`"`) around the url? What OS are you using?

soredake:
@beardypig tried, error, Gentoo stable (with some unstable packages).

soredake:
Seems streamlink cannot handle country blocks, youtube-dl returns 
`[ytdl_hook] ERROR: Access Denied. This room is not available to your region or gender.`
streamlink returns
`Invalid URL '': No schema supplied. Perhaps you meant http://?`

beardypig:
Yep, sounds like a bug then, or at least a missing error message :-)

foodtooth:
Same here when trying `$ streamlink -p vlc "http://www.huya.com/gege07" live`

- Latest Arch Linux
- streamlink 0.7.0

gravyboat:
@foodtooth That should be a new issue if one doesn't exist since this is a different plugin.

PR #1708, opened 2018-05-29
plugins.chaturbate: only open a stream if the url is not empty

back-to:
- `""` is not a valid url anymore
- Added **log info** "Stream status: public,offline,hidden,private" ...
- the api doesn't return a geo blocked msg (it's just public),
  so it can't be added but streamlink won't start a invalid stream now

Fixed https://github.com/streamlink/streamlink/issues/1179


beardypig:
Looks fine, I re-ran the AppVeyor build too.

gravyboat:
Thanks @back-to.

Issue #1709, opened 2018-05-30
Possible to automatically stop capturing Twitch live stream after a period of time?


pcolmer:
This is more of a question (in case I've overlooked/missed something) than a feature request … unless I haven't missed something :).

If I'm using streamlink to capture a live Twitch stream, using -o, is there an option that I can use that will get streamlink to stop the capture after a set period of time? I was hoping that the stream itself would stop but it seems to go into a loop, which means that if I'm leaving streamlink to run unattended, I could end up filling up a lot of disc space.

Thanks.

beardypig:
You could try the `--hls-duration` option, eg. `--hls-duration 01:00:00` to stop after 1 hour.

pcolmer:
Unfortunately, that doesn't work on live streams. I think I'll have to do something kludgy like stopping the process at a certain time.

craftwar:
It should stop, the maximum stream time limit is 48hr. I think it may be auto-host or vod.
You can try `--twitch-disable-hosting` to stop when auto-hosting.

beardypig:
@pcolmer, `--hls-duration` should work on live streams - if not, then it's a bug.

pcolmer:
@beardypig if you do `streamlink --help`, it says that `--hls-duration` has no effect on live streams.

beardypig:
@pcolmer I stand corrected - perhaps it can be made to work.

walker9:
@pcolmer I'm using a small GUI to control Streamlink. With it it's possible to make scheduled records like an old videorecorder :) So if you are on Windows you could use it https://www.reddit.com/r/Twitch/comments/4ng19j/how_to_watch_twitch_streams_in_vlc_mediaplayer/

beardypig:
@pcolmer I have submitted a PR that will bring support for `--hls-duration` for live streams. Hopefully it will be merged quickly, and be available soon in the nightly version.

gravyboat:
Going to close this since I've merged #1710. Should be available for testing when the nightly builds, or the next stable release.

PR #1710, opened 2018-05-30
hls: support for live streams when using --hls-duration

beardypig:
@pcolmer asked if it was possible to stop a live stream after a given time in #1709. This PR enhances the functionality of the existing `--hls-duration` option to include support for live streams. 

Had to fix a test too, it didn't seem to be working as intended anyway.


gravyboat:

PR #1712, opened 2018-05-30
win-installer: add missing isodate module

beardypig:
My bad, didn't update the windows installer... thanks @mmetak


gravyboat:

PR #1714, opened 2018-05-30
plugins.onetv: fix issues with ctc channels and add DASH support

beardypig:
The "ctc" channels weren't working, and now DASH streams are supported.


gravyboat:
Thanks @beardypig.

PR #1715, opened 2018-05-30
logging: set the log level once the plugin config files have been loaded

beardypig:
As @back-to to pointed out (https://github.com/streamlink/streamlink/pull/1690#pullrequestreview-124594681) there was an issues with changing the `loglevel` when using plugin specific configs. This PR resolves that by resetting the `loglevel` after the plugin configs have been loaded, before that point the `loglevel` will still be the default/set by command line.


gravyboat:
Good catch @back-to, thanks for the quick fix @beardypig.

PR #1716, opened 2018-05-30
plugins.dailymotion: fix error logging for stream errors

beardypig:
Should give better error logging for offline channels, and other errors.


gravyboat:
Looks good, thanks.

PR #1718, opened 2018-05-31
stream.dash: fix stuttering streams & maybe high CPU load

Billy2011:


beardypig:
Looks good, can't remember why that's in there in the first place - probably a copy-paste. Can remove the import too `from streamlink.stream.wrappers import StreamIOIterWrapper`.

Billy2011:
OK :)

gravyboat:
Good stuff, thanks @Billy2011!

PR #1720, opened 2018-05-31
logging: fixed issue with logging from plugins using logging module

beardypig:
Previously plugins would be loaded with the module name set to the plugin name, eg. `twitch`, this PR changes that so that plugins are loaded within the `streamlink.plugin` hierarchy. 
I also tweaked the way that the plugin load setup is called (I am slowly chipping away at the global variables). There is also now a debugging log message when a plugin is overridden by an external plugin file - this should help us when debugging issues with plugins, in case the user is using a non-standard plugin version.


gravyboat:
This is a pretty nice improvement, good idea @beardypig.

PR #1722, opened 2018-05-31
plugins.reshet: support for reshet.tv live and VOD streams

beardypig:
Includes a small enhancement to the brightcove player class.


gravyboat:
I honestly don't even know where you are finding these sites at this point. Good contribution though! Thanks @beardypig.

PR #1721, opened 2018-05-31
plugins.funimationnow: login support and bug fixes

beardypig:
This PR adds the ability to login. Premium content should be available, however I do not have a subscription so I cannot test it. 

Fixes #789, pinging @vinnyA3. 

If #1724 is merged, I'd like to update this plugin to store the incap cookies.

PR #1724, opened 2018-06-01
plugins: store cookies between sessions

beardypig:
Adds some new methods to Plugin to simplify the process of storing and loading cookies between sessions. It makes use of the existing cache mechanism, cookies can be filtered before storing and clearing, and are loaded automatically. 

I wanted this feature to store longer lived cookies for anti-robot scripts, eg. in #1721. It can also be used to store auth cookies.

### DIFF
--- a/src/streamlink/cache.py
+++ b/src/streamlink/cache.py
@@ -90,5 +90,23 @@ class Cache(object):
         else:
             return default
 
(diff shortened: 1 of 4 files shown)


gravyboat:
Looks good, thanks @beardypig and thanks for the review @back-to.

Issue #1567, opened 2018-03-22
questions about commands to close streamlink asap after error


CplSgt82:
Hello, after I have read through the command-line interface tutorial, I have some questions about some of the commands in streamlink.

Situation: youtube live streams will get a 403 error after streaming for 6hrs, so I tried to write a bat file to automatically launch a new streamlink. I've succeeded in doing that, but it takes 1min 10s for the new streamlink to launch because the old streamlink keeps reloading the playlist.

Question:  if i want the 'old' streamlink to close asap after it encounter the 403 error at 5:59:59, which one of the following should i include in the command line? or is it some other commands ?

--hls-playlist-reload-attempts=0 (can i set to 0,or is 1 smallest?)

--hls-timeout=1 (means 1 sec right?)

sorry if i am asking noob questions, because its my 1st time trying out all this.

back-to:
isn't it easier to close streamlink after 5:59:59 and reopen it with a script,
instead of waiting for streamlink to close itself.

sbernsteinHBO:
Or even easier -- why not just launch a new streamlink in a different process/batch file at 5:59 before the other one even has a chance to close???

johnthecracker:
the issue is this:

I run streamlink, stream ends and streamlink is still trying for 1 minute to access
no longer existing stream.

What switch do I have to use to terminate streamlink when the stream comes to an end,
instead of waiting for streamlink for 1 minute ?

After 1 minute batch file continues running, 1 minute lost.

This is how my batch file looks like:

:start
set url=http://x.x.x.x/website/playlist.m3u8
streamlink.exe --http-header %agent% hlsvariant://%url% best --stdout | ffmpeg.exe -hide_banner -i pipe:0 -c:v copy -c:a aac c:\%stream%_%d%.flv
timeout /t 30
goto start

back-to:
#2626

Issue #1231, opened 2017-09-18
HLS plugin is failing on multiple streams


sou7611:
Seeing read timeout errors, forcing streamlink to exit across multiples of HLS sources.

Been able to verify on foxsportsgo, watchespn, nbcsports, mls-live, and many others.

Run flawlessly for a period of time, then abruptly exits with "Read timeout; exiting"

There is no network interruption at all, nothing on the PC, nothing to interrupt the stream.

This happens with latest 0.8.1 version.

gravyboat:
We're going to need more details before we can provide any support on this (as noted in the template), can you rerun the commands with verbose output and debug logging please? Add  `-v` and debug `-l debug` to your command.

back-to:
@sou7611 
you could try to use `--hls-timeout` and all the other commands https://github.com/streamlink/streamlink/issues/1146#issuecomment-320471910
https://streamlink.github.io/cli.html#cmdoption-hls-timeout

but you need to provide **more details**

- do you give streamlink all the required cookies?
- does your url expire after some seconds?

sou7611:
Still seeing this even with latest version.

Yes all required cookies present, the URL does not expire.

You can start multiple copies and watch in MPC-HC and the stream keeps going, while it times out and stops dead while recording.

Doesn't matter the source, happens on EVERY SINGLE hls source ... espn, nbc, you name it, this happens on all of them.

Streamlink is not correctly restarting when whatever happens, happens, no mater what you do with hls-timeout or any of the other commands.

sou7611:
For example, saw this multiple times tonight, on NBC ... which on Comcast/Xfinity doesn't even require a cookie (it authenticates from your IP) this is adding every flag I can think of to try and buffer it through whatever's causing the problem.  Over the course of a 4 hour football game, had a dozen + read timeouts.

Could plainly see other windows playing the same exact stream not affected, so it wasn't network or source related.  Was clearly streamlink failing.

[...]

Issue #664, opened 2017-03-03
Question About Tvplayer


willbrow:
Hi i have two questions please forgive me as i am new to streamlink:

Ive noticed that after a few hours that the stream stops piping to VLC player and in CLI it is showing by the looks of it that the KEY has expired by the looks of it ofwitch makes sence for them to expire your session after a certain amount of time.

Anyway my question - Is there anyway to set streamlink so that if a connection fails or expires to run the CLI command again that was entered to get a new keyid and start the pipe again to VLC with same port number.I have seen option where if connection is lost to keep trying but not one to submit the streamlink --tvplayer-email --tvplayer-password tvplayer.com/watch/bbc best and pipe the stream on the same port.

My second question is if i had other devices in my house on the same network and wanted to watch the stream on them via restreaming is there anyway to do this,  I have tried VLC setting source as IP:PORT but does not seem to work so maybe there is setting i need to change or different player will re-stream better.I have also tried connection the devies directly to 127.0.0.1:port 

Thanks for everyones help

beardypig:
We can create a new `Stream` implementation that refreshes the URL when it expires. Alternatively `--player-continuous-http` might work for you. 

VLC supports creating an RTSP for streaming video to multiple clients, there is a lot of documentation online about doing that :)

willbrow:
Hi 

Is it possible to create new stream implementation automatically when the old one expires or is the only way manually ?
With VLC you need to select the source to stream via rtsp or other protocols and when you enter the source IP:Port it is not able to stream maybe this is becuase the video is being piped im not sure but i will keep trying to find a solution 

thanks

beardypig:
By new `Stream` implementation, I mean write some new code to handle it automatically :) 

I thought there was a way to start a stream while it was playing, but seems like I am mistaken.

willbrow:
I have tried everything i can think off to stream through vlc .

It must be something to do with rtmpdump or whatever pipes the video to vlc

back-to:
#2626

Issue #1380, opened 2017-12-17
YouTube stream recording freezes after exactly 6 hours


Matasx:
Hi,

YouTube stream recording freezes after exactly 6 hours. I happened to me twice already. As you can imagine, it is difficult to investigate in reasonable ammount of time. Unfortunatelly I do not have any logs handy that might help. The recorded video is exactly 6h long which is strange, but the recording process does not quit, it rather hangs.

Any ideas? Am I missing some settings? I tried to go through all the settings but nothing seemed to be related.

Any help would be much appreciated. Thanks.

gravyboat:
This isn't something on our end. We've had some previous discussion on this before: https://github.com/streamlink/streamlink/issues/1058 with another service and it's most likely that Youtube is just invalidating some key that is associated with the stream after 6 hours per connection, they close connections that have been open for 6 hours, or that the URL of the stream changes after 6 hours (there are a few youtube livestreams that do this) and the one you have is no longer live. I'm going to mark this as out of scope and won't fix since we can't do anything about it. You could try to write a small wrapper around Streamlink so that it just runs another Streamlink command call to the stream when it is finished and stacks them up as an option though.

Matasx:
I have such wrapper in place, the problem is that streamlink does not quit when this occurs so detecting this case on my end is difficult.

gravyboat:
What does the log output say at the end of the 6 hours if you run the command with `-l debug` and `-v` for debug output and verbose logging?

Matasx:
Ok I will try to do that, but will take me a while to get back with results...

Joshfindit:
Spitballing: is it possible your drive is running in to a max file-size problem?

At common bitrates, 6hr is very close to 4GB.

Matasx:
No, it is already over 4GB and I am using NTFS.

back-to:
type `streamlink https://www.youtube.com/watch?v=9c_Bac-17Rk best --json`

inspect the url, there should be something like this.
`expire/1515120411`

https://duckduckgo.com/?q=unix+time+1515120411&ia=answer

the youtube url is just **6h valid**

---

requesting a new and replacing the old url after **X time** could solve this,
 but i don't know if there is a decent way to do it in streamlink.

Matasx:
It turned out that streamlink is exiting "properly" after 6h. The issue was on my part. So no issue here, if exiting after source expiration is considered to be correct.

PR #1725, opened 2018-06-01
hls: new option session reload

back-to:
**TL;DR**

New command to replace the current hls stream with a new one.

`--hls-session-reload-time` New session after a given time.

`--hls-session-reload-segment` New session if a normal playlist fails twice.

`--hls-segment-ignore-number` Allow invalid segment numbers

also reload on **StreamError** 

---

`--hls-session-reload-time`

It will **replace** the current stream with a new **stream**
after the given time, **self.session** will be **updated** at the same time,

The command must **cache** the stream **url** and the stream **name** somewhere in the self.session,
as it's only available in **streamlink_cli**

If someone want's to reuse this command from an API, it must be done there aswell
or it will print a warning message.

**Example**

HLS-URL timesout after 10 minutes
SESSION-RELOAD is set to 9 minutes

The stream will be reloaded/replaced every 9 minutes,
it can never hit the 10 minutes timeout.

---

`--hls-session-reload-segment`

This can be used as a fallback, for wrong set-up time with the other command

and also for errors like this:

```
error: Error when reading from stream: Read timeout, exiting
[console][info] Stream ended
[console][info] Closing currently open stream...
```

when the stream gets replaces by an invalid stream

normal playlist

```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-ALLOW-CACHE:YES
#EXT-X-TARGETDURATION:5
#EXT-X-MEDIA-SEQUENCE:41031
#EXTINF:5.000000,
44-41031.ts
#EXTINF:5.000000,
44-41032.ts
#EXTINF:5.000000,
44-41033.ts
```

invalid replacement

```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-ALLOW-CACHE:NO
#EXT-X-TARGETDURATION:10
#EXT-X-MEDIA-SEQUENCE:0
#EXTINF:10.000000,
errors/403/403-000.ts
#EXTINF:10.000000,
errors/403/403-001.ts
#EXTINF:10.000000,
errors/403/403-002.ts
#EXTINF:0.280000,
errors/403/403-003.ts
#EXT-X-ENDLIST
```

---

`--hls-segment-ignore-number`

This allows invalid segment numbers.

required for some streams
if they get different segment numbers on different playlists.

---

allow **seconds passthrough** with **hours_minutes_seconds**

---

**Known issues**

1. plugins with authentication

  It might error because you are already logged in,
  the plugin can't find items that are only presented for guests.

2. MuxedStream

  not implemented, lack of test streams and might not work the same way like this.

3. might not work because of different segment numbers

  use `--hls-segment-ignore-number`

---

I tried to create tests with unittest/requests_mock,

but it does not really work with it.

So for now there are no tests.

---

It's completely build as **opt-in**, so if it's not used it can't error anything.

---

it might need some namechanges, if these are not wanted

---

can be used for this
Fixed https://github.com/streamlink/streamlink/issues/1567

Not directly what was asked but it can be done with this differently
Fixed https://github.com/streamlink/streamlink/issues/664

should allow a reload here
Fixed https://github.com/streamlink/streamlink/issues/1231

ref https://github.com/streamlink/streamlink/issues/1058
ref https://github.com/streamlink/streamlink/issues/1370
ref https://github.com/streamlink/streamlink/issues/1380

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -146,6 +146,12 @@ def hours_minutes_seconds(value):
     :param value: hh:mm:ss
     :return: seconds
     """
+    try:
+        if int(value):
+            return value
+    except ValueError:
+        pass
+
     match = _hours_minutes_seconds_re.match(value)
     if not match:
         raise ValueError
@@ -762,6 +768,48 @@ def build_parser():
         Note: The --hls-timeout must be increased, to a time that is longer than the ignored break.
         """
     )
+    transport.add_argument(
+        "--hls-segment-ignore-number",
+        type=num(int, min=5),
+        metavar="SEGMENTS",
+        help="""
+        Ignore invalid segment numbers,
+        this option is the max. difference between the valid and invalid number.
+
+        If the valid segment is 100 and this option is set to 20,
+
+        only a segment of 1-80 will be allowed and added,
+        everything between 81-99 will be invalid and not added.
+
+        Default is Disabled.
+        """
+    )
+    transport.add_argument(
+        "--hls-session-reload-segment",
+        action="store_true",
+        help="""
+        Reloads a Streamlink session, if the playlist reload fails twice.
+
+        Default is False.
+
+        Note: This command is meant as a fallback for --hls-session-reload-time
+        if the time is set incorrectly, it might not work for every stream.
+        """)
+    transport.add_argument(
+        "--hls-session-reload-time",
+        type=hours_minutes_seconds,
+        metavar="HH:MM:SS",
+        default=None,
+        help="""
+        Reloads a Streamlink session after the given time.
+
+        Useful for playlists that expire after an amount of time.
+
+        Default is Disabled.
+
+        Note: --hls-segment-ignore-number can be used for new playlists
+        that contain different segment numbers
+        """)
     transport.add_argument(
         "--hls-audio-select",
         type=comma_list,
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -147,8 +147,7 @@ def hours_minutes_seconds(value):
     :return: seconds
     """
     try:
-        if int(value):
-            return value
+        return int(value)
     except ValueError:
         pass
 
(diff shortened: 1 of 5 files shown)


sbernsteinHBO:
As someone who understands Streamlink very well and has a pretty deep understanding of all of the arguments, I find the description of these new arguments to be very confusing/vague - especially "--hls-session-reload-time" -- I don't quite understand what that would do/be useful for?

beardypig:
Definitely a useful feature, I wonder if it is something that we could generalise a bit more. That's something that requires a bit of planning though...

src/streamlink_cli/argparser.py
```diff
@@ -146,6 +146,12 @@ def hours_minutes_seconds(value):
     :param value: hh:mm:ss
     :return: seconds
     """
+    try:
+        if int(value):
```
beardypig: I think you could just do `return int(value)` here, normally it returns an `int` and 0 would be a valid result too.

back-to:
@sbernsteinHBO 
_of course everything could be renamed, to something different_

**_but the feature of this command is_**

that streamlink will **restart/reload** your terminal command

this will
- update your `session`
- overwrite your streaming url

It's usefull for streams that expires after a short time,
they would timeout and streamlink would close them.

But with this command it **won't timeout** it will replace the expired stream before it can timeout,
if you provide a timestamp before the timeout.

---

@beardypig 

> generalise

well everything can be renamed to something else

back-to:
> we could generalise a bit more. That's something that requires a bit of planning though...

**let's do it, like you wanted it to**

streamlink needs some different api endpoints for this though

as this might take longer and it will probably incompatible, **I will close this**.

---

I recreated this feature as a custom plugin with the new Plugin Arguments API,
so I don't really need it in the master branch right now

beardypig:
@back-to OK, the `hours_minutes_seconds` change was worth including though - if you want to push that to master. 

I don't know how to organise it, but I think we need to do some planning for the `Stream` API and how to solve problems like #1048 and #1055.

Issue #1723, opened 2018-06-01
Request: Add more functions to facebook plugin


roshavagarga:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description
Reminder that with the new initial support of Mpeg Dash #880 and #990 might be fixable now, depending on what streamlink supports and how Facebook's videos and livestreaming has changed since this was last looked it.

back-to:
there dash stream does not work with streamlink

```
[console][info] Found matching plugin dash for URL https://video.fhaj1-1.fna.fbcdn.net/.../live-dash/dash-abr4/000.mpd?...
[plugin.dash][debug] Parsing MPD URL: https://video.fhaj1-1.fna.fbcdn.net/.../live-dash/dash-abr4/000.mpd?...
[console][info] Available streams: 240p (worst), 360p, 480p, 720p (best)
[console][info] Opening stream: 720p (dash)
[stream.dash][debug] Opening DASH reader for: live-hd1-v
[stream.dash][debug] Opening DASH reader for: live-md-a
[stream.dash_manifest][debug] Generating segment timeline for dynamic playlist (id=live-hd1-v))
[stream.dash][debug] Reloading manifest (live-hd1-v)
[stream.dash_manifest][debug] Generating segment timeline for dynamic playlist (id=live-md-a))
[stream.dash][debug] Reloading manifest (live-md-a)
[stream.mp4mux-ffmpeg][debug] ffmpeg command: /usr/bin/ffmpeg -nostats -y -i /tmp/ffmpeg-18118-462 -i /tmp/ffmpeg-18118-804 -c:v copy -c:a copy -copyts -f matroska pipe:1
[stream.ffmpegmux][debug] Starting copy to pipe: /tmp/ffmpeg-18118-462
[stream.ffmpegmux][debug] Starting copy to pipe: /tmp/ffmpeg-18118-804
[console][debug] Pre-buffering 8192 bytes
[stream.dash][debug] Download of segment: .../live-dash/live-hd1-v/000-init.m4v complete
[stream.dash][debug] Download of segment: .../live-dash/live-md-a/000-init.m4a complete
[stream.dash][debug] Download of segment: .../live-dash/live-hd1-v/000-11894667.m4v complete
[stream.dash][debug] Download of segment: .../live-dash/live-md-a/000-11894667.m4a complete
[...]

PR #1726, opened 2018-06-01
dash: fix for manifest reload - should be more reliable

beardypig:
Should fix the issue @back-to pointed out here: https://github.com/streamlink/streamlink/issues/1723#issuecomment-393961866

It would be good to get some people to test dash streams with this change, I tested a few :)


gravyboat:
Nice, thanks @beardypig.

PR #1729, opened 2018-06-02
stream.akamaihd: fix some log.debug... issues

Billy2011:


beardypig:
Good catch @Billy2011 :) I thought I had checked for any using kwargs...

PR #1732, opened 2018-06-02
tests: coverage on src instead of the modules

beardypig:
This appears to fix an issue with counting the coverage of the tests.


gravyboat:
Good catch @beardypig.

beardypig:
@back-to is due the credit for this one

Issue #1551, opened 2018-03-14
PotPlayer can handle stdin mode


CrendKing:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

I noticed that the [Players](https://streamlink.github.io/players.html) page lists PotPlayer unable to handle stdin pipe mode. This is not true. PotPlayer does have a built-in pipe streaming source filter.

There is a trick though. I found that the stdin ( - ) can only be used as input if the working directory is user's home directory (`$USERPROFILE`).

Please update the page for this information.

### Versions

Streamlink: 0.11.0
PotPlayer: 1.7.8557

### Reproduction steps / Explicit stream URLs to test

Example PowerShell command to launch the URL from clipboard:

`powershell -ExecutionPolicy Bypass -Command "Start-Process -NoNewWindow -WorkingDirectory $env:USERPROFILE \"<Streamlink_path>\Streamlink.exe\" \"--player <PotPlayer_path>\PotPlayerMini64.exe $(Get-Clipboard) best,source\""`

### Comments, screenshots, etc.

![player](https://user-images.githubusercontent.com/975235/37387936-b08e60a2-271b-11e8-840d-6901614f5e75.png)
![command](https://user-images.githubusercontent.com/975235/37387938-b1b7d436-271b-11e8-9157-b6fc4ee5d927.png)

bastimeyer:
@CrendKing is there an official documentation for this or can someone else confirm this please?
In either case, please open a PR with the documentation changes. The file you're looking for is `docs/players.rst`.

Thanks!

CrendKing:
I could not find document. I found this with some trial and errors. Would be nice if someone can confirm.

bastimeyer:
> I could not find document.

I was talking about the file of Streamlink's player documentation:
https://github.com/streamlink/streamlink/blob/master/docs/players.rst
If you want it to be improved, submit your changes in a new pull request...

CrendKing:
I meant I could not find official documentation from PotPlayer about the strange behavior. Getting a second confirmation from someone else would be nice.

Ro-Den:
Thanks, for me (Windows XP SP3 x86) PotPlayer v1.7.8557 STDIN works without additional settings.

PR #1737, opened 2018-06-03
docs: PotPlayer Stdin Pipe

back-to:
It works **without any special settings**,
even on older versions.

Fixed https://github.com/streamlink/streamlink/issues/1551


gravyboat:
Good stuff, thanks @back-to.

fireattack:
Just fyi, it seems Potplayer ver. 201209 brok stdin pipe, or at least streamlink.

PR #1748, opened 2018-06-04
dash: fix --json for dash streams and allow custom headers

beardypig:
Fixes the `--json` option when using `DASHStream`s. Enables custom headers to be added to the DASHStream request, eg. `User-Agent`, as per the other `Stream` classes.


beardypig:
I should add some tests :)

gravyboat:
Thanks for adding the tests @beardypig!

PR #1752, opened 2018-06-05
logging: when using the trace level, log the timestamp

beardypig:
Add a timestamp to the logs when using `--loglevel trace`.


gravyboat:
This is a good addition, thanks @beardypig.

PR #1754, opened 2018-06-06
help: reformat all the help text so that it is <80 chars

beardypig:
On smaller consoles I noticed that the help text gets a bit garbled with text wrapping, so I fixed it up a bit :)

And I noticed a mistake here: https://github.com/streamlink/streamlink/compare/master...beardypig:docs-80-chars?expand=1#diff-2b258c59b2cac0b01e6240cd4024be63R244

There are probably other mistakes, I might spend some time reading the argument help again :)

### DIFF
--- a/src/streamlink/plugins/abweb.py
+++ b/src/streamlink/plugins/abweb.py
@@ -34,7 +34,10 @@ class ABweb(Plugin):
             required=True,
             requires=["password"],
             metavar="USERNAME",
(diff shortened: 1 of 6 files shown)


gravyboat:
Took a quick look over everything and it looks good.

beardypig:
@gravyboat I think this is OK to merge.

gravyboat:
@beardypig Yep thanks for the reminder and the contribution.

PR #1756, opened 2018-06-06
Release 0.13.0

cdrage:


cdrage:
@beardypig So steps: 

1. Merge this
2. Create tag?

beardypig:
Exactamundo :) Did you enjoy the script?

cdrage:
Way easier! :+1:

beardypig:
Now you can create the tag, and the script will automatically update the change log :)

cdrage:
@beardypig https://github.com/streamlink/streamlink/releases/tag/0.13.0 there we go!

beardypig:
Failed :(
Bug in the release script... boo

beardypig:
We can try again, it didn't deploy :p

amurzeau:
Hi,

I see the streamlink-0.13.0.tar.gz was signed with 0xE3DB9E282E390FA0 key, but I cannot find any information about that one. It seems missing on key servers.
Before 0.13.0, the used key was 0xDA227403C037D617 (from Charlie Drage)

Where can in find 0xE3DB9E282E390FA0 key if it is the correct one ?

Note: I use this to verify the downloaded .tar.gz when updating the Debian package.

amurzeau:
Ah just found it in https://github.com/streamlink/streamlink/commit/c408ec6deec0abd28d98a2f64f674a77c4807ff9#diff-3dbec4e55416f7a240f3902b6e4250f7
:)

amurzeau:
Just found that this one is a encrypted private key.
Is its public key available to check signatures ?

beardypig:
@amurzeau I will publish the public key to the MIT key server, and to this issue.

gravyboat:
@amurzeau Hmm does it? I wasn't aware of that and they don't have any details on the FAQ. If that's the case we should provide links to the other key servers that it will be hosted on.

beardypig:
Yep, it will propagate to a few key servers, the Ubuntu Key Sever too. We can add the public in the repo too.

amurzeau:
Did you already add it ? MIT pgp website still says key not found.

beardypig:
Not yet, I won’t have access to the key until tomorrow, or maybe late tonight.

gravyboat:
@amurzeau Good catch, I skipped that item in the FAQ since it was related to removal. Let's just make sure we note multiple key servers so that we're covered and people won't get confused if they can't reach it for some reason.

beardypig:
@amurzeau I have published it to the MIT and Ubuntu key servers.

gravyboat:
It's `2E390FA0`: https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xE3DB9E282E390FA0 for anyone that may need to look it up. Thanks for uploading to the Ubuntu keyserver @beardypig. I couldn't even perform a search on the MIT keyserver without getting a 502 proxy error to show the actual key.

amurzeau:
thanks @beardypig :)

PR #1757, opened 2018-06-06
scripts: fix bug in the release for generating the CHANGELOG

beardypig:
The changelogs should have `##`, there was a bug in the release script where it only generated one. 

@cdrage we could merge this, delete the old tag and then retag.


cdrage:
Got it! :+1:

beardypig:
Still need to update the tag (delete and retag).

PR #1760, opened 2018-06-07
plugins.bbciplayer: fix bug is DASH for VOD

beardypig:
Updated all the logging too.

Bug was here: https://github.com/streamlink/streamlink/compare/master...beardypig:bbciplayer-fix?expand=1#diff-9255f1b591e686fe40d95231404a723bL162


gravyboat:
Nice catch, thanks!

PR #1764, opened 2018-06-08
plugins.earthcam: Fixed hls_url - No schema supplied.

back-to:


gravyboat:
Good stuff, thanks @back-to!

Issue #1223, opened 2017-09-15
windows - pynsist: old files won't be removed after an update.


sou7611:
Streamlink is mis-identifying hls streams.

Given the same exact hls stream (tested with foxsportsgo and watchespn), Streamlink will sometimes select "stream" for the plugin, other times it will select "hls" for the plugin.

The command line option given for URL starts with "hls://urltostream", so it should always be using the hls plugin ... right?

this is causing stream failures when the wrong plugin is selected instead, and it read-timeouts and exits.

example:

Try 1 : [cli][info] Found matching plugin stream for URL hls://hlslinear-akc.med1.foxsportsgo.com/{lotsofstuff}/5000K/5000_slide_ads.m3u8
[cli][info] Available streams: live (worst, best)
[cli][info] Opening stream: live (hls)

Try 2 : [cli][info] Found matching plugin hls for URL hls://hlslinear-akc.med1.foxsportsgo.com/{lotsofstuff}/5000K/5000_slide_ads.m3u8
[cli][info] Available streams: live (worst, best)
[cli][info] Opening stream: live (hls)

beardypig:
The `stream` plugin was removed, it possible that it's still kicking around your installation. I would recommend that you uninstall streamlink then reinstall it to make sure you get a clean install :-)

sou7611:
Ok, that explains that.  Not good, however, that the installers are failing to remove obsolete files.

Streamlink blindly loads EVERYTHING in the plugins directory, so if an obsolete file is left there on upgrade, bad things -- like this -- can occur.

gravyboat:
I double checked my install as well and the `stream` plugin did still exist, on Windows at least.

sou7611:
Yes, and I'm also using windows (and it was still in there on all of mine)

back-to:
it's a known issue from https://github.com/takluyver/pynsist/issues/66

**old files won't be removed after an update**

PR #1766, opened 2018-06-08
plugins: marked or removed broken plugins

back-to:
created a dummy file for deleted plugins #1223

_Removed plugins_
- are dead
- redirect to a different website
- are marked as broken for +1 year
- agreed to delete it in some other issue

---

some of the marked plugins could also be deleted


gravyboat:
This is a really nice addition, thanks @back-to. I'll leave it open in case anyone else has input then merge it.

beardypig:
Looks good @back-to :)

PR #1771, opened 2018-06-09
[plugins.rtbf] Fix radio streams + DASH support

melmorabity:
This PR:
* brings DASH stream support
* fixes radio stream retrieving


gravyboat:
Looks good, thanks @melmorabity.

PR #1773, opened 2018-06-11
logging: rename logger for main back to 'cli'

beardypig:
Perhaps we should do a `0.13.1` release with just this fix - can branch from `0.13.0` (`release-0.13.x`?), apply this fix and tag on the release branch.


gravyboat:
We haven't had many items merged since 0.13.0 so I don't know if it's worth the extra effort to cherry pick this one PR. We should definitely do a new release though.

gravyboat:
@cdrage Would you mind running another release now that this is in? We need this merged back in so @bastimeyer isn't suffering with a bunch of users getting errors on the GUI.

PR #1782, opened 2018-06-12
plugins.cnews : separate CNEWS data from CanalPlus plugin.

BZHDeveloper:
Changes in the CanalPlus plugin would be too ugly and complex, so I split plugin and move the CNEWS part in a new plugin.

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -43,8 +43,8 @@ btsports                sport.bt.com         Yes   Yes   Requires subscription a
 btv                     btv.bg               Yes   No    Requires login, and geo-restricted to Bulgaria.
 cam4                    cam4.com             Yes   No
 camsoda                 camsoda.com          Yes   No
-canalplus               - mycanal.fr         No    Yes   Streams may be geo-restricted to France.
-                        - cnews.fr
+canalplus               mycanal.fr           No    Yes   Streams may be geo-restricted to France.
+cnews                   cnews.fr             Yes   Yes
 cdnbg                   - tv.bnt.bg          Yes   No    Streams may be geo-restricted to Bulgaria.
                         - bgonair.bg
                         - kanal3.bg
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -44,7 +44,6 @@ btv                     btv.bg               Yes   No    Requires login, and geo
 cam4                    cam4.com             Yes   No
 camsoda                 camsoda.com          Yes   No
 canalplus               mycanal.fr           No    Yes   Streams may be geo-restricted to France.
-cnews                   cnews.fr             Yes   Yes
 cdnbg                   - tv.bnt.bg          Yes   No    Streams may be geo-restricted to Bulgaria.
                         - bgonair.bg
                         - kanal3.bg
@@ -57,6 +56,7 @@ cinergroup              - showtv.com.tr      Yes   No
                         - showmax.com.tr
                         - showturk.com.tr
                         - bloomberght.com
+cnews                   cnews.fr             Yes   Yes
 crunchyroll             crunchyroll.com      --    Yes
 cybergame               cybergame.tv         Yes   Yes
 dailymotion             dailymotion.com      Yes   Yes
--- a/src/streamlink/plugins/canalplus.py
+++ b/src/streamlink/plugins/canalplus.py
@@ -15,8 +15,7 @@ class CanalPlus(Plugin):
     _url_re = re.compile(r'''
         (https|http)://
         (
-            www.mycanal.fr/(.*)/(.*)/p/(?P<video_id>[0-9]+) |
-            www\.cnews\.fr/.+
+            www.mycanal.fr/(.*)/(.*)/p/(?P<video_id>[0-9]+)
         )
 ''', re.VERBOSE)
     _video_id_re = re.compile(r'(\bdata-video="|<meta property="og:video" content=".+?&videoId=)(?P<video_id>[0-9]+)"')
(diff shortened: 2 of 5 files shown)


src/streamlink/plugins/cnews.py
```diff
@@ -0,0 +1,26 @@
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import http, useragents, validate
+from streamlink.stream import HDSStream, HLSStream, HTTPStream
+
+
+class CNEWS(Plugin):
+    _url_re = re.compile(r'(https|http)://www.cnews.fr/(.*)')
```
beardypig: `(.+)` would fix the failed test, assuming you don’t want to match nothing?

BZHDeveloper:
@beardypig : fixed :+1:

docs/plugin_matrix.rst
```diff
@@ -43,8 +43,8 @@ btsports                sport.bt.com         Yes   Yes   Requires subscription a
 btv                     btv.bg               Yes   No    Requires login, and geo-restricted to Bulgaria.
 cam4                    cam4.com             Yes   No
 camsoda                 camsoda.com          Yes   No
-canalplus               - mycanal.fr         No    Yes   Streams may be geo-restricted to France.
-                        - cnews.fr
+canalplus               mycanal.fr           No    Yes   Streams may be geo-restricted to France.
+cnews                   cnews.fr             Yes   Yes
```
back-to: sort it correctly

src/streamlink/plugins/cnews.py
```diff
@@ -0,0 +1,26 @@
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import http, useragents, validate
+from streamlink.stream import HDSStream, HLSStream, HTTPStream
```
back-to: don't import unused stuff `validate HDSStream, HLSStream, HTTPStream`

src/streamlink/plugins/cnews.py
```diff
@@ -0,0 +1,26 @@
...
+    _embed_video_url_re = re.compile(r'class="dm-video-embed_video" src="(?P<dm_url>.*)"')
+    _embed_live_url_re = re.compile(r'class="wrapper-live-player main-live-player"><iframe src="(?P<dm_url>.*)"')
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return CNEWS._url_re.match(url)
+
+    def _get_streams(self):
+        # Retrieve URL page and search for Dailymotion URL
+        res = http.get(self.url, headers={'User-Agent' : useragents.CHROME})
```
back-to: could you remove the whitespace before `:`
for flake8 `E203 — whitespace before ':'`

src/streamlink/plugins/cnews.py
```diff
@@ -0,0 +1,26 @@
...
+        match = self._embed_live_url_re.search(res.text) or self._embed_video_url_re.search(res.text)
+        if match is not None:
+            for stream in self.session.streams(match.group('dm_url')).items():
```
back-to: you could just return it `return self.session.streams(match.group('dm_url'))`

src/streamlink/plugins/cnews.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class CNEWS(Plugin):
+    _url_re = re.compile(r'(https|http)://www.cnews.fr(/\w)+')
+    _embed_video_url_re = re.compile(r'class="dm-video-embed_video" src="(?P<dm_url>.*)"')
+    _embed_live_url_re = re.compile(r'class="wrapper-live-player main-live-player"><iframe src="(?P<dm_url>.*)"')
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return CNEWS._url_re.match(url)
```
back-to: not important, but you could just use `cls` instead of `CNEWS` here

beardypig: More important when subclassing, so a good habit to get in to :)

src/streamlink/plugins/canalplus.py
```diff
@@ -15,8 +15,7 @@ class CanalPlus(Plugin):
     _url_re = re.compile(r'''
         (https|http)://
```
beardypig: `https?` will match both these two, same for cnews :)

src/streamlink/plugins/cnews.py
```diff
@@ -1,26 +1,24 @@
...
 
 from streamlink.plugin import Plugin
-from streamlink.plugin.api import http, useragents, validate
-from streamlink.stream import HDSStream, HLSStream, HTTPStream
+from streamlink.plugin.api import http, useragents
 
 
 class CNEWS(Plugin):
-    _url_re = re.compile(r'(https|http)://www.cnews.fr(/\w)+')
+    _url_re = re.compile(r'https?://www.cnews.fr(/\w)+')
```
beardypig: Looks like there is a mistake in the regex... You have a repeating group of `/\w` so it will only match a single character after a `/` (repeated) eg. `/f`, `/f/x`, `/f/x/w/x/t/t`.. which is probably not what you intended :) 
Since you are using `re.match` for the URL test, and if you want to match any page on `cnews` you could do `r'https?://www.cnews.fr/'` and it'll match everything.

beardypig: I didn't see you tests just now :) 

I see that you don't want to match the root path. I would do r'https?://www.cnews.fr/.+', which should match any non-root path on the site.

beardypig:
@BZHDeveloper can you rebase with master?

BZHDeveloper:
done

beardypig:
@gravyboat this is good to merge, if you wanna do the honours.

campones:
hi there

sorry but how is the usage for this plugin?

PR #1784, opened 2018-06-12
[plugins.ert] New plugin, but youtube based

Twilight0:
Why this plugin?
1) I 'd like to practice making sl plugins, hopefully it will be accepted.
2) Youtube ids change frequently
3) Two different streams for Greece and outside Greece, ERT World's stream however is universal.
4) Can be improved to facilitate VOD playback later from the rest of the network (which are ordinary hls streams).

### DIFF
--- a/src/streamlink/plugins/ert.py
+++ b/src/streamlink/plugins/ert.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+
+import re
+import requests
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import http, validate, useragents
+
+YOUTUBE_URL = "https://www.youtube.com/watch?v={0}"
+req = requests.get('http://extreme-ip-lookup.com/json/', params={'User-Agent': useragents.CHROME})
+j_son = req.json()
+
+_youtube_id = re.compile(r'/embed/([\w-]+)(?:.+?/embed/([\w-]+)|)', re.M | re.S)
+
+_url_re = re.compile(r'http(s)?://webtv\.ert\.gr/ert.+?/')
+
+
+class Ert(Plugin):
+
+    @classmethod
+    def can_handle_url(cls, url):
+
+        return _url_re.match(url)
+
+    def _yt_id(self):
+
+        r1 = requests.get(self.url, params={'User-Agent': useragents.CHROME})
+
+        iframe_url = re.search(r'iframe src="(.+?)"', r1.content, re.U).group(1)
+
+        return iframe_url
+
+    def _youtube_url_schema(self, link):
+
+        _youtube_url_schema = validate.Schema(
+            validate.all(
+                validate.transform(_youtube_id.search),
+                validate.any(
+                    None,
+                    validate.all(
+                        validate.get(
+                            2 if j_son['countryCode'] == 'GR' and 'ertworld-live' not in link else 1
+                        ),
+                        validate.text
+                    )
+                )
+            )
+        )
+
+        return _youtube_url_schema
+
+    def _get_streams(self):
+
+        channel_id = http.get(self._yt_id(), schema=self._youtube_url_schema(self.url))
+
+        if channel_id:
+            return self.session.streams(YOUTUBE_URL.format(channel_id))
+
+
+__plugin__ = Ert


Twilight0:
P.S. I am almost certain at this point I am doing something wrong.

src/streamlink/plugins/ert.py
```diff
@@ -0,0 +1,60 @@
+# -*- coding: utf-8 -*-
+
+import re
+import requests
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import http, validate, useragents
+
+YOUTUBE_URL = "https://www.youtube.com/watch?v={0}"
+req = requests.get('http://extreme-ip-lookup.com/json/', params={'User-Agent': useragents.CHROME})
```
back-to: I dislike this random website call, it should be done without this.

Twilight0: I am using it purely for locating user's countrycode, nothing more.

tests/test_plugin_ert.py
```diff
@@ -0,0 +1,23 @@
+import unittest
+
+from streamlink.plugins.ert import Ert
+
+
+class TestPluginErt(unittest.TestCase):
+    def test_can_handle_url(self):
+        should_match = [
+            'http://webtv.ert.gr/ert1/',
```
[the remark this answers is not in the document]
Twilight0: Correct, will fix.

src/streamlink/plugins/ert.py
beardypig: This will be run on import, which is not a good idea. Can this lookup be avoided as well?

Twilight0: Even original website does this lookup via javascript to display certain iframe, geolocked stream. It can be avoided with a try-except block by prioritizing greek geo-locked stream but it will have significant overhead... i.e. trying playing the stream only to fail due to Youtube's denial.

beardypig: If you are going to call that IP lookup service, it must be done on-demand. 

In the case of geo-blocked videos, the YouTube will fail before returning any streams and fairly quickly. I agree with @back-to that the IP lookup should be avoided, if that service goes away or doesn't allow connections from proxy servers then the plugin may fail.

src/streamlink/plugins/ert.py
```diff
@@ -0,0 +1,60 @@
...
+    @classmethod
+    def can_handle_url(cls, url):
+
+        return _url_re.match(url)
+
+    def _yt_id(self):
+
+        r1 = requests.get(self.url, params={'User-Agent': useragents.CHROME})
+
+        iframe_url = re.search(r'iframe src="(.+?)"', r1.content, re.U).group(1)
```
beardypig: There is a method `streamlink.plugin.api.utils.itertags` that is more robust than this regex, you might find it nicer to use that.

beardypig:
Did FIFA DMCA their YouTube channel?!

Vangelis66:
> Did FIFA DMCA their YouTube channel?!

I'm inside Greece and can verify that all other live channels 
```
http://webtv.ert.gr/ert2-live/
http://webtv.ert.gr/ert3-live/
http://webtv.ert.gr/ertworld-live/
http://webtv.ert.gr/ert-play-live/
http://webtv.ert.gr/ert-play-live/
```
still employ the Youtube player...

However, the 1st national channel 
`http://webtv.ert.gr/ert1-live/`
has now switched to a geo-blocked player: 
```
http://webtv.ert.gr/webtv/ert/tv/ytb/ert1_ipinfo-geo.html
http://player.pstatic.gr/phaistos-player/phaistosplayer.js?t=20180617_1830
http://player.pstatic.gr/bitmovin-dev/bitmovinplayer.js
```
... that uses a geo-blocked AppleHLS playlist: 
`http://ertlive.mdc.akamaized.net/ertlive/live1/playlist.m3u8` 
(Akamai CDN, best quality = 848x480p @1400kbps)

If you're not inside Greece, you are being redirected to 
the International variant of ERT1, ERT1-Worldwide: 
`https://www.youtube.com/embed/SNxttZ8YMFo`
which has no rights-protected Sports content...

I suspect `ert1-live` will again revert to Youtube hosting once the World Cup is over... 😜

Twilight0:
@beardypig : They always change some of their channels every summer (3rd time in row) into another CDN other than Youtube when events like Mundial or Olympics are in progress. A good chance for me to try improving the plugin.
@Vangelis66 : Yes as a matter of fact the stream geo-blocked, but I was told there is a way to play it abroad in a way other than deploying a VPN passthrough.

gravyboat:
@Twilight0 Let's be careful about circumventing that. FIFA is an organization of assholes and I REALLY don't feel like dealing with a DMCA notice from them. Is there seriously no way to legally watch this content?

src/streamlink/plugins/ert.py
back-to: some other solution

---

use the IP lookup but use the logger with a **warning** about it
and use it somewhere in `class Ert(Plugin):` not at the top of the ert.py file

**cache the countrycode** for 30 days or more

---

use a command for the  countrycode `--ert-country XX`

Twilight0: I am not sure how to do that.

PR #1785, opened 2018-06-12
sdist: include tests resources

amurzeau:
Tests resources files were not included in the source archive before, while tests might require
them. For example dash *.mpd files in tests/resources.


beardypig:
@amurzeau thanks, good catch.

PR #1786, opened 2018-06-12
rtlxl: use raw string to fix escape sequences

amurzeau:
With only """, the string is still not a raw string (just a multiline
string) and regex escape sequences were read a string escape sequences
by python, with warnings about unknown escape sequences (like \w).


beardypig:
Those pesky deprecated warnings :)

PR #1787, opened 2018-06-12
tests: freezegun: use object instead of lambda

amurzeau:
Lambda support is a recent addition in freezegun (in the latest release, 0.3.10) and
might not be available everywhere yet (as of now: Debian stable and sid)


beardypig:
Another good fix :)

PR #1794, opened 2018-06-13
Addition of random user agent string function

Twilight0:
Can be utilized to masquerade streamlink and python installation to multiple browser possibilities.

### DIFF
--- a/src/streamlink/plugin/api/useragents.py
+++ b/src/streamlink/plugin/api/useragents.py
@@ -22,3 +22,41 @@ SAFARI_7 = ("Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_3) "
 FIREFOX = "Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:52.0) Gecko/20100101 Firefox/52.0"
 FIREFOX_MAC = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10; rv:33.0) Gecko/20100101 Firefox/33.0"
 OPERA = "Opera/9.80 (Windows NT 6.0) Presto/2.12.388 Version/12.14"
+
+
+def random_agent():
+
+    import random
+    from streamlink.compat import range
+
+    BR_VERS = [
+        ['%s.0' % i for i in range(18, 50)],
+        ['37.0.2062.103', '37.0.2062.120', '37.0.2062.124', '38.0.2125.101', '38.0.2125.104', '38.0.2125.111',
+         '39.0.2171.71', '39.0.2171.95', '39.0.2171.99', '40.0.2214.93', '40.0.2214.111', '40.0.2214.115',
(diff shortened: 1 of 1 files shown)


src/streamlink/plugin/api/useragents.py
```diff
@@ -22,3 +22,41 @@
 FIREFOX = "Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:52.0) Gecko/20100101 Firefox/52.0"
 FIREFOX_MAC = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10; rv:33.0) Gecko/20100101 Firefox/33.0"
 OPERA = "Opera/9.80 (Windows NT 6.0) Presto/2.12.388 Version/12.14"
+
+
+def random_agent():
+
+    import random
```
beardypig: Should import modules at the top, unless there is a specific reason not to?

bastimeyer:
I think the idea of having better UA strings is not bad, but I'm not a fan of this particular solution, as it only provides a set of statically generated UAs, which doesn't solve the issue of using Streamlink "anonymously" / "transparently". The current `CHROME` UA is pretty much outdated and all of Streamlink's HTTP requests using this UA can be identified pretty easily. Same would be true for this suggested approach, but for a set of static UA strings instead of one.

Having something like `CHROME_LATEST` that gets included during build time by some external library like [fake-useragent](https://github.com/hellysmile/fake-useragent) and changes in each Streamlink release would be much better, since those UAs are not unique to Streamlink and are based on up2date real world usage statistics. A random UA string could also be generated by using this lib, if needed.

src/streamlink/plugin/api/useragents.py
Twilight0: Lazy imports tend to work faster and since they are needed only locally in the function that why I put them there. This is only example however because the difference will be negligible anyway.

Twilight0:
@bastimeyer Why add yet another dependency for build or core one when you can add a simple one and update manually from time to time.

back-to:
using random **old** versions makes you **more unique** than the latest stable version

which is exactly the opposite of what you probably wanted

---

https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugin/api/useragents.py

the streamlink useragents should be changed once a month or so,
they are very old which means they are very unique right now.

---

using a dependency is maybe a bit to much for this.

---

`python-requests/2.18.4` is currently the most used User-Agent of Streamlink,
which is fine as it's used very often but it could be also changed to Firefox or Chrome

beardypig:
It might be nice to have a few more choices - but you'd want it to be consistent across requests in the same session. We could randomly assign a default, then plugins can change it as needed.

I don't think we'd want to add `fake-useragent` as a dependency, but you could probably use the output file from that to update the useragents list semi-automatically. Taking the most popular version of Chrome, Firefox, Opera, etc.

PR #1795, opened 2018-06-13
docs.thirdparty: Added LiveProxy

back-to:
LiveProxy allows Streamlink to be easy accessible from **m3u** playlists,
it is also available for **Kodi Leia** and **Enigma2** devices.

---

It is only meant for a local network.

---

The argparser is copied from **streamlink_cli**,
this makes it easier to maintain and also allows the usage of all Streamlink commands.

Most of them will work, some not because they make no sense for LiveProxy.

**supported**

- Sideloading Plugins
- config files
- HLS, HTTP, HDS are supported.

**not supported**

- RTMP is not supported
- Dash will be supported at some point, but not right now.

It can play the stream or redirect the streaming url.

---

Example of URLs

```
http://127.0.0.1:53422/play/?url=https%3A%2F%2Fwww.youtube.com%2Fuser%2Ffrance24

http://127.0.0.1:53422/301/?url=https%253A%252F%252Fwww.euronews.com%252Flive
```

A detailed guide can be found on the website.

### DIFF
--- a/docs/thirdparty.rst
+++ b/docs/thirdparty.rst
@@ -29,6 +29,10 @@
 
 .. include:: _applications.rst
 
+.. |br| raw:: html
+
+  <br />
+
 
 Third Party Applications
 ========================
@@ -36,9 +40,17 @@ Third Party Applications
 
 .. content list start
 
+LiveProxy
+---------
 
-.. replace this line with the first entry
-
+:Description: Local proxy server between Streamlink and an URL
+:OS: |Windows| |MacOS| |Linux|
+:Author: `back-to <https://github.com/back-to>`_
+:Website: https://liveproxy.github.io
+:Source code: https://github.com/back-to/liveproxy
+:Info: LiveProxy allows Streamlink to be easy accessible from **m3u** playlists, |br|
+  it is also available for **Kodi Leia** and **Enigma2** devices. |br|
+  A detailed guide can be found on the website.
 
 .. content list end
 


beardypig:
Nice :)

gravyboat:
Good stuff, thanks @back-to.

Twilight0:
Perhaps it 'd be nice if we can add my streamlink tester and the packages I prepared as well.

beardypig:
@Twilight0, of course. Prepare a PR and we'll get it merged :)

Twilight0:
I 'd like to have a chat session with @back-to if possible in order to combine our efforts into maintaining a single streamlink package for Kodi.

beardypig:
I wouldn't be against streamlink having an official Kodi repo and Kodi package for streamlink 👍 If there are enough developers interested in maintaining it - I might even get back in to it :)

back-to:
well the `streamlink` package (**Streamlink API**) has not that many issues with Kodi

there are only 3 custom patches that I need for it.

https://github.com/back-to/script.module.streamlink/blob/master/replace.py#L7-L36

also **pycryptodome** / **pycryptodomex** must be installed,
every other solution is to slow or does not work correctly.

---

how the **Streamlink API** is used, that is another story.

This depends on the user, what will be used or not.

beardypig:
@back-to do you know if the `cryptography` package is available? I was thinking it would be nicer to switch to that, as it has a better API. I think it's available in most distros, but I dunno about Kodi.

back-to:
@beardypig I don't see `cryptography` anywhere in Kodi, so it is not available by default

beardypig:
Damn, oh well.

PR #1796, opened 2018-06-13
Add arguments to __init__ - super to work on Python 2.7.X

Twilight0:
Super() without arguments in only supported on Python 3.


beardypig:
Looks good

Issue #1516, opened 2018-02-27
[Plugin Request] https://www.tlctv.com.tr


mike8032:
*Thanks for reporting an issue!*
*Please read the contribution guidelines first! (see the link above)*
*Also check the list of known issues before reporting an issue!*

*Feel free to use the following template. Be as detailed as possible.*
*Please see the text preview to avoid unnecessary formatting errors.*
*Don't forget to remove this text before submitting.*

----

### Checklist

- [] This is a bug report.
- [ ] This is a feature request.
- [ x] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description

...

### Expected / Actual behavior

...

### Reproduction steps / Explicit stream URLs to test

1. https://www.tlctv.com.tr/canli-izle
2. ...
3. ...

### Logs

_Logs are always required for a bug report, use `-l debug` [(help)](https://streamlink.github.io/cli.html#cmdoption-l)
Make sure to **remove username and password**
You can upload your logs to https://gist.github.com/ or_

```
REPLACE THIS TEXT WITH YOUR LOG
```

### Comments, screenshots, etc.
would be great to get a plugin for this tv channel
...

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

dreamboxco:
you dont need a plugin for tlc , 
you can see tlc with the original link 

https://pa9agxjg.rocketcdn.com/tlctv.smil/chunklist_b2328000_sltur.m3u8

iyi seyirler

PR #1798, opened 2018-06-13
plugins.tlctr: New Plugin for tlctv.com.tr/canli-izle

back-to:
TV channel from Turkey

https://en.wikipedia.org/wiki/TLC_(Turkey)

Fixed https://github.com/streamlink/streamlink/issues/1516


gravyboat:
Thanks @back-to.

PR #1802, opened 2018-06-15
Move docs version selection to sidebar

### LINKED ISSUE #254: Documentation doesn't match stable version

bastimeyer:
Also:
- Remove home icon from "Streamlink" title
- Fix spacing and text layout in sidebar header and menu
- Remove border from active main menu list item
- Rename CSS classes (wy-side-nav-search -> wy-side-nav-header)
- Remove some unneeded CSS rules

~~**DON'T MERGE**
This is a follow up to #1684 and includes all of its current commits. I will rebase this PR once #1684 gets merged.~~

Btw, the docs theme is an unmaintainable mess and the templates and stylesheets are outdated and just terrible. Upgrading it would break all custom changes which have been applied over the time, though.

Fixes #254 and closes #1684


beardypig:
Cool - I will test it out. TBH, this could be merged instead of the original - doesn't make any difference doing one at a time (I don't think).

amurzeau:
I believe most theme files are generated from https://github.com/chrippa/sphinx_rtd_theme_violet/ which is a modified version of the read-the-docs sphinx theme.
Maybe retrieving this to remove generated files and modify .sass and other sources instead might be better while keeping some compatibility
(I don't know much about sass and other things sphinx theme use but it's an idea ^^)

beardypig:
@amurzeau, ideally we would do that - but it's quite a bit of work. We should create an issue to track it though, so maybe someone will work on it :)

beardypig:
@gravyboat, just merge this one and close #1684 :)

beardypig:
@bastimeyer I messed with your original post a bit... :]

gravyboat:
Looks great @bastimeyer, thanks!

bastimeyer:
Now we need a new stable release to make it fully work:
https://streamlink.github.io/
https://streamlink.github.io/latest/

I'd appreciate a new release anyway because of the issue with the changed log output that can break the error detection of the Twitch GUI :)

gravyboat:
@cdrage Would you mind doing a stable release today if you have time?

cdrage:
@gravyboat Didn't get this message until now, but yes! I'll do a release today. :+1:

PR #1792, opened 2018-06-13
Convert literal comprehensive dicts to dict contructs

Twilight0:
Including beardypig's fix and a couple of cosmetics

PR #1804, opened 2018-06-15
api.utils: hours_minutes_seconds update, twitch automate time offset

back-to:
**hours_minutes_seconds**:

allow more timestamps
- hours:minutes:seconds to seconds
- 11h22m33s to seconds
- 11h to seconds
- 20h15m to seconds
- seconds to seconds

**twitch**:
- automate `hls-start-offset` for urls with timestamps
- update for the old vod url's, which I don't think exist anymore

**youtube**:
- tested on youtube, but did not add it because there are to many http streams

---

I put it into `api/utils.py` but could also move it somewhere else.

---

will break https://github.com/streamlink/streamlink/pull/1792 as it uses the same file

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -13,6 +13,7 @@ from streamlink.plugin.api.utils import parse_json, parse_query
 from streamlink.stream import (
     HTTPStream, HLSStream, FLVPlaylist, extract_flv_header_tags
 )
+from streamlink.utils.times import hours_minutes_seconds
 
 try:
     from itertools import izip as zip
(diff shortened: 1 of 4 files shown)


beardypig:
Looks good to me. I restarted the Travis job, it failed due to network errors on their side.

beardypig:
Maybe this is better in `streamlink.utils`?

back-to:
should it be in a newfile
such as `src/streamlink/utils/times.py`
or
in `src/streamlink/utils/__init__.py` ?

beardypig:
I was debating that with myself too... I'm not sure, maybe nicer to have a separate file :) Which do you prefer?

back-to:
I would go with a **separate file** `times.py`,
the `__init__` file is already too packed.

beardypig:
Agreed.

back-to:
moved it into a new file :octocat:

beardypig:
@gravyboat this is good to go.

gravyboat:
Looks good, thanks @back-to!

PR #1822, opened 2018-06-18
[plugins.pandatv]fix indentation bug and add support for new url pattern

fozzysec:
fix indentation bug and add support for new url pattern 'https://xingxiu.panda.tv/<channel id>'

### DIFF
--- a/src/streamlink/plugins/pandatv.py
+++ b/src/streamlink/plugins/pandatv.py
@@ -6,7 +6,7 @@ import json
 from streamlink.compat import quote
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import http, validate
-from streamlink.stream import HTTPStream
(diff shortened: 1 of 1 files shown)


gravyboat:
Thanks @fozzysec.

stefan327:
I think since some days it does not work anymore with this solution.

Example:
```
streamlink.exe" -o test.flv "https://xingyan.panda.tv/52168" best
[cli][info] Found matching plugin pandatv for URL https://xingyan.panda.tv/52168

error: Unable to open URL: https://xingyan.panda.tv/52168 (HTTPSConnectionPool(host='xingyan.panda.tv', port=443): Read timed out. (read timeout=20.0))
```

Still works without subdomain.
(example: https://www.panda.tv/2201108)

PR #1717, opened 2018-05-30
Steam Broadcast Plugin

beardypig:
The long awaited Steam Broadcast plugin... I wrote this a long time ago (Nov 2016), but never completed the DASH support :)

The only outstanding issue is how to de-couple the 2FA requests from `streamlink_cli`, currently they use the `ConsoleOutput` class to ask the user for the codes. This means importing from `streamlink_cli`, I'd like to find a better solution that this. 

Fixes #173

PR #1827, opened 2018-06-19
API for plugins to request input from the user

beardypig:
I have taken a bash at creating an API for plugins to request input from a user. I have only implemented in solution for the standard `streamlink_cli`, but I have tried to design it in a way that other front-ends can use. 

During the `Plugin.bind` call and additional (optional) parameter is passed - an instance of `UserInputRequester`, the job of which is to provide a way for the user to give an input. Implementations of `UserInputRequester` must implement two methods, `ask` and `ask_password`. `ask` is for users to input non-sensitive data, eg. username. `ask_password` is for users to input sensitive data, eg. password. For example, in the `ConsoleUserInputRequester` the two methods (`ask`, and `ask_password`) are handled by `Console.ask` and `Console.askpass`. 

For a `Plugin` to request input from the user, there are two new methods `Plugin.input_ask` and `Plugin.input_ask_password` that call `UserInputRequester.ask` and `UserInputRequester.ask_password` respectively. If no `UserInputRequester` is given and a `Plugin` tries to request user input, it will raise a `PluginError`. 

As a user of the `Streamlink` API, the `UserInputRequester` instance can set by setting the option `user-input-requester`, using `Sreamlink` options constructor parameter. It must be set when instance of `Streamlink` is created to affect built-in plugins.

I mainly created this as some way is needed for the `SteamBroadcast` plugin (#1717) to request input from the user for 2FA, and we need to maintain the disconnect between `streamlink` and `streamlink_cli`.

### DIFF
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -12,6 +12,8 @@ from streamlink.cache import Cache
 from streamlink.exceptions import PluginError, NoStreamsError
 from streamlink.options import Options, Arguments
 
+log = logging.getLogger(__name__)
+
 # FIXME: This is a crude attempt at making a bitrate's
 # weight end up similar to the weight of a resolution.
 # Someone who knows math, please fix.
@@ -155,6 +157,33 @@ def parse_params(params):
     return rval
 
 
+class UserInputRequester(object):
+    """
+    Base Class / Interface for requesting user input
+
+    eg. From the console
+    """
+    def ask(self, prompt):
+        """
+        Ask the user for a text input, the input is not sensitive
+        and can be echoed to the user
+
+        :param prompt: message to display when asking for the input
+        :return: the value the user input
+        """
+        raise NotImplementedError
+
+    def ask_password(self, prompt):
+        """
+        Ask the user for a text input, the input _is_ sensitive
+        and should be masked as the user gives the input
+
+        :param prompt: message to display when asking for the input
+        :return: the value the user input
+        """
+        raise NotImplementedError
+
+
 class Plugin(object):
     """A plugin can retrieve stream information from the URL specified.
 
@@ -167,14 +196,20 @@ class Plugin(object):
     options = Options()
     arguments = Arguments()
     session = None
+    _user_input_requester = None
 
     @classmethod
-    def bind(cls, session, module):
+    def bind(cls, session, module, user_input_requester=None):
         cls.cache = Cache(filename="plugin-cache.json",
                           key_prefix=module)
         cls.logger = logging.getLogger("streamlink.plugin." + module)
         cls.module = module
         cls.session = session
+        if user_input_requester is not None:
+            if isinstance(user_input_requester, UserInputRequester):
+                cls._user_input_requester = user_input_requester
+            else:
+                raise RuntimeError("user-input-requester must be an instance of UserInputRequester")
 
     def __init__(self, url):
         self.url = url
@@ -482,6 +517,23 @@ class Plugin(object):
                     removed.append(key)
 
         return removed
+
+    def input_ask(self, prompt):
+        if self._user_input_requester:
+            try:
+                return self._user_input_requester.ask(prompt)
+            except NotImplementedError:  # ignore this and raise a PluginError
+                pass
+        raise PluginError("This plugin requires user interaction, however it is not supported on this platform")
+
+    def input_ask_password(self, prompt):
+        if self._user_input_requester:
+            try:
+                return self._user_input_requester.ask_password(prompt)
+            except NotImplementedError:  # ignore this and raise a PluginError
+                pass
+        raise PluginError("This plugin requires user interaction, however it is not supported on this platform")
+
 
 
 __all__ = ["Plugin"]
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -9,7 +9,7 @@ from functools import partial
 from collections import OrderedDict
 
 from streamlink.cache import Cache
-from streamlink.exceptions import PluginError, NoStreamsError
+from streamlink.exceptions import PluginError, NoStreamsError, FatalPluginError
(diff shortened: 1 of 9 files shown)


gravyboat:
This looks pretty good to me @beardypig. I'd like to get @bastimeyer's feedback as well considering the scope.

Twilight0:
I suppose it will able to pass username/password into plugins like zattoo as well. That's great.

beardypig:
@Twilight0 you could, but if you are implementing it as part of a Kodi plugin - that is up to you.

Twilight0:
@beardypig : I was talking about the cli utility as well, ask user input at runtime instead of adding user/pass as arguments. But yeah... Kodi will definitely need some deviation and ask input with the appropriate dialog.

beardypig:
@Twilight0 there is already the feature to request the username/password on the CLI - the Zattoo plugin isn't using it though (only for password, if username is given). The Zattoo plugin could be updated to require a username.

```diff
diff --git a/src/streamlink/plugins/zattoo.py b/src/streamlink/plugins/zattoo.py
index b4ae2d4..403341d 100644
--- a/src/streamlink/plugins/zattoo.py
+++ b/src/streamlink/plugins/zattoo.py
@@ -56,6 +56,7 @@ class Zattoo(Plugin):
         PluginArgument(
             "email",
             requires=["password"],
+            required=True,
             metavar="EMAIL",
             help="""
             The email associated with your zattoo account, required to access
```

src/streamlink/plugin/plugin.py
```diff
@@ -483,5 +518,22 @@ def clear_cookies(self, cookie_filter=None):
 
         return removed
 
+    def input_ask(self, prompt):
+        if self._user_input_requester:
```
back-to: if someone uses **streamlink as a subprocess**,
there might be no console to actually type the value in.

There might be a cancel button if the set up is nicely _(like in https://github.com/streamlink/streamlink-twitch-gui)_,
if not it will run forever in the background until someone cancels it.

---

**_some solutions_**

1. Add an option so this can be disabled,
so it will always return an Error or it will be skipped and the Plugin must handle the Error.

2. Timeout the prompt after X minutes.

beardypig: I guess that is an issue that exists already in `Console.ask` and `Console.askpass`. I think the best thing to do in this case is test to see if there is a `tty` available, and if not raise an exception.

back-to: not sure if this is how it should work with `--retry-open`

or if this is an issue

```
Found matching plugin twitch for URL twitch.tv/...
Plugin specific arguments:
--twitch-disable-hosting=True (disable_hosting)
User input error: no TTY available
Waiting for streams, retrying every 1.0 second(s)
User input error: no TTY available
User input error: no TTY available
User input error: no TTY available
User input error: no TTY available
...
```

it won't stop

beardypig: Probably shouldn’t print that so much, I guess it’s a side effect - which input is it waiting for though?

back-to: this is my testcase with https://github.com/streamlink/streamlink-twitch-gui git version

```diff
diff --git a/src/streamlink/plugins/twitch.py b/src/streamlink/plugins/twitch.py
index d42cba2..65bced6 100644
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -631,6 +631,7 @@ class Twitch(Plugin):
         return streams
 
     def _get_streams(self):
+        self.input_ask('123')
         if self.video_id:
             if self.video_type == "v":
                 return self._get_hls_streams("video")
```

beardypig: Ah ok. Cool, I will take a look - I guess a effect of retry open is that you will get that error repeatedly. Not sure if it can be avoided, if you need to give that input and you cannot pass it via the command line then maybe the plugin cannot run - if it can run without the input then the plugin should handle the error... maybe.

beardypig: I added a `FatalPluginError`, which by default cannot be recovered from - unless the Plugin overrides this behaviour (see https://github.com/streamlink/streamlink/pull/1827#issuecomment-398707513 for an example).

back-to:
your zattoo example might not work as expected,
as there is a _minor timestamp issue_ on the first login with zattoo

but yes it could be used for a relog on a failed login

it could also be used for the stations code in ustvnow
https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugins/ustvnow.py#L77-L78

---

the `FatalPluginError` is a good solution

beardypig:
@back-to I don't know the Zattoo plugin, but it's meant to be the same as using `get_option` - up to you if you want to do it though :)

tests/test_plugins_input.py
```diff
@@ -0,0 +1,59 @@
...
+        p = _TestPlugin("http://example.com/stream")
+        p.bind(self.session, 'test_plugin', UserInputRequester())
+        self.assertRaises(PluginError, p.input_ask, 'test')
+        self.assertRaises(PluginError, p.input_ask_password, 'test')
+
+    def test_user_input_console(self):
+        p = _TestPlugin("http://example.com/stream")
+        with self._mock_console_input() as console_input:
+            p.bind(self.session, 'test_plugin', console_input)
+            self.assertEquals("username", p.input_ask("username"))
```
back-to: `DeprecationWarning: Please use assertEqual instead.`

beardypig:
I think this is good to go. @back-to, did you have any other comments?

gravyboat:
Can you rebase @beardypig? Then I can get this merged.

beardypig:
@gravyboat rebase, all tests should pass :)

gravyboat:
Good stuff, thanks @beardypig, and thanks for the review @back-to.

gravyboat:
@beardypig I cut down the commit message simply because it would have mucked up the history just as a heads up in case you look back at this and wonder where everything went.

Issue #1772, opened 2018-06-09
Usage of ustv plugin ?


aykuter:
Hey guys! I'm struggling with ustvnow plugin and i couldn't figure it out. Username checked password checked but what is station code ? How can i find it ? Is that enough to use it ? Thanks.

beardypig:
Unfortunately you need to know the station code, which is the internal name for the channel. They are a bit cryptic, but if you the url and don’t give a station code it should print them out.

aykuter:
Thanks for the reply. I don't get it should i enter url to streamlink for print them out ? I can't actually get any link ustvnow made you watch all channels in home screen. Can you please give me a little more detail ?

beardypig:
Did you try anything? `streamlink http://watch.ustvnow.com/` with the username and password?

Ordval:
O.K. , really, really sorry to bother you **beardypig**,  but put in a command line (argument?) 
how would it (precisely) look ?
```
C:\Users\Ordval>call streamlink.exe "http://watch.ustvnow.com/" best
Enter ustvnow username: xxxxx@xxx.xx
Enter USTV Now account password:
[console][info] Frdvalound matching plugin ustvnow for URL http://watch.ustvnow.com/
[plugin.ustvnow][error] Station code not provided, use --ustvnow-station-code.
[plugin.ustvnow][error] Available stations are: discovery, whtm, animalplanet, tlchd, espntwo, whvl, sndnc, foxnews, wlyh, nickelodeon, whphd, syfy, natgeo, bbca, wpmt, fxhd, usahd, wpsu, frfrm, amchd, bravo, comedycentral, spiketv
error: No playable streams found on this URL: http://watch.ustvnow.com/ 
```
The input in cmd won't accept a password so I had to leave the field empty.

--ustvnow-email=xxx --ustvnow-password=xxx 
 ustvnow username: xxx  USTV Now account password: xxx  etc., etc. won't work, 
 so how ?

beardypig:
@Ordval that should work, could you file a bug ?

Ordval:
I've never written a bug file nor a report, however, the "Submit new issue"-button is greyed out,
also for the form in its original state.

aykuter:
I tried this one (i see this arguments on streamlink's webpage) :
`streamlink http://watch.ustvnow.com/ --ustvnow-username xxxx@xxxx.com --ustvnow-password xxxxx`
but it's giving me error :
[...]

PR #1840, opened 2018-06-23
USTV Now

beardypig:
Fix for missing NBC channel and better logging of the station codes. 

As discussed in #1772.

@Ordval @aykuter


gravyboat:
Thanks @beardypig, not many changes required which is always nice.

PR #1846, opened 2018-06-25
stream.dash: use the ID and mime-type to identify a representation

beardypig:
This is a work-around for some DASH streams that have duplicate IDs for representations. By using the mimeType as well as the ID, we can better identify the different streams. Naughty Valve.


gravyboat:
Thanks @beardypig, changes look good.

PR #1852, opened 2018-06-26
stream.dash: support for SegmentList streams with ranged segments

beardypig:
Support for `SegmentList` type DASH streams. 

Fixes #1849

PR #1853, opened 2018-06-26
plugins.atresplayer: update to support VOD streams

beardypig:
Support for VOD streams on `atresplayer`.

Requires #1852 for the DASH streams to work.

### DIFF
--- a/src/streamlink/plugins/atresplayer.py
+++ b/src/streamlink/plugins/atresplayer.py
@@ -2,17 +2,18 @@ from __future__ import print_function
 
 import logging
 import re
+from functools import partial
 
 from streamlink.plugin import Plugin
-from streamlink.plugin.api import http, validate
-from streamlink.stream import HLSStream
-from streamlink.utils import parse_json, update_scheme
+from streamlink.plugin.api import http, validate, StreamMapper
+from streamlink.stream import HLSStream, DASHStream
+from streamlink.utils import parse_json, update_scheme, search_dict
 
 log = logging.getLogger(__name__)
 
 
 class AtresPlayer(Plugin):
-    url_re = re.compile(r"https?://(?:www.)?atresplayer.com/directos/([\w-]+)/?")
+    url_re = re.compile(r"https?://(?:www.)?atresplayer.com/")
     state_re = re.compile(r"""window.__PRELOADED_STATE__\s*=\s*({.*?});""", re.DOTALL)
     channel_id_schema = validate.Schema(
         validate.transform(state_re.search),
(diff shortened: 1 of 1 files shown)


gravyboat:

beardypig:
@jalba69 I will take a look. In the meantime, you can use the option `--stream-types dash` to only return DASH streams and choose "best" quality.

jalba69:
Hello, i had edited my post and added this too, but seeing your answer now only.

I tried this option but it returns me that there is no best streams when using this option.

Thank you.

beardypig:
@jalba69 you're right, there is a bug there :) I'll create a fix for that. 

The HLS stream plays fine for me using `ffmpeg 4.0.1` (`ffmpeg` is used for `hls-multi`) but it does not play using `ffmpeg 2.8.14`. It seems there is an issue with the audio stream. Unfortunately this isn't something we can fix. If you can, try using a newer version of `ffmpeg`, or stick the to the DASH streams.

beardypig:
ffiw, I was using `mpv 0.28.2` :)

jalba69:
Tried a version of MPV ( 0.28.0 / Nightly 20180624 (June 24, 2018) , did not found 0.28.2 for windows after a very quick search ) and it seems like it works somehow... but there is a huge delay before starting to play, a bit like if it was getting both video + audio downloaded then serving it muxed to the player.

Spent quite some time banging my head against the wall trying various things and then grabbed directly the "Best" hls-multi file to analyze it a bit, mediainfo reports audio + video tracks but my usual players can't read audio on it...

FFplay reports this : 

Will try with suggestions for analyzeduration & probesize on the message later at night...

Tried to demux the audio track using MP4box, got an error message, 
`[iso file] Unknown box type .B.%
[iso file] Incomplete box UNKN
[iso file] Incomplete file while reading for dump - aborting parsing
[iso file] Unknown box type .B.%
Truncated file - missing 1138843956 bytes
Error opening file H:\vid.mp4: IsoMedia File is truncated
`

Then used eac3to.exe to get audio track, this time it worked but on the log i get this : 
`[a02] The last AAC frame is incomplete and thus gets skipped.  <WARNING>`

Once demuxed i can load external audio track obtained on top of video and it works but seems like something is messed up with streams... and with my eternal good luck i picked the very best link to try this plugin, sorry for that :/  

Well, at least _720p+a128k, 480p+a128k_ etc will be parsed now if i understood the changes you done so it's not a total waste (i hope).

Noticed there is also 1080_dash, 720_dash etc on some other videos but using "Best" combined with `stream-priority=dash`  located inside `streamlinkrc.atresplayer` config file seems to work and deliver the dash file correctly so i guess you already had this covered :) 

Thanks a lot for your help and work on this.

Now i need a a little break and i'll ask later in a dedicated thread about the plugins specific configuration files and why they do not take the priority above the other settings and if something is possible to be done for that so the parameters inside those plugin settings files always get top priority even if the parameter have already been set to the command or is present in the general _streamlinkrc_ settings file and will apologize for bothering you guys ^^ 

Thank you very much ;)

PR #1860, opened 2018-06-27
plugins.tv8cat: plugin removed - the live broadcast is no longer available

### LINKED ISSUE #1858: 8tv.cat live broadcast gone
### Description

8tv.cat no longer broadcast live streams. We should remove the plugin as it no longer works, and is not fixable. 

It never actually worked quite right, the URL regex was wrong - if you used the wrong URL it worked fine, but with the actual URL it did not work. 

beardypig:
Fixes #1858


gravyboat:

PR #1863, opened 2018-06-27
plugins.nineanime: no longer supported

### LINKED ISSUE #1862: Remove 9anime.to
As discussed over here: https://github.com/streamlink/streamlink/issues/1110#issuecomment-400687075 9anime.to isn't worth supporting at this point and is broken so I'm proposing we remove it.

beardypig:
Fixes #1862


gravyboat:
Thanks @beardypig.

PR #1868, opened 2018-06-28
plugin: support stream weights returned by DASHStream.parse_manifest

beardypig:
`DASHStream.parse_manifest` returns stream qualities that sometimes include the audio bitrate, eg. `720p+a128k` and `720p+a256k`. This adds support in the stream weight parser for these types of stream names, so they can be properly sorted and the `"best"`/`"worst"` options will work. 

See https://github.com/streamlink/streamlink/pull/1853#issuecomment-400981479

### DIFF
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -56,7 +56,7 @@ def stream_weight(stream):
         if stream in weights:
             return weights[stream], group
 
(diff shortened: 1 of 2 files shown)


back-to:
some **tests** would be nice

https://github.com/streamlink/streamlink/blob/master/tests/test_plugin_stream.py#L140

the **testfile name** is a bit **outdated**, but there are some tests for `stream_weight`

beardypig:
Added a few tests, and updated the `asserts` for the older ones.

gravyboat:
Nice thanks @beardypig, thanks for adding tests as well.

PR #1871, opened 2018-06-28
Release 0.14.1

beardypig:


cdrage:
LGTM

gravyboat:

beardypig:
@gravyboat just the tag.

back-to:
this changelog is now outdated

beardypig:
Merged in the wrong order :)

gravyboat:
@back-to Yeah there have been other merged PRs since then which is a problem. @beardypig What do we want to do in situations like this? Seems like we should generate a new updated changelog, but will that break things? I should have merged this first but opened about 20 things in my notifications and just went through them in order.

beardypig:
Well, I guess we could skip 0.14.1 and just have 0.14.2 with the changes from both.

beardypig:
@gravyboat are you able to jump on gitter for 10 mins?

PR #1885, opened 2018-06-30
 tests.plugins: Test that each plugin has a test file

back-to:
#### tests.plugins: Added URL tests for every plugin

- broken plugins might have only negative tests
- some URLs are only derived from the _url_re

#### tests.plugins: New test if a plugin has a test file

- errors when there is no test file for a plugin
  (New Plugin without Tests)
- errors when there is a test file but no plugin
  (Removed old Plugin, but not the tests)


gravyboat:
Looks good to me @back-to, I'll leave it open in case anyone else wants to weigh in. My only concern is over the naming of `tests/test_tests_plugins.py` but I also don't have a better suggestion at this time.

beardypig:
I’d say test_plugins_meta, maybe the tests for the docs could be moved there too ;)

back-to:
I will move it to `test_plugins_meta`

gravyboat:
Thanks @back-to, this is a great addition.

PR #1886, opened 2018-06-30
Handle unicode log message in Python 2

beardypig:
Using the `maybe_{en,de}code` methods from #1576 to handle log messages that might be unicode in Python 2.7.


gravyboat:
This looks good. Eventually we should look at ending 2.7 support...

beardypig:
@gravyboat, never going to happen hehe :)

gravyboat:
@beardypig Let me dream!

PR #1887, opened 2018-06-30
plugin args: if args are suppressed, ignore them

beardypig:
If a `PluginArgument` has been suppress (deprecated) then it should be ignored when collecting the options.


gravyboat:

PR #1889, opened 2018-06-30
Add tvtoya plugin

beardypig:
Support for the live stream on http://tvtoya.pl/live. 

Most of the work done by @remek30.


gravyboat:
No wonder I was confused on the issue, I didn't look and thought this was an existing plugin that was acting up, not a new one.

PR #1894, opened 2018-07-02
stream.hls_playlist: improve parse speed

Billy2011:
speeds up hls-playlist processing up to 2 times

### DIFF
--- a/src/streamlink/stream/hls_playlist.py
+++ b/src/streamlink/stream/hls_playlist.py
@@ -38,9 +38,6 @@ Resolution = namedtuple("Resolution", "width height")
 Segment = namedtuple("Segment", "uri duration title key discontinuity "
                                 "byterange date map")
 
-ATTRIBUTE_REGEX = (r"([A-Z\-]+)=(\d+\.\d+|0x[0-9A-z]+|\d+x\d+|\d+|"
-                   r"\"(.+?)\"|[0-9A-z\-]+)")
-
 
 class M3U8(object):
     def __init__(self):
@@ -62,6 +59,12 @@ class M3U8(object):
 
 
 class M3U8Parser(object):
+    _extinf_re = re.compile("(?P<duration>\d+(\.\d+)?)(,(?P<title>.+))?")
+    _attr_re = re.compile(r"([A-Z\-]+)=(\d+\.\d+|0x[0-9A-z]+|\d+x\d+|\d+|"
+                   r"\"(.+?)\"|[0-9A-z\-]+)")
+    _range_re = re.compile("(?P<range>\d+)(@(?P<offset>.+))?")
+    _line_re_ = re.compile(r'(.*)[\n|\r]')
+
     def __init__(self, base_uri=None):
         self.base_uri = base_uri
 
@@ -102,7 +105,7 @@ class M3U8Parser(object):
         def map_attribute(key, value, quoted):
             return (key, quoted or value)
 
-        attr = re.findall(ATTRIBUTE_REGEX, value)
+        attr = M3U8Parser._attr_re.findall(value)
 
         return dict(starmap(map_attribute, attr))
 
@@ -110,14 +113,14 @@ class M3U8Parser(object):
         return value == "YES"
 
     def parse_byterange(self, value):
-        match = re.match("(?P<range>\d+)(@(?P<offset>.+))?", value)
+        match = M3U8Parser._range_re.match(value)
 
         if match:
             return ByteRange(int(match.group("range")),
                              int(match.group("offset") or 0))
 
     def parse_extinf(self, value):
-        match = re.match("(?P<duration>\d+(\.\d+)?)(,(?P<title>.+))?", value)
+        match = M3U8Parser._extinf_re.match(value)
         if match:
             return float(match.group("duration")), match.group("title")
         return (0, None)
@@ -147,10 +150,7 @@ class M3U8Parser(object):
 
         return value
 
-    def parse_line(self, lineno, line):
-        if lineno == 0 and not line.startswith("#EXTM3U"):
-            raise ValueError("Missing #EXTM3U header")
-
+    def parse_line(self, line):
         if not line.startswith("#"):
             if self.state.pop("expect_segment", None):
                 byterange = self.state.pop("byterange", None)
@@ -238,9 +238,15 @@ class M3U8Parser(object):
         self.state = {}
         self.m3u8 = M3U8()
 
-        for lineno, line in enumerate(filter(bool, data.splitlines())):
-            self.parse_line(lineno, line)
+        lines = enumerate(filter(bool, data.splitlines()))
+        _, line = next(lines)
+        if not line.startswith("#EXTM3U"):
+            raise ValueError("Missing #EXTM3U header")
 
+        parse_line = self.parse_line
+        for _, line in lines:
+            parse_line(line)
+            
         # Associate Media entries with each Playlist
         for playlist in self.m3u8.playlists:
             for media_type in ("audio", "video", "subtitles"):


src/streamlink/stream/hls_playlist.py
```diff
@@ -238,9 +239,16 @@ def parse(self, data):
         self.state = {}
         self.m3u8 = M3U8()
 
-        for lineno, line in enumerate(filter(bool, data.splitlines())):
-            self.parse_line(lineno, line)
+        lines = (l.rstrip() for l in StringIO(data))
+        line = lines.next()
```
beardypig: Should be `next(lines)` I think.

beardypig:
I ran some benchmarks of your version vs. the version in master but I did not see an improvement in performance.

I ran the `load` method 1,000 times with an in-memory M3U8 string that had been loaded from a file, the M3U8 had ~6,800 lines. Average run times are given here.

| Python   | master   | hls_speedup | % diff   |
|----------|----------|-------------|----------|
| `3.4.8`  | `0.0589` | `0.0575`    | `-2.4%`  |
| `3.5.5`  | `0.0340` | `0.0348`    | `+2.9%`  | 
| `3.6.5`  | `0.0618` | `0.0615`    | `-0.4%`  |
| `2.7.14` | `0.0494` | `0.0483`    | `+2.2%`  |
| `2.7.5`  | `0.0488` | `0.0509`    | `-4.1%`  |

Billy2011:
I did the tests on my VU + Zero Box.
I measure the latency in stream.hls:
Log:
```
[16:50:02,418][stream.hls][debug] Reloading playlist                                                    
[16:50:07,974][stream.hls][debug] HLS PL Latency: 5.55420207977s                                        
[16:50:07,999][stream.hls][debug] HLS PL Reload-tm: 2s                   
```             
```
    def reload_playlist(self):
        if self.closed:
            return

        self.reader.buffer.wait_free()
        log.debug("Reloading playlist")
        self.start_time = time()
        res = self.session.http.get(self.stream.url,
                                    exception=StreamError,
                                    retries=self.playlist_reload_retries,
                                    **self.reader.request_params)
        try:
            playlist = hls_playlist.load(res.text, res.url)
        except ValueError as err:
            raise StreamError(err)

        if playlist.is_master:
            raise StreamError("Attempted to play a variant playlist, use "
                              "'hlsvariant://{0}' instead".format(self.stream.url))

        if playlist.iframes_only:
            raise StreamError("Streams containing I-frames only is not playable")

        media_sequence = playlist.media_sequence or 0
        sequences = [Sequence(media_sequence + i, s)
                     for i, s in enumerate(playlist.segments)]

        if sequences:

    def process_sequences(self, playlist, sequences):
        first_sequence, last_sequence = sequences[0], sequences[-1]

        if first_sequence.segment.key and first_sequence.segment.key.method != "NONE":
            log.debug("Segments in this playlist are encrypted")

        self.playlist_changed = ([s.num for s in self.playlist_sequences] !=
                                 [s.num for s in sequences])
        self.playlist_reload_time = (playlist.target_duration or
                                     last_sequence.segment.duration)
        hls_latency = time() - self.start_time
        log.debug("HLS PL Latency: {}s", hls_latency)
```

beardypig:
@Billy2011 can you run this script a few times on your branch and on master?

https://gist.github.com/beardypig/f3eb28b6bbac95f6f897ddd272492da5 

You'll need a playlist.m3u8 file too. You can change `N` to 100, or 500 if it takes a very long time to run. 

The timing you did includes the network stuff, which can be quite variable, so I'm not sure it's a fair test. Which version of Python runs on your VU+ Zero box?

Billy2011:
On the Box is Python 2.7.9

Billy2011:
1000 times took avg. 0.1058s (improved version for my Linux box)
1000 times took avg. 0.0986s
I've downloaded the playlist file from the same Rocket Beans stream.

I repeated the test on the VU+ Zero.
I'm only measuring the load () time now and the latency seems to spread more between 5 - 9s and between 10 - 15s with the orig. version.

The tests on the pc and the linux box are unfortunately not comparable.

beardypig:
Are you not able to run that script on the VU+ boxes? I’m unable to see any advantage in these changes.

src/streamlink/stream/hls_playlist.py
Billy2011: No, lines is a generator obj :)

Billy2011:
I can try it, but it makes no sense without a real stream playback. The playlist processing has a great influence on the segment download speed. I've tried this test yesterday and today many times and it gives always comparable results.

beardypig:
This test only test the `load` method so it won’t be affected by anything else. The changes you made only affect the `load` method. If there is some improvement on the VU+ box then it is worth investigating. I cannot are any reason why this would be faster, but maybe it’s non-obvious.

Billy2011:
You're right, but it's only a static test without threading and so on.

This are the test's on the box:
10 times took avg. 3.5413s (improved)
10 times took avg. 3.6185s

This test results are meaningless; in real use the times are about 2 times greater with great scattering and without my changes much greater. I think the great scattering depends on thread overlapping, the python threading is more useable for waiting IO's and not for great time consuming loop processing.

beardypig:
OK, I get your point. So where do you think the problem is? Seems like it could be `splitlines` locking the GIL. 
As this is a non-obvious change I was put a bit more detail in explaining the rationale.

beardypig:
@Billy2011 will you fix the tests?

beardypig:
@Billy2011, it _cannot_ be ignored. There is a `maybe_encode` method that will do it, or you can try with `BytesIO`. 

This change **must** support Python 2 and Python 3, and the tests **must** pass.

Billy2011:
This issue is fixed in python > 2.7.2 and I can make a wrapper for the older versions, OK?

src/streamlink/stream/hls_playlist.py
```diff
@@ -239,15 +239,14 @@ def parse(self, data):
         self.state = {}
         self.m3u8 = M3U8()
 
-        lines = (l.rstrip() for l in StringIO(data.encode('utf-8')))
-        line = next(lines)
+        lines = enumerate(filter(bool, data.splitlines()))
```
beardypig: You don't need to use `enumerate` here if you are just ignoring the enumeration :)

src/streamlink/stream/hls_playlist.py
```diff
@@ -238,9 +238,15 @@ def parse(self, data):
         self.state = {}
         self.m3u8 = M3U8()
 
-        for lineno, line in enumerate(filter(bool, data.splitlines())):
-            self.parse_line(lineno, line)
+        lines = enumerate(filter(bool, data.splitlines()))
```
beardypig: You don't need to use enumerate here if you are just ignoring the enumeration :)

beardypig:
I finally understand why this is better, taken me a while - must be my thick head.

src/streamlink/compat.py
```diff
@@ -55,8 +55,8 @@ def bytes(b, enc="ascii"):
 
 
 
-getargspec = getattr(inspect, "getfullargspec", inspect.getargspec)
 
+getargspec = getattr(inspect, "getfullargspec", inspect.getargspec)
```
back-to: you did not change anything here, so it should not be in the history

src/streamlink/stream/hls_playlist.py
```diff
@@ -62,6 +59,12 @@ def __init__(self):
 
 
 class M3U8Parser(object):
+    _extinf_re = re.compile("(?P<duration>\d+(\.\d+)?)(,(?P<title>.+))?")
```
back-to: `W605 invalid escape sequence` use `r""`

src/streamlink/stream/hls_playlist.py
```diff
@@ -62,6 +59,12 @@ def __init__(self):
...
+    _attr_re = re.compile(r"([A-Z\-]+)=(\d+\.\d+|0x[0-9A-z]+|\d+x\d+|\d+|"
+                   r"\"(.+?)\"|[0-9A-z\-]+)")
```
back-to: `E128 continuation line under-indented for visual indent`

src/streamlink/stream/hls_playlist.py
```diff
@@ -62,6 +59,12 @@ def __init__(self):
...
+    _range_re = re.compile("(?P<range>\d+)(@(?P<offset>.+))?")
```
back-to: `W605 invalid escape sequence`

src/streamlink/stream/hls_playlist.py
```diff
@@ -62,6 +59,12 @@ def __init__(self):
...
+    _line_re_ = re.compile(r'(.*)[\n|\r]')
```
back-to: `_line_re_` is not used

src/streamlink/stream/hls_playlist.py
```diff
@@ -102,22 +105,22 @@ def parse_attributes(self, value):
         def map_attribute(key, value, quoted):
             return (key, quoted or value)
 
-        attr = re.findall(ATTRIBUTE_REGEX, value)
+        attr = M3U8Parser._attr_re.findall(value)
```
back-to: shouldn't this be `self.`

src/streamlink/stream/hls_playlist.py
```diff
@@ -102,22 +105,22 @@ def parse_attributes(self, value):
...
 
         return dict(starmap(map_attribute, attr))
 
     def parse_bool(self, value):
         return value == "YES"
 
     def parse_byterange(self, value):
-        match = re.match("(?P<range>\d+)(@(?P<offset>.+))?", value)
+        match = M3U8Parser._range_re.match(value)
```
back-to: shouldn't this be `self.`

src/streamlink/stream/hls_playlist.py
```diff
@@ -102,22 +105,22 @@ def parse_attributes(self, value):
...
 
         if match:
             return ByteRange(int(match.group("range")),
                              int(match.group("offset") or 0))
 
     def parse_extinf(self, value):
-        match = re.match("(?P<duration>\d+(\.\d+)?)(,(?P<title>.+))?", value)
+        match = M3U8Parser._extinf_re.match(value)
```
back-to: shouldn't this be `self.`

src/streamlink/stream/hls_playlist.py
```diff
@@ -238,9 +238,15 @@ def parse(self, data):
...
-        for lineno, line in enumerate(filter(bool, data.splitlines())):
-            self.parse_line(lineno, line)
+        lines = enumerate(filter(bool, data.splitlines()))
+        _, line = next(lines)
+        if not line.startswith("#EXTM3U"):
+            raise ValueError("Missing #EXTM3U header")
 
+        parse_line = self.parse_line
+        for _, line in lines:
+            parse_line(line)
```
back-to: why not `self.parse_line` as before

beardypig: That is actually a slight optimisation as member look ups are slower - however, I'm not convinced it makes any measurable difference in this case.

src/streamlink/stream/hls_playlist.py
beardypig: Perhaps you are using `enumerate` where it would be more suitable to use `iter`. Of course, in Python 3, `filter` returns a `filter` object which it's iterable, and you could use `itertools.ifilter` in Python 2 to make it a generator. 

I assume you tried using `re.finditer` based on `_line_re_`, but didn't find any improvement?

src/streamlink/compat.py
Billy2011: How can I do that?

src/streamlink/stream/hls_playlist.py
Billy2011: It's not from me, but OK.

src/streamlink/stream/hls_playlist.py
Billy2011: I think both are possible

src/streamlink/stream/hls_playlist.py
beardypig: Needs to be `r".."` also.

src/streamlink/stream/hls_playlist.py
beardypig: Both are correct, it's an issue if you want to subclass and change the regex - which we might want to do in the future for LHLS (or similar).

src/streamlink/compat.py
beardypig: Make a commit to undo the change, then [squash your commits](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing).

src/streamlink/stream/hls_playlist.py
beardypig: While you're touching that code you may as well fix it too :D

src/streamlink/stream/hls_playlist.py
Billy2011: It's no from me, but what's wrong, the regex works?

back-to: change
`re.compile("`
to
`re.compile(r"`

see https://lintlyci.github.io/Flake8Rules/rules/W605.html

beardypig: In future versions of Python it might generate an Exception - they keep deprecating it. I imagine in Python 4.0 it will not work.

src/streamlink/stream/hls_playlist.py
beardypig: Does this still have the same improvement for you as the last version?

Billy2011: Right, the best solution is with 'cStringIO' and I'll use it in my version.

src/streamlink/compat.py
Billy2011: Sorry, I've 'Zero Bock' for long attempts, do you have detailed instructions for that?

beardypig: Just commit the change to revert the change and we'll squash it at merge time :)

src/streamlink/stream/hls_playlist.py
beardypig: You'll experience problems if you encounter a playlist with an non-ascii characters using that. 

You could try
```python
if is_py2:
    try:
        from cStringIO import StringIO as BytesIO
    except ImportError:
        from StringIO import StringIO as BytesIO
else:
    from io import BytesIO

...

lines = (l.decode("utf8") for l in BytesIO(data.encode("utf8")
```

src/streamlink/compat.py
```diff
@@ -26,11 +26,17 @@ def devnull():
     def bytes(b, enc="ascii"):
         return _str(b)
 
+    try:
+        from cStringIO import StringIO
```
beardypig: This is more like `BytesIO` than `StringIO`, it doesn't support unicode.

Billy2011: It supports unicode, but only in one direction.

Do you need a revert of this to remove all compat changes?

Issue #1895, opened 2018-07-02
DASHStream - TypeError - 'NoneType' and 'int' - t += tsegment.d


back-to:
## Bug Report

- [x] I have read the contribution guidelines.

### Description

DASHStream is **not valid for streamlink**,
the stream **works with vlc and mpv**

### Reproduction steps / Explicit stream URLs to test

1. https://zattoo.com/ondemand any vod there
    free account required,
    but the url is probably different in the webbrowser (drm)
    use the logs.

### Logs

**error**

https://gist.github.com/back-to/5ce300951242872552c21ebb2591c67d#file-error-txt

**mpd**

https://gist.github.com/back-to/5ce300951242872552c21ebb2591c67d#file-zattoo_vod-mpd-xml

beardypig:
I’d say the problem is on line 633. Should check `if tsegment.t is not None` before swapping them :)

back-to:
yes this works.

PR #1897, opened 2018-07-03
stream.dash: fix bug where timeline_segments.t can be None

beardypig:
Fixes the issue reported by @back-to in #1895


gravyboat:
Good stuff.

PR #1903, opened 2018-07-05
utils.args: Moved streamlink_cli utils.args into streamlink package

back-to:
make it usable for plugin args

added tests

---

ref https://github.com/streamlink/streamlink/pull/1902#discussion_r200304653


gravyboat:
Nice @back-to, I really like this. If @beardypig also approves we can get this merged in.

PR #1913, opened 2018-07-08
streamlink.session: add option "no-plugin-reload"

Billy2011:
Improves the time to repeatedly initialize a new Streamlink session, for
example on a Vu+ Zero set-top box by more than 5s.

### DIFF
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -18,6 +18,7 @@ from .plugin import api
 # Ensure that the Logger class returned is Streamslink's for using the API (for backwards compatibility)
 logging.setLoggerClass(StreamlinkLogger)
 log = logging.getLogger(__name__)
+_loaded_plugins = None
 
 
 def print_small_exception(start_after):
@@ -75,14 +76,32 @@ class Streamlink(object):
             "ffmpeg-video-transcode": "copy",
             "ffmpeg-audio-transcode": "copy",
             "locale": None,
-            "user-input-requester": None
+            "user-input-requester": None,
+            "no-plugin-reload": False
         })
         if options:
             self.options.update(options)
+ 
+        global _loaded_plugins
         self.plugins = {}
-        self.load_builtin_plugins()
+        if _loaded_plugins and self.get_option("no-plugin-reload"):
+            self._update_loaded_plugins()
+        else:
+            self.load_builtin_plugins()
+            if self.get_option("no-plugin-reload"):
+                _loaded_plugins = self.plugins.copy()
+        
         self._logger = None
 
+    def _update_loaded_plugins(self):
+        # Set the global http session
+        api.http = self.http
+        # Reload plugins from saved ones
+        self.plugins = _loaded_plugins.copy()
+        # Update plugins session
+        for plugin in self.plugins.itervalues():
+            plugin.session = self
+            
     @property
     def logger(self):
         """


back-to:
They don't reload, they are only loaded once.

```diff
diff --git a/src/streamlink/session.py b/src/streamlink/session.py
index 79bd856..9baadfe 100644
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -80,6 +80,7 @@ class Streamlink(object):
         if options:
             self.options.update(options)
         self.plugins = {}
+        print('---- Streamlink ----')
         self.load_builtin_plugins()
         self._logger = None
 
@@ -442,6 +443,7 @@ class Streamlink(object):
         return self.plugins
 
     def load_builtin_plugins(self):
+        print('------ load_builtin_plugins ------')
         self.load_plugins(plugins.__path__[0])
 
     def load_plugins(self, path):
```

```
$ streamlink https://www.cnews.fr/le-direct -l info
---- Streamlink ----
------ load_builtin_plugins ------
[cli][info] Found matching plugin cnews for URL https://www.cnews.fr/le-direct
[cli][info] Available streams: 180p_alt (worst), 180p, 288p_alt, 288p, 477p_alt, 477p, 720p_alt, 720p, 1080p_alt, 1080p (best)
[cli][info] Player closed
[cli][info] Stream ended
```

Billy2011:
@back-to, this works in conjunction with a server, such as your liveproxy and my streamlinksrv.

@beardypig, because several streams can run at the same time.

beardypig:
`api.http` is bad and should be removed, all the global variables are `livestreamer` legacy. 
The better thing to do would be to reuse the same `Session` instance, if this cannot be done we should fix that.

Billy2011:
I think that can give issues, if for example always the same http session is used.

beardypig:
We could add a reset session method to deal with that sort of thing.

back-to:
@Billy2011 

yes I know,
but I think this special use case should be handled by the third party

the normal streamlink api would probably never use it

---

you could just reuse the `Streamlink` class 

**EXAMPLE** of your script

used `values` (python3) in this example for `intervalues`

```py
from streamlink.plugin import api
from streamlink import plugins

_loaded_plugins = None


class Streamlink2(Streamlink):

    def load_builtin_plugins(self):
        global _loaded_plugins
        print("-----")
        if not _loaded_plugins:
            print(_loaded_plugins)
        else:
            print(len(_loaded_plugins))
        if _loaded_plugins and True:  # self.get_option("no-plugin-reload")
            self._update_loaded_plugins()
        else:
            self.load_plugins(plugins.__path__[0])
            if True:  # self.get_option("no-plugin-reload")
                _loaded_plugins = self.plugins.copy()
        if not _loaded_plugins:
            print(_loaded_plugins)
        else:
            print(len(_loaded_plugins))

    def _update_loaded_plugins(self):
        # Set the global http session
        api.http = self.http
        # Reload plugins from saved ones
        self.plugins = _loaded_plugins.copy()
        # Update plugins session
        for plugin in self.plugins.values():
            plugin.session = self
```

```diff
-    session = Streamlink()
+    session = Streamlink2()
```

Billy2011:
@back-to, of course it works too, but as the saying goes, "Several roads lead to Rome" :)

beardypig:
@back-to it might be useful for reloading the plugin streams, like in your PR from a while back?

PR #1916, opened 2018-07-09
Fix ssh101

Twilight0:
Also converted delimiters to single quotes
Test url: https://www.ssh101.com/live/m1gacytv54201

### DIFF
--- a/src/streamlink/plugins/ssh101.py
+++ b/src/streamlink/plugins/ssh101.py
@@ -7,15 +7,14 @@ from streamlink.compat import urljoin
 
 
 class SSH101(Plugin):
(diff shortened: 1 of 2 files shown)


back-to:
Doesn't work for normal HLSStreams

https://github.com/streamlink/streamlink/blob/5790bd2434ab32d720b6e6914988f1ed64ad93b7/src/streamlink/plugins/ssh101.py#L31

you must use something like this

https://github.com/streamlink/streamlink/blob/5a36e512e5adb5b4f8056eda9e4290ffebf8a299/src/streamlink/plugins/tlctr.py#L33-L37

---

also if you open an offline stream, you will **stuck for 60 seconds**,

you should add some online check.

Twilight0:
I simply fixed the regex pattern so far, but can add additional conditionals for what you say.

PR #1918, opened 2018-07-10
stream.hls_playlist: fix some regex pattern & code optimization

Billy2011:
- declares explicit all pattern strings as raw strings
- move global & all other regex pattern to precompiled regex objects to `M3U8Parser`
- code optimization for parser loop

### DIFF
--- a/src/streamlink/stream/hls_playlist.py
+++ b/src/streamlink/stream/hls_playlist.py
@@ -63,6 +63,7 @@ class M3U8Parser(object):
     _attr_re = re.compile(r"([A-Z\-]+)=(\d+\.\d+|0x[0-9A-z]+|\d+x\d+|\d+|\"(.+?)\"|[0-9A-z\-]+)")
     _range_re = re.compile(r"(?P<range>\d+)(@(?P<offset>.+))?")
     _tag_re = re.compile(r"#(?P<tag>[\w-]+)(:(?P<value>.+))?")
+    _res_re = re.compile(r"(\d+)x(\d+)")
 
     def __init__(self, base_uri=None):
         self.base_uri = base_uri
@@ -132,7 +133,7 @@ class M3U8Parser(object):
         return unhexlify(value)
 
     def parse_resolution(self, value):
-        match = re.match(r"(\d+)x(\d+)", value)
+        match = self._res_re.match(value)
 
         if match:
             width, height = int(match.group(1)), int(match.group(2))
--- a/src/streamlink/stream/hls_playlist.py
+++ b/src/streamlink/stream/hls_playlist.py
@@ -103,7 +103,7 @@ class M3U8Parser(object):
         def map_attribute(key, value, quoted):
             return (key, quoted or value)
 
-        attr = M3U8Parser._attr_re.findall(value)
+        attr = self._attr_re.findall(value)
 
         return dict(starmap(map_attribute, attr))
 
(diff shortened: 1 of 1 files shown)


beardypig:
Why not use `iter` in place of `enumerate`?

src/streamlink/stream/hls_playlist.py
```diff
@@ -102,22 +103,22 @@ def parse_attributes(self, value):
         def map_attribute(key, value, quoted):
             return (key, quoted or value)
 
-        attr = re.findall(ATTRIBUTE_REGEX, value)
+        attr = M3U8Parser._attr_re.findall(value)
```
beardypig: Can you make these `self.` instead of `M3U8Parser.` - for subclassing, as we discussed in the previous PR?

Billy2011: Done.

beardypig:
@Billy2011 does this version speed it up for you on the VU+ boxes?

Billy2011:
On the VU+ boxes I use the version with `cStringIO` - I have not tested this last version.

beardypig:
OK, it would be good to know if this does speed it up at all - otherwise the commit messages won't be accurate :)

back-to:
lets say I open an **invalid hls stream** or the **stream expires** where the **.m3u8 file is empty**

with the current master version it will timeout after x seconds which is also not optimal

with your change **it will crash**, a clean close would be better.

```
$ streamlink hls://https://ssh101.com/m3u8/dyn/test/index.m3u8
[cli][debug] Streamlink: 0.14.2+29.gc2ca180.dirty
[cli][debug] Requests(2.19.1), Socks(1.6.7), Websocket(0.48.0)
[cli][info] Found matching plugin hls for URL hls://https://ssh101.com/m3u8/dyn/random/index.m3u8
[plugin.hls][debug] URL=https://ssh101.com/m3u8/dyn/random/index.m3u8; params={}
-- data - def parse(self, data): --


  File "env/bin/streamlink", line 11, in <module>
  File "src/streamlink_cli/main.py", line 1027, in main
    handle_url()
  File "src/streamlink_cli/main.py", line 557, in handle_url
  File "src/streamlink_cli/main.py", line 437, in fetch_streams
  File "src/streamlink/plugin/plugin.py", line 431, in get_streams
    return self.streams(*args, **kwargs)
  File "src/streamlink/plugin/plugin.py", line 333, in streams
  File "src/streamlink/plugins/hls.py", line 44, in _get_streams
    streams = HLSStream.parse_variant_playlist(self.session, urlnoproto, **params)
  File "src/streamlink/stream/hls.py", line 378, in parse_variant_playlist
    parser = hls_playlist.load(res.text, base_uri=res.url)
  File "src/streamlink/stream/hls_playlist.py", line 282, in load
    return parser(base_uri).parse(data)
  File "src/streamlink/stream/hls_playlist.py", line 242, in parse
    line = next(lines)
StopIteration
```

Billy2011:
I do not know what you mean by `clean close` - with the change it should now behave like the master.

src/streamlink/stream/hls_playlist.py
```diff
@@ -130,7 +132,7 @@ def parse_hex(self, value):
         return unhexlify(value)
 
     def parse_resolution(self, value):
-        match = re.match("(\d+)x(\d+)", value)
+        match = re.match(r"(\d+)x(\d+)", value)
```
back-to: since you moved all other lines, you could also move this one.

src/streamlink/stream/hls_playlist.py
```diff
@@ -238,9 +238,19 @@ def parse(self, data):
...
+        except StopIteration:
+            return self.m3u8
+        else:
+            if not line.startswith("#EXTM3U"):
+                raise ValueError("Missing #EXTM3U header")
 
+        parse_line = self.parse_line
+        for line in lines:
+            parse_line(line)
+
```
back-to: @Billy2011 
**can you remove these spaces here,**
did not see it before because my editor removes them automatically.

This is the only issue left and should be good to merge after this.

I did a small `speed` test on an old FireTV,
it has still some small speed improvements in thousandths of a second not much but still.

Billy2011:
@back-to which editor do you use - Notepad++ does not do that automatically?

Issue #1882, opened 2018-06-29
[Update] TV4 Play.


Ordval:
- [x] This is a plugin (improvement) request. 

### Description
tv4play.py needs an update.  
They now separate audio and picture. 
### Log
```
C:\Users\Ordval>call streamlink.exe -l debug "https://www.tv4play.se/program/monk/3938989" best -o "E:\Documents\Monk S 3 E  2.mp4"
[console][debug] OS:         Windows 10
[console][debug] Python:     3.5.2
[console][debug] Streamlink: 0.13.0
[console][debug] Requests(2.18.4), Socks(1.6.7), Websocket(0.48.0)
error: No plugin can handle URL: https://www.tv4play.se/program/monk/3938989
```
also tried with Streamlink: 0.14.2
but could this easily be accomplished in the old Plug-In with the [`--ffmpeg-ffmpeg FILENAME`](https://streamlink.github.io/cli.html#cmdoption-ffmpeg-ffmpeg) commands ?
If so as usual, I need a very hands on command line example to it:
```
C:\Users\Ordval>call streamlink.exe --ffmpeg-ffmpeg hls://"https://lbs-usp-hls-vod.cmore.se/vod/efeaa/pid500519856_3939062(3939062_ISMUSP).ism/pid500519856_3939062(3939062_ISMUSP)-video_eng=4340000.m3u8" best
error: No plugin can handle URL: best -o "E:\Documents\Monk S 3 E  1.mp4"
```
```
call streamlink.exe hls://"https://lbs-usp-hls-vod.cmore.se/vod/efeaa/pid500519856_3939062(3939062_ISMUSP).ism/pid500519856_3939062(3939062_ISMUSP)-video_eng=4340000.m3u8" best --ffmpeg-ffmpeg
usage: streamlink-script.py [OPTIONS] <URL> [STREAM]
streamlink-script.py: error: argument --ffmpeg-ffmpeg: expected one argument
```
`call streamlink.exe --ffmpeg-ffmpeg hls://"https://lbs-usp-hls-vod.cmore.se/vod/efeaa/pid500519856_3939062(3939062_ISMUSP).ism/pid500519856_3939062(3939062_ISMUSP)-video_eng=4340000.m3u8" best -o --ffmpeg-video-transcode mp4 --ffmpeg-audio-transcode aac --ffmpeg-verbose-path "E:\Documents\Monk S 3 E  1.mp4"`
```
[...]

PR #1919, opened 2018-07-10
plugins.tv4play: Fix for updated website

back-to:
VOD and Livestream fix

closes https://github.com/streamlink/streamlink/issues/1882

ping @Ordval

---

also support for `self.get_title()`


gravyboat:
Thanks @back-to.

PR #469, opened 2017-01-23
Swiss SRG SSR Fix 

beardypig:
Here's a fixed version of the srgssr plugin along with support for the swisstxt hosted sports streams. 

The PR fixes #403 

Should now support the sports live streams. You can find the sports streams on the following URLs (however, there are not always any streams available):

- srf.ch/sport/resultcenter/results
- rts.ch/sport/programmes/
- live.rsi.ch/results.html

In addition to the sports streams, there are some extra live streams listed on rsi.ch/live-streaming/

@karlo2105 also confirmed that the plugins were working as expected :-)

PR #1925, opened 2018-07-12
streamlink.plugins: replace global http session by self.session.http

Billy2011:
Global http session obj should not be used because there may be problems
with other server/proxy applications.


beardypig:
I like it. But, why is this commit 32e2458954beabb8fc84d47ee6800ccd685a10e9 included? Shouldn't you rebase on the upstream master to get this included.

Billy2011:
All PR's end up with me - it is no longer possible to create a PR for streamlink???

back-to:
I think it should work now, if you push something to `Billy2011:api-http-fix`

beardypig:
What's happening here? :D

Billy2011:
I've been wondering the same thing all day (•ิ_•ิ)?

Billy2011:
How can I fix the test plugins garena + tvplayer?
That, for example, no longer exists: `streamlink.plugins.garena.http`

beardypig:
@Billy2011 https://github.com/Billy2011/streamlink/pull/10 :)

back-to:
Missed 2 plugins.

https://github.com/Billy2011/streamlink/blob/api-http-fix/src/streamlink/plugins/bongacams.py#L48-L49
https://github.com/Billy2011/streamlink/blob/api-http-fix/src/streamlink/plugins/tvrplus.py#L30

Billy2011:
I guess I was too frustrated yesterday :-D

Billy2011:
Must the global var. `api.http` and all remaining references to it be preserved?

beardypig:
Anything that referrers to the global should be removed. While it won’t help with the thread safety, we will have more control - for example, we could create one session per thread ID (not sure if that’s actually a workable idea).

Billy2011:
I think the last reference to api.http is in streamlink.session, should I remove that?

beardypig:
>I think the last reference to api.http is in streamlink.session, should I remove that?

@Billy2011 Can you post a line reference? Maybe that one will have to be done separately :)

Billy2011:
@beardypig, in line #469    api.http = self.http

beardypig:
@Billy2011 I believe that would make it a breaking change. We should open a separate PR for that, and probably keep it open until we are ready for `1.0.0` :)

Billy2011:
OK, then I hope that this PR is ready to merge and no other member raises an objection :)

gravyboat:
Alright let's get this in then, thanks for the hard work @Billy2011, and for the feedback everyone!

PR #1934, opened 2018-07-13
plugins.pluzz : Hide bugged streams

Urgau:
This fix hide temporary bugged streams

### DIFF
--- a/src/streamlink/plugins/pluzz.py
+++ b/src/streamlink/plugins/pluzz.py
@@ -169,14 +169,6 @@ class Pluzz(Plugin):
                 res = http.get(self.TOKEN_URL.format(video_url))
                 video_url = res.text
 
-            """if '.mpd' in video_url:
-                # Get redirect video URL
-                res = http.get(res.text)
-                video_url = res.url
-                for bitrate, stream in DASHStream.parse_manifest(self.session,
-                                                                 video_url).items():
-                    streams.append((bitrate, stream))
-            el"""
             if '.f4m' in video_url and swf_url is not None:
                 for bitrate, stream in HDSStream.parse_manifest(self.session,
                                                                 video_url,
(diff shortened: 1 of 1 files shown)


Urgau:
The problem is that the video stream takes 5 seconds to start and then a few seconds later nothing more. I have a closer look and I think the problem comes from the website and not from streamlink.

back-to:
this could be a streamlink issue rather than a plugin issue

only the init video will be played,
everything else will be downloaded in the background but won't be played.

it works in the webbrowser

https://www.france.tv/france-3/direct.html
https://gist.github.com/back-to/31e8367af1b9a177054e28de72bc59f4/revisions

Urgau:
Okay, however, if it's a streamlink problem I don't have enough knowledge to fix it or even find it.
However, I recommend that you accept my pull request until you find and fix the problem.

What are you going to do to fix the problem?

back-to:
I don't know how much work it is to fix the dash stream,
but if you want it removed temporarily,
you should remove all related dash code

see when it was added https://github.com/streamlink/streamlink/commit/84a2f90c8c9ff93b876c1f9bb04a586de8b146ea#diff-57d44e5f969dec7c7a4f48de4c3f2c8b

**or** you can ignore the dash streams with `--stream-types hls,hds`

https://streamlink.github.io/cli.html#cmdoption-stream-types

---

also the issue is related to

https://github.com/streamlink/streamlink/blob/cac78e43e0f1ab632cb6796a758c10d876e36ed5/src/streamlink/stream/dash.py#L245-L246

when the audio and video gets muxed

---

when `self.video_representation` or `self.audio_representation` gets played alone,
without calling `open` on the other type, they work.

https://github.com/streamlink/streamlink/blob/cac78e43e0f1ab632cb6796a758c10d876e36ed5/src/streamlink/stream/dash.py#L237-L243

https://github.com/streamlink/streamlink/blob/cac78e43e0f1ab632cb6796a758c10d876e36ed5/src/streamlink/stream/dash.py#L247-L250

PR #1935, opened 2018-07-13
[picarto.tv] Fix regarding changed URL

### LINKED ISSUE #1932: Can't Download Picarto Video
## Bug Report

- [X ] I have read the contribution guidelines. 

### Description

Picarto.tv plugin doesn't seem to download any recorded streams anymore

Expected it to download video, didn't

1. streamlink -o Amarynceus.flv https://picarto.tv/videopopout/Amarynceus_2018.07.04.22.55.48.flv

### Logs

```
streamlink -o Amarynceus.flv https://picarto.tv/videopopout/Amarynceus_2018.07.04.22.55.48.flv -l debug
[cli][debug] OS:         Linux-4.15.0-24-generic-x86_64-with-Ubuntu-18.04-bionic
[cli][debug] Python:     2.7.15rc1
[cli][debug] Streamlink: 0.14.2+24.g6bf654a2
[cli][debug] Requests(2.19.1), Socks(1.6.7), Websocket(0.48.0)
[cli][info] Found matching plugin picarto for URL https://picarto.tv/videopopout/Amarynceus_2018.07.04.22.55.48.flv
[plugin.picarto][debug] Possible VOD stream...
[plugin.picarto][warning] Probably a VOD stream but no VOD found?
[plugin.picarto][error] The channel Amarynceus_2018.07.04.22.55.48.flv does not exist
error: No playable streams found on this URL: https://picarto.tv/videopopout/Amarynceus_2018.07.04.22.55.48.flv
```

skulblakka:
Looks like picarto started escaping the slashes after the https in their VOD URL. 

This adjusts the regex accordingly and removes the slashes (I hope this is the correct way to do that).

Fixes #1932

### DIFF
--- a/src/streamlink/plugins/picarto.py
+++ b/src/streamlink/plugins/picarto.py
@@ -25,7 +25,7 @@ class Picarto(Plugin):
     """, re.VERBOSE)
 
     # Regex for VOD extraction
-    _vod_re = re.compile(r'''(?<=[#vod-player", ])(\{.*?\})''')
+    _vod_re = re.compile(r'''(?<=#vod\-player", )(\{.*?\})''')
 
     data_schema = validate.Schema(
         validate.transform(_vod_re.search),
--- a/src/streamlink/plugins/picarto.py
+++ b/src/streamlink/plugins/picarto.py
@@ -22,7 +22,7 @@ class Picarto(Plugin):
     """, re.VERBOSE)
 
     # Regex for VOD extraction
-    _vod_re = re.compile(r'''vod: "(https?://[\S]+?/index.m3u8)",''')
+    _vod_re = re.compile(r'''"vod":"(https?:[\S]+?/index.m3u8)",''')
 
     @classmethod
     def can_handle_url(cls, url):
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/picarto.py
```diff
@@ -55,7 +55,7 @@ def _create_flash_stream(self, server, channel, token):
     def _get_vod_stream(self, page):
         m = self._vod_re.search(page.text)
         if m:
-            return HLSStream.parse_variant_playlist(self.session, m.group(1))
+            return HLSStream.parse_variant_playlist(self.session, m.group(1).replace('\\/', '/'))
```
beardypig: You could use `parse_json` if you included the `"` characters in the match group. It looks like JSON so there might be other things escaped for other streams...

skulblakka:
@beardypig Is this what you meant? If yes, is the new regex ok this way?

beardypig:
Yeah like that. But, I’m not sure the regex is exactly how you meant it - or I don’t understand it :) 

Should it match `#vod-player", {foo}`? I think at the moment it will match things like `v {foo}`, which might not be how you wanted it. You can check it on regex101.com, which has some nice features for debugging regex :)

skulblakka:
>But, I’m not sure the regex is exactly how you meant it

Not unlikely. I can't say I have much experience using regular expressions.

>Should it match #vod-player", {foo}?

I want it to match `{foo}` after `#vod-player", ` (ight now you could also match `{foo}` by just matching everything between two curly braces as they this is the only occurrence in the page but I thought that would be to generic).

Looking at the explanation of the expression on regex101.com (thanks for the suggestion btw 👍 ) I see what you mean. I updated the regex hoping it does the right thing now.

beardypig:
No problem. I’m not sure you need to escape the `-` in this case, otherwise it looks good. I believe you can change the regex engine to Python on regex101 which might highlight any unnecessary escaping, or if you use a decent IDE it’ll do the same.

PR #1926, opened 2018-07-12
stream.hls: only include audio/video streams in MuxedHLSStreams

beardypig:
If streams that are being muxed include data streams, the stream mapping will not be correct. This PR adds a filter so that only `video` and `audio` streams are mapped. 

Fixes #1898

PR #1941, opened 2018-07-17
stream.hls: if the primary video stream has audio then include it

### LINKED ISSUE #1938: hls-audio-select option preclude downloading of default audio
This does not happen in streamlink 0.14.2, but issue is present in git version of streamlink.

I am downloading mlb games using streamlink. There are usually 3 audio streams. One default called "English" and two additional ones called "English radio" and "Natural sound". Without using hls-audio-select it downloads video with default stream as expected. If I add --hls-audio-select "English Radio", it does not download default stream. Using --hls-audio-select "*" downloads English Radio and Natural Sound audio. Default audio is still not present.

beardypig:
Due to the changes in #1926, if the primary video stream included an audio stream as well it would not be included in the output. This change restores that audio stream. 

Fixes #1938

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -295,7 +295,7 @@ class MuxedHLSStream(MuxedStream):
 
     def __init__(self, session, video, audio, force_restart=False, ffmpeg_options=None, **args):
         tracks = [video]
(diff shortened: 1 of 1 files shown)


gravyboat:
Nice, thanks @beardypig.

PR #1944, opened 2018-07-19
plugins.youtube: Added support for {author} and {title}

back-to:
`streamlink --title "{author} - {title}" URL best`

---

Use `PluginError` instead of `log.error` with `return` for video_id

---

removed not used import and schema


back-to:
@beardypig 
do you want some other changes or is it good enough?

beardypig:
LGTM 🎉

gravyboat:
Looks good @back-to, thanks!

PR #1946, opened 2018-07-19
plugins.twitch: Fixed AttributeError and Flake8

back-to:
`AttributeError: 'TwitchAPI' object has no attribute 'session'`
`AttributeError: 'UsherService' object has no attribute 'session'`

---

error is related to #1925


gravyboat:
Thanks @back-to.

PR #1947, opened 2018-07-19
plugins.filmon: Fixed AttributeError

back-to:
`AttributeError: 'FilmOnAPI' object has no attribute 'session'`

error is related to #1925


gravyboat:
Good stuff, thanks @back-to.

PR #1948, opened 2018-07-19
plugins.crunchyroll: Fixed AttributeError and Flake8

back-to:
`AttributeError: 'CrunchyrollAPI' object has no attribute 'session'`

error is related to #1925


gravyboat:
Thanks @back-to.

PR #1952, opened 2018-07-22
plugins.media_ccc_de: removed plugin

back-to:
I don't think anyone is using this plugin
at least the VOD part is broken for two years

`display_mime_type` was droped 2 years ago and would always raise a 
KeyError in Streamlink

Example when it changed in 2016 for the Kodi Addon

https://github.com/cccc/plugin.video.media-ccc-de/pull/7/files
https://github.com/voc/voctoweb/commit/32e27e6814e1435e903f2b4a6ccecd729742327c

https://github.com/streamlink/streamlink/blob/2122ebf401eb868718a0a15c7167c83d4b0b6467/src/streamlink/plugins/media_ccc_de.py#L85

The livestreams might work, but they are not available right now.

---

The `self.session.http` part would need an update,
but since I don't know if someone is using this plugin it is not worth 
it.

---

If someone is still using the livestream part,
leave a comment here and I would fix it.


gravyboat:
Sounds reasonable to me. Thanks!

PR #1953, opened 2018-07-22
plugins.younow: Fix for session error and plugin cleanup

back-to:
- Fixed `self.session.http` issues
- added title and author
- better error msg
- plugin cleanup / flake8


gravyboat:
Looks good, thanks @back-to.

PR #1966, opened 2018-07-27
Reword+reformat issue templates for consistency

### LINKED ISSUE #1961: Issue template improvements
Ever since the issue template has been split into multiple topic-templates, it feels like even more *bad issues* or issues without any useful information at all have been submitted than before. I didn't count them, so I may be wrong here, because a couple of *bad issues* have been created by a few people who were spamming the tracker lately, but still, I hope everyone can agree that the quality of new issues is not what was expected from this change. What is the reason for this?

**Is the template list too confusing?**
I'm not sure, but there are still lots of issues which are using the wrong template. Maybe the order should be changed and the descriptions be improved. It could also be the fault of some popular mobile apps. FastHub(-Libre) for example doesn't support multiple issue templates.

**Are the individual templates too complicated and overwhelming?**
I think a couple of things need to be removed, so the overall template size can be reduced, which will make it easier to understand. Maybe it's the HTML comment tags which are confusing. The first comment block on each template is pretty big. This could make people think that the whole template is just a single large comment and can be ignored. People who don't know HTML comments/tags or Markdown could also be intimidated or confused by this.

It's not a dramatic issue, but since there are a couple of new threads each day, I think this should be addressed, so that the amount of time unnecessarily spent on the issue tracker here can be reduced.

bastimeyer:
Resolves #1961

Further discussion is welcome.

- Made all templates as consistent as possible
- Removed a couple of lines and tried to use short and precise sentences
- Fixed left over spelling/grammar errors
- Reworded the header and other comments
- Changed a few links

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_Report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_Report.md
@@ -5,7 +5,7 @@ about: Create a bug report to help us improve Streamlink
 
 <!--
 Thanks for reporting a bug!
-USE THE TEMPLATE, otherwise, your bug report may be rejected.
+USE THE TEMPLATE. Otherwise your bug report may be rejected.
 
 First, see the contribution guidelines:
 https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink
@@ -25,7 +25,7 @@ Please see the text preview to avoid unnecessary formatting errors.
 
 ### Description
 
-<!-- Explain the bug as detailed as you can. -->
+<!-- Explain the bug as thoroughly as you can. Don't leave out information which is necessary for us to reproduce and debug this issue. -->
 
 
 ### Expected / Actual behavior
--- a/.github/ISSUE_TEMPLATE/Bug_Report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_Report.md
@@ -5,25 +5,27 @@ about: Create a bug report to help us improve Streamlink
 
 <!--
 Thanks for reporting a bug!
+USE THE TEMPLATE, otherwise, your bug report may be rejected.
 
-Please read the contribution guidelines before creating a bug report:
+First, see the contribution guidelines:
+https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink
 
-(https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
-
-Also check the list of known issues before reporting an issue!
-
-Please fill out the following template. Be as detailed as possible.
+Also check the list of open and closed bug reports:
+https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22
 
 Please see the text preview to avoid unnecessary formatting errors.
 -->
 
+
 ## Bug Report
 
-- [ ] I have read the contribution guidelines. <!-- Replace the space with an x to check the box: [x] -->
+<!-- Replace [ ] with [x] in order to check the box -->
+- [ ] This is a bug report and I have read the contribution guidelines.
+
 
 ### Description
 
-<!-- Explain the bug as clearly as you can. How is functionality impacted? Which OS/Player/Streamlink version are you using? Etc. -->
+<!-- Explain the bug as detailed as you can. -->
 
 
 ### Expected / Actual behavior
@@ -39,19 +41,22 @@ Please see the text preview to avoid unnecessary formatting errors.
 2. ...
 3. ...
 
-### Logs
+
+### Log output
 
 <!--
-Logs are always required for a bug report, use `-l debug` [(help)](https://streamlink.github.io/cli.html#cmdoption-l) and avoid using `--stream-url` as that will suppress log output.
+TEXT LOG OUTPUT IS REQUIRED for a bug report!
+Use the `--loglevel debug` parameter and avoid using parameters which suppress log output.
+https://streamlink.github.io/cli.html#cmdoption-l
 
-Make sure to **remove username and password**
-
-You can upload your logs to https://gist.github.com/ or place your logs below.
+Make sure to **remove usernames and passwords**
+You can copy the output to https://gist.github.com/ or paste it below.
 -->
 
 ```
-REPLACE THIS TEXT WITH YOUR LOG
+REPLACE THIS TEXT WITH THE LOG OUTPUT
 ```
+
 
 ### Additional comments, screenshots, etc.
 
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/Bug_Report.md
```diff
@@ -5,25 +5,27 @@ about: Create a bug report to help us improve Streamlink
...
 
-- [ ] I have read the contribution guidelines. <!-- Replace the space with an x to check the box: [x] -->
+<!-- Replace [ ] with [x] in order to check the box -->
+- [ ] This is a bug report and I have read the contribution guidelines.
+
 
 ### Description
 
-<!-- Explain the bug as clearly as you can. How is functionality impacted? Which OS/Player/Streamlink version are you using? Etc. -->
+<!-- Explain the bug as detailed as you can. -->
```
gravyboat: Grammatically this is a weird sentence, maybe something like `Explain the bug as thoroughly as you can`? Also since we're dropping the requirement of player and such does this open us up to fewer details? Granted people weren't doing it any way...

.github/ISSUE_TEMPLATE/Bug_Report.md
```diff
@@ -39,20 +41,23 @@ Please see the text preview to avoid unnecessary formatting errors.
...
-### Logs
 
-<!--
-Logs are always required for a bug report, use `-l debug` [(help)](https://streamlink.github.io/cli.html#cmdoption-l) and avoid using `--stream-url` as that will suppress log output.
+### Log output
 
-Make sure to **remove username and password**
+<!--
+TEXT LOG OUTPUT IS REQUIRED for a bug report!
+Use the `--loglevel debug` parameter and avoid using parameters which suppress log output.
```
gravyboat: Should we add an example of what the streamlink command looks like when running with `-l debug` here? Some people have been confused about this lately.

.github/ISSUE_TEMPLATE/Feature_Request.md
```diff
@@ -5,17 +5,25 @@ about: Create a feature request to help us add additional functionality to Strea
...
 <!-- 
 Thanks for filing a feature request!
+USE THE TEMPLATE, otherwise, your feature request may be rejected.
 
-Please read the contribution guidelines (https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink) first!
+First, see the contribution guidelines:
+https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink
 
-Please fill out the following template. Be as detailed as possible.
+Open a plugin request, if you're requesting a new plugin instead of a new feature.
```
gravyboat: No need for the `,` here.

.github/ISSUE_TEMPLATE/Bug_Report.md
```diff
@@ -5,25 +5,27 @@ about: Create a bug report to help us improve Streamlink
 
 <!--
 Thanks for reporting a bug!
+USE THE TEMPLATE, otherwise, your bug report may be rejected.
```
gravyboat: I don't think we need these commas.

.github/ISSUE_TEMPLATE/Feature_Request.md
```diff
@@ -5,17 +5,25 @@ about: Create a feature request to help us add additional functionality to Strea
 
 <!-- 
 Thanks for filing a feature request!
+USE THE TEMPLATE, otherwise, your feature request may be rejected.
```
gravyboat: Same as the other one, I don't think we need these commas.

.github/ISSUE_TEMPLATE/Plugin_Issue.md
```diff
@@ -5,24 +5,27 @@ about: Create a plugin issue report
 
 <!--
 Thanks for reporting a plugin issue!
+USE THE TEMPLATE, otherwise, your plugin issue may be rejected.
```
gravyboat: Same as before, don't think we need these commas.

.github/ISSUE_TEMPLATE/Plugin_Issue.md
```diff
@@ -5,24 +5,27 @@ about: Create a plugin issue report
...
 
-- [ ] This is a plugin issue and I have read the contribution guidelines. <!-- Replace the space with an x to check the box: [x] -->
+<!-- Replace [ ] with [x] in order to check the box -->
+- [ ] This is a plugin issue and I have read the contribution guidelines.
+
 
 ### Description
 
-<!-- Explain the plugin issue as clearly as you can. Which plugin is causing isuses? what exactly is the issue? Etc. -->
+<!-- Explain the plugin issue as detailed as you can. -->
```
gravyboat: Similarly to above the grammar here is weird. I think my proposal from earlier would resolve this, otherwise we can do something else.

.github/ISSUE_TEMPLATE/Plugin_Request.md
```diff
@@ -5,36 +5,41 @@ about: Create a plugin request to discuss adding a new plugin to Streamlink
 
 <!--
 Thanks for requesting a plugin!
+USE THE TEMPLATE, otherwise, your plugin request may be rejected.
```
gravyboat: Same thing here, no commas.

gravyboat:
What do you think about creating closed example issues and linking them? Is that adding too much or overwhelming? At the very least I think we should have examples to link when someone asks about them.

.github/ISSUE_TEMPLATE/Bug_Report.md
bastimeyer: I think that would make it a bit too long

.github/ISSUE_TEMPLATE/Bug_Report.md
bastimeyer: Before I change this, does
"USE THE TEMPLATE! Otherwise your bug report may be rejected."
sound better? The exclamation mark may sound a bit harsh.

gravyboat: I'd avoid the exclamation mark, I agree it sounds harsh.

bastimeyer:
> closed example issues

I'm not really a fan of this idea. Let's just wait and see what these changes will do.

beardypig:
These changes look good to me

gravyboat:
@bastimeyer I'm not really a fan of the closed example issues either, but I'm getting pretty tired of people asking "well what should my issue look like?" when it doesn't meet the requirements simply because they didn't read when they opened the issue initially.

bastimeyer:
We could add something to the repo wiki or repo itself and then link to it. Example issue threads would get spammed with (bad) references.

gravyboat:
I'm going to merge this. We can always open a new issue if we want to discuss things further. Thanks @bastimeyer!

Issue #1969, opened 2018-07-30
mjunoon plugin authorization error


maanich:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

mjunoon plugin 401 Client Error before it was working fine

### Reproduction steps / Explicit stream URLs to test

1. https://www.mjunoon.tv/a-plus-live
2. https://www.mjunoon.tv/channel-24-live
3. https://www.mjunoon.tv/filmax-live

### Log output

```
$ streamlink -l debug https://www.mjunoon.tv/a-plus-live 480p
[cli][info] streamlink is running as root! Be careful!
[cli][debug] OS:         Linux-3.13.0-151-generic-x86_64-with-Ubuntu-14.04-trusty
[cli][debug] Python:     3.4.3
[cli][debug] Streamlink: 0.14.2+42.gb3c7648
[cli][debug] Requests(2.2.1), Socks(1.6.7), Websocket(0.48.0)
[cli][info] Found matching plugin mjunoon for URL https://www.mjunoon.tv/a-plus-live
[plugin.mjunoon][debug] Found the playerScript script tag
[utils.l10n][debug] Language code: en_US
[cli][info] Available streams: 270p (worst), 480p (best)
[cli][info] Opening stream: 480p (hls)
[stream.hls][debug] Reloading playlist
[cli][error] Try 1/1: Could not open stream <HLSStream('https://cdn2.mjunoon.tv:25463/streaming/clients_live.php?username=xxxxx&password=xxxxx&stream=129&extension=m3u8&id=1')> (Could not open stream: Unable to open URL: https://cdn2.mjunoon.tv:25463/streaming/clients_live.php?username=xxxxxGMh7&password=xxxx&stream=129&extension=m3u8&id=1 (401 Client Error: Unauthorized))
error: Could not open stream <HLSStream('https://cdn2.mjunoon.tv:25463/streaming/clients_live.php?username=xxxx&password=xxxxx&stream=129&extension=m3u8&id=1')>, tried 1 times, exiting
```

### Additional comments, screenshots, etc.
before it was working without any issue.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

PR #1971, opened 2018-07-31
plugins.mjunoon: use a User-Agent

back-to:
Fixed https://github.com/streamlink/streamlink/issues/1969


gravyboat:
Looks good, thanks @back-to.

Issue #1967, opened 2018-07-27
Camsoda Error - Unable to validate JSON


ChamJay:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.
### Description

Plugin for camsoda.com seems to be broken probably due to some site update or something. Gets the below error,
error: Unable to validate JSON: Unable to validate key 'user': Key 'online' not found in {…}

### Reproduction steps / Explicit stream URLs to test

1. just use any camsoda link with any command (I tried with and without '--http-no-ssl-verify' and 'best -o output.mp4')

### Logs

```
[Streamlink for Windows v0.14.2 - Git 215d789]
[cli][info] Found matching plugin camsoda for URL https://www.camsoda.com/kateyjlil
[...]

PR #1972, opened 2018-07-31
plugins.camsoda: Fix user status

back-to:
Stream online

```
$ streamlink https://www.camsoda.com/x
[cli][info] Found matching plugin camsoda for URL https://www.camsoda.com/x
[cli][info] Available streams: 600p (worst, best)
[cli][info] Opening stream: 600p (hls)
```

Stream offline

```
$ streamlink https://www.camsoda.com/x/
[cli][info] Found matching plugin camsoda for URL https://www.camsoda.com/x/
error: status - offline
```

Stream hidden

```
$ streamlink https://www.camsoda.com/x/
[cli][info] Found matching plugin camsoda for URL https://www.camsoda.com/x/
error: status - connected
```

Stream ticket

```
$ streamlink https://www.camsoda.com/x/
[cli][info] Found matching plugin camsoda for URL https://www.camsoda.com/x/
error: status - limited
```

Stream private

```
$ streamlink https://www.camsoda.com/x
[cli][info] Found matching plugin camsoda for URL https://www.camsoda.com/x
error: status - private
```

Invalid username

```
$ streamlink https://www.camsoda.com/x
[cli][info] Found matching plugin camsoda for URL https://www.camsoda.com/x
error: No username found
```

Fixed https://github.com/streamlink/streamlink/issues/1967

### DIFF
--- a/src/streamlink/plugins/camsoda.py
+++ b/src/streamlink/plugins/camsoda.py
@@ -1,89 +1,91 @@
+import logging
 import random
 import re
 
+from streamlink.exceptions import PluginError
 from streamlink.plugin import Plugin
-from streamlink.plugin.api import validate
 from streamlink.plugin.api import useragents
+from streamlink.plugin.api import validate
 from streamlink.stream import HLSStream
 
-_url_re = re.compile(r"http(s)?://(www\.)?camsoda\.com/(?P<username>[^\"\']+)")
-
-_api_user_schema = validate.Schema(
-    {
-        "status": validate.any(int, validate.text),
-        validate.optional("user"): {
-            "online": validate.any(int, validate.text),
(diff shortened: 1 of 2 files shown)


gravyboat:
Do we even want to fix this instead of just removing the plugin?

back-to:
I don't care if it is removed or not,
but removing a plugin that doesn't meet the new requirements **one by one** is a **very slow** process.

If some existing plugins should be removed,
they should be removed all at the same time instead of waiting until they break.

It makes people a false hope,
that some plugin are supported when they are not getting updated when they break.

There are also some other plugins that does not match this

johnthecracker:

if you carefully rad all ot that, streamlink will evenyually be a project that works only for Twitch.

PR #1973, opened 2018-07-31
plugins.facebook: support for videos in posts

beardypig:
Support for videos posts. eg. https://www.facebook.com/119555411802156/posts/500665313691162/


gravyboat:
Thanks @beardypig, looks good.

PR #1974, opened 2018-07-31
api.useragents: use newer and more common User-Agent's

back-to:
Most of them are very old and outdated, use some recent versions.

Did not change `Safari7/8, IE6/7/8/9, WindowsPhone8`

Renamed `IPHONE_6` to `IPHONE`, but it can still be used as 6

---

Ref some comments on https://github.com/streamlink/streamlink/pull/1794


beardypig:
Nice. It must have been a pain to update all those. I started, but I didn’t have the will to finish :)

gravyboat:
Thanks @back-to.

PR #1975, opened 2018-07-31
script.makeinstaller: use a more recent version of Python and Pycryptodome

back-to:
Python = 3.6.6
Pycryptodome = 3.6.4

**Known Issue:**
Old Python 3.5 files from a previous installation, won't be removed.
(because of the limitations of pynsist)

### DIFF
--- a/script/makeinstaller.sh
+++ b/script/makeinstaller.sh
@@ -33,7 +33,7 @@ entry_point=streamlink_cli.main:main
 icon=../win32/doggo.ico
 
 [Python]
-version=3.5.2
+version=3.6.6
 format=bundled
 
 [Include]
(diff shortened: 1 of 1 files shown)


beardypig:
Anything significant that will be left behind from 3.5?

back-to:
I think only these files

```
C:\Program Files (x86)\Streamlink\Python\python35.dll
C:\Program Files (x86)\Streamlink\Python\python35.zip
```

everything else has a different timestamp and looks overwritten.

beardypig:
That sounds fine I think.

gravyboat:
Thanks @back-to and sorry about the delay on this, been a bit busy.

Issue #1141, opened 2017-08-04
Leftover version numbers from Livestreamer docs


MineRobber9000:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [x] I have read the contribution guidelines.

### Description

Some documentation still contains livestreamer version numbers, including the API reference. (things implemented in 1.6.0 when streamlink isn't even to 1.0.0)

### Suggested solution

Instead of removing these version numbers, we should prefix them so as to say that they are livestreamer versions. (i.e; "1.6.0"->"ls1.6.0" or "livestreamer 1.6.0")

back-to:
removing it **completely** might be better than `ls tags`,
since tags like this are **not made for any streamlink changes**.

since running an **old version** of streamlink is **not** that **efficient**.

PR #1987, opened 2018-08-04
Removed old Livestreamer deprecations and API references

back-to:
- removed **versionchange references** as Streamlink does not use it.
- removed **old code** that was deprecated in Livestreamer
- removed DeprecatedWarning for **2.6** as it is already mentioned in 
`setup.py`
  
https://github.com/streamlink/streamlink/blob/a14f170a04252d4275de417c13dcc2a19bf269c9/setup.py#L82
- removed **unittest2**
- **Flake8** for some lines/files that I changed.

---

Fixed https://github.com/streamlink/streamlink/issues/1141

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -7,4 +7,3 @@ mock
 requests-mock
 pynsist
 freezegun
-unittest2; python_version < '2.7'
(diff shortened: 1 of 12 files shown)


gravyboat:
Looks great @back-to, thanks for cleaning things up.

Issue #1982, opened 2018-08-04
Twitch has removed the "/api/v2/clips/" endpoint - VODs broken


Tomalak:
## Plugin Issue

- [X] This is a plugin issue and I have read the contribution guidelines.

### Description

When given a video (VOD) URL like https://www.twitch.tv/entropicj/video/12345678, the Twitch plugin checks the status and sources of that video by calling

    return http.json(self.call_subdomain("clips", "/api/v2/clips/" + clip_name + "/status", format="")

Unfortunately this API endpoint recently got removed, so that watching VODs is currently impossible.

### Reproduction steps / Explicit stream URLs to test

- go to the "videos" list of any Twitch streamer
- copy the URL of one of the VODs
- start streamlink with that URL

### Log output

```
[cli][info] Found matching plugin twitch for URL https://www.twitch.tv/user/video/12345678
error: Unable to open URL: https://clips.twitch.tv/api/v2/clips/video/status (404 Client Error: Not Found for url: https://clips.twitch.tv/api/v2/clips/video/status?as3=t)

Tomalak:
Unfortunately this seems to break HLS streaming. I can't scrub backwards/forwards in VLC with videos opened that way.

back-to:
@Tomalak 

your issue is **not related to the change**, it is an **VLC** issue.

[mpv](https://mpv.io/ ) will work as expected.

Tomalak:
@gravyboat You're right, it might have been a glitch with the video I've tested with. Tried a few others now, it actually works in VLC.

gravyboat:
@Tomalak You're tagging the wrong person. @back-to has been the one kind enough to propose the PR to fix this.

Tomalak:
Whoops, sorry. :)

PR #1988, opened 2018-08-04
plugins.twitch: new video URL regex

back-to:
https://www.twitch.tv/twitch/video/292713971

---

Fixed https://github.com/streamlink/streamlink/issues/1982

@Tomalak


gravyboat:
Good stuff, thanks @back-to.

PR #1994, opened 2018-08-05
Update and reactivate plugin for dr.dk

fuglede:
The main reason for the existing plugin failing was the URL for live
streams being outdated; no general list appears to exist, so instead we
rely on those provided for the individual channels, cf. also the approach
used for VODs.


fuglede:
Looks like there's actually an overview available at https://www.dr.dk/mu-online-stage/api/1.4/channel/all-active-dr-tv-channels still; in any case, this ends up doing the same thing.

gravyboat:
Thanks @fuglede, looks good!

PR #2001, opened 2018-08-07
Eleven Sports Taiwan Plugin

stecooper:
Channels are:

https://www.elevensports.tw/channel/eleven-zh
https://www.elevensports.tw/channel/eleven-sports-en
https://www.elevensports.tw/channel/eleven-plus-en


gravyboat:
We can't support this due to a Neulion DMCA takedown: #1493

PR #1620, opened 2018-04-25
plugins.filmon: more robust channel_id extraction

beardypig:
Fixes #1575, a continuation of @Billy2011's PR #1610.

This changes the plugin to look for the `channel_id` in the page when given a `tv` URL, this makes it more robust and fixes the issue with Sat1. 

Thanks to @bastimeyer for giving me the copy-paste quote regex ;)

PR #2005, opened 2018-08-09
plugins.filmon: fix NoPluginError, if channel is not an ID

Billy2011:
this error is related to #1620


src/streamlink/plugins/filmon.py
```diff
@@ -138,7 +139,7 @@ def _get_streams(self):
                 yield stream["quality"], FilmOnHLS(self.session, vod_id=vod_id, quality=stream["quality"])
 
         else:
-            if not channel:
+            if not self._is_channel_id_re.match(channel):
```
beardypig: Here you might be able to use `channel.isdigit()` instead of the regex.

gravyboat:
Are we good to go here @beardypig?

Issue #2020, opened 2018-08-21
Add support for steam.tv


lunks:
## Feature Request

- [x] This is a feature request and I have read the contribution guidelines.

### Description

steam has launched its own streaming website for Dota 2 TI. I'm unsure if it relates to [Steam Brodcast](https://github.com/streamlink/streamlink/pull/1717).

### Expected / Actual behavior

Allow steam.tv links to be handled by streamlink.

Currently it does not.

acz13:
I think this is just a fancy page for steam broadcast

acz13:
Close this

PR #2030, opened 2018-08-25
Update steam plugin to work with steam.tv urls

nickgal:
See also:  #2020


gravyboat:
Looks good to me, thanks. Fixes #2020.

PR #2032, opened 2018-08-25
Improve terminal progress display for  wide characters

jackyzy823:
1. Calculate string width  in terminal  instead of string length for progress output in order to support East asian Characters(CJK) better and avoid terminal display overflow.

Overflow sample:
```
[download][..ちはアロハabcd.mp4] Written 143.1 KB (0s @ 228.7 KB/s)     
     [download][..ちはアロハabcd.mp4] Written 280.5 KB (1s @ 223.9 KB/s)
          [download][..ちはアロハabcd.mp4] Written 280.5 KB (1s @ 223.9 KB/s)
``` 
2. Cut string according to unicode not bytes to avoid malformed character (in python 2)

Malformed sample
```
[download][..�好はロハabcd.mp4] Written 143.1 KB (0s @ 228.7 KB/s) 
```

Some details

|Type|Display|Length
|----|---|--|
|Word| 你好| 2
|Utf-8(Python2 bytes)| \xe4\xbd\xa0\xe5\xa5\xbd| 6|
|Unicode(Python3 str)| \u4f60\u597d| 2|
|Utf-8 [1:]|��好|5|
|Unicode [1:]|好|1|
|Terminal | 你好 |4 cells  

`1 Wide character takes 2 cells in terminal`

### DIFF
--- a/src/streamlink_cli/utils/progress.py
+++ b/src/streamlink_cli/utils/progress.py
@@ -2,6 +2,7 @@ import sys
 
 from collections import deque
 from time import time
+from unicodedata import east_asian_width
 
 from ..compat import is_win32, get_terminal_size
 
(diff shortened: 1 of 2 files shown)


src/streamlink_cli/utils/progress.py
```diff
@@ -13,22 +13,51 @@
...
 
+# widths generated from
+# http://www.unicode.org/Public/4.0-Update/EastAsianWidth-4.0.0.txt
 
-def terminal_len(value):
-    """Returns the length of the string it would be when displayed.
 
-    Attempts to decode the string as UTF-8 first if it's a bytestring.
-    """
+widths = [
```
beardypig: `unicodedata.east_asian_width` might be a better way to get this data?

beardypig: https://stackoverflow.com/questions/48598304/width-of-a-string-with-zero-width-and-two-width-characters-in-python-3-in-a-ter

jackyzy823: Using `east_asian_width` has some issues. (https://stackoverflow.com/a/31666966)

The width table in this Pull Request is based on https://github.com/urwid/urwid/blob/master/urwid/old_str_util.py#L38

Another better way is using library `wcwidth`(https://github.com/jquast/wcwidth)

Thanks for reviewing.

gravyboat:
Are we good to go here @beardypig or is there further feedback?

gravyboat:
Thanks @jackyzy823, and thanks for the review @beardypig.

PR #2070, opened 2018-09-17
plugin request: new plugin for egame.qq.com

fozzysec:
https://egame.qq.com/ is a Chinese game living sites.


back-to:
I updated your branch.

`git diff` won't show it correctly because the file format was changed to `LF`,
for more information see the commit message.

Issue #2034, opened 2018-08-26
Problems with filmon


LustigesBassgebrumm:
## Feature Request

- [x] This is a feature request and I have read the contribution guidelines.

### Description

Hello,
I have some problems with the plugin Filmon. All streams end after a few minutes. The length to standstill is different. Can someone tell me what I need to do to fix the bugs? The debug output provides the following:

root@vusolo2:~# /usr/sbin/streamlinksrv manualstart
[streamlinksrv][info] Sun Aug 26 16:12:33 2018 Server (1.1.0) started
[streamlinksrv][info] Host:          vusolo2
[streamlinksrv][info] Port:          8088
[streamlinksrv][info] OS:            Linux-3.13.5-mips-with-glibc2.0
[streamlinksrv][info] Python:        2.7.9
[streamlinksrv][info] Streamlink:    0.14.2 / 06.07.2018
[streamlinksrv][info] Requests(2.12.4), Websocket(0.48.0)
[streamlinksrv][info] Processing URL: https://www.filmon.com/tv/itv1
[streamlinksrv][info] Found matching plugin filmon for URL https://www.filmon.com/tv/itv1
[streamlinksrv][info] Available streams:
low (worst), high (best)
[streamlinksrv][info] Opening stream: high (hls-filmon)
[streamlinksrv][error] Got exception: Read timeout
['Traceback (most recent call last):', '  File "/usr/sbin/streamlinksrv", line 207, in Stream', '    shutil.copyfileobj(fd, http.wfile)', '  File "/usr/lib/python2.7/shutil.py", line 49, in copyfileobj', '    buf = fsrc.read(length)', '  File "/usr/lib/python2.7/site-packages/streamlink/stream/segmented.py", line 207, in read', '    timeout=self.timeout)', '  File "/usr/lib/python2.7/site-packages/streamlink/buffers.py", line 111, in read', '    raise IOError("Read timeout")', 'IOError: Read timeout']
[streamlinksrv][info] Stream ended
[streamlinksrv][info] Closing currently open stream...

[streamlinksrv][info] Processing URL: https://www.filmon.com/tv/itv2
[streamlinksrv][info] Found matching plugin filmon for URL https://www.filmon.com/tv/itv2
[streamlinksrv][info] Available streams:
low (worst), high (best)
[streamlinksrv][info] Opening stream: high (hls-filmon)
[...]

PR #2074, opened 2018-09-20
plugins.filmon: use the same cdn for a playlist url

back-to:
do not change the **cdn** when a playlist gets reloaded,
avoids loops with the same segment and read timeout.

better handle of url get error on playlist reload

use normal HLSStream with parse_variant_playlist for VODs

---

PR is semi based of https://github.com/streamlink/streamlink/pull/2037
without the `process_sequences` part, instead the same **cdn** will be used
and the code is in the **plugin file** instead of the global **hls file**.

closes https://github.com/streamlink/streamlink/pull/2037
closes https://github.com/streamlink/streamlink/issues/2034

### DIFF
--- a/src/streamlink/plugins/filmon.py
+++ b/src/streamlink/plugins/filmon.py
@@ -3,11 +3,71 @@ import logging
 import time
 
 from streamlink import StreamError
+from streamlink.compat import urlparse
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
-from streamlink.stream import HLSStream
+from streamlink.stream import HLSStream, hls_playlist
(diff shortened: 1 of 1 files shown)


Billy2011:
@back-to I think one `urlparse` is enough:
```
    if self.stream.channel:
            parsed = urlparse(self.stream.url)
            if self.stream._first_netloc is None:
                # save the first netloc
                self.stream._first_netloc = parsed.netloc
            # always use the first saved netloc
            new_stream_url = parsed._replace(netloc=self.stream._first_netloc).geturl()
     else:
            new_stream_url = self.stream.url
```

back-to:
yes, one `urlparse` is enough.

PR #2086, opened 2018-10-02
cli: make the HH part of HH:MM:SS options optional

beardypig:
Options `HH` part in the `HH:MM:SS` options, so that you can do `10:00` for 10 minutes. 

Also gives extra utility to the `hours_minutes_seconds` method for parsing time offsets from streaming sites, etc.


gravyboat:
Thanks @beardypig.

PR #2090, opened 2018-10-03
cli: disable progress output for -o when no tty is available

### LINKED ISSUE #1770: Add an option to disable the download progress log message
### Checklist

- [ ] This is a bug report.
- [x] I have read the contribution guidelines.

### Description

Apart from regular console use, I also run Streamlink from Java (through a Java Process) and redirect the system output & error streams to Java. If I tell Streamlink to use a player, all is fine. However, if I tell Streamlink to save the file, it stops writing to the file after less than 1 minute. It appears to still be capturing as there are no new log messages to indicate an error but the file doesn't get any bigger.

After some investigating, it turns out to be the "[download] ??? Written ???" log message that clears the current line in order to print the new value that is causing the problem (the print_inplace function in progress.py). That log message isn't printed in Java which was a clue to the problem.

I've tried the none loglevel option but the download progress log message still appears. My only option ATM without modifying the Streamlink code myself is to turn the loglevel up to debug level causing no line clearing, just a whole lot of logspam, BUT a successful recording.

TLDR;
When I specify -Q option for quiet logging, can it NOT print the download progress as well please.
Or maybe an extra CLI argument to disable the download progress message altogether (or disable current line clearing and just write the new value on a new line). Ideally, I'd still like to get the error/warning/info log messages and just disable the constantly updating download progress.

### Comments, screenshots, etc.

Thanks to all the contributors for this awesome software.

beardypig:
Changes to the behaviour of the progress bar when using -o.

If there is no tty available for output (eg. when piping, or running a subprocess), then no progress bar is output. This is because the progress bar rewrites the console, which is only possible on interactive terminals not for a pipe.

If there is no tty available for input and the file already exists an error is given instead of a prompt to the user.

Should fix #1770.

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -57,10 +57,14 @@ def check_file_output(filename, force):
     log.debug("Checking file output")
 
     if os.path.isfile(filename) and not force:
-        answer = console.ask("File {0} already exists! Overwrite it? [y/N] ",
-                             filename)
+        if sys.stdin.isatty():
+            answer = console.ask("File {0} already exists! Overwrite it? [y/N] ",
+                                 filename)
 
-        if answer.lower() != "y":
+            if answer.lower() != "y":
+                sys.exit()
+        else:
+            log.error("File {0} already exists, use --force to overwrite it.".format(filename))
             sys.exit()
 
     return FileOutput(filename)
@@ -322,7 +326,7 @@ def read_stream(stream, output, prebuffer, chunk_size=8192):
     is_player = isinstance(output, PlayerOutput)
     is_http = isinstance(output, HTTPServer)
     is_fifo = is_player and output.namedpipe
-    show_progress = isinstance(output, FileOutput) and output.fd is not stdout
+    show_progress = isinstance(output, FileOutput) and output.fd is not stdout and sys.stdout.isatty()
 
     stream_iterator = chain(
         [prebuffer],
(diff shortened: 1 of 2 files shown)


src/streamlink_cli/main.py
```diff
@@ -56,12 +56,16 @@ def check_file_output(filename, force):
...
-        if answer.lower() != "y":
-            sys.exit()
+    if sys.stdin.isatty():
+        if os.path.isfile(filename) and not force:
+            answer = console.ask("File {0} already exists! Overwrite it? [y/N] ",
+                                 filename)
+
+            if answer.lower() != "y":
+                sys.exit()
+    elif os.path.isfile(filename):
```
back-to: `--force` won't work, ` and not force` is required

instead of **repeating** it, it could be called first and isatty could be used after it

```py
    if os.path.isfile(filename) and not force:
        if sys.stdin.isatty():
            answer = console.ask("File {0} already exists! Overwrite it? [y/N] ",
                                 filename)

            if answer.lower() != "y":
                sys.exit()
        else:
            log.error("File {0} already exists, use --force to overwrite it.".format(filename))
            sys.exit()
```

beardypig: Yep. My bad :)

beardypig:
Hmm. I pushed some changes to my branch but it doesn't seem to have updated.

bastimeyer:
@beardypig https://status.github.com/messages

beardypig:
That's a bunch of red :)

gravyboat:

Quaraxkad:
With this change, how can an external process monitor the progress of streamlink while it's running?

PR #2091, opened 2018-10-03
docs: add appropriate funimation docs

vinnyA3:
I added documentation and notes for the Funimation plugin

**cli.rst**:
  - added authentication and `incap_ses` cookie information

**plugin_matrix.rst**:
  - added note about the required cookie

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -245,7 +245,7 @@ Authenticating with FunimationNow
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Like Crunchyroll, the FunimationNow plugin requires authenticating with a premium account to access some
 content: :option:`--funimation-email`, :option:`--funimation-password`. In addition, this plugin requires a ``incap_ses`` cookie to be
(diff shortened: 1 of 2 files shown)


gravyboat:
Looks great. @beardypig any feedback? My only notes would be do we need to title it as FunimationNow instead of just Funimation? Would this lead to confusion, and should we link to the docs directly from the note in the plugin matrix? If that happens we should probably do the same for crunchyroll.

vinnyA3:
@bastimeyer  I could definitely see some confusion arising from the explicit cookie value, thanks for the review!  A link to #2091 would be a good addition as well.

@beardypig  I think FunimationNow would be best to avoid some confusion;  I'll make the updates!  As **bastimeyer** suggested above, I'll link to the issue for some further clarification.  If you would like, I could link to the funimation and crunchyroll docs in the plugin_matrix.

As for the digital ocean t-shirt:  I completely forgot about that!  That's really cool, I'm going to sign up :smile:

beardypig:
Good idea @vinnyA3, I didn't think to add documentation ;) Thanks!

Also, hope you get the hacktoberfest t-shirt...

vinnyA3:
Glad I could help a bit @beardypig :)  I hope to get the shirt too lol

PR #2092, opened 2018-10-04
[plugins.egame]update regex to support new pattern

fozzysec:
test url: https://egame.qq.com/497383565


gravyboat:
Thanks @fozzysec.

PR #2094, opened 2018-10-04
travis: increase git clone depth

back-to:
There are not enough commits for Versioneer `streamlink-0+untagged`,
with more commits the Streamlink version should be fixed for travis.

Fixes Windows build.

Error Example:
https://travis-ci.org/streamlink/streamlink/jobs/436299646#L1165
https://github.com/streamlink/streamlink.github.io/commit/42e74cce133a0dfcc6c40a9a71ebbaa336b3285f#diff-1438650dd2508a32fb38f7b8aa3c8e19

Ref:
https://docs.travis-ci.com/user/customizing-the-build/#git-clone-depth


gravyboat:
Nice catch @back-to.

PR #2096, opened 2018-10-05
stream.dash_manifest: Fixed bug for dynamic playlists when parent.id is None

back-to:
if `self.parent.id` is None, use `self.parent.mimeType` instead.

This will create two timelines for video and audio,
instead of one for both which won't work.

---

I don't know that much about the MPD class,
maybe it could be solved differently.

---

Test URLs:

https://www.france.tv/france-5/direct.html

http://livesim.dashif.org/livesim/segtimeline_1/testpic_2s/Manifest.mpd
from https://github.com/Dash-Industry-Forum/dash-live-source-simulator/wiki/Test-URLs

---

closes https://github.com/streamlink/streamlink/pull/1934

### DIFF
--- a/src/streamlink/stream/dash_manifest.py
+++ b/src/streamlink/stream/dash_manifest.py
@@ -509,6 +509,10 @@ class SegmentTemplate(MPDNode):
 
     def format_media(self, **kwargs):
         if self.segmentTimeline:
+            if self.parent.id is None:
+                # workaround for invalid `self.root.timelines[self.parent.id]`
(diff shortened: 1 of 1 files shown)


bastimeyer:
@back-to, if you rebase to master, the build should succeed

gravyboat:
Thanks @back-to!

beardypig:
Another way to do it, would have been to _generate_ an `id` when none exists, for example using the md5sum of the content on the tag :)

PR #2107, opened 2018-10-10
Added --default-plugin option

Talarion291:
If default-plugin is used, you are able to use "streamlink --default-plugin twitch.tv imaqtpie" or if config file is used, just "streamlink imaqtpie"

### DIFF
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -77,7 +77,8 @@ class Streamlink(object):
             "ffmpeg-video-transcode": "copy",
             "ffmpeg-audio-transcode": "copy",
             "locale": None,
-            "user-input-requester": None
+            "user-input-requester": None,
+            "default-plugin": None
         })
         if options:
             self.options.update(options)
@@ -241,6 +242,10 @@ class Streamlink(object):
                                  to collect input from the user at runtime. Must be
                                  set before the plugins are loaded.
                                  default: ``UserInputRequester``.
+
+        default-plugin           (str) If no plugin could be found, instead of
+                                raising NoPluginError, tries to use default-plugin
+                                first.
         ======================== =========================================
 
         """
@@ -387,12 +392,21 @@ class Streamlink(object):
         :param follow_redirect: follow redirects
 
         """
+        default_plugin_url = url
+        default_plugin = self.get_option("default_plugin")
+        if default_plugin is not None:
+            default_plugin_url = default_plugin + "/" + default_plugin_url
+            default_plugin_url = update_scheme("http://", default_plugin_url)
         url = update_scheme("http://", url)
 
         available_plugins = []
         for name, plugin in self.plugins.items():
             if plugin.can_handle_url(url):
                 available_plugins.append(plugin)
+            if default_plugin is not None:
+                if plugin.can_handle_url(default_plugin_url):
+                    url = default_plugin_url
+                    available_plugins.append(plugin)
 
         available_plugins.sort(key=lambda x: x.priority(url), reverse=True)
         if available_plugins:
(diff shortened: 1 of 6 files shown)


back-to:
the PR itself is a bit to customized like above described

---

but as a **side note**

if the goal was to open streams without the usage of copy and paste
you might enjoy [Open With](https://github.com/darktrojan/openwith)

it can open Streamlink directly from your Browser URL

https://addons.mozilla.org/en-US/firefox/addon/open-with/
https://chrome.google.com/webstore/detail/open-with/cogjlncmljjnjpbgppagklanlcbchlno

small guide

https://gist.github.com/back-to/191f522011deb19b0ab2d70aa5f52c12#file-readme-rst

Talarion291:
New commit enables following options:
``streamlink --default-plugin twitch channel   => opens twitch.tv/channel``
``streamlink --default-plugin twitch vodnr => opens https://www.twitch.tv/videos/vodnr``
``streamlink --default-plugin youtube watchnr => opens https://www.youtube.com/watch?v=EU90h239832``
As well as enables plugin-creator to implement plugin.get_urls to support default-plugin functionallity (if no other options are required for plugin to fetch stream)

src/streamlink/session.py
```diff
@@ -387,13 +392,40 @@ def resolve_url(self, url, follow_redirect=True):
...
+            plugin_found = False
+
+            for name, plugin in self.plugins.items():
+                if plugin.__name__.lower() == default_plugin.lower():
+                    plugin_found = True
+                    for string in plugin.get_urls():
+                        if plugin.can_handle_url(string + default_plugin_url_copy):
+                            try:
+                                plugininstance = plugin(string + default_plugin_url_copy)
+                                streams = plugininstance.streams(None, None)
```
bastimeyer: Plugin instantiation doesn't belong into the `resolve_url` method...

src/streamlink/session.py
```diff
@@ -387,13 +392,40 @@ def resolve_url(self, url, follow_redirect=True):
...
                 available_plugins.append(plugin)
 
+        if len(available_plugins) == 0 and default_plugin is not None:
+            plugin_found = False
+
+            for name, plugin in self.plugins.items():
+                if plugin.__name__.lower() == default_plugin.lower():
+                    plugin_found = True
+                    for string in plugin.get_urls():
+                        if plugin.can_handle_url(string + default_plugin_url_copy):
```
bastimeyer: This is what I meant with the URL schemes. You can't expect every plugin/website to have their URL formatted like this where the channel name or video ID or something else gets added to the end of the defined plugin URLs. You need to use a URL template that replaces a `{value}` variable for example.

**Example:**

```python
# string = "https://twitch.tv/{value}"
# string = "https://youtube.com/watch?v={value}"
# string = "https://some.streaming.website/{value}/live"
string.format(value=default_plugin_url_copy)
```

This should also be done only once (you are currently concatenating the URL parts 3 times).

src/streamlink/session.py
```diff
@@ -387,13 +392,40 @@ def resolve_url(self, url, follow_redirect=True):
...
+                                plugininstance = plugin(string + default_plugin_url_copy)
+                                streams = plugininstance.streams(None, None)
+                                available_plugins.append(plugin)
+                                url = string + default_plugin_url_copy
+                                break
+                            except (NoPluginError, PluginError) as e:
+                                pass
+
+                    if len(available_plugins) == 0:
+                        print("No streams with default plugin " + default_plugin + " found, plugin either doesnt support url shortening or url is not valid. Try without --default-plugin.")
```
bastimeyer: Please use the appropriate logging methods and watch the max line length.
https://github.com/streamlink/streamlink/blob/d022e9b8fde1570727267ff0e39c5cf6f6c87e2c/src/streamlink/session.py#L22

Is it even necessary to log something here? I think a general log message which says that it is looking for a fallback plugin now when the parameter is set and the `available_plugins` is empty is more fitting. The message below could then also be removed, as it is unnecessary as well.

back-to:
see above, if you want something like this it should be done totally differently.

PR #1888, opened 2018-06-30
Update method for finding YouTube videoIds

beardypig:
Uses an alternative method to find the videoId for YouTube videos. First the canonical page is loaded (if different from the current page, eg. embed pages), then we use `currentVideoEndpoint.watchEndpoint.videoId` (credits to @back-to) from `ytInitialData` to find the videoId. This worked on all the pages I tested.

The decoding of the video info has been improved to support unicode error messages. This requires the logging fix from #1886 to work correctly. 

Blocked by #1886 

Fixes #1876
Fixes #1865

PR #2113, opened 2018-10-12
plugins.youtube: Added support for embedded livestream urls

back-to:
URL regex update.

use canonical url for embedded livestreams,
this was not added before because it is not needed for normal videos.

Ref https://github.com/streamlink/streamlink/pull/1888#discussion_r199504011

Added some more regex group tests.

---

Example where this url can be found

`https://www.aljazeera.com/live/` with `Having trouble watching?`

The youtube url is currently not used for Streamlink,
but might be useful for other future websites.


gravyboat:
Looks good, thanks @back-to.

PR #2127, opened 2018-10-22
feature: {best,worst}-unfiltered stream synonyms

### LINKED ISSUE #1055: Fallback selection for --stream-sorting-excludes
### Checklist

- [ ] This is a bug report.

### Description

Related issue:
https://github.com/streamlink/streamlink-twitch-gui/issues/481

In certain situations, the usage of `--stream-sorting-excludes` can lead to an empty stream selection. This is an issue for applications which try to map the available streams to a custom list of qualities.

The Streamlink Twitch GUI needs to do this stream quality mapping because of Twitch's inconsistent quality naming. This is an example of trying to find the "low" quality:

```
# available: 160p, 360p, 480p, 720p, 1080p
streamlink --stream-sorting-excludes ">360p" twitch.tv/CHANNEL low,best
# will open 360p

# available: low, medium, high, source
streamlink --stream-sorting-excludes ">360p" twitch.tv/CHANNEL low,best
# will open low

# available: 720p
streamlink --stream-sorting-excludes ">360p" twitch.tv/CHANNEL low,best
# returns an empty selection (channel doesn't have re-encoded qualities)
```

Having a fallback parameter (eg. `--stream-sorting-excludes-fallback`) with a list of streams that will be opened if no matching stream could be found would fix this situation.

Another solution could be having a different notation in the actual quality selection, but this is probably a bit too complex.

```
# 1. try to find the quality labeled "medium"
# 2. look for qualities named by video resolution matching the defined range
# 3. fall back to lower qualities
# 4. use any available quality
streamlink twitch.tv/CHANNEL "medium||>360p&&<=480p||<=360p||best"
streamlink twitch.tv/CHANNEL "medium,>360p<=480p,<=360p,best"
```

bastimeyer:
Add special stream synonyms "best-unfiltered" and "worst-unfiltered" in
cases when all available streams have been filtered out by the
`--stream-sorting-excludes` parameter and the "best" and "worst" stream
synonyms don't exist. These new synonyms point towards the respective
streams of the unfiltered streams list so that the user is able to
select a fallback stream.

Example:
streamlink --stream-sorting-excludes '>=480p' URL best,best-unfiltered
Will try to open the best stream below 480p, but if none is available
will continue with the best of all streams as a fallback selection.

Resolves #1055

### DIFF
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -384,6 +384,7 @@ class Plugin(object):
 
         stream_names = filter(stream_weight_only, streams.keys())
         sorted_streams = sorted(stream_names, key=stream_weight_only)
+        unfiltered_sorted_streams = sorted_streams
 
         if isinstance(sorting_excludes, list):
             for expr in sorting_excludes:
@@ -402,6 +403,11 @@ class Plugin(object):
             worst = sorted_streams[0]
             final_sorted_streams["worst"] = streams[worst]
             final_sorted_streams["best"] = streams[best]
+        elif len(unfiltered_sorted_streams) > 0:
+            best = unfiltered_sorted_streams[-1]
+            worst = unfiltered_sorted_streams[0]
+            final_sorted_streams["worst-unfiltered"] = streams[worst]
+            final_sorted_streams["best-unfiltered"] = streams[best]
 
         return final_sorted_streams
 
--- a/tests/plugins/testplugin.py
+++ b/tests/plugins/testplugin.py
@@ -37,6 +37,14 @@ class TestPlugin(Plugin):
     def _get_streams(self):
         if "empty" in self.url:
             return
+
+        if "UnsortableStreamNames" in self.url:
+            def gen():
+                for i in range(3):
+                    yield "vod", HTTPStream(self.session, "http://test.se/stream")
+
+            return gen()
+
         if "NoStreamsError" in self.url:
             raise NoStreamsError(self.url)
 
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -107,7 +107,7 @@ def build_parser():
         help="""
         Stream to play.
 
-        Use "best" or "worst" for selecting the highest or lowest available
+        Use ``best`` or ``worst`` for selecting the highest or lowest available
         quality.
 
         Fallback streams can be specified by using a comma-separated list:
@@ -523,7 +523,7 @@ def build_parser():
         help="""
         Stream to play.
 
-        Use "best" or "worst" for selecting the highest or lowest available
+        Use ``best`` or ``worst`` for selecting the highest or lowest available
         quality.
 
         Fallback streams can be specified by using a comma-separated list:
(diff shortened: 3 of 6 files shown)


bastimeyer:
Something on GH seems to be broken right now, oh boy... https://status.github.com/messages :confused:

bastimeyer:
> will decrease coverage by 0.16%

?????
https://codecov.io/gh/streamlink/streamlink/commit/0aad083fdc23ed649b736f29cccab1b2ae25e5be
https://codecov.io/gh/streamlink/streamlink/compare/b21df07689023f1f6fc79b228da4faa76c617667...0aad083fdc23ed649b736f29cccab1b2ae25e5be/changes

gravyboat:
Talk about an edge case. This looks good to me @bastimeyer. I'll leave it open for a bit longer in case anyone else wants to weigh in, but if not we'll get it merged.

bastimeyer:
Let's tag it as enhancement...

> edge case

The main problem and motivation for this change is that people who are using the Twitch GUI can set different stream quality selections which filter out unwanted streams by using the stream-sorting-excludes parameter. This has to be done because Twitch doesn't provide an API for knowing the available qualities, so you have to define the selection beforehand (or manually parse the HLS playlist in a couple of redundant requests, which is kinda stupid when using Streamlink).
Now If a user has selected a medium quality and thus filters out everything above the specified medium qualities and a streamer on Twitch doesn't have a full partnership and only a certain quality above "medium" is available, this stream-sorting-excludes filter fails. By having fallback stream name synonyms, this won't happen.

src/streamlink/plugin/plugin.py
```diff
@@ -402,6 +403,11 @@ def stream_weight_only(s):
             worst = sorted_streams[0]
             final_sorted_streams["worst"] = streams[worst]
             final_sorted_streams["best"] = streams[best]
+        else:
+            best = unfiltered_sorted_streams[-1]
```
back-to: current behavior **without this PR**

```
error: The specified stream(s) 'best' could not be found.
       Available streams: vod, vod_alt, vod_alt2
```

**with this PR**

as `vod` is not a valid sorted stream name for Streamlink,
**sorted_streams (unfiltered_sorted_streams)** is empty here and it will error

```
  File "venv/bin/streamlink", line 11, in <module>
  File "src/streamlink_cli/main.py", line 1007, in main
    handle_url()
  File "src/streamlink_cli/main.py", line 557, in handle_url
  File "src/streamlink_cli/main.py", line 437, in fetch_streams
  File "src/streamlink/plugin/plugin.py", line 407, in streams
    best = unfiltered_sorted_streams[-1]
IndexError: list index out of range
```

it shouldn't error and it should return the same as before or handle it with some unfiltered worst/best

---

this can be used to test it without a valid URL / Plugin
```diff
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -382,9 +382,11 @@ class Plugin(object):
             return (self.stream_weight(s)[0] or
                     (len(streams) == 1 and 1))
 
+        streams = {'vod': '', 'vod_alt': '', 'vod_alt2': ''}
         stream_names = filter(stream_weight_only, streams.keys())
         sorted_streams = sorted(stream_names, **key=stream_weight_only)**
```

bastimeyer:
Interesting... There's something going on with parsing the `vod_alt` stream name on Python<=3.5. This is unrelated to the PR because I was using the suggested stream names by @back-to.

tests/plugins/testplugin.py
```diff
@@ -37,6 +37,12 @@ def get_category(self):
     def _get_streams(self):
         if "empty" in self.url:
             return
+        if "UnsortableStreamNames" in self.url:
+            return {
+                "vod": HTTPStream(self.session, "http://test.se/stream"),
+                "vod_alt": HTTPStream(self.session, "http://test.se/stream"),
```
back-to: the `_alt` tag is to early

you are here

https://github.com/streamlink/streamlink/blob/cd6f94a5a4ee83cb5b61a295df9e13956bcfcb27/src/streamlink/plugin/plugin.py#L317

and they are all called `vod` here

it will be added later

https://github.com/streamlink/streamlink/blob/cd6f94a5a4ee83cb5b61a295df9e13956bcfcb27/src/streamlink/plugin/plugin.py#L358-L366

---

you might use a generator instead of a dict here

```py
        if "UnsortableStreamNames" in self.url:
            def gen():
                for i in range(3):
                    yield "vod", HTTPStream(self.session, "http://test.se/stream")
            return gen()
```

Issue #2126, opened 2018-10-22
Bilibili Live error


aliofwind:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description

Bilibili Live error

### Reproduction steps / Explicit stream URLs to test

1. https://live.bilibili.com/3
2. https://live.bilibili.com/14047
3. https://live.bilibili.com/*

### Log output

```
C:\Users\aliof>streamlink -p potplayermini64 https://live.bilibili.com/3 best
[cli][info] Found matching plugin bilibili for URL https://live.bilibili.com/3
error: Unable to parse JSON: Expecting value: line 1 column 1 (char 0) ()
```

### Additional comments, screenshots, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

qkolj:
I cannot reproduce your issue, `streamlink https://live.bilibili.com/3 best` works for me without any problems. Are you sure you are using the latest version?

aliofwind:
bililive team said that they modified the address of the playurl yesterday, but the original address was restored after the test. thanks

undeadking:
same problem again

```
streamlink https://live.bilibili.com/3 best -l debug
[cli][debug] OS:         macOS 10.13.6
[cli][debug] Python:     3.7.0
[cli][debug] Streamlink: 0.14.2
[cli][debug] Requests(2.19.1), Socks(1.6.7), Websocket(0.48.0)
[cli][info] Found matching plugin bilibili for URL https://live.bilibili.com/3
error: Unable to parse JSON: Expecting value: line 1 column 1 (char 0) ()
```

PR #2129, opened 2018-10-23
plugins.bilibili: Fix plugin after API change

qkolj:
#2126 was closed because we couldn't reproduce the issue OP had, but it appears that the website indeed changed its API as @undeadking pointed out. With these changes, the plugin works again, but I am unsure if they will revert back API changes again (as it already happened when #2126 was closed).


gravyboat:
Talk about annoying to flip flop back and forth between APIs. Thanks for the fix @qkolj.

PR #2131, opened 2018-10-23
plugins.vimeo: add support for vimeo.com

melmorabity:
This plugin adds support for Vimeo. It supports both VOD and live streams. It also supports subtitles.

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -245,6 +245,7 @@ viasat                  - juicyplay.dk       Yes   Yes   Streams may be geo-rest
                         - viafree.no
                         - viafree.se
 vidio                   vidio.com            Yes   Yes
+vimeo                   vimeo.com            Yes   Yes   Password-protected videos are not supported.
 vinhlongtv              thvli.vn             Yes   No    Streams are geo-restricted to Vietnam
 vk                      vk.com               Yes   Yes
 vrtbe                   vrt.be/vrtnu         Yes   Yes
@@ -273,7 +274,6 @@ zattoo                  - zattoo.com         Yes   Yes
 zdf_mediathek           zdf.de               Yes   Yes   Streams may be geo-restricted to Germany.
 zengatv                 zengatv.com          Yes   No
 zhanqi                  zhanqi.tv            Yes   No
-vimeo                   vimeo.com            Yes   Yes   Password-protected videos are not supported.
 ======================= ==================== ===== ===== ===========================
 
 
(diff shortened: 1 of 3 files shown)


gravyboat:
Since when does Vimeo do live streams? Can you link to some of them? We've been trying to take the stance of not allowing sites which are mostly VODs since we updated the plugin request requirements a while back so I'm hesitant to add Vimeo.

back-to:
they got live streams

https://vimeo.com/blog/post/vimeo-live-is-here-professional-live-streaming

but they are rare because a Premium Plan is required

https://vimeo.com/upgrade?vcid=31608

docs/plugin_matrix.rst
```diff
@@ -273,6 +273,7 @@ zattoo                  - zattoo.com         Yes   Yes
 zdf_mediathek           zdf.de               Yes   Yes   Streams may be geo-restricted to Germany.
 zengatv                 zengatv.com          Yes   No
 zhanqi                  zhanqi.tv            Yes   No
+vimeo                   vimeo.com            Yes   Yes   Password-protected videos are not supported.
```
back-to: alphabetical order

src/streamlink/plugins/vimeo.py
```diff
@@ -0,0 +1,101 @@
...
+        PluginArgument(
+            "mux-subtitles",
+            action="store_true",
+            help="Automatically mux available subtitles in to the output stream.",
+        )
+    )
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return Vimeo._url_re.match(url)
```
back-to:
```py
return cls._url_re.match(url)
```

docs/plugin_matrix.rst
melmorabity: Good catch, fixed

src/streamlink/plugins/vimeo.py
melmorabity: Fixed too

src/streamlink/plugins/vimeo.py
```diff
@@ -0,0 +1,101 @@
...
+        for stream_type in ("hls", "dash"):
+            if not stream_type in videos:
+                continue
+            for _, video_data in videos[stream_type]["cdns"].items():
+                url = video_data.get("url")
+                if stream_type == "hls":
+                    for stream in HLSStream.parse_variant_playlist(self.session, url).items():
+                        streams.append(stream)
+                elif stream_type == "dash":
+                    url = url.replace("master.json", "master.mpd")
```
back-to: this doesn't work with the livestream

```
error: Unable to parse XML: not well-formed (invalid token): line 1, column 0 (b'{"url":"https://live.vimeocdn.com ...)
```

url is `https://live-api.cloud.vimeo.com/sessions/.../playlist/dash.mpd?json=1&~id=player&~exp=&~sig=`

and the output is a json file with a dash url

`{"url":"https://live.vimeocdn.com/.../.../dash.mpd"}`

gravyboat:
Thanks @melmorabity and @back-to!

PR #2137, opened 2018-10-24
Add a note about specifying the full player path

gravyboat:
I added a note under the configuration file contents about adding the full path because I realized we don't have one in the config file example. I also updated the APPDATA note since with Windows 10 it's now `AppData`

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -106,7 +106,7 @@ You can also specify the location yourself using the :option:`--config` option.
 .. note::
 
   - `$XDG_CONFIG_HOME` is ``~/.config`` if it has not been overridden
-  - `%APPDATA%` is usually ``<your user directory>\Application Data``
+  - `%APPDATA%` is usually ``<your user directory>\AppData``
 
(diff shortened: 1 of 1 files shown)


docs/cli.rst
```diff
@@ -143,6 +143,9 @@ Example
     # Authenticate with Twitch
     twitch-oauth-token=mytoken
 
+.. note::
+    Full player paths are supported via configuration file options such as
+    ``player="C:\mpv-x86_64\mpv"``
```
bastimeyer: Line breaks don't work like this in ReST. You have to define `|br|` at the top of the document like in install.rst:
https://github.com/streamlink/streamlink/blame/0.14.2/docs/install.rst#L3-L6
And then use `|br|` like an HTML line break element:
https://github.com/streamlink/streamlink/blame/0.14.2/docs/install.rst#L293

gravyboat: I don't actually want there to be a line break in the note there, it was just over the 79 character line length limit.

PR #2144, opened 2018-10-29
Use mpv as default player

gravyboat:
VLC has been pretty bad the past few releases for compatibility (generating a lot of issues) as well as functionality wise where as mpv has continued working without issue. I'm proposing that we swap over to mpv for the default player going forward.

Edit: This should not simply be merged without review and significant discussion as it would only be able to go into 1.0 as it's a breaking change. I'm not 100% sure on the mac paths but I believe these should be okay (I don't have a mac currently to test it on a live system).

### DIFF
--- a/src/streamlink_cli/utils/player.py
+++ b/src/streamlink_cli/utils/player.py
@@ -15,23 +15,23 @@ def check_paths(exes, paths):
 def find_default_player():
     if "darwin" in sys.platform:
         paths = os.environ.get("PATH", "").split(":")
-        paths += ["/Applications/VLC.app/Contents/MacOS/"]
-        paths += ["~/Applications/VLC.app/Contents/MacOS/"]
-        path = check_paths(("VLC", "vlc"), paths)
+        paths += ["/Applications/MPV.app/Contents/MacOS/"]
+        paths += ["~/Applications/MPV.app/Contents/MacOS/"]
+        path = check_paths(("MPV", "mpv"), paths)
     elif "win32" in sys.platform:
-        exename = "vlc.exe"
+        exename = "mpv.exe"
         paths = os.environ.get("PATH", "").split(";")
         path = check_paths((exename,), paths)
 
         if not path:
-            subpath = "VideoLAN\\VLC\\"
+            subpath = "mpv-x86_64\\mpv\\"
             envvars = ("PROGRAMFILES", "PROGRAMFILES(X86)", "PROGRAMW6432")
             paths = filter(None, (os.environ.get(var) for var in envvars))
             paths = (os.path.join(p, subpath) for p in paths)
             path = check_paths((exename,), paths)
     else:
         paths = os.environ.get("PATH", "").split(":")
-        path = check_paths(("vlc",), paths)
+        path = check_paths(("mpv",), paths)
 
     if path:
         # Quote command because it can contain space
--- a/src/streamlink_cli/utils/player.py
+++ b/src/streamlink_cli/utils/player.py
@@ -15,8 +15,8 @@ def check_paths(exes, paths):
 def find_default_player():
     if "darwin" in sys.platform:
         paths = os.environ.get("PATH", "").split(":")
-        paths += ["/Applications/MPV.app/Contents/MacOS/"]
-        paths += ["~/Applications/MPV.app/Contents/MacOS/"]
+        paths += ["/Applications/mpv.app/Contents/MacOS/"]
+        paths += ["~/Applications/mpv.app/Contents/MacOS/"]
         path = check_paths(("MPV", "mpv"), paths)
     elif "win32" in sys.platform:
         exename = "mpv.exe"
--- a/src/streamlink_cli/utils/player.py
+++ b/src/streamlink_cli/utils/player.py
@@ -24,7 +24,7 @@ def find_default_player():
         path = check_paths((exename,), paths)
 
         if not path:
-            subpath = "mpv-x86_64\\mpv\\"
+            subpath = "mpv-x86_64\\"
             envvars = ("PROGRAMFILES", "PROGRAMFILES(X86)", "PROGRAMW6432")
             paths = filter(None, (os.environ.get(var) for var in envvars))
             paths = (os.path.join(p, subpath) for p in paths)
--- a/win32/streamlinkrc
+++ b/win32/streamlinkrc
@@ -1,7 +1,7 @@
 # Format is option=value. Lines starting with a # is considered comments
 # and are ignored.
 
-# By default streamlink will attempt to locate MPV on your system
+# By default streamlink will attempt to locate mpv on your system
 # and use that, but you can also specify the location of a player
 # yourself.
 
--- a/win32/streamlinkrc
+++ b/win32/streamlinkrc
@@ -1,7 +1,7 @@
 # Format is option=value. Lines starting with a # is considered comments
 # and are ignored.
 
-# By default streamlink will attempt to locate VLC on your system
+# By default streamlink will attempt to locate MPV on your system
 # and use that, but you can also specify the location of a player
 # yourself.
 
@@ -9,7 +9,12 @@
 # is because the player command is parsed like a shell command to allow
 # parameters to be passed to the player.
 
-# Here is a few examples of players:
+# Here are a few examples of players:
+
+# mpv
+
+#player="C:\Program Files (x86)\mpv-x86_64\mpv.exe"
+#player="C:\Program Files\mpv-x86_64\mpv.exe"
 
 # VLC
 #player="C:\Program Files (x86)\VideoLAN\VLC\vlc.exe"
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![TravisCI build status][travisci-build-status-badge]][travisci-build-status]
 [![codecov.io][codecov-coverage-badge]][codecov-coverage] [![Backers on Open Collective][opencollective-backers-badge]](#backers) [![Sponsors on Open Collective][opencollective-sponsors-badge]](#sponsors)
 
-Streamlink is a CLI utility that pipes flash videos from online streaming services to a variety of video players such as VLC.
+Streamlink is a CLI utility that pipes flash videos from online streaming services to a variety of video players such as mpv.
 
 The main purpose of streamlink is to convert CPU-heavy flash plugins to a less CPU-intensive format.
 
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -13,7 +13,7 @@ and if you're on Linux or BSD you probably already know the drill.
 
 The way Streamlink works is that it's only a means to extract and transport
 the streams, and the playback is done by an external video player. Streamlink
-works best with `VLC`_ or `mpv`_, which are also cross-platform, but other players
+works best with `mpv`_ or `VLC`_, which are also cross-platform, but other players
 may be compatible too, see the :ref:`Players` page for a complete overview.
 
 Now to get into actually using Streamlink, let's say you want to watch the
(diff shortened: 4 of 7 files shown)


karlo2105:
Since VideoLan team released VLC 3.0.x branch, new features were added, many issues were fixed so far I didn't notice any major issues with this player. VLC is daily fixed through nightly releases.

bastimeyer:
I agree with your points and the overall idea, @gravyboat, but there are some things we should think about before changing the default player.

- MPV doesn't have official builds for Windows and users have to rely on 3rd party builds. In comparison to VLC, this will sound *untrustworthy* to many users who expect something that's *officially supported*.
- MPV is mostly a CLI and config file oriented player and doesn't have a *real* GUI. This does and will scare off many people, especially users of Windows and probably most users of macOS. In my eyes, MPV is the better player by far, but the default player should be one which won't cause UX problems for the average user.
- I've heard that MPV is about to get a default GUI, so maybe we should wait for that to happen first. There are also a couple of MPV forks which already have a GUI, like [`gnome-mpv`](https://github.com/gnome-mpv/gnome-mpv) on Linux or [`IINA`](https://github.com/lhc70000/iina) on macOS. Not sure though how useful it would be to add examples to the config for these as well.
- Yes, this should require a major version bump of Streamlink.
- Regarding app paths on macOS, the homebrew cask and all of the other listed binary distributions are using a lowercased app name:
  https://github.com/Homebrew/homebrew-cask/blob/893a1462b37950f969d3bdbf56807bd817e8ca48/Casks/mpv.rb#L14
- Regarding Windows, since there are no official binaries and no official installer and instead just simple 7zip archives, there is no way to reliably guess a default path. Users will probably be required to set custom configs, which is kind of a nuisance.
  This is how the Twitch GUI defines its default MPV paths:
  https://github.com/streamlink/streamlink-twitch-gui/blob/ed91353cbf9f41129036ec3a9762ed244b4b9bd6/src/config/players.json#L60-L80
  Maybe the Streamlink Windows installer could be improved by prompting for the player path which will then be added to the config...

gravyboat:
@karlo2105 On initial release for 3.x there were a fair number of issues people attributed to Streamlink that were actually caused by VLC. While I understand bugs (we have a fair number of them on some releases too) it generates a lot of work to review these issues and have users confirm whether it's really a player issue (especially when different users have different 3.x releases like recently) or an issue with Streamlink. This happened on the last big release as well. I also just think that VLC is an inferior player. A lot of issues occur with it even on stable releases for absolutely no reason that users encounter which never happen with mpv from audio/video syncing issues, to video randomly closing or not playing all the way through. 

@bastimeyer These are all good points and I agree with you across the board, especially about mpv not having official builds and that being a problem (this is probably my biggest issue because then we could technically be introducing our users to a malicious vector we would be blamed for in the event of compromise). Your point about mpv not having a real GUI is a good one. I rarely use any of the GUI options when I do use VLC and exclusively launch streams via the CLI so it doesn't come up for me much but it is something that needs to be considered as a big issue and will look like a major downgrade to some users when VLC has been working fine for them and they aren't encountering the issues I noted above.

Do we know what the timeline is for their GUI? Or have they even ever discussed having official builds? I guess we could always make a PR and propose automated release builds like we use that have actual installers (https://github.com/mpv-player/mpv/releases are just archives) but that would be a lot of work and we might potentially annoy people who maintain those releases. We could start doing our own release, but that seems like it's just stretching ourselves even thinner on this project and I'm not really interested in that.

I like the idea of the installer requesting path information and adding it to the config file as a potential option. I wish we had usage stats for the percentage of people using the Streamlink GUI versus people using Streamlink past the download statistics. That would help a lot in determining what sort of impact this change might have in the daily usage of users but since we don't have that we'll have to discuss other options. I appreciate the feedback and points of discussion. They are all extremely valid and impactful.

back-to:
> VLC has been pretty bad the past few releases for compatibility (generating a lot of issues) 

**normal users don't know what mpv is,**
this change will probably generate more issues
as the user is now required to change the config file to VLC

johnthecracker:
I would go for player that is the best for delivering streams.

If player doesn't work for streams, what do you do with menus, buttons, etc.

Here is about audio equalizer with MPV if somebody wants that:
https://github.com/mpv-player/mpv/issues/6210

win32/streamlinkrc
```diff
@@ -1,15 +1,20 @@
...
 
 # Important: You must use a quoted path if there are spaces in the path. This
 # is because the player command is parsed like a shell command to allow
 # parameters to be passed to the player.
 
-# Here is a few examples of players:
+# Here are a few examples of players:
+
+# mpv
+
```
back-to: the other players don't have an empty line here

src/streamlink_cli/utils/player.py
```diff
@@ -15,23 +15,23 @@ def check_paths(exes, paths):
...
-            subpath = "VideoLAN\\VLC\\"
+            subpath = "mpv-x86_64\\"
             envvars = ("PROGRAMFILES", "PROGRAMFILES(X86)", "PROGRAMW6432")
             paths = filter(None, (os.environ.get(var) for var in envvars))
             paths = (os.path.join(p, subpath) for p in paths)
             path = check_paths((exename,), paths)
     else:
         paths = os.environ.get("PATH", "").split(":")
-        path = check_paths(("vlc",), paths)
+        path = check_paths(("mpv",), paths)
```
back-to: maybe this could be used for vlc and mpv,
so the behavior for a default vlc don't change.

example for _linux_

```py
path = check_paths(("vlc",), paths)
if not path:
    path = check_paths(("mpv",), paths)
```

johnthecracker:
Here is the same youtube stream played with streamlink + MPV, VLC only and MPV only.
I removed VLC and don't use it, screenshots are for VLC and MPV comparison.

```
C:\>youtube-dl.exe --user-agent "Mozilla/5.0 (Linux; Android 8.1.0; Pixel) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.98 Mobile Safari/537.36" --verbose --list-formats --no-check-certificate "https://www.youtube.com/watch?v=fO8x9MZ8m9g"
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['--user-agent', 'Mozilla/5.0 (Linux; Android 8.1.0; Pixel) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.98 Mobile Safari/537.36', '--verbose', '--list-formats', '--no-check-certificate', 'https://www.youtube.com/watch?v=fO8x9MZ8m9g']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2018.12.17
[debug] Python version 3.4.4 (CPython) - Windows-10-10.0.17763
[debug] exe versions: ffmpeg N-92801-g7efe84aebd, ffprobe N-92801-g7efe84aebd, rtmpdump 2.4
[debug] Proxy map: {}
[youtube] fO8x9MZ8m9g: Downloading webpage
[youtube] fO8x9MZ8m9g: Downloading video info webpage
[youtube] fO8x9MZ8m9g: Downloading embed webpage
WARNING: unable to extract uploader nickname
WARNING: unable to extract channel id; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
[info] Available formats for fO8x9MZ8m9g:
format code  extension  resolution note
249          webm       audio only DASH audio   58k , opus @ 50k, 28.26MiB
250          webm       audio only DASH audio   70k , opus @ 70k, 34.95MiB
171          webm       audio only DASH audio  117k , vorbis@128k, 56.30MiB
251          webm       audio only DASH audio  131k , opus @160k, 63.69MiB
140          m4a        audio only DASH audio  134k , m4a_dash container, mp4a.40.2@128k, 77.24MiB
160          mp4        256x144    144p  101k , avc1.4d400c, 30fps, video only, 30.75MiB
278          webm       256x144    144p  103k , webm container, vp9, 30fps, video only, 51.18MiB
133          mp4        426x240    240p  182k , avc1.4d4015, 30fps, video only, 55.22MiB
242          webm       426x240    240p  225k , vp9, 30fps, video only, 93.18MiB
134          mp4        640x360    360p  406k , avc1.4d401e, 30fps, video only, 124.02MiB
243          webm       640x360    360p  410k , vp9, 30fps, video only, 157.39MiB
244          webm       854x480    480p  757k , vp9, 30fps, video only, 267.12MiB
135          mp4        854x480    480p  792k , avc1.4d401f, 30fps, video only, 233.25MiB
17           3gp        176x144    small , mp4v.20.3, mp4a.40.2@ 24k, 50.15MiB
36           3gp        320x180    small , mp4v.20.3, mp4a.40.2, 151.27MiB
18           mp4        640x360    medium , avc1.42001E, mp4a.40.2@ 96k, 303.11MiB
43           webm       640x360    medium , vp8.0, vorbis@128k, 383.56MiB (best)

```

<a href='https://postimg.cc/0bBjhPnV' target='_blank'>Screen-Shot</a>

<a href='https://postimg.cc/V59H7Nfc' target='_blank'>Screen-Shot2</a>

<a href='https://postimg.cc/JDKyYJtF' target='_blank'>Screen-Shot3</a>

Voultapher:
Personally I've switched some time ago to mpv, as it has been both more stable and given me drastically better performance, ~10% cpu usage for mpv vs ~30% cpu usage for vlc when streaming 1080p60 with hardware decoding enabled for both players on an i75500U.

Seeing the discussion it seems there are 2 main audiences for streamlink with different needs. The 'linux' users that mostly use the command line, install mpv via their package manager and cherish its advanced features. On the other side there are windows users that install software by downloading things from their browsers, are accustomed to vlc specifically and gui use in general.

How about instead of migrating the whole user base at once, we migrate only those whose needs match more closely with what mpv has to offer, eg 'linux' users. Obviously that does not solve problems introduced by vlc, as a big portion of the user base will still be using that. However this gradual switch could allow us to gain more experience with mpv as a default player and could already address the needs of the first group more closely than has been done so far.

bastimeyer:
https://github.com/mpv-player/mpv/commit/a20ae0417f2d1e1a2c173f5eaf66a81974df0008

> Note: "Direct" Windows support is deprecated, pending complete GL/Vulkan support
> by Microsoft's WSL 2. This will enable running Linux binaries under win32,
> and, owing to Microsoft's competence, it will be *faster* than native.

This makes the decision of switching to MPV as default player even more complicated.

Just in case it's unclear what is meant by this:
https://www.phoronix.com/scan.php?page=news_item&px=Linux-GUI-Apps-GPU-WSL2

src/streamlink_cli/utils/player.py
beardypig: I think having a list of possible players by default would be advantageous to the user. If they have a specific player then they set it in the config, same as it ever was. 

I would say that prioritising mpv would be a good idea too, but that should probably done in a major version bump.

bastimeyer:
> MPV Devs Consider Blocking MPV From Running On Gnome

https://old.reddit.com/r/linux/comments/hnoksv/mpv_devs_consider_blocking_mpv_from_running_on/?sort=top

> mpv is not anymore supporting gnome. and the owner reverted the commit again shortly after and then again made a new one, to add the changes

https://old.reddit.com/r/gnome/comments/hn1s3r/mpv_is_not_anymore_supporting_gnome_and_the_owner/?sort=top

----

This looks like MPV development is in a very weird state right now and may get forked by other people. See the linked commits and their crude commit messages and also the linked issue/pr threads, as I don't want to post them here.

gravyboat:
@bastimeyer That maintainer has made similarly inflammatory comments before and it's pretty normal. I agree the decisions that have been made recently do make it harder. This has been open for over a year, let's just close it. Can always re-open if we want.

PR #2147, opened 2018-10-31
Update to the NPO plugin

vincentrozenberg:
Added support for the new standard domain "npostart.nl".


gravyboat:
Thanks @vincentrozenberg!

PR #2152, opened 2018-11-04
 Fix recording added in #920

qkolj:
Fix recording added in #920 by @grazzolini

 - Remove `-r` as short option for `--rtmp-rtmpdump`

 - Add options `--record` (`-r`) and `--record-and-pipe` (`-R`) as suggested by @bastimeyer in discussion on #920

 - Add info message that the download started as suggested by @back-to in discussion on #920

 - Add tests for `streamlink_cli.main.create_output` where these arguments are used

---

Since original author abandoned his PR, and I think this is a really useful feature, I have gone ahead and changed their code to implement suggested changes and add necessary tests.


beardypig:
As this is a breaking change (changing the function of `-r`) it would require a major release.

grazzolini:
@karlo2105, this assumes using VLC. I have written this option directly into streamlink for several reasons, including being player agnostic. But also because when doing the recording through a pipe, I had run through some buffering issues were the recording would be ok, but the playback would suffer. I have been patching my local streamlink installation for some time now to add this functionality and I would continue to do so. But it would be good to have this added to streamlink directly.

src/streamlink_cli/argparser.py
```diff
@@ -500,6 +500,24 @@ def build_parser():
...
+        Write stream data to FILENAME while also playing it.
+
+        You will be prompted if the file already exists.
+        """
+    )
+    output.add_argument(
+        "-R", "--record-and-pipe",
+        metavar="FILENAME",
+        help="""
+        Write stream data to FILENAME while also playing it.
```
back-to: same description as `-r`
both should have more details.

qkolj:
Thanks for the review! I fixed the duplicate help string, but I am out of ideas how to expand them to include more details. If you, or anyone else has a better idea what to write there, please feel free to suggest.

src/streamlink_cli/main.py
```diff
@@ -327,6 +340,7 @@ def read_stream(stream, output, prebuffer, chunk_size=8192):
     is_http = isinstance(output, HTTPServer)
     is_fifo = is_player and output.namedpipe
     show_progress = isinstance(output, FileOutput) and output.fd is not stdout and sys.stdout.isatty()
+    show_record_progress = isinstance(output.record, FileOutput) and output.fd is not stdout and sys.stdout.isatty()
```
back-to: > AttributeError: 'PlayerOutput' object has no attribute 'fd'

should be `output.record.fd`

qkolj:
This broken pipe error:

```
error: Error when writing to output: [Errno 32] Broken pipe, exiting
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>
BrokenPipeError: [Errno 32] Broken pipe
```

can be reproduced even with the existing `-O` command. For example:

```
$ streamlink -O https://www.twitch.tv/yogscast 360p | mpv -
Playing: -
[file] Reading from stdin...
 (+) Video --vid=1 (h264 640x360)
 (+) Audio --aid=1 (aac 2ch 48000Hz)
AO: [pulse] 48000Hz stereo 2ch float
VO: [gpu] 640x360 yuv420p
AV: 00:00:00 / 00:00:11 (2%) A-V: -0.004 DS: 2.083/0 Cache: 11s+1MB


Exiting... (Quit)
error: Error when writing to output: [Errno 32] Broken pipe, exiting
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>
BrokenPipeError: [Errno 32] Broken pipe
```

I am not sure what's the best way to handle this. Closing `stderr` on `SIGPIPE` would certainly prevent the error message from showing up, but I am not sure if that would prevent some other useful output.

back-to:
> can be reproduced even with the existing -O

as this is the same behavior, I think it can be ignored for now.

---

also there is a `AttributeError`, see my other comment

src/streamlink_cli/main.py
qkolj:
```suggestion
    show_record_progress = isinstance(output.record, FileOutput) and output.record.fd is not stdout and sys.stdout.isatty()
```

gravyboat:
Since we're doing a major release I'm going to merge this in.

bastimeyer:
This merge has caused issues with the `--player-continuous-http` and `--player-external-http` parameters.

@back-to, since you reviewed the changes of this PR, could you please see if there are more regressions other than #2234 when you get the time? Thanks!

PR #2160, opened 2018-11-09
plugins.skai: Fix plugin after site update

### LINKED ISSUE #2159: Skai plugin broken
## Plugin Issue

### Description

Skai plugin is broken since yesterday, but actually it is no longer needed because they provide a lot more stable stream (they don't change stream three or so times a day). **Imho it can be removed.**

New live url as follows:

http://www.skaitv.gr/live

qkolj:
Closes #2159

 - Updated URL regex to support new valid URL

 - Plugin is now extracting stream URL from an API call instead of searching the HTML document

 - Added a test for the new URL

### DIFF
--- a/src/streamlink/plugins/skai.py
+++ b/src/streamlink/plugins/skai.py
@@ -3,20 +3,15 @@ import re
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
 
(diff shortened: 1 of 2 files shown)


gravyboat:
Looks good, thanks @qkolj.

Issue #1123, opened 2017-07-27
New Streamlink logo


bastimeyer:
We had this discussion before in November 2016 (#179), but now since we're about to release `1.0.0` (#1115), we need a non-temporary and professional logo that fits Streamlink.

This is the current, temporary one:
![current Streamlink logo](https://avatars0.githubusercontent.com/u/22243823?v=4&s=200)
https://github.com/streamlink/streamlink/blame/0.7.0/docs/index.rst#L17
https://github.com/streamlink/streamlink/blob/0.7.0/NOTICE#L51-L102 

Suggestions and examples are welcome, but please only post stuff that is reasonable and doesn't violate any licenses.

If you are a graphics designer with a little bit of experience who wants to help out, please let us know.

Thanks a lot!

gravyboat:
It's worth noting we should definitely pay whoever we get to do this art. We have some money available via Bountysource we can cash out to whoever does the work.

bastimeyer:
@gravyboat 
I've just asked a friend who's a graphics designer and who I have worked with for several years. He would be interested in creating a new logo, but this depends on the time frame and payment. He asked me to get back to him once I know how much we are willing to spend and that the payment depends on the time frame. If we don't find anybody else, he'd be willing to help us with a new logo.

gravyboat:
Sounds good, I'll approach my friend only if we really need him, he's super busy so I don't know if he realistically has time to take on this kind of project. We've got ~$70 USD to cash out from bountysource after fees and everything, which is definitely low for logo design.

cdrage:
i like doggo

j/k

Yeah, a new logo would be better to reflect the project and show our professionalism, of course.. logo's are a hell-of-a-lot for design. If someone is willing to donate their time and help-out that'd be great! :100:

c-r-r1983:
font and image are free for commercial use

http://i.imgur.com/Sf92y9x.jpg
http://i.imgur.com/VYL83pF.jpg

its a simple design and not too busy. Colours and font can be changed to suit whatever style you want

i personally like the 2nd logo the best

gravyboat:
I've responded with this tweet for a potential lead for logo design, with an offer to pay of course: https://twitter.com/failvarez/status/900538321910210561

jimaek:
Hey, if would be very helpful if you wrote here the requirements for your logo. What style you want, text, colors, examples of logos you like...

gravyboat:
[...]

PR #2165, opened 2018-11-14
Add icon, modify installer, update docs

gravyboat:
Fixes https://github.com/streamlink/streamlink/issues/1123


script/makeinstaller.sh
```diff
@@ -43,14 +43,23 @@ do
     touch "${build_dir_plugins}/$i.py"
 done
 
+echo "Creating images"
```
beardypig:
```suggestion
echo "Creating images" 1>&2
```

Send the output to `stderr`, better for logging :)

script/makeinstaller.sh
```diff
@@ -43,14 +43,23 @@ do
...
+# Create all of our various images
+for size in 16 32 48 256; do
+  inkscape --no-gui --export-png="icon-${size}.png" -w ${size} -h ${size} icon.svg
```
beardypig: This will require `inkscape` to be install on the TravisCI machine doing the build... :(

script/makeinstaller.sh
```diff
@@ -43,14 +43,23 @@ do
...
+  optipng -o7 -out "optimized-${size}.png" "icon-${size}.png"
```
beardypig: Ditto for `optipng`.

script/makeinstaller.sh
beardypig: I would actually replace the whole loop with 

```
convert -density 256x256 -background transparent icon.svg -define icon:auto-resize -colors 256 icon.ico
```

ImageMagick can do it all. Using ImageMagick 7.0.8, the ico looks decent and you get all the sizes generated for you. The `optipng` stage does nothing in the end as the `ico` file ends up the same size either way :)

bastimeyer:
@gravyboat could you please add the necessary TravisCI package dependencies here
https://github.com/streamlink/streamlink/blob/30043408c74ddbb152e694b5afe1f185e6664b26/.travis.yml#L55

That's the available package whitelist for Ubuntu trusty:
https://raw.githubusercontent.com/travis-ci/apt-package-safelist/master/ubuntu-trusty

package names are
```
imagemagick
optipng
```

beardypig:
Incase my inline comment got lost:
> ImageMagick can do it all. Using ImageMagick 7.0.8, the ico looks decent and you get all the sizes generated for you. The optipng stage does nothing in the end as the ico file ends up the same size either way :)
```
convert -density 256x256 -background transparent icon.svg -define icon:auto-resize -colors 256 icon.ico
```

gravyboat:
I'll try to get around to this over the weekend, been a bit busy lately unfortunately.

script/makeinstaller.sh
gravyboat: My main issue with imagemagick is that it doesn't (unless they fixed it recently) do a very good job with svg -> png conversion. We still want the `optipng` stage as other teams (https://github.com/streamlink/streamlink-chocolatey/issues/19) either might need it. I've pushed a new commit, let me know what you think if travis passes on it.

gravyboat:
Hmm, seems like there is an issue with the `ico` conversion: `convert.im6: width or height exceeds limit `icon.ico' @ error/icon.c/WriteICONImage/833.` Looks like this is due to `icon:auto-resize` requiring a 256x256 input image.

gravyboat:
I messed with the image conversion but it wasn't working properly so I've gone back to the simplified convert and mv. If you want to mess with it some more feel free @beardypig, but I'm not even sure if we want to use imagemagick for the svg -> ico conversion as it seems like people have had pretty mixed results. If anyone else had feedback let me know, otherwise let's get this merged in so we can do a 1.0 release. This is the last hurdle to do so I believe.

gravyboat:
@beardypig The `optipng` command is further compressing the icon by between 6%-28% https://travis-ci.org/streamlink/streamlink/jobs/467841224#L1742 depending on the original png size. It's not a huge gain, but the smaller we can keep the files the better. If you want me to remove it however I can. If not let's get this approved so we can move along the 1.0 release :)

beardypig:
@gravyboat sure it compresses the pngs, but then the pngs are turned in to an ico which (in my testing) ended up the same size wether the input pngs were compressed or not :)

gravyboat:
@beardypig Fair enough! I've removed it and updated the ico creation command.

.travis.yml
```diff
@@ -53,6 +53,9 @@ addons:
   apt:
     packages:
     - nsis
+    - imagemagick
+    - optipng
```
beardypig:
```suggestion

```

No need to install it now :)

gravyboat: Bah this is what I get for trying to crunch this in before going somewhere! Fixed.

beardypig:
LGTM.

back-to:
the old file `doggo.ico` and its references could be removed at the same time.

https://github.com/streamlink/streamlink/blob/87a55fe4a026c5673f4135ddb6a9f03cd9a1c5b4/win32/doggo.ico
https://github.com/streamlink/streamlink/blob/87a55fe4a026c5673f4135ddb6a9f03cd9a1c5b4/NOTICE#L2
https://github.com/streamlink/streamlink/blob/87a55fe4a026c5673f4135ddb6a9f03cd9a1c5b4/NOTICE#L51-L102
https://github.com/streamlink/streamlink/blob/87a55fe4a026c5673f4135ddb6a9f03cd9a1c5b4/win32/LICENSE.txt#L2
https://github.com/streamlink/streamlink/blob/87a55fe4a026c5673f4135ddb6a9f03cd9a1c5b4/win32/LICENSE.txt#L666-L717

gravyboat:
@back-to Good catch, removed the mentions and the ico itself. @beardypig If there's nothing else let's merge this so we can proceed on to 1.0. We really need to do a new release.

bastimeyer:
Can you please move the generated files into the `build/files/icons` directory instead of just writing them into `$PWD`? Otherwise it'll pollute the repo's working tree when building the installer locally. The `build` directory is already being ignored by `.gitignore` and `build/files` already contains other files for building the installer.

```bash
$ ./scripts/makeinstaller.sh 1>&2 >/dev/null
$ git status
HEAD detached at refs/pull/upstream/2165
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	icon-16.png
	icon-256.png
	icon-32.png
	icon-48.png
	win32/icon.ico
```

As a suggestion:
```diff
diff --git a/script/makeinstaller.sh b/script/makeinstaller.sh
index b1610069..712195c4 100755
--- a/script/makeinstaller.sh
+++ b/script/makeinstaller.sh
@@ -18,9 +18,10 @@ build_dir="$(pwd)/build"
 build_dir_plugins="${build_dir}/lib/streamlink/plugins"
 nsis_dir="${build_dir}/nsis"
 files_dir="${build_dir}/files"
+icons_dir="${files_dir}/icons"
 # get the dist directory from an environment variable, but default to the build/nsis directory
 dist_dir="${STREAMLINK_INSTALLER_DIST_DIR:-$nsis_dir}"
-mkdir -p "${build_dir}" "${dist_dir}" "${nsis_dir}" "${files_dir}"
+mkdir -p "${build_dir}" "${dist_dir}" "${nsis_dir}" "${files_dir}" "${icons_dir}"
 
 echo "Building streamlink-${STREAMLINK_VERSION} package..." 1>&2
 python setup.py build 1>&2
@@ -46,10 +47,9 @@ done
 echo "Creating images" 1>&2
 # Create images
 for size in 16 32 48 256; do
-  inkscape --without-gui --export-png="icon-${size}.png" -w ${size} -h ${size} icon.svg
+  inkscape --without-gui --export-png="${icons_dir}/icon-${size}.png" -w ${size} -h ${size} icon.svg
 done
-convert icon-{16,32,48,256}.png icon.ico
-mv icon.ico ./win32/icon.ico
+convert "${icons_dir}"/icon-{16,32,48,256}.png "${icons_dir}/icon.ico"
 
 
 echo "Building ${STREAMLINK_INSTALLER} installer..." 1>&2
@@ -59,7 +59,7 @@ cat > "${build_dir}/streamlink.cfg" <<EOF
 name=Streamlink
 version=${STREAMLINK_VERSION}
 entry_point=streamlink_cli.main:main
-icon=../win32/icon.ico
+icon=${icons_dir}/icon.ico
 
 [Python]
 version=3.6.6
```

gravyboat:
@bastimeyer Good suggestion, should be good to merge now.

.travis.yml
```diff
@@ -53,6 +53,8 @@ addons:
   apt:
     packages:
     - nsis
+    - imagemagick
```
back-to: I don't think `imagemagick` is used anymore and it could be removed here
https://github.com/streamlink/streamlink/pull/2165/commits/c0b0ac68bfe23f86d34dab1fca3406e5ac228894

.travis.yml
```diff
@@ -53,6 +53,8 @@ addons:
...
+    - inkscape
```
back-to: a test if `inkscape` is installed should be added in `script/makeinstaller.sh`

https://github.com/streamlink/streamlink/blob/2c6e4924b4fc1a567dbdffdfb2df71f6aa7add66/script/makeinstaller.sh#L6-L7


```sh
command -v inkscape > /dev/null 2>&1 || { echo >&2 "inkscape is required to build the installer. Aborting."; exit 1; }
```

gravyboat: Done, I also added one for imagemagick. Good idea.

gravyboat:
Any other comments? Otherwise let's get this merged so we can do a 1.0 release!

gravyboat:
@bastimeyer Looks like everyone has approved it, can you merge (with a squash from the interface is fine)? I don't like to merge my own PRs.

bastimeyer:
I usually don't like using the Github UI for merging PRs, because it uses a different committer name... I tried to merge it locally with the correct commit author applied and it was all fine, but it didn't accept it when I tried to push it because of the CI config of the repo and the protected master branch. Whatever...

Issue #1106, opened 2017-07-19
Plugin request: SBS  Play


skilleek:
### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

A plugin to catch streams from: http://play.sbs.co.kr/onair/pc/index.html

As always, many thanks for all your help!

johnthecracker:
This is tall order.

All of their streams come from original source, this is like writting 50 plugins and putting them all into one.

It would be handy plugin.

skilleek:
The problem of your url is that it expires quickly. At the moment, I would need only a plugin for SBS Power-FM-Audio.

From what I have seen, the token is generated using JWT (json web token). I could not figure out how to automaticaly renew the url stream after it expires.

Hope we will find a solution

johnthecracker:
There is another problem.

With that url I get sound only.

skilleek:
@johnthecracker  that's completely fine, since the online stream is only audio

johnthecracker:
I find rtmp stream on your new site http://play.sbs.co.kr/onair/pc/index.html

Here is how you can play this with VLC. 

I get bad broadcast in Europe, maybe because resolution is 1920 x 1800.

`rtmpdump -v -r "rtmp://live.sbs.co.kr/sbsch6/_definst_/sbsch60.stream" ^
-a "sbsch6/_definst_/sbsch60.stream?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjE1MDA0NzUxNjMsInBhdGgiOiIvc2JzY2g2L19kZWZpbnN0Xy9zYnNjaDYwLnN0cmVhbSIsImR1cmF0aW9uIjoxODB9.obT2gHOmx8N5i2u-UVXuT7ZsGglBxnzsFnLrpMHhao0" ^
-f "WIN 26,0,0,137" ^
-W "http://play.sbs.co.kr/onair/pc/img/SBSOnAirPlayer.swf" ^
-p "http://play.sbs.co.kr/onair/pc/index.html" ^
-y "sbsch60.stream" | vlc.exe -
`

Pzx5073:
Got some old data for SBS, from a old proyect I started a few months ago and I never finished

Example code:

[...]

PR #2187, opened 2018-12-01
plugins.sbscokr: New Plugin for http://play.sbs.co.kr/onair/pc/index.html

back-to:
SBS is a national South Korean television and radio network company.
- https://en.wikipedia.org/wiki/Seoul_Broadcasting_System

closes https://github.com/streamlink/streamlink/issues/1106


gravyboat:
Thanks @back-to.

Issue #1536, opened 2018-03-08
[Plugin Request]  http://vtvgo.vn/xem-truc-tuyen-kenh-vtv3-3.html 


314pi:
### Checklist

- [ ] This is a bug report.
- [ ] This is a feature request.
- [x] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description
This is the official website of Vietnam Television

### Expected / Actual behavior

...

### Reproduction steps / Explicit stream URLs to test

1. ...
2. ...
3. ...

### Logs

_Logs are always required for a bug report, use `-l debug` [(help)](https://streamlink.github.io/cli.html#cmdoption-l)
Make sure to **remove username and password**
You can upload your logs to https://gist.github.com/ or_

```
REPLACE THIS TEXT WITH YOUR LOG
```

### Comments, screenshots, etc.

...

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

PR #2188, opened 2018-12-01
plugins.vtvgo: New plugin for https://vtvgo.vn/

back-to:
VTV is the national television broadcaster of Vietnam.
https://en.wikipedia.org/wiki/Vietnam_Television

closes https://github.com/streamlink/streamlink/issues/1536


gravyboat:
Thanks @back-to.

Issue #2038, opened 2018-08-28
Plugin STV


chrisarnott86:
## Plugin Request

- [x] This is a plugin request and I have read the contribution guidelines and plugin request requirements.

### Description
STV is the Scottish version of ITV in the UK. ITV streams do not allow viewers in Scotland to stream so must use STV. A user login is required.

### Example stream URLs

1. https://player.stv.tv/live

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

PR #2189, opened 2018-12-01
plugins.stv: New Plugin for https://player.stv.tv/live

back-to:
STV is a television channel serving Scotland.
- https://en.wikipedia.org/wiki/STV_(TV_channel)

closes https://github.com/streamlink/streamlink/issues/2038


gravyboat:
Thanks @back-to.

PR #2194, opened 2018-12-12
api.useragents: update User-Agent list

back-to:
use some more common / new User Agents


gravyboat:
Good stuff, thanks @back-to.

Issue #2203, opened 2018-12-19
Afreecatv not working again?


tenminute:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

Seems to have stopped working some odd hours ago. It was working this morning but I've been getting an error the past hour. Maybe they're redoing stuff again?

### Reproduction steps / Explicit stream URLs to test
http://play.afreeca.com/user best

### Log output

```
C:\Users\User>streamlink http://play.afreeca.com/user best
[cli][info] Found matching plugin afreeca for URL http://play.afreeca.com/user
[cli][info] Available streams: sd (worst), hd, original (best)
[cli][info] Opening stream: original (hls)
[cli][error] Try 1/1: Could not open stream <HLSStream('http://live-hls-local-cf.afreecatv.com/livestream-east-04/auth_playlist.m3u8?aid=.A32.7bbT56vyHM9fKZk.SwKSrXv6ItOFi5pR1fJb8AIpawXGUhzWZWS0TZhvegqQUewAVNAFbTZYH0uIKZaRe9KCWUo--B9Cu7aZnQAyN1OqyPT4DTB362QBwR6wrVQ')> (Could not open stream: Unable to open URL: http://live-hls-local-cf.afreecatv.com/livestream-east-04/auth_playlist.m3u8?aid=.A32.7bbT56vyHM9fKZk.SwKSrXv6ItOFi5pR1fJb8AIpawXGUhzWZWS0TZhvegqQUewAVNAFbTZYH0uIKZaRe9KCWUo--B9Cu7aZnQAyN1OqyPT4DTB362QBwR6wrVQ (403 Client Error: Forbidden for url: http://live-hls-local-cf.afreecatv.com/livestream-east-04/auth_playlist.m3u8?aid=.A32.7bbT56vyHM9fKZk.SwKSrXv6ItOFi5pR1fJb8AIpawXGUhzWZWS0TZhvegqQUewAVNAFbTZYH0uIKZaRe9KCWUo--B9Cu7aZnQAyN1OqyPT4DTB362QBwR6wrVQ&aid=.A32.7bbT56vyHM9fKZk.SwKSrXv6ItOFi5pR1fJb8AIpawXGUhzWZWS0TZhvegqQUewAVNAFbTZYH0uIKZaRe9KCWUo--B9Cu7aZnQAyN1OqyPT4DTB362QBwR6wrVQ))
[...]

PR #2204, opened 2018-12-19
plugins.afreeca: use Referer for every connection

back-to:
closes https://github.com/streamlink/streamlink/issues/2203


gravyboat:
Looks good, thanks @back-to.

PR #2209, opened 2018-12-24
Update tvrby.py

catvlad:

PR #2210, opened 2018-12-24
Update test_tvrby.py

catvlad:


back-to:
use only one PR instead of two,

also this url test changes might not be needed
see #2209

PR #2223, opened 2019-01-01
Fix subprocess argument debug log on Windows

RomanKornev:
Due to #1576, `args` are `string` on Windows but `list` on other OS.

https://github.com/streamlink/streamlink/blob/94a8a7301c45dfdee89dd7d4d14043438dda4b38/src/streamlink_cli/output.py#L188-L194

A later debug call then converts `string` into space-separated characters.
https://github.com/streamlink/streamlink/blob/94a8a7301c45dfdee89dd7d4d14043438dda4b38/src/streamlink_cli/output.py#L209-L210

### DIFF
--- a/src/streamlink_cli/output.py
+++ b/src/streamlink_cli/output.py
@@ -207,7 +207,7 @@ class PlayerOutput(Output):
 
     def _open_call(self):
         args = self._create_arguments()
-        log.debug(u"Calling: {0}".format(subprocess.list2cmdline(args)))
+        log.debug(u"Calling: {0}".format(args))
         subprocess.call(args,
                         stdout=self.stdout,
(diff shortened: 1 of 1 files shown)


beardypig:
Thanks for the fix @RomanKornev. I'd prefer it if the debug messages were no duplicated (so they could be changed more easily), could you format the string depending on the OS then log the message? eg.

```python
if is_win32:
  fargs = args
else:
  fargs = subprocess.list2cmdline(args)

log.debug(u"....".format(fargs))
```

RomanKornev:
Done.

gravyboat:
Thanks @RomanKornev and thanks for the review @beardypig.

PR #2224, opened 2019-01-01
Add window title to PotPlayer

RomanKornev:
After #1576 we can add window title support to other players.

Due to limitations of PotPlayer, titles can only be added to URL streams and there is no separate title argument.
PotPlayer command line documentation in `PotPlayer - About - Command Line`:
- You can specify titles for URLs by separating them with a backslash (\\) at the end of URLs. ("http://...\title of this url")

Because of this, the title needs to be appended at the end of the filename so args need to be formatted after player check.

Here's an example of it working on Windows 10:
potplayer-title

### DIFF
--- a/src/streamlink_cli/constants.py
+++ b/src/streamlink_cli/constants.py
@@ -13,7 +13,8 @@ DEFAULT_STREAM_METADATA = {
 SUPPORTED_PLAYERS = {  # these are the players that streamlink knows how to set the window title for with `--title`. key names are used in help text
     # name: possible binary names (linux/mac and windows)
     "vlc": ["vlc", "vlc.exe"],
-    "mpv": ["mpv", "mpv.exe"]
+    "mpv": ["mpv", "mpv.exe"],
+    "pot": ["potplayer", "potplayermini64.exe", "potplayermini.exe"]
 }
 
 if is_win32:
--- a/src/streamlink_cli/constants.py
+++ b/src/streamlink_cli/constants.py
@@ -14,7 +14,7 @@ SUPPORTED_PLAYERS = {  # these are the players that streamlink knows how to set 
     # name: possible binary names (linux/mac and windows)
     "vlc": ["vlc", "vlc.exe"],
     "mpv": ["mpv", "mpv.exe"],
-    "potplayer": ["potplayermini64.exe", "potplayermini.exe"]
+    "potplayer": ["potplayer", "potplayermini64.exe", "potplayermini.exe"]
 }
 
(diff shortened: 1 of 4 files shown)


RomanKornev:
Added tests

beardypig:
Thanks for the PR @RomanKornev. Does the titles work with named pipes too? Maybe PotPlayer doesn't work with named pipes anyway. Can you test it with a title that has some quote characters? eg. `--title "A \"{title}\""`?

RomanKornev:
> Does the titles work with named pipes too? Maybe PotPlayer doesn't work with named pipes anyway.

It doesn't work with named pipes: https://streamlink.github.io/players.html.
Is `filename` always `"-"` for regular stdin pipes?
In that case, title should remain `Streaming media (stdin)`. Sadly, I haven't found a way to feed a title using stdin pipe to PotPlayer.

> Can you test it with a title that has some quote characters? eg. `--title "A \"{title}\""`?

Upon further examination, PotPlayer just deletes any quotes anyway and doesn't recognize anything after closing quote. So `before \"STREAMTITLE\" after` would be `before STREAMTITLE`.
In that case, it might be better to just explicitly delete double quotes before passing the title.
Then `--title "before \"{title}\" \"{author}\" after"` would be `before STREAMTITLE AUTHOR after`.

src/streamlink_cli/constants.py
```diff
@@ -13,7 +13,8 @@
 SUPPORTED_PLAYERS = {  # these are the players that streamlink knows how to set the window title for with `--title`. key names are used in help text
     # name: possible binary names (linux/mac and windows)
     "vlc": ["vlc", "vlc.exe"],
-    "mpv": ["mpv", "mpv.exe"]
+    "mpv": ["mpv", "mpv.exe"],
+    "pot": ["potplayer", "potplayermini64.exe", "potplayermini.exe"]
```
back-to: is there even a `potplayer` binary? as this player is windows only.

back-to: Potplayer is always used as Potplayer

I can't find any references with just the name `pot`
so it should be named `potplayer` instead of `pot`

RomanKornev: Removed

RomanKornev: Renamed

RomanKornev: Now tests are failing

RomanKornev: Reverted back to previous binary name.

Issue #2225, opened 2019-01-02
Add {url} argument to window --title 


RomanKornev:
## Feature Request

- [x] This is a feature request and I have read the contribution guidelines.

### Description
We have {title}, {author}, {category}, {game}.
https://github.com/streamlink/streamlink/blob/94a8a7301c45dfdee89dd7d4d14043438dda4b38/src/streamlink_cli/argparser.py#L409
If no title format was provided it defaults to `args.url` so it clearly is supported.
https://github.com/streamlink/streamlink/blob/94a8a7301c45dfdee89dd7d4d14043438dda4b38/src/streamlink_cli/main.py#L137-L148
But what if I want a custom title and a url?

### Expected / Actual behavior
`$ streamlink twitch.tv/CHANNEL best --title "{title} - {game} - {url}"`
Actual: `KeyError: 'url'`
Expected: Window title to change to `STREAM_NAME - GAME - twitch.tv/CHANNEL`

### Additional comments, screenshots, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

john2545:
how about this?
streamlink -p vlc -a "--title abc" twitch.tv/CHANNEL best

PR #2232, opened 2019-01-04
Add {url} argument to window --title

RomanKornev:
For #2225 
Adds `{url}` formatting variable to player window `--title`.

Can there be a situation where `args.url` doesn't exist?
If so, maybe I should add a `DEFAULT_STREAM_METADATA["url"]`?
https://github.com/streamlink/streamlink/blob/07387e22d567994f8fe7eaba79638e22cb2be24b/src/streamlink_cli/constants.py#L7-L12

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -465,6 +465,9 @@ def build_parser():
             gaming oriented platforms. "Game being played" is a way to categorize
             the stream, so it doesn't need its own separate handling.
 
+        {{url}}
+            URL of the stream.
+
         Examples:
 
             %(prog)s -p vlc --title "{{title}} -!- {{author}} -!- {{category}} \\$A" <url> [stream]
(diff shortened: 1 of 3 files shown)


beardypig:
`args.url` can be `None`, but never when launching the player, so I think you'd be OK here.

back-to:
`plugin.url` is also available there, instead of `args.url`

RomanKornev:
I haven't seen any problems using `url=args.url`.

beardypig:
Do urls get unshortened, E.g. bitly, when using args.url?

RomanKornev:
No, with `args.url` they do not. With `plugin.url` they do. So I guess `plugin.url` is more robust.

Another thing I noticed when using `plugin.url` is that now every URL has a scheme `http://` prefixed. That might not be the most readable representation. Maybe we should drop the scheme for title using: 
```python
url=''.join(urlsplit(plugin.url)[1:])
```

Issue #2231, opened 2019-01-03
ZDF Mediathek-Plugin: JSON-Validation error


keydon:
## Bug Report

- [X ] This is a bug report and I have read the contribution guidelines.

### Description

ZDF Mediathek Plugin broken.

### Expected / Actual behavior

Trying to play a movie from the ZDF Mediathek.
But it gives a json validation error instead :/

### Reproduction steps / Explicit stream URLs to test

```
streamlink \
  --loglevel=debug \
  --player-http \
  https://www.zdf.de/filme/der-fernsehfilm-der-woche/die-lichtenbergs---zwei-brueder-drei-frauen-und-jede-menge---zoff-100.html
```
Every movie i tried from https://www.zdf.de/filme did not work.

### Log output

```
[cli][info] streamlink is running as root! Be careful!
[cli][debug] OS:         Linux-4.9.125-linuxkit-x86_64-with
[cli][debug] Python:     3.7.2
[cli][debug] Streamlink: 0.14.2
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.54.0)
[cli][info] Found matching plugin zdf_mediathek for URL https://www.zdf.de/filme/der-fernsehfilm-der-woche/die-lichtenbergs---zwei-brueder-drei-frauen-und-jede-menge---zoff-100.html
error: Unable to validate JSON: 
  Unable to validate key 'mainVideoContent': 
  Unable to validate key 'http://zdf.de/rels/target': 
  Key 'http://zdf.de/rels/streams/ptmd' not found in 
{
	'title': 'Die Lichtenbergs - - zwei Br├╝der, drei Frauen und jede Menge Zoff',
	'duration': 5324, 
	'visible': True, 
	'visibleFrom': '2018-12-31T10:00:00.000+01:00',
	'visibleTo': '2019-03-30T23:59:00.000+01:00',
	'profile': 'http://zdf.de/rels/content/content-video-vod-player',
	'self': '/content/documents/zdf/filme/der-fernsehfilm-der-woche/videos/vod-die-lichtenbergs-----zwei-brueder-drei-frauen-und-jede-menge-zoff-100.json?profile=player',
	'canonical': '/content/documents/vod-die-lichtenbergs-----zwei-brueder-drei-frauen-und-jede-menge-zoff-100.json',
	'http://zdf.de/rels/streams/ptmd-template': '/tmd/2/{playerId}/vod/ptmd/mediathek/181231_2015_sendung_fdw'
[...]

PR #2233, opened 2019-01-04
fix zdf mediathek plugin

keydon:
* json schema changed slightly
* new playId constant was introduced
* added required "Accept"-header

Closes streamlink/streamlink#2231

### DIFF
--- a/src/streamlink/plugins/zdf_mediathek.py
+++ b/src/streamlink/plugins/zdf_mediathek.py
@@ -48,7 +48,7 @@ _documents_schema = validate.Schema(
     {
         "mainVideoContent": {
             "http://zdf.de/rels/target": {
-                "http://zdf.de/rels/streams/ptmd": validate.text
+                "http://zdf.de/rels/streams/ptmd-template": validate.text
             },
         },
     }
@@ -128,6 +128,7 @@ class zdf_mediathek(Plugin):
             return
 
         headers = {
+            "Accept": "application/vnd.de.zdf.v1.0+json",
             "Api-Auth": "Bearer {0}".format(zdf_json['apiToken']),
             "Referer": self.url
         }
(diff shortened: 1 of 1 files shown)


skulblakka:
As I was just about the create my own pull request with this fix, I can confirm that this should resolve the issue. 

Only thing is for me it worked with just the changes to the schema and replacing the playerId. No changes to the headers were required.

keydon:
@skulblakka 
hmm seems you are right, I tried it now without the new header and it worked this time.
But I could have sworn I got a 403 previously! o.O

bastimeyer:
The header gets sent by the web browser and is used to select the API version, which defines the structure of the JSON response. Unless they don't shut down old API endpoints when they introduce changes, including such a header is the right way to ensure that a plugin like this stays stable.

And in regards to this, wouldn't it make sense to extract the `playerId` instead of using a static string here? Since they are providing a stable API with the included header, the only dynamic thing I can see that could change is the playerId. You can get it from here:
`https://www.zdf.de/ZDFplayer/latest-v2/skins/zdf/zdf-player.js`
```js
this.ptmd_player_id="ngplayer_2_3"
```
How likely it is for them to change their player code is another question, though.

I'm not merging this PR yet in case such a change is desired. If not, we can merge it later. Thanks!

src/streamlink/plugins/zdf_mediathek.py
```diff
@@ -128,15 +128,16 @@ def _get_streams(self):
...
             "Api-Auth": "Bearer {0}".format(zdf_json['apiToken']),
             "Referer": self.url
         }
 
         res = self.session.http.get(zdf_json['content'], headers=headers)
         document = self.session.http.json(res, schema=_documents_schema)
 
-        stream_request_url = document["mainVideoContent"]["http://zdf.de/rels/target"]["http://zdf.de/rels/streams/ptmd"]
-        stream_request_url = API_URL + stream_request_url
+        stream_request_url = document["mainVideoContent"]["http://zdf.de/rels/target"]["http://zdf.de/rels/streams/ptmd-template"]
```
back-to: Kika has an invalid space at the end,
https://www.zdf.de/sender/kika/kika-live-beitrag-100.html

```
'/tmd/2/{playerId}/live/ptmd/247onAir-15 '
```

streamlink should probably remove it

PR #2234, opened 2019-01-07
Fix crashing on missing output.record when using player-continous-http option

xerodotc:
Add a checking for record attribute in output in read_stream function to prevent crashing
when start with --player-continuous-http or --player-external-http option


xerodotc:
Thanks, I've rebased the commit over the new one now.

gravyboat:
Good stuff, thanks @xerodotc.

PR #2235, opened 2019-01-08
CI: make sure all the dev dependencies are up to date

beardypig:
It seems like some environments ship with an old version of `pytest` that in now incompatible with the latest version of `pytest-cov`. This change should mean that all the packages are at the latest version.


bastimeyer:
Ok, so Travis has changed its env configs. I was already wondering why the other builds were failing. Good catch!

Issue #2214, opened 2018-12-29
trt.py not  No plugin can handle URL


dreamboxco:
## Bug Report
- [x] This is a bug report and I have read the contribution guidelines.

### Description

I cant see TRT Live Streams with Streamlink

### Reproduction steps / Explicit stream URLs to test

1. http://www.trt.net.tr/anasayfa/canli.aspx?y=tv&k=trt1

### Log output

[streamlinksrv][info] OS:            Linux-4.10.6-mips-with-glibc2.0           
[streamlinksrv][info] Python:        2.7.13                                    
[streamlinksrv][info] Streamlink:    0.14.2 / 27.10.2018                       
[streamlinksrv][debug] Options Parser: 0.1.7                                   
[streamlinksrv][info] Requests(2.13.0), Websocket(0.48.0)                      
[streamlinksrv][info] Processing URL: https://www.trt.net.tr/anasayfa/canli.asp
?y=tv&k=trt1                                                                   
[streamlinksrv][error] No plugin can handle URL: https://www.trt.net.tr/anasayf
/canli.aspx?y=tv&k=trt1                                                        
127.0.0.1 - - [30/Dec/2018 00:02:26] "GET /https://www.trt.net.tr/anasayfa/canl
.aspx?y=tv&k=trt1 HTTP/1.1" 200 -                                              
[streamlinksrv][info] Processing URL: https://www.trt.net.tr/anasayfa/canli.asp
?y=tv&k=trt1                                                                   
[streamlinksrv][error] No plugin can handle URL: https://www.trt.net.tr/anasayf
/canli.aspx?y=tv&k=trt1                                                        
127.0.0.1 - - [30/Dec/2018 00:02:26] "GET /https://www.trt.net.tr/anasayfa/canl
.aspx?y=tv&k=trt1 HTTP/1.1" 200 -

PR #2241, opened 2019-01-11
plugins.trt: Added support for https url, fixed logger error

back-to:
Fixed https://github.com/streamlink/streamlink/issues/2214


gravyboat:
Thanks @back-to.

Issue #2216, opened 2018-12-29
A Para Livestream (Wish)


dreamboxco:
## Plugin Request

- [x] This is a plugin request and I have read the contribution guidelines and plugin request requirements.

### Description

It is about the Channel A Para, this sends the livestream on the side of the newspaper Sabah. It is a legal stream.

### Example stream URLs

1.https://www.sabah.com.tr/apara/canli-yayin

### Additional comments, screenshots, etc.

The transmitter belongs to the ATV group, this is via the plugin turkuvaz.py

PR #2242, opened 2019-01-11
plugins.turkuvaz: Added support for channel 'apara'

back-to:
- use https for api urls
- added missing url tests and plugin matrix entry for other domains

closes https://github.com/streamlink/streamlink/issues/2216


gravyboat:
Looks good, thank you.

Issue #2245, opened 2019-01-12
plugins.youtube: 'gaming' subdomain (Could not find a video on this page)


back-to:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

The html code on `gaming` subdomains changed,
the changes in [1746](https://github.com/streamlink/streamlink/pull/1746) are not compatible anymore

### Reproduction steps / Explicit stream URLs to test

- https://gaming.youtube.com/channel/UCq9e_hCv2jvgck8WowW1NXg
- https://gaming.youtube.com/channel/UCq9e_hCv2jvgck8WowW1NXg/live
- https://gaming.youtube.com/user/rocketbeanstv
- https://gaming.youtube.com/user/rocketbeanstv/live

### Log output

```
[cli][debug] OS:         Linux
[cli][debug] Python:     3.6.7
[cli][debug] Streamlink: 0.14.2+197.g42b8edd
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.54.0)
[cli][info] Found matching plugin youtube for URL https://gaming.youtube.com/channel/UCq9e_hCv2jvgck8WowW1NXg/live
error: Could not find a video on this page
```

### Additional comments, screenshots, etc.

**possible solutions** 

a second `_ytdata_re` for the changes on `gaming` subdomains

https://github.com/streamlink/streamlink/blob/42b8edd663e3f3aafecf4f0a39694e1e2d047c0c/src/streamlink/plugins/youtube.py#L100

or **easier** change `gaming` to `www` in `self.url`

PR #2247, opened 2019-01-13
plugins.youtube: replace "gaming" with "www" subdomain

back-to:
closes https://github.com/streamlink/streamlink/issues/2245


gravyboat:
Good stuff, thank you.

PR #2252, opened 2019-01-14
Fix HLS "Could not open stream"

alan7s:
CHANNEL_API_URL changed. Fix this issue [#2203](https://github.com/streamlink/streamlink/issues/2203)


src/streamlink/plugins/afreeca.py
```diff
@@ -5,7 +5,7 @@
 from streamlink.plugin.api import validate
 from streamlink.stream import HLSStream
 
-CHANNEL_API_URL = "http://live.afreecatv.com:8057/afreeca/player_live_api.php"
+CHANNEL_API_URL = "http://live.afreecatv.com/afreeca/player_live_api.php"
```
back-to: the old url still works, are you sure that this is needed?

PR #2274, opened 2019-01-27
Added support for 1und1.tv

SteveOswald:
Added support for the German TV provider 1und1.tv.

### DIFF
--- a/src/streamlink/plugins/zattoo.py
+++ b/src/streamlink/plugins/zattoo.py
@@ -43,7 +43,7 @@ class Zattoo(Plugin):
                 tvonline\.ewe|nettv\.netcologne|tvplus\.m-net
             )\.de
             )|(?:(?:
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/zattoo.py
```diff
@@ -43,7 +43,7 @@ class Zattoo(Plugin):
                 tvonline\.ewe|nettv\.netcologne|tvplus\.m-net
             )\.de
             )|(?:(?:
-                player\.waly|www\.netplus
+                player\.waly|www\.netplus|www\.1und1
```
back-to: reuse www

```regex
player\.waly|www\.(?:1und1|netplus)
```

SteveOswald:
Hi @back-to,

thank you for your suggestions. I've updated the corresponding files.

Best regards,
Steve

PR #2285, opened 2019-02-01
streams.dash: Support manifest strings in addition to manifest urls

pmrowla:
Modifies `parse_manifest()` to accept a `url_or_manifest` parameter.

See also: #2239 

This does not include any of the changes for supporting DASH SegmentBase.

### DIFF
--- a/tests/streams/test_dash.py
+++ b/tests/streams/test_dash.py
@@ -5,7 +5,7 @@ from streamlink.stream import *
 from streamlink.stream.dash import DASHStreamWorker
 from streamlink.stream.dash_manifest import MPD
 from tests.mock import MagicMock, patch, ANY, Mock, call
-from tests.resources import xml
+from tests.resources import text, xml
 
 
 class TestDASHStream(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/streams/test_dash.py
```diff
@@ -208,6 +208,26 @@ def test_parse_manifest_drm(self, mpdClass):
                           self.session, self.test_url)
         mpdClass.assert_called_with(ANY, base_url="http://test.bar", url="http://test.bar/foo.mpd")
 
+    @patch('streamlink.stream.dash.MPD')
```
beardypig: I don't think this test is quite valid, it doesn't test that the correct thing is passed to the MPD class. I think parsing one of the included resources might be a better test.

```python
    def test_parse_manifest_string(self):
        with text("dash/test_9.mpd") as mpd_txt:
            test_manifest = mpd_txt.read()

        streams = DASHStream.parse_manifest(self.session, test_manifest)

        self.assertSequenceEqual(
            sorted(list(streams.keys())),
            sorted(["2500k"])
        )
```

pmrowla: fixed in 6aea0f4

gravyboat:
Thanks @pmrowla, and thanks for the reviews @back-to and @beardypig!

PR #2292, opened 2019-02-01
Add OpenBSD to the installation docs

ibara:
We've had a package of streamlink since the fork.
Thanks!


gravyboat:
Thanks for the addition!

PR #2302, opened 2019-02-04
Fill out matrix with more information

Chazzz:
Since there's an open issue with an explanation for why afreeca vods isn't implemented and other ways to stream/download them, link to it in the documentation.


back-to:
the website https://streamlink.github.io/ is meant for Streamlink,
in my opinion adding notes about not related Software should be avoided.

Chazzz:
Hi @back-to,

I think the additional information provided by the link covers several things:

1. There's already been a request for Streamlink to support this feature
2. Why the feature hasn't been implemented in Streamlink
3. A non-Streamlink way to acquire the files if necessary

In my opinion, points 1 and 2 are very relevant to Streamlink, and point 3 is relevant to the Streamlink community for whom getting it done is more important than using Streamlink to do so. All 3 are beneficial to be included in the Streamlink documentation. That said, emphasis on point 3 is unnecessary; if you would prefer wording along the lines of "Afreeca vod discussion: [1]", it would still be a positive contribution to the Streamlink documentation.

back-to:
see above.

PR #2304, opened 2019-02-05
plugins.crunchyroll: Allow CR's multilingual URLs to be handled

rlko:
Multilingual URLs shoud be allowed to be handled by the crunchyroll plugin.

Using [crunchyroll's rss](https://www.crunchyroll.com/rss), I've seen that the `<link></link>` given is formatted depending your localization, probably IP based. Since I am currently living in France I always get the /fr language path

```python
import feedparser
url = 'https://www.crunchyroll.com/rss'
f = feedparser.parse(url)
print(f['entries'][0]['link'])
```
```
# gives:
http://www.crunchyroll.com/fr/shonen-ashibe-go-go-goma-chan/episode-94-graduating-from-goma-chan-780795
```

### DIFF
--- a/src/streamlink/plugins/crunchyroll.py
+++ b/src/streamlink/plugins/crunchyroll.py
@@ -38,6 +38,9 @@ _url_re = re.compile(r"""
     (?:
         com|de|es|fr|co.jp
     )
+    (?:
+        /(en-gb|es|es-es|pt-pt|pt-br|fr|de|ar|it|ru)
+    )?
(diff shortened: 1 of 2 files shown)


back-to:
Add a `can_handle_url` test

https://github.com/streamlink/streamlink/blob/f69a18fc7a977d72c56b357609043c26c817cf6f/tests/plugins/test_crunchyroll.py#L8-L12

Issue #2281, opened 2019-01-30
[MITELE] New Channels


An0nima:
```
streamlink https://www.mitele.es/directo/mtmad-24h best -l debug
[cli][debug] OS:         Windows 7
[cli][debug] Python:     3.6.6
[cli][debug] Streamlink: 0.14.2+205.g7dbda7b
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.54.0)
error: No plugin can handle URL: https://www.mitele.es/directo/mtmad-24h
```

```
streamlink https://www.mitele.es/directo/gh-duo-24h-senal-1 best -l debug
[cli][debug] OS:         Windows 7
[cli][debug] Python:     3.6.6
[cli][debug] Streamlink: 0.14.2+205.g7dbda7b
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.54.0)
error: No plugin can handle URL: https://www.mitele.es/directo/gh-duo-24h-senal-1
```

```
streamlink https://www.mitele.es/directo/gh-duo-24h-senal-2 best -l debug
[cli][debug] OS:         Windows 7
[cli][debug] Python:     3.6.6
[cli][debug] Streamlink: 0.14.2+205.g7dbda7b
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.54.0)
error: No plugin can handle URL: https://www.mitele.es/directo/gh-duo-24h-senal-2
```

bastimeyer:
This looks like a simple regexp problem:
https://github.com/streamlink/streamlink/blob/42c34ca104f9a1761164dfce6c3ebabea984a823/src/streamlink/plugins/mitele.py#L15

An0nima:
The new channels work, thankssssssssss ;) , you can close the topic now.

PR #2310, opened 2019-02-07
plugins.mitele: plugin update

back-to:
- plugin cleanup: removed hls url build (Ref #1337),
  the api returns a correct url at the moment and this is not needed.
- better error messages
- added support for new channels

closes https://github.com/streamlink/streamlink/issues/2281


gravyboat:
Looks good, thanks @back-to.

PR #2307, opened 2019-02-06
plugins.streamme: Fixed source quality, added title and author

back-to:
```
$ streamlink https://www.stream.me/... source
[cli][info] Found matching plugin streamme for URL https://www.stream.me/...
[cli][info] Available streams: 144p (worst), 368p, 576p, source (best)
[cli][info] Opening stream: source (hls)
```

closes https://github.com/streamlink/streamlink/issues/2254

PR #2313, opened 2019-02-10
plugins.streamme: Fixed null error for 'origin'

back-to:
Ref https://github.com/streamlink/streamlink/pull/2307#issuecomment-462113332

```
error: Unable to validate response text: Unable to validate key 'formats': Unable to validate key 'mp4-hls':
Unable to validate key 'origin': Type of None should be 'dict' but is 'NoneType'
```


gravyboat:
Looks good, thanks @back-to.

PR #2318, opened 2019-02-16
Handle keyboard interrupts in can_handle_url checks

aqxa1:
These checks can take a long time in the case of a typo in the url, for example.
This sets the return code to 130 so that external scripts can handle it
correctly.


beardypig:
I see this was already merged, but shouldn't the exit code have been `-2`? :)

Issue #1979, opened 2018-08-01
itertags has issues on Python 3.7


beardypig:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description

With the tests for `itertags` added in #1675 an issue with cpython 3.7.0+ `finditer` was exposed. There is a possibility for `itertags` to give incorrect results in some cases. This likely _does not_ actually affect any Plugins, however, there is the possibility of Plugins not working as expected on Python 3.7+. 

There is a bug report against cpython to investigate this issue: [bpo-34294](https://bugs.python.org/issue34294). The tests that do not pass on Python 3.7.0 have been marked as xfail. To resolve this issue we xneed to track [bpo-34294](https://bugs.python.org/issue34294) and wait for the bug to be fixed, or not, in Python. If the bug is not fixed, or deemed not a bug, we will need to implement a workaround (which we tried in various ways, see discussion in #1675). 

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

back-to:
@beardypig 

there is also some other solution,

instead of searching for **every tag**, it will only search for the **wanted tag**.

The issue with reusing the `inner` is still there but it would reduce the failure
and the current tests would pass.

```diff
commit 57bedbd6187109067602e3520e98630639b45e24 (HEAD -> master)
Author: back-to <backto@protonmail.ch>
Date:   Wed Aug 8 14:54:17 2018 +0200

    commit

diff --git a/src/streamlink/plugin/api/utils.py b/src/streamlink/plugin/api/utils.py
index f6f3f121..7869e874 100644
--- a/src/streamlink/plugin/api/utils.py
+++ b/src/streamlink/plugin/api/utils.py
@@ -6,10 +6,7 @@ from ...utils import parse_qsd as parse_query, parse_json, parse_xml
 
 __all__ = ["parse_json", "parse_xml", "parse_query"]
 
-
-tag_re = re.compile('''(?=<(?P<tag>[a-zA-Z]+)(?P<attr>.*?)(?P<end>/)?>(?:(?P<inner>.*?)</\s*(?P=tag)\s*>)?)''',
-                    re.MULTILINE | re.DOTALL)
-attr_re = re.compile('''\s*(?P<key>[\w-]+)\s*(?:=\s*(?P<quote>["']?)(?P<value>.*?)(?P=quote)\s*)?''')
[...]

PR #2325, opened 2019-02-25
tests: detect unsupported versions for itertags

back-to:
> 3.7.0 - 3.7.2 and 3.8.0a1

skip only unsupported versions,
test it on future supported updates

---

3.7.3 final: 2019-03-25 (expected)
https://www.python.org/dev/peps/pep-0537/#id3

3.8.0 alpha 2: Sunday, 2019-02-24 (expected)
https://www.python.org/dev/peps/pep-0569/#schedule

---

Ref https://bugs.python.org/issue34294
Fixed https://github.com/streamlink/streamlink/issues/1979

### DIFF
--- a/tests/test_plugin_utils.py
+++ b/tests/test_plugin_utils.py
@@ -5,6 +5,24 @@ import unittest
 from streamlink.plugin.api.utils import itertags
 
 
+def unsupported_versions_1979():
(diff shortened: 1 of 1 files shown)


gravyboat:
Looks good, thanks @back-to.

Issue #2308, opened 2019-02-06
`streamlink-script.py` is not found on windows installer installation.


Dinir:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description

I want to use Streamlink Twitch GUI and I get an error playing a stream there, but it says it didn't find `streamlink-scripts.py` that apparently should be provided by Streamlink so I am submitting this issue here in Streamlink repository.

I installed Streamlink and GUI on Windows 10 x64, and reinstalled them after installing Python 3.7 x64, because I wanted to ensure that it finds Python, although I am aware that Streamlink installer comes with a local python environment.

I also tried using the solution of setting the python script to the exe file (https://github.com/streamlink/streamlink/issues/2287#issuecomment-460010782) as follows:
- Python executable: `C:\Program Files\Python37\pythonw.exe`
- Streamlink python script: `C:\Program Files (x86)\Streamlink\bin\streamlink.exe`

### Expected / Actual behavior

When I was trying to play a stream without the setting for stream not filled up, I get this error:

> **Error: Streamlink misconfiguration**  
> Couldn't find python script  
> Executables were not found

and when I filled up the setting page as I written above, I get this one: 

> **Error: Couldn't launch the stream**  
> Unexpected version check output

### Reproduction steps / Explicit stream URLs to test

1. Install Streamlink with `streamlink-1.0.0.exe` in the release page, for all users.
2. Install Streamlink Twitch GUI with `streamlink-twitch-gui-v1.6.0-win64-installer.exe`, for all users.
3. With or without setting these values in `Settings - Streaming - Streamlink / Livestreamer` in Streamlink Twitch GUI, try watching a stream by clicking any stream that's live.

https://twitch.tv/TwitchDefrags always stream, so this should be a great explicit stream url to test.

### Log output

I can't provide the log output, I can use the CLI one without a problem, but the lack of the py file causes the problem on the GUI.

### Additional comments, screenshots, etc.

my Path as system variable contains
```
C:\Program Files\Python37\Scripts\
C:\Program Files\Python37\
C:\Program Files (x86)\Streamlink\bin
```

bastimeyer:
> Unexpected version check output
> * Python executable: `C:\Program Files\Python37\pythonw.exe`
> * Streamlink python script: `C:\Program Files (x86)\Streamlink\bin\streamlink.exe`

[...]

PR #2326, opened 2019-02-25
Build additional "streamlinkw" launcher on Windows

### LINKED ISSUE #2287: Streamlink no longer provides streamlink-script.py
As of version 0.14.0 streamlink does not install the streamlink-script.py script,
because of this streamlink-twitch-gui does not work as it cannot load the script.
This may an issue in streamlink, as there is no mention of its removal in the changelog.

bastimeyer:
This adds an additional `streamlink-noconsole.exe` "GUI" launcher to the Windows installer and the Streamlink python wheel. This launcher uses `pythonw.exe` under the hood, so that a new terminal window doesn't get opened if executed from a non-terminal context on Windows.

Resolves #2287, #2308 

----

**Motivation of this PR and why this change is needed**

A couple of months ago, both `pip` and `pynsist` changed their entry point launchers on Windows from [`setuptools`](https://github.com/pypa/setuptools/blob/v40.8.0/launcher.c) to [`simple_launcher`](https://bitbucket.org/vinay.sajip/simple_launcher). This change broke the Streamlink resolver in Streamlink Twitch GUI, as these new launchers are lacking the additional `streamlink-script.py` file, which is used for parsing its shebang and getting the path to `pythonw.exe`, which is needed to suppress the terminal window from being opened. Unfortunately, it is impossible to work around the new launchers without parsing them or asking the user to manually set custom paths. That is however not a real solution.

Fortunately though, `simple_launcher` does provide two different versions of the launchers, a "console" one and a "GUI" one. Setuptools does this as well, but their GUI launcher is useless, as its process terminates immediately and therefore can't be used. The new `simple_launcher`-based launchers however work exactly as they should and the GUI launcher can be used by GUI applications like Streamlink Twitch GUI without any issues.

That's why I'm submitting this pull request, which adds `streamlink-noconsole` as a new additional GUI launcher.

I've already submitted a PR to pynist (takluyver/pynsist#179) a couple of days ago which adds a new `console=false` option to the installer's "command" section, so that a GUI launcher can be built. My PR was merged a couple of hours ago, but since a new release hasn't been published yet, it will be installed from the master branch for now (see the TravisCI config changes).

Regarding the Streamlink python wheel, an additional [`gui_scripts` entry point](https://packaging.python.org/specifications/entry-points/#use-for-scripts) can be added to `setup.py`, so that pip will also build a GUI launcher.

----

**Issues / Open questions**

- Since I don't know much about python wheels and python packaging in general, I guess that this change will also add a `streamlink-noconsole` on Linux and macOS. Is it possible to change that?
- Is the name `streamlink-noconsole` okay? Or does anybody have a better idea?
- Regular installs via `python setup.py install` will still build the useless GUI launcher.
- ~~I've noticed some issues with certain plugins while using the GUI launcher, but I don't know if it was because of the build env in my Windows VM or something else. This has to be resolved before this PR can get merged. (I will check this tomorrow)~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,9 @@ before_install:
       pip install -r docs-requirements.txt;
       pip install doctr;
     fi
+  - if [[ $BUILD_INSTALLER == 'yes' ]]; then
+      pip install git+https://github.com/takluyver/pynsist.git@88f56f9e86af9c55522147a67f8b7806ac2ca55b;
+    fi
 
 install:
   - pip install -e .
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,5 +5,4 @@ codecov
 coverage
 mock
 requests-mock
-pynsist
 freezegun
--- a/script/makeinstaller.sh
+++ b/script/makeinstaller.sh
@@ -106,6 +106,10 @@ files=../win32/LICENSE.txt > \$INSTDIR
 [Command streamlink]
 entry_point=streamlink_cli.main:main
 
+[Command streamlinkw]
+entry_point=streamlink_cli.main:main
+console=false
+
 [Build]
 directory=nsis
 nsi_template=installer_tmpl.nsi
--- a/script/sdistsign.sh
+++ b/script/sdistsign.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+shopt -s nullglob
 set -e
 
 KEY_ID=2E390FA0
@@ -8,6 +9,8 @@ version=$(python setup.py --version)
 dist_dir=${STREAMLINK_DIST_DIR:-dist}
 temp_keyring=$(mktemp -d) && trap "rm -rf ${temp_keyring}" EXIT || exit 255
 
+wheel_platforms_windows=("win32" "win-amd64" "cygwin")
+
 if [[ -n "${TRAVIS}" ]]; then
       openssl aes-256-cbc -K ${encrypted_eeb8b970d3a3_key} -iv ${encrypted_eeb8b970d3a3_iv} -in signing.key.enc -out "${SDIST_KEY_FILE}" -d
 fi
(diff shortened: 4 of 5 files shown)


beardypig:
What does `console=false` actually do? Can we just set it for the main `streamlink` binary?

bastimeyer:
> What does `console=false` actually do?

What it *actually* does? See takluyver/pynsist#179
TL;DR it builds the *same* so-called "GUI"/"no console" launcher as the `gui_scripts` entry points when a wheel gets installed via pip.
https://packaging.python.org/specifications/entry-points/#use-for-scripts

Again, and I don't want to repeat myself over and over :weary:, this is necessary to prevent a terminal window from being opened in a non-terminal context on Windows. This "GUI" launcher uses `pythonw.exe` instead of `python.exe` and properly redirects the i/o streams.

> Can we just set it for the main `streamlink` binary?

No, because that would break situations where a terminal window is desired/needed, like for example when launching `streamlink.exe` from shortcuts. We would also have to switch the `console_scripts` entry point to `gui_scripts`.

**If python on Windows wouldn't have this stupid distinction between python/pythonw, this all wouldn't be necessary.**

----

I've been trying to solve this mess for the past two weeks now and adding an additional "GUI" launcher is the only way to properly and sensibly fix this. Otherwise, every non-CLI application using Streamlink that doesn't want a terminal window from being shown would have to parse the **binary** launcher executable for the included shebang, so that the path to the python executable can be found and replaced with pythonw. This is nonsense.

There is no harm in adding an additional executable to the Windows installer. The wheel however is a bit different, because it also affects Linux and macOS. So if a different *flavor* can't be built and published explicitly for Windows, we can remove the `gui_scripts` entry point from setup.py again if you want. This isn't too much of a problem for me and the Twitch GUI, because the incompatible launcher can be ignored in pip via `--no-binary streamlink` and an old regular entry script will be created instead.

beardypig:
Yeah, I got all that from the previous issues, I was just wondering if a regular `python.exe` console script was even required - sounds like it is.

If we build separate wheels for each platform we could probably include it just for Windows, but as we are building universal wheels I don't think we can do it as-is. Maybe @back-to knows more. 

I'm fine with the name, but if you want another suggestion `streamlinkw` :p

bastimeyer:
https://packaging.python.org/guides/distributing-packages-using-setuptools/#platform-wheels

We just need to build an additional Windows specific wheel. That will be selected first on Windows, so there's no need to change the current config.

Would it make sense copying setup.py to a temporary location, modifying it and renaming the output wheel name? Or is there an easier/better way for building platform specific wheels from one location?

> streamlinkw

That's probably better :)

bastimeyer:
Looks like we need to build Windows specific wheels for `win32` and `win-amd64` (`python setup.py bdist_wheel --plat-name=XYZ`). Probably also `cygwin`. :confused: 

https://www.python.org/dev/peps/pep-0425/#id1

https://github.com/pypa/wheel/blob/0.33.1/wheel/bdist_wheel.py#L49-L51
https://github.com/pypa/wheel/blob/0.33.1/wheel/bdist_wheel.py#L111-L119
https://github.com/pypa/wheel/blob/0.33.1/wheel/bdist_wheel.py#L148-L156
https://github.com/python/cpython/blob/v3.7.2/Lib/distutils/util.py#L18-L91

Regarding the actual build procedure, would it make sense adding a diff/patch file for setup.py that gets applied after building the non-platform-specific wheel and then building the specific ones? Before I update this pull request, I'd like to get some feedback first whether this is the right path to take here.

Btw, the [`scripts/sdistsign.sh`](https://github.com/streamlink/streamlink/blob/1.0.0/script/sdistsign.sh) script name doesn't make sense if a wheel also gets built :sweat_smile:

beardypig:
I'm not too fussed having the `streamlink-noconsole` console script in the universal wheel, I imagine no one will notice :)

bastimeyer:
Let's not add this to Linux and macOS. Remember that this will add a duplicate entry script to the user's PATH and will make tab-completion a bit annoying.

Regarding my patch file approach, this unfortunately won't work because of versioneer. The resulted version string will have a `dirty` tag added because of the modified setup file. This means that the file either has to be duplicated (which is kinda stupid), or modified with custom logic. Since the `--plat-name` parameter has to be set for the bdist_wheel command, it would make sense making use of it in the entry_points field in order to see whether the wheel gets built for Windows or not.

bastimeyer:
@beardypig Could you please review these changes when you get the time? Since you're the one who wrote the sdistsign script, I would like to have it reviewed by you.

I could split the PR into two, one for the installer and one for the wheels if that's better or if you don't have the time right now. The installer changes should be a no-brainer and could get merged immediately.

I would really like to have this merged and a new release published, because the Twitch GUI is in an awkward situation on Windows since Streamlink's 1.0.0 release.

gravyboat:
Looks fine to me as well on a quick review so I'm going to merge, thanks @bastimeyer.

Issue #971, opened 2017-05-31
Parsing issue in --player-args


bastimeyer:
### Checklist

- [x] This is a bug report.
- [ ] This is a feature request.
- [ ] This is a plugin (improvement) request.
- [ ] I have read the contribution guidelines.

### Description

Original bug report by @404NetworkError here:
streamlink/streamlink-twitch-gui#464

Streamlink is exiting with an error if the first argument of the `--player-args` param is a single hyphen parameter. Defining the same player parameters in the `--player` param is working fine.

### Environment details

Operating system and version: tested on Linux, MacOS and Windows
Streamlink and Python version: latest

### Comments, logs, screenshots, etc.

```bash
# just some example commands that don't necessarily make sense

# single hyphen params at the beginning of --player-args are not working
$ streamlink -p vlc -a "-v {filename}" twitch.tv/CHANNEL best
usage: streamlink [OPTIONS] <URL> [STREAM]
streamlink: error: argument -a/--player-args: expected one argument

# no hyphen params at the beginning of --player-args are working fine
$ streamlink -p vlc -a "v {filename}" twitch.tv/CHANNEL best
[cli][info] Found matching plugin twitch for URL twitch.tv/CHANNEL
^CInterrupted! Exiting...

# double hyphen params at the beginning of --player-args are working fine
$ streamlink -p vlc -a "--v {filename}" twitch.tv/CHANNEL best
[cli][info] Found matching plugin twitch for URL twitch.tv/CHANNEL
^CInterrupted! Exiting...

# Windows style params at the beginning of --player-args are working fine
$ streamlink -p vlc -a "/v {filename}" twitch.tv/CHANNEL best
[cli][info] Found matching plugin twitch for URL twitch.tv/CHANNEL
^CInterrupted! Exiting...
```

beardypig:
This is a limitation of `argparse`. It can be worked around by using a `=` to set the value. 
eg. `-a="-v {filename}"` or `--player-args="-v {filename}"`.

Here is a small snippet to recreate the issue:
```python
import argparse
args = argparse.ArgumentParser()
args.add_argument("-t", "--test")
print(args.parse_args())
```
There was a [bug report](http://bugs.python.org/issue9334) raised around 10 years ago for this issue :)

back-to:
the last message on the bug report has a workaround https://bugs.python.org/issue9334

not sure if there is any downside of it.

```
[...]

PR #2333, opened 2019-03-01
cli.argparser: Fix single hyphen params at the beginning of --player-args

back-to:
use workaround solution from

https://bugs.python.org/issue9334

don't know if there is any downside,
but seems like it could work

it can be reversed, if there are some issue

closes https://github.com/streamlink/streamlink/issues/971

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -1,4 +1,5 @@
 import argparse
+import numbers
 import re
 from string import printable
 from textwrap import dedent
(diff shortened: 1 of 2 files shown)


bastimeyer:
Is this compatible with all versions?

back-to:
@bastimeyer 
travis did not fail, so it might be.

bastimeyer:
I was just wondering because all that's being tested here is this special case and not the entire argparser testsuite.

Just had a look at it again and didn't realize that all that was added here is the first `if match is None` block, at least compared to 3.7 (you posted [this](https://gist.github.com/back-to/d940a2273a7e58152f4fb9fca79cf454/revisions#diff-df389dc0f92c4a0567555d1f96ff2bf9) on the other issue). So yeah, looks like it's fine. :sweat_smile:

back-to:
as you are fine with it @bastimeyer ,
I merge it and lets see if something breaks :rabbit:

PR #2340, opened 2019-03-06
Add force_verify=true to Twitch OAuth URL

bastimeyer:
Forces the authentication confirm dialog on Twitch, in case the user is already logged in. This lets them switch to a different account or review the OAuth permissions.


bastimeyer:
This can be merged

PR #2342, opened 2019-03-07
TravisCI: Windows + build stages

### LINKED ISSUE #2128: TravisCI reconfiguration
We need to reconfigure the TravisCI hooks, because GH services have been deprecated since the beginning of this month and TravisCI is still registered as one of these here in the Streamlink repo. Old integrations will be removed at the end of January 2019 and need to be replaced with webhooks.
https://developer.github.com/changes/2018-04-25-github-services-deprecation/

What needs to be done now?
Will simply removing and adding TravisCI again work? What will happen to the build history? I couldn't find any answers to these questions in the TravisCI docs. Maybe I'm just blind...

Since the TravisCI GH service was registered by @gravyboat when Streamlink was forked, I decided to not mess with it on my own and open this thread instead.

----

Also regarding CI tools, I've seen that Windows is now supported by TravisCI.
https://blog.travis-ci.com/2018-10-11-windows-early-release

Python environments don't seem to be available in the Windows VMs/containers yet, but once they become available, I'd say that we should switch and remove AppVeyor, as it is redundant. AppVeyor is so utterly slow and bad, replacing it could speed up the overall CI process. It's ridiculous that several builds are still queued since almost an entire day after the database issues of GH yesterday, even though the issue has been resolved and the webhooks have been re-triggered. TravisCI by comparison has caught up quickly in just a few hours.

Also, I've noticed that the Streamlink builds on AppVeyor are registered on @beardypig's account. It's not really a problem, but other people don't have access to the account, so this might cause problems in the future if we don't intend to remove the AppVeyor integration.

bastimeyer:
AppVeyor is failing once again with random network stuff and is breaking the builds. Let's see if Windows on TravisCI is already good enough for Streamlink (still in beta).

Resolves #2128

## Changes

- Added Windows VMs to the build matrix
  Based on @back-to's config mentioned [here](https://github.com/streamlink/streamlink/issues/2128#issuecomment-462141708).
- Added build stages
  This splits the test, documentation and deploy configs into consecutive and conditional stages instead of using one single python 3.5 VM with special build env vars. It should also improve the readability of the config. The deploy stage will only be run if all previous tests have passed in all environments. This wasn't the case previously.
- Split `sdistsign.sh` into `build-and-sign.sh` and `deploy-pypi.sh`

## To do

- [ ] Remove appveyor.yml once webhook has been removed
- [ ] Test deploy configs in a side-repo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,19 @@ script:
   - codecov
 
 
+x-windows-installer: &x-windows-installer
+  python: '3.5'
+  addons:
+    apt:
+      packages:
+        - nsis
+        - imagemagick
+        - inkscape
+  before_script:
+    - python -m pip install git+https://github.com/takluyver/pynsist.git@88f56f9e86af9c55522147a67f8b7806ac2ca55b
+  script: ./script/makeinstaller.sh
+
+
 jobs:
   allow_failures:
     - stage: test
@@ -82,7 +95,11 @@ jobs:
         - export PATH="/c/Python37:/c/Python37/Scripts:$PATH"
         - export TRAVIS_PYTHON_VERSION="3.7"
 
-    # test documentation
+    # build Windows installer
+    - name: installer
+      <<: *x-windows-installer
+
+    # build documentation
     - name: documentation
       python: '3.5'
       before_script:
@@ -124,15 +141,7 @@ jobs:
     # tagged release: Windows installer + Github
     - stage: deploy
       if: repo = streamlink/streamlink AND branch = master AND type = push AND tag IS present
-      python: '3.5'
-      addons:
-        apt:
-          packages:
-            - nsis
-            - imagemagick
-            - inkscape
-      before_script: python -m pip install git+https://github.com/takluyver/pynsist.git@88f56f9e86af9c55522147a67f8b7806ac2ca55b
-      script: ./script/makeinstaller.sh
+      <<: *x-windows-installer
       deploy:
         - provider: releases
           api_key: "${RELEASES_API_KEY}"
@@ -146,15 +155,7 @@ jobs:
     # cron: Windows installer + Bintray
     - stage: deploy
       if: repo = streamlink/streamlink AND branch = master AND type = cron
-      python: '3.5'
-      addons:
-        apt:
-          packages:
-            - nsis
-            - imagemagick
-            - inkscape
-      before_script: python -m pip install git+https://github.com/takluyver/pynsist.git@88f56f9e86af9c55522147a67f8b7806ac2ca55b
-      script: ./script/makeinstaller.sh
+      <<: *x-windows-installer
       before_deploy: ./script/bintrayconfig.sh
       deploy:
         - provider: bintray
--- a/.travis.yml
+++ b/.travis.yml
@@ -109,7 +109,7 @@ jobs:
     # deploy documentation
     - stage: deploy
       name: documentation
-      if: repo = streamlink/streamlink AND branch = master AND type = push
+      if: repo = streamlink/streamlink AND ( branch = master AND type = push OR tag IS present )
       python: '3.5'
       before_script:
         - python -m pip install -r docs-requirements.txt
@@ -126,7 +126,7 @@ jobs:
     # tagged release
     - stage: deploy
       name: release
-      if: repo = streamlink/streamlink AND branch = master AND type = push AND tag IS present
+      if: repo = streamlink/streamlink AND tag IS present
       python: '3.5'
       before_script: python -m pip install --upgrade wheel twine
       script: ./script/build-and-sign.sh
@@ -145,7 +145,7 @@ jobs:
 
     - stage: deploy
       name: "installer (release)"
-      if: repo = streamlink/streamlink AND branch = master AND type = push AND tag IS present
+      if: repo = streamlink/streamlink AND tag IS present
       <<: *x-windows-installer
       deploy:
         - provider: releases
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
   global:
     - STREAMLINK_DIST_DIR=${TRAVIS_BUILD_DIR}/dist
     - STREAMLINK_INSTALLER_DIST_DIR=${STREAMLINK_DIST_DIR}/nsis
-    - SDIST_KEY_FILE="${TRAVIS_BUILD_DIR}/signing.key"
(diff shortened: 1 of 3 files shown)


bastimeyer:
Why is Github not triggering the build queue on Travis? I've just tried it again with a force-push... Is this a security restriction because of the modified config or are there technical issues right now? What is going on?

bastimeyer:
> Build config did not create any jobs.

https://travis-ci.org/streamlink/streamlink/requests

hm...

Is this because of the explicit build matrix (`include` keyword)?
https://docs.travis-ci.com/user/build-stages/#build-stages-and-build-matrix-expansion

bastimeyer:
Travis has already finished the 11 test stages (only 6 ran in parallel) while AppVeyor is still running the 5th one. :smile: 

One issue I've noticed was that `TRAVIS_PYTHON_VERSION` was not set on the Windows VMs due to `language: shell`. This env var is missing in the coverage report and needs to be fixed.

bastimeyer:
Apart from having checked the new build config in a separate test-repo, I'm quite happy with the changes. A review would be appreciated. In case something's not clear, please see the individual commits. I can clean up my PR a bit, if it's needed.

gravyboat:
Looks good to me @bastimeyer! I'll leave it open for a bit longer in case anyone else has opinions, but if not we'll get it merged in. Thanks!

bastimeyer:
Let's wait for a third (or fourth) person to confirm these changes first. It's a complete rewrite of the CI config and that shouldn't be merged without a proper review. I've already missed at least three critical things in the first commits and needed to fix that, which means that there's a chance I've missed something else, too.

----

Btw, another thing I just thought about: why does the wheel need to get uploaded to the Github releases page? How important is it to have it there when it's available on PyPI anyway? Yes, the 1.0.0 wheel has been downloaded ~400 times so far, but that's only a tiny fraction compared to the other release assets and stats from PyPI:
- https://api.github.com/repos/streamlink/streamlink/releases/latest
- https://pypistats.org/packages/streamlink

The problem is that with the new special wheels for Windows (win32, win-amd64 and cygwin) this will probably confuse many people, because the Windows installer which is only being uploaded and available there doesn't have a platform name included in its file name, whereas the new special wheels for Windows do.

Even without these CI changes, the next release page will include 1 platform independent wheel, 3 Windows specific wheels, 1 source archive, 5 pgp signatures and the Windows installer. A total number of 11 files, plus the two archives Github lists of the repo's tagged commit.

- **current** github-releases deploy config 
  https://github.com/streamlink/streamlink/blob/e96c725cb553027dd1ca7b1c278211b57947b05c/.travis.yml#L100
- **unchanged** github-releases deploy config
  https://github.com/streamlink/streamlink/pull/2342/files#diff-354f30a63fb0907d4ad57269548329e3R139

beardypig:
@bastimeyer I'd be in favour of reducing that list down to the Windows installer ;) That's the important one for the GitHub releases, everyone else will use PyPI - although might have to check with the package maintainers :)

bastimeyer:
I've left the tarball and its signature file. I think that's fair.

I'll test the deploy configs in a side-repo later today. If nothing goes wrong, this should be good to go.

bastimeyer:
Ok, did a bit of testing and learned a couple of things

https://travis-ci.org/bastimeyer-test/streamlink-test/builds/506686333
https://github.com/bastimeyer-test/streamlink-test/releases

- `allow_failures` does not work, neither does specifying `allow_failure: true` on an individual job. This must be set on the global build matrix and **all** fields must match. I've googled this for at least 30 mins and only found a couple of other people having the same issue. That's why I'm ignoring this for now, since `python: 3.8-dev` seems to be working fine.
- All Windows VMs failed in my tests during the dependency install via chocolatey for no apparent reason (I've removed them in my side-repo in the second test-release)
  https://travis-ci.org/bastimeyer-test/streamlink-test/builds/506676918
- Travis does a terrible job at documenting the build stage conditions. There's an issue with the `tag` and `type`/`branch` variables where either `type` or `branch` have a different value when `tag` is set. This needs a fix here. Glad I've tested it.
- Didn't get doctr to work with the `--token` option and didn't want to read about creating a doctr deploy key, hence the deploy failures. Replaced the gpg signing key and openssl encryption key/iv, though (working fine).

The "release" deploy job failed because I've messed up the changelog, but it's not relevant to the CI changes, same as the "documentation" deploy job.
The failure of the Windows VMs is a bit weird, though, and I'd like to know why this happened.

bastimeyer:
Damn it, the Windows VMs are failing when secure env vars are set:
https://travis-ci.community/t/current-known-issues-please-read-this-before-posting-a-new-topic/264/10
:angry: :angry: :angry:

PRs don't have any secure vars, which is the reason why the VMs were working here. @back-to's build config also didn't have any secure env vars.

I guess we'll have to wait for a fix on Travis' side then. Or we could move the secure env vars into the config file itself and then only apply them to the deploy jobs running on Linux. I don't think it's worth it, though, as it would only bloat it up more.

bastimeyer:
Github's own CI will probably leave beta before Travis fixes their Windows stuff.
https://github.com/features/actions

Github actions are looking very good and promising. Let's hope it's fast, too.

gravyboat:
@bastimeyer Yep saw this today and was thinking it might be a nice option for us.

gravyboat:
@bastimeyer We should just move to github actions. The CI/CD is really good. We're using it where I work and minus a few interface issues it's much better than the alternatives.

bastimeyer:
@gravyboat Sure, but remember that Github actions are still in beta.

I have a config stashed here in my local repo, but there are a couple of issues which I need to resolve first. I can open a WIP PR though.

1. I don't have the secret keys which are set on TravisCI, so these have to be set in the repo's Github settings first.
2. Then there's the issue with the deploy scripts. TravisCI has their dpl library which Github does not have and you have to rely on 3rd party stuff instead. These 3rd party scripts are unfortunately (still) in a bad shape. I've looked through some Github releases deploy scripts and they all were kinda bad. And for Bintray, there don't seem to be any available. This means that we have to write our own scripts for that and do the necessary API requests and file uploads via curl.
3. Another thing I currently don't like about Github actions are the configs themselves. For example, you need to have a dedicated config file for cron jobs, which means that you have to duplicate a lot of stuff instead of having it in one single config.

Before we switch over to Github actions, we should keep things running in parallel for a while and keep the deploy stuff on TravisCI until all issues have been resolved. I also don't know how fast it is compared to TravisCI. It should be faster than AppVeyor though.

Issue #2346, opened 2019-03-07
plugin for ntv.ru


vstavrinov:
## Plugin Request

- [x] This is a plugin request and I have read the contribution guidelines and plugin request requirements.

### Description

This is one of three main russian tv channels: НТВ. They are streaming their own contents on their official site https://www.ntv.ru/ 
### Example stream URLs

1. https://www.ntv.ru/air/
2. ...
3. ...

### Additional comments, screenshots, etc.
I have already written and tested plugin for this stream - it's working as expected. See attachments. So You can include it in the project and I am willing to maintain it.

[ntv.zip](https://github.com/streamlink/streamlink/files/2943428/ntv.zip)

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

PR #2351, opened 2019-03-09
plugins.ntv: new Plugin for ntv.ru (#2346)

vstavrinov:
NTV is russian tv channel (#2346): 

https://en.wikipedia.org/wiki/NTV_(Russia)

### DIFF
--- a/src/streamlink/plugins/ntv.py
+++ b/src/streamlink/plugins/ntv.py
@@ -1,0 +1,21 @@
+import re
+from streamlink.plugin import Plugin
+from streamlink.stream import HLSStream
+
+
+class NTV(Plugin):
+    url_re = re.compile(r"https?://(?:www\.)?ntv.ru/air/$")
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls.url_re.match(url) is not None
+
+    def _get_streams(self):
+        tns = 'http://www.ntv.ru/services/live/tns/'
+        body = self.session.http.get(tns).text
+        d = {}
+        exec(body, d)
+        mrl = d['hls']
+        return HLSStream.parse_variant_playlist(self.session, mrl)
+
+__plugin__ = NTV
--- a/src/streamlink/plugins/ntv.py
+++ b/src/streamlink/plugins/ntv.py
@@ -4,7 +4,7 @@ from streamlink.stream import HLSStream
 
 
 class NTV(Plugin):
-    url_re = re.compile(r'https://www.ntv.ru/air/.*')
+    url_re = re.compile(r'https?://www.ntv.ru/air/.*')
 
     @classmethod
     def can_handle_url(cls, url):
--- a/src/streamlink/plugins/ntv.py
+++ b/src/streamlink/plugins/ntv.py
@@ -12,12 +12,15 @@ class NTV(Plugin):
 
     def _get_streams(self):
         body = self.session.http.get(self.url).text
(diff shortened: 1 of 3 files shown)


src/streamlink/plugins/ntv.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls.url_re.match(url) is not None
+
+    def _get_streams(self):
+        tns = 'http://www.ntv.ru/services/live/tns/'
+        body = self.session.http.get(tns).text
+        d = {}
+        exec(body, d)
```
beardypig: The use of `exec` is not permitted, it's very insecure to `exec` unvalidated input. Instead you could do 
```suggestion
        d = dict(re.findall(r'(\w+)=\'(.*?)\'', body))
```

vstavrinov: > The use of `exec` is not permitted, it's very insecure to `exec` unvalidated input. Instead you 

Already fixed. Thank you @beardypig

src/streamlink/plugins/ntv.py
```diff
@@ -0,0 +1,23 @@
...
+class NTV(Plugin):
+    url_re = re.compile(r'https?://www.ntv.ru/air/.*')
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls.url_re.match(url) is not None
+
+    def _get_streams(self):
+        body = self.session.http.get(self.url).text
+        mrl_re = re.compile(r'var camHlsURL = \'(.*)\'')
```
beardypig: You could make the regex more generic and use the same one for both, and only compile it one. If you only use a regex once, there is no need to compile it :)

This would probably serve for both. 
```python
mrl_re = re.compile(r'var (?:camH|h)lsURL = \'(.*)\'')
```

beardypig: In fact, you could just do

```python
body = self.session.http.get(self.url).text
match = re.search(r'var (?:camH|h)lsURL = \'(.*)\'', body)
if match:
  mrl = match.group(1)
  return HLSStream.parse_variant_playlist(self.session, mrl)
```

And it would match both just fine - obviously you need to test it, cos I just wrote it directly in the comments...

vstavrinov: I had the same idea before writing the code for last commit. I like compact and simple code and that was reason for exec () that You fairly criticized.  But...  In this case problem is that both pattern will match  in the same request (body) for any channels other than main one. Thus merging both pattern is not enough. Further we need additional code to distinguish requested channel. This requires more complex code than actual one. So I prefer this solution.

beardypig: I undertand :)

I think you add some more error handling, what if neither regex matches? Also, you can avoid performing the regex twice.

```python
    def _get_streams(self):
        body = self.session.http.get(self.url).text
        mrl = None
        match = re.search(r'var camHlsURL = \'(.*)\'', body)
        if match:
            mrl = 'http:' + match.group(1)
        else:
            match = re.search(r'var hlsURL = \'(.*)\'', body)
			if match:
	            mrl = match.group(1)
		if mrl:
	        return HLSStream.parse_variant_playlist(self.session, mrl)
```

vstavrinov: Done. Thanks, @beardypig.

vstavrinov: Are there something missing to get merging  completed?

beardypig:
Looks fine to me :)

PR #2358, opened 2019-03-16
plugins.twitch: platform=_ in access_token request

### LINKED ISSUE #2357: pre-roll ads playing before twitch stream
## Bug Report

- [x ] This is a bug report and I have read the contribution guidelines.

### Description

streamlink 1.0.0 on windows 10 64bit (latest updates) has a chance of playing pre-roll ads before loading the twitch stream. sometimes it will partially play the end of an ad (i.e. 1 or 2 seconds), sometimes the entire ad. sometimes the ad is low resolution (480/720), sometimes 1080p. sometimes you get expected behavior (no ads).

expected: play the stream straight away
actual: sporadic pre-rolls play, then the stream.

1. use streamlink 1.0.0
2. execute streamlink using cmd line "streamlink twitch.tv/sodapoppin best"
3. other twitch streams can exhibit the same issue, but not always

### Log output

```
C:\WINDOWS\system32>streamlink twitch.tv/sodapoppin best -l debug
[cli][info] Found matching plugin twitch for URL twitch.tv/sodapoppin
[plugin.twitch][debug] Getting live HLS streams for sodapoppin
[cli][info] Starting player: "C:\mpv-x86_64-20190224\mpv.exe"
```

configuration file is default save for player-passthrough enabled/default player changed to mpv.
video of the issue in action - https://mega.nz/#!HEBn1K4S!titCic2lUv4Hsz2c5LjY4TDvK2cjrHRS_Y_BbyO8COk
i'll stress that this is sporadic - sometimes i'll get an ad no matter how many times i load the stream. sometimes i won't get an ad at all. this first started happening for me on march 14th.
bonus: alternate player for twitch.tv plugin (chrome/firefox) has the same issue.

bastimeyer:
Resolves #2357
Sets the `platform=_` GET parameter in the stream access token request, which results in `server_ads` to be set to `false` in the response (at least for now - might get changed/fixed in the future).

As I've said in #2357, I'm not sure how to feel about actively suppressing ads in Streamlink and its plugins. Yes, this is what probably most - if not all - of the users want, but this violates the ToS of Twitch and may put us at risk of losing our API client ID in the future. Discussion is welcome.

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -222,7 +222,7 @@ class TwitchAPI(object):
     # Private API calls
 
     def access_token(self, endpoint, asset, **params):
-        return self.call("/api/{0}/{1}/access_token".format(endpoint, asset), **params)
+        return self.call("/api/{0}/{1}/access_token".format(endpoint, asset), **dict(platform="_", **params))
 
     def token(self, **params):
         return self.call("/api/viewer/token", **params)


gravyboat:
Thanks for adding this @bastimeyer. Which portion of the ToS are we explicitly violating by doing this? In the past Twitch has said they are okay with people bypassing ads, and tons of browser plugins and other tools have been doing so for quite some time with no repercussions. Does this specific change impact us? We're simply modifying the request we send to the server and it's responding to that request with what we've asked for. I don't think I'd be so against the ads on Twitch if they weren't complete cancer (like most video ads).

bastimeyer:
Hm, I should have said that I'm not sure whether this violates ToS or not.
What I mean is that it's not exactly clear. We're obviously not using any of their player tech, but the following point is listed under "Requirements for Specific Features and APIs" and the embeddable player is the only offical way to view streams/vods/clips. A custom HLS client/player may fall into the same category, although not explicitly mentioned.

> Don’t modify, replace, interfere with or block (a) functionality of the embeddable player, including advertisements placed by Twitch; and (b) the Twitch Marks.

https://www.twitch.tv/p/legal/developer-agreement/

But you're probably right and it shouldn't be a big deal. I'm just saying that we should be careful when adding this kind of stuff, nothing more :)

gravyboat:
@bastimeyer I don't think we fall into that category at all as we aren't embedding or even using their player. I agree with you that we should be careful and apologize if my comment came off as anything other than acknowledging and agreeing with that fact

MancV21:
Is this branch available yet?

bastimeyer:
@MancV21 
You can clone my repo and build from the branch of this PR if you don't want to wait for it to be merged into Streamlink's master branch. Or just apply this simple change manually.

gravyboat:
I'm going to merge this in @bastimeyer.

Issue #2344, opened 2019-03-07
Pluzz plugin : Broken


karlo2105:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

Pluzz plugin is unable to open French TV public channels.

### Reproduction steps / Explicit stream URLs to test

1. `streamlink "https://www.france.tv/franceinfo/direct.html" best`
2. `streamlink "https://www.france.tv/france-2/direct.html" best`
3. ...

### Log output

```
streamlink "https://www.france.tv/franceinfo/direct.html" best
[cli][info] Found matching plugin pluzz for URL https://www.france.tv/franceinfo/direct.html
error: No playable streams found on this URL: https://www.france.tv/franceinfo/direct.html
```

### Additional comments, screenshots, etc.

This plugin doesn't open any channels from French Public TV. 
I use the latest streamlink build `streamlink 1.0.0+34.ge96c725` with Python 2.7.15 on Win 7 Pro. 
Tested channel France Info isn't geoblocked. 
Thanks for patching.

tmatthews5511:
I would like to confirm this issue. 

streamlink https://www.france.tv/france-2/direct.html best
[cli][info] Found matching plugin pluzz for URL https://www.france.tv/france-2/direct.html
error: No playable streams found on this URL: https://www.france.tv/france-2/direct.html

streamlink https://www.france.tv/france-2/envoye-special/913213-envoye-special.html best
[cli][info] Found matching plugin pluzz for URL https://www.france.tv/france-2/envoye-special/913213-envoye-special.html
error: No playable streams found on this URL: https://www.france.tv/france-2/envoye-special/913213-envoye-special.html

streamlink 1.0.0+34.ge96c725 / Python 2.7.15rc1 (Linux)

campones:
I had opened  a thread about it and I thought it was fixed but it's still not working
https://github.com/streamlink/streamlink/issues/2332

tmatthews5511:
With latest streamlink, same problem:

streamlink -l debug https://www.france.tv/france-3/direct.html best
[cli][debug] OS:         Linux-4.15.0-45-generic-x86_64-with-LinuxMint-19.1-tessa
[cli][debug] Python:     2.7.15rc1
[cli][debug] Streamlink: 1.0.0+38.g9ae6756
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.54.0)
[...]

PR #2361, opened 2019-03-18
plugins.pluzz: Fixed regex, they use quotes now.

back-to:
Fixed https://github.com/streamlink/streamlink/issues/2344


gravyboat:
Thanks @back-to.

Issue #2343, opened 2019-03-07
BSTV.bg


anchev:
## Plugin Request

- [x] This is a plugin request and I have read the contribution guidelines and plugin request requirements.

### Description

There is a new free Bulgarian TV with live stream on https://bstv.bg/

### Example stream URLs

Looking at the HTML source of the page I see (in the javascript code) those which seem to be for various resolutions:

1. http://213.91.254.84:80/hls/stream.m3u8
2. http://213.91.254.84:80/hvod1/start.mp4/index.m3u8
3. http://213.91.254.84:80/hvod2/razlichna.mp4/index.m3u8

However the JS player seems to add additional GET parameters. Using those URLs directly in streamlink doesn't work (403).

In case you can't access it internationally you can try proxies listed in [this issue](https://github.com/streamlink/streamlink/issues/374#issuecomment-271077357).

### Additional comments, screenshots, etc.

None.

roshavagarga:
Domain seems to have shifted to https://live.bstv.bg/

bastimeyer:
Please open a new issue.

PR #2362, opened 2019-03-18
plugins.cdnbg: New domain videochanel.bstv.bg

back-to:
closes https://github.com/streamlink/streamlink/issues/2343


gravyboat:
Thanks @back-to.

PR #2367, opened 2019-03-20
Release 1.1.0

### LINKED ISSUE #2349: 1.1.0 release
With #2326 being merged, I would appreciate if we could have a new minor release in the next couple of days. This is important for streamlink/streamlink-twitch-gui#618, which relies on a new Streamlink release.

Another thing to consider is the EOL of Python 3.4 next week (2019-03-16). Would it make sense waiting for that to happen first and then removing its support from Streamlink? Or are there any reasons for keeping it?

> DEPRECATION: Python 3.4 support has been deprecated. pip 19.1 will be the last one supporting it. Please upgrade your Python as Python 3.4 won't be maintained after March 2019 (cf PEP 429).

https://www.python.org/dev/peps/pep-0429/

bastimeyer:
Here's the changelog for the next release. Edit rights are granted, so if anything is wrong, feel free to fix it.

Resolves #2349

~~Allegedly there are still issues with the Twitch plugin and embedded ads. Maybe we should wait for that to resolve first if it's not a false alarm. Still waiting for a proper bug report by one of the people who are running into this issue...~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,80 @@
 # Changelog
 
+## streamlink 1.1.0 (2019-03-31)
+
+These are the highlights of Streamlink's first minor release after the 1.0.0 milestone:
+
+- several plugin fixes, improvements and new plugin implementations
+- addition of the `--twitch-disable-ads` parameter for filtering out advertisement segments from Twitch.tv streams (#2372)
+- DASH stream improvements (#2285)
+- documentation enhancements (#2292, #2293)
+- addition of the `{url}` player title variable (#2232)
+- default player title config for PotPlayer (#2224)
+- new `streamlinkw` executable on Windows (wheels + installer) (#2326)
+- Github release assets simplification (#2360)
+
+
+```text
+Brian Callahan <bcallah@openbsd.org> (1):
+      Add OpenBSD to the installation docs
+
+Peter Rowlands (변기호) <peter@pmrowla.com> (2):
+      streams.dash: Support manifest strings in addition to manifest urls (#2285)
+      plugins.facebook: Support manifest strings and tahoe player urls (#2286)
+
+Roman Kornev <w.romankornev@gmail.com> (2):
+      cli.main: Add {url} argument to window --title (#2232)
+      cli.output: Add window title for PotPlayer (#2224)
+
+Sebastian Meyer <mail@bastimeyer.de> (1):
+      Build additional "streamlinkw" launcher on Windows (#2326)
+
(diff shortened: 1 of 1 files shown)


gravyboat:
@bastimeyer Thanks for putting this together. I saw the report about people still having issues as well but it seems like that might be due to people using out of date plugins? It's hard to tell. Did you manually generate lines 14-61? That was the portion that wasn't being created for me.

bastimeyer:
> Did you manually generate lines 14-61

Yes, I didn't run that script, because I have my remotes named differently and it's hardcoded. It simply runs
```bash
git shortlog --email --no-merges --pretty=%s ${LAST_TAG_NAME}..
```

laurencee:
The twitch ads PR is probably the highlight for this release, lots of people (myself included) are excited for that 👍

bastimeyer:
Rebased the PR and updated the changelog...
This release doesn't necessarily have to happen today, but everything should be ready so far (except the empty stream player issue when filtering out twitch ads)

bastimeyer:
The codecov spam and mislabeled coverage reduction is getting really annoying. This is caused by Appveyor, btw...

bastimeyer:
Any concerns here? If not, I'll update the release date once more later today, then merge the changes and push the `1.1.0` tag.

gravyboat:
@bastimeyer No I think we are good, do you want to get in https://github.com/streamlink/streamlink/pull/2387/files and https://github.com/streamlink/streamlink/pull/2386? If not we can do them on the next release.

bastimeyer:
The plugin bugfix can/should be merged. I'll wait for that until this evening or tomorrow, in case somebody else want to verify the changes first. The Twitch plugin cleanup can be done after the release (with some more cleanups).

If you want to do it instead, go head. I've updated the date of the changelog, so all you'll have to do is merge this PR and add+push the `1.1.0` tag afterwards.

gravyboat:
Merged this then opened up https://github.com/streamlink/streamlink/pull/2391 because of the changelog append (couldn't push directly to master), then tagged 1.1.0 and pushed that, should be all set to go. Thanks for your hard work @bastimeyer.

bastimeyer:
The PyPI deploy job failed:
https://travis-ci.org/streamlink/streamlink/jobs/513828143#L3637
https://pypi.org/project/streamlink/#files

This means that there are no `win32` and `win-amd64` wheels. The `cygwin` wheel can be removed from the next build, but now I'm not sure whether the other ones will be accepted.

Issue #2368, opened 2019-03-20
Twitch Pre-Roll ads (even on audio_only) -- Yes, tested using git snapshot post "fix"


mjevans:
- [x] This is a bug report and I have read the contribution guidelines.

The previously closed issue #2357 is still happening and I have reproduced it after manually installing a git checkout from about 10 min before this bug report.

Current test URL:

Even audio only (though the example was captured at another resolution)

```
streamlink twitch.tv/zisteau audio_only
```

Will display an actual VIDEO ad, also completely ignoring resolution or bandwidth use desires.

Occasionally the mismatch between the ad and the stream also causes the audio to playback in an oddly corrupted way (the streamer sounded down-pitch shifted, and the audio was very studdery, but that was a result of the ad invasion).

```
streamlink --loglevel=debug twitch.tv/zisteau 480p 2>&1 | tee streamlog.txt
[cli][debug] OS:         Linux-4.9.0-8-amd64-x86_64-with-debian-9.8
[cli][debug] Python:     2.7.13
[cli][debug] Streamlink: 1.0.0+46.g42105d8
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.55.0)
[cli][info] Found matching plugin twitch for URL twitch.tv/zisteau
[plugin.twitch][debug] Getting live HLS streams for zisteau
[utils.l10n][debug] Language code: en_US
[cli][info] Available streams: audio_only, 160p (worst), 360p, 480p, 720p, 720p60, 900p60 (best)
[cli][info] Opening stream: 480p (hls)
[stream.hls][debug] Reloading playlist
[stream.hls][debug] First Sequence: 0; Last Sequence: 4
[stream.hls][debug] Start offset: 0; Duration: None; Start Sequence: 2; End Sequence: None
[stream.hls][debug] Adding segment 2 to queue
[cli][debug] Pre-buffering 8192 bytes
[stream.hls][debug] Adding segment 3 to queue
[stream.hls][debug] Adding segment 4 to queue
[stream.hls][debug] Download of segment 2 complete
[cli][info] Starting player: /usr/bin/vlc
[cli.output][debug] Opening subprocess: /usr/bin/vlc --input-title-format twitch.tv/zisteau -
[stream.hls][debug] Download of segment 3 complete
[stream.hls][debug] Download of segment 4 complete
[cli][debug] Writing stream to output
[stream.hls][debug] Reloading playlist
[stream.hls][debug] Adding segment 5 to queue
[...]

PR #2369, opened 2019-03-20
stream.hls: add option to skip discontinuous segments (eg. ads)

beardypig:
Adds a new option `--hls-segment-skip-discontinuity`, that will cause Streamlink to skip any part of the stream that is between two `#EXT-X-DISCONTINUITY` markers. 

This can be used to skip ads that are injected in to the stream. 

Can be used to skip the pre-roll and in-stream ads add by Twitch, as noted in #2368. 

For example 

```
streamlink --hls-segment-skip-discontinuity https://www.twitch.tv/esl_csgo best
```

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -763,6 +763,21 @@ def build_parser():
         """
     )
     transport.add_argument(
+        "--hls-segment-skip-discontinuity",
+        action="store_true",
+        help="""
+        Ignore segments in the playlist that occur during a discontinuity. This
+        option can be used to skip ads that are inserted in to a stream and marked
+        as a discontinuity.
+
+        Default is False.
+
+        Note: This option should be used with care, if the stream is running an ad
+        when Streamlink connects the ad will play and the actual stream will be
+        considered as an ad, and only ads will play.
+        """
+    )
+    transport.add_argument(
         "--hls-segment-key-uri",
         metavar="URI",
         type=str,
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -22,9 +22,9 @@ def num_to_iv(n):
 
 
 def pkcs7_decode(paddedData, keySize=16):
-    '''
+    """
     Remove the PKCS#7 padding
-    '''
+    """
     # Use ord + [-1:] to support both python 2 and 3
     val = ord(paddedData[-1:])
     if val > keySize:
@@ -46,6 +46,7 @@ class HLSStreamWriter(SegmentedStreamWriter):
         self.key_data = None
         self.key_uri = None
         self.key_uri_override = options.get("hls-segment-key-uri")
+        self.skip_discontinuity = options.get("hls-segment-skip-discontinuity")
 
         if self.ignore_names:
             # creates a regex from a list of segment names,
@@ -105,7 +106,10 @@ class HLSStreamWriter(SegmentedStreamWriter):
             request_params = self.create_request_params(sequence)
             # skip ignored segment names
             if self.ignore_names and self.ignore_names_re.search(sequence.segment.uri):
-                log.debug("Skipping segment {0}".format(sequence.num))
+                log.debug("Skipping segment {0} due to ignore name match".format(sequence.num))
+                return
+            if self.skip_discontinuity and sequence.segment.discontinuity:
+                log.debug("Skipping discontinuous segment {0}".format(sequence.num))
                 return
 
             return self.session.http.get(sequence.segment.uri,
--- a/src/streamlink/stream/hls_playlist.py
+++ b/src/streamlink/stream/hls_playlist.py
@@ -1,10 +1,13 @@
 import re
+import logging
 
 from binascii import unhexlify
 from collections import namedtuple
 from itertools import starmap
 
 from streamlink.compat import urljoin, urlparse
+
+log = logging.getLogger(__name__)
 
 __all__ = ["load", "M3U8Parser"]
 
@@ -158,11 +161,13 @@ class M3U8Parser(object):
                 date = self.state.pop("date", None)
                 map_ = self.state.get("map")
                 key = self.state.get("key")
+                discontinuity = self.state.get("discontinuity", False)
 
                 segment = Segment(self.uri(line), extinf[0],
                                   extinf[1], key,
-                                  self.state.pop("discontinuity", False),
+                                  discontinuity,
                                   byterange, date, map_)
+
                 self.m3u8.segments.append(segment)
             elif self.state.pop("expect_playlist", None):
                 streaminf = self.state.pop("streaminf", {})
@@ -210,7 +215,7 @@ class M3U8Parser(object):
                           attr.get("CHARACTERISTICS"))
             self.m3u8.media.append(media)
         elif line.startswith("#EXT-X-DISCONTINUITY"):
-            self.state["discontinuity"] = True
+            self.state["discontinuity"] = not self.state["discontinuity"]
             self.state["map"] = None
         elif line.startswith("#EXT-X-DISCONTINUITY-SEQUENCE"):
             self.m3u8.discontinuity_sequence = self.parse_tag(line, int)
@@ -235,7 +240,7 @@ class M3U8Parser(object):
             self.m3u8.start = start
 
     def parse(self, data):
-        self.state = {}
+        self.state = {"discontinuity": False}
         self.m3u8 = M3U8()
 
         lines = iter(filter(bool, data.splitlines()))
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -796,6 +796,9 @@ def setup_options():
     if args.hls_segment_ignore_names:
         streamlink.set_option("hls-segment-ignore-names", args.hls_segment_ignore_names)
 
+    if args.hls_segment_skip_discontinuity:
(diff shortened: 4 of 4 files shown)


Mathijsz:
Hi! There seem to be some NULL characters at the end of `argparser.py`. Removing them (e.g. by running `sed -i 's/\x0//g' src/streamlink_cli/argparser.py`) fixes the build errors :)

beardypig:
Ah damn. I’m having some issues with files getting corrupted on my VMs...

bastimeyer:
Would it make sense splitting this into two parameters, one which ignores only segments at the beginning of the stream and one which ignores all of these segments?

In case of Twitch, you'd be able to skip the pre-roll ads or all ads, which is quite a difference.

beardypig:
I made it for twitch, but the idea is for skipping any discontinuity, to avoid borking the output stream. 

The true purpose of this option is to avoid discontinuity in the stream output. In the case of twitch, it will drop the ads. But by dropping the ads you don't gain anything in terms of stream latency, or content that might be hidden by ads. You won't see the ads, nor the stream - I imagine you'll see the last frame of the stream before the ad started. Although I didn't encounter any mid-stream ads while testings, so it would be nice to know what actually happens :)

c133:
@beardypig Do you think it is possible to both block the pre-roll ad and get back the stream content that was lost because of that pre-roll ad?

bastimeyer:
@beardypig yes, I fully understand what it does and how it works... My idea was to add an additional parameter for only blocking initial pre-roll ads when launching the stream, because that is somewhat more acceptable to block than those kinds of ads during the stream. The ignored ads at the beginning will only cause a delay of the stream start, but if you're also blocking regular ads, then the stream will halt, just like when having a timeout, and will continue when the ads are over. Bridging this time can be problematic, depending on the player and its config, or the user expectation, so giving the user a choice here whether to block all ads or just the ones at the beginning sounds better to me. 

@c133 No, this doesn't seem to be possible. Just think of ads during a live TV broadcast. Nobody has found a way yet to trick Twitch into listing the live steam segments instead of the ads. And it would be silly of them if they had implemented a way around that system. You'd need a *magic* request parameter, just like the workaround from a couple of days ago.

beardypig:
@bastimeyer this is a rather clean way to implement it, if we start messing with pre-roll vs. mid-stream it's going to get ugly. Not for this PR, perhaps a separate one.

DeVaulted:
@beardypig I tested it for quite a while, you will simply see the last frame of the stream before the ad until the ad is over. Works like a charm, thank you!

bastimeyer:
@beardypig Hm, is it really messy? Wouldn't it be as simple as checking whether the discontinuity tag was set in the first playlist? But yeah, sure, this can be added later of course :)

src/streamlink_cli/argparser.py
```diff
@@ -762,6 +762,21 @@ def build_parser():
...
+        "--hls-segment-skip-discontinuity",
+        action="store_true",
+        help="""
+        Ignore segments in the playlist that occur during a discontinuity. This
+        option can be used to skip ads that are inserted in to a stream and marked
+        as a discontinuity.
+
+        Default is False.
+
+        Note: This option should be used with care, if the stream is running an ad
```
bastimeyer: @beardypig do you think it would make sense filtering plugin specific tags/attributes for detecting this case?
See https://github.com/streamlink/streamlink/issues/2357#issuecomment-473501203
Twitch live segments are annotated in the format `#EXTINF:2.000,live` and ad segments in the format `#EXTINF:2.002,87d22d0b208d6a47c3bba6f412f7fb43-03-04-2019-05.03.22`

Or are the metadata tags included at the beginning of all playlists during stream ads and not just the first one when the channel starts running ads?

src/streamlink_cli/argparser.py
```diff
@@ -762,6 +762,21 @@ def build_parser():
...
+        when Streamlink connects the ad will play and the actual stream will be
```
bastimeyer: missing comma after "connects"

src/streamlink_cli/argparser.py
beardypig: I didn't want it to be enabled by default in any plugin, but it could be. I didn't check, but I assumed that the format for VOD would be similar to the ads...

bastimeyer:
Btw, shouldn't the discontinuity state be kept between playlist reloads? I wrote some tests for your changes, and I'm not sure if I'm misinterpreting something here or did something else wrong, but it looks like it doesn't keep it.
https://github.com/bastimeyer/streamlink/commit/384eadc1d09433b8a37595f1f2511ea7038db494

src/streamlink_cli/argparser.py
bastimeyer: Hm, yeah, maybe, I didn't think about that. :confused:

beardypig:
Yep, needs more work :D

bastimeyer:
If you move the discontinuity state to the worker and pass it to the playlist parser so that it can toggle the state, it should be working fine. I've updated the tests again and added a commit which includes these changes. It's probably a bit messy and not done the right way, but at least you can see what I mean.

src/streamlink_cli/argparser.py
itsotter: Would appreciate this option — I only watch live streams and right now it's a crapshoot if I get an ad on launch.

bastimeyer:
I don't think this boolean `discontinuity` approach works... I just got two ads in a row. The first one was properly skipped, the second one was shown, and then the regular live stream got skipped again.

This is what needs to be done: https://github.com/streamlink/streamlink/issues/2357#issuecomment-474116392
`#EXT-X-SCTE35-OUT` and `#EXT-X-SCTE35-IN` mark the start and end of an ad.

bastimeyer:
Since I don't know much about the actual HLS specs (especially regarding these tags), I did a little bit of more research and from how I understand this, it looks like Twitch is using their own implementation here which differs from other standards.

While Twitch is using the `EXT-X-SCTE35-OUT` and `EXT-X-SCTE35-IN` tags, data about the start and length of the ad segments can also be defined in the attributes of the `EXT-X-DATERANGE` tag via `START-DATE` and `DURATION`/`PLANNED-DURATION`, which they are also including. How useful it is for skipping the segments is questionable, though.
Then there are also the `EXT-X-CUE-OUT`, `EXT-X-CUE-OUT-CONT` and `EXT-X-CUE-IN` tags, which are basically the same as the `EXT-X-SCTE35-{IN,OUT}` ones, but with an additional `-CONT` tag meant to be included in following media sequences.

This makes me think whether this kind of stuff belongs into the twitch plugin with a custom HLSStream implementation and whether the CLI parameter should be specific to the twitch plugin.

I've changed my branch to the SCTE35 stuff now (not yet pushed to my fork though), but since I haven't got any ads on Twitch, I can't verify if it's actually working. The tests alone don't say much.

I would appreciate if we could find a solution here as quickly as possible, because this is blocking the 1.1.0 release which I have been waiting for for far too long now.

Zero3K:
Any news regarding it?

bastimeyer:
@beardypig Can you take a look at this please before I open a pull request? (see each individual commit)
https://github.com/streamlink/streamlink/compare/master...bastimeyer:plugins/twitch/disable-ads

This adds the `--twitch-disable-ads` parameter, is based on the `SCTE35-{OUT,IN}` HLS tags and implements a custom HLSStream solution for the Twitch plugin. Required a bit of refactoring here and there and the code is probably a bit messy. I haven't got any ads on Twitch since 2 days unfortunately, so I can't test this on real streams. The unittests however should be fine for the custom HLSStream* classes.

Zero3K:
@bastimeyer Why not just have it disable Twitch ads automatically without having to use the parameter that you mentioned?

bastimeyer:
Because it's skipping segments and that will cause the stream launch to be delayed. And ads during the stream will cause it to freeze until there's regular stream data again. This will probably confuse everyone who's not aware of the embedded streams situation.
The default value can still be debated (not in this PR though), but I think this implementation should be behind a parameter.

Zero3K:
There's no way to prevent those issues?

bastimeyer:
https://github.com/streamlink/streamlink/pull/2369#issuecomment-475056675

> No, this doesn't seem to be possible. Just think of ads during a live TV broadcast. Nobody has found a way yet to trick Twitch into listing the live steam segments instead of the ads. And it would be silly of them if they had implemented a way around that system. You'd need a magic request parameter, just like the workaround from a couple of days ago.

dojima:
> Nobody has found a way yet to trick Twitch into listing the live steam segments instead of the ads. And it would be silly of them if they had implemented a way around that system. You'd need a _magic_ request parameter, just like the workaround from a couple of days ago.

I wanted to test it more thoroughly before I mentioned it, but a few days ago when they were vigorously testing direct ad injection, I did notice that I didn't receive any ads on the official Twitch app on iOS when using an adblocking DNS.  This leads me to believe that they were still using the old system of displaying ads for some platforms.  It's not likely to matter much if their testing was a success, since they'd eventually move over to direct injection for all platforms, but it's interesting nonetheless.

beardypig:
@bastimeyer, yeah not so simple as I hoped. Your approach sounds much better, right now I don’t have time to review it - I will on Monday/Tuesday, it looks quite substantial. I’d like to get @back-to to take a look too.  

Create a PR, and we’ll move discussion over there as this PR is dead :)

Issue #2364, opened 2019-03-18
ok.ru - does not work!


dmitrylc:
![image](https://user-images.githubusercontent.com/2939460/54556728-9398d900-49ca-11e9-93cb-a55856e07882.png)

ok.ru - does not work!

alex-zed:
[cli][debug] OS:         Linux-4.15.0-46-generic-x86_64-with-Ubuntu-18.04-bionic
[cli][debug] Python:     2.7.15rc1
[cli][debug] Streamlink: 1.0.0+46.g42105d8
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.55.0)
[cli][info] Found matching plugin ok_live for URL ok.ru/live/1149468679758
[utils.l10n][debug] Language code: ru_RU
error: Unable to open URL: https://vsd24.mycdn.me/hls/542637296206.m3u8/sig/RxGJVgZoWNQ/expires/1553187068279/srcIp/x.x.x.x/clientType/0/srcAg/CHROME/video.m3u8?p (403 Client Error: Forbidden for url: https://vsd24.mycdn.me/hls/542637296206.m3u8/sig/RxGJVgZoWNQ/expires/1553187068279/srcIp/x.x.x.x/clientType/0/srcAg/CHROME/video.m3u8?p)

Bluebuster116:
@dmitrylc I'm also having this problem. Tried to update to the nightly build but still the problem persists. So far the plug in has not been updated. I'm still unsure what might be happening.

c769214:
@dmitrylc I have the same issue, waiting for any updates for this plugin

PR #2374, opened 2019-03-24
plugins.okru: Fixed Plugin

back-to:
- renamed plugin from `ok_live` to `okru`
- use a better version of this plugin
- added support for `HTTP` + `RTMP`

Fixed https://github.com/streamlink/streamlink/issues/2364


karlo2105:
@back-to 
Thanks for fix. ;-)

Bluebuster116:
@back-to Installation instructions for this update?. Just deleting the old file in plugins directory and replacing it with the current? I did just that and it works. The file which I deleted was ok_live.py and replaced it with okru.py I think this change should be included in the nigthly build.

beardypig:
@Bluebuster116 it own't be in the nightly build, until it's merged in to master. 

@back-to, can you rebase on to master and push it again - I tried to resolve the conflict, but GitHub make terrible commit messages :D

Bluebuster116:
@beardypig Okay. But how do I know that its been merged in the master?. Sorry I don't know where to follow those updates.
@back-to I had confused a bit in the beginning for why there were two different filenames for ok.ru plugin?. ok_live.py and okru.py wouldn't be much better to just update the first one?. Anyways I'm hoping this gets merged in the master as it is difficult to catch up with changes every time issues arisen in the log.

beardypig:
@Bluebuster116 once this PR is merged it will be closed as "merged", it will say at the bottom of the feed. Something like this:
merged

PR #2382, opened 2019-03-28
plugins.channel_one: new Plugin for 1tv.ru

vstavrinov:
Channel One is russian tv channel: 

https://en.wikipedia.org/wiki/Channel_One_Russia

1tv.ru is it's official site providing live stream.


vstavrinov:
stream example:

https://www.1tv.ru/live

back-to:
https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugins/onetv.py

vstavrinov:
Sorry for missing it.

Issue #2379, opened 2019-03-27
RTP Play 'Directo' stream not working with Streamlink


sebastiandoe5:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description

On the RTP Play live stream website, when I try to add any of the RTP channels to Streamlink it says that it can find the plugin but it can't find any actual streams. On the actual website there is a working stream, so I'm not sure what's happening here - maybe they changed the way their streaming works.

### Expected / Actual behavior

I expected it to open the stream in VLC (my default player) as usual, but it displayed the following when it was open in the Windows 10 CMD version:

```
C:\Users\username>streamlink https://www.rtp.pt/play/direto/rtp1 best
[cli][info] Found matching plugin rtpplay for URL https://www.rtp.pt/play/direto/rtp1
error: No playable streams found on this URL: https://www.rtp.pt/play/direto/rtp1
```

### Reproduction steps / Explicit stream URLs to test

1. In the Windows 10 CLI (I'm not sure about other operating systems but I'd assume they'd be the same), if you type in `streamlink https://www.rtp.pt/play/direto/rtp1 best` it should come up with the error.

### Log output

```
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.6.6
[cli][debug] Streamlink: 1.0.0
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.54.0)
[cli][info] Found matching plugin rtpplay for URL https://www.rtp.pt/play/direto/rtp1
error: No playable streams found on this URL: https://www.rtp.pt/play/direto/rtp1
```

### Additional comments, screenshots, etc.

Nothing else, but I would just like to say that I love what you are doing here and I find it very useful. Thanks!

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

PR #2386, opened 2019-03-30
plugins.rtpplay: fix _m3u8_re and plugin cleanup

back-to:
closes https://github.com/streamlink/streamlink/issues/2379


gravyboat:
Looks good, thanks @back-to.

PR #2387, opened 2019-03-30
plugins.twitch: cleanup old b and c vods

back-to:


src/streamlink/plugins/twitch.py
```diff
@@ -42,19 +34,19 @@
...
-    twitch.tv/
+    twitch\.tv/
     (?:
         videos/(?P<videos_id>\d+)|
         (?P<channel>[^/]+)
     )
     (?:
         /
-        (?P<video_type>[bcv])(?:ideo)?
+        v(?:ideo)?
```
bastimeyer: Do these URLs even exist anymore? I'm getting a 404. The forward-slash above is also wrong here. I think the `video_id` group can be removed in favor of `videos_id`, although not 100% sure.

back-to: most of it is gone,

`https://www.twitch.tv/twitch/v/292713971` just redirects to
`https://www.twitch.tv/videos/292713971`

back-to:
> like for example the Requests 2.0 prepare_request method check from the UserService

this is also used in `http.py`, not sure if this is still needed or if it can be removed

https://github.com/streamlink/streamlink/blob/87a55fe4a026c5673f4135ddb6a9f03cd9a1c5b4/src/streamlink/stream/http.py#L48-L52

---

> cookie PluginArgument

well I don't know if this works or not,
but if you want it dropped it could be removed

src/streamlink/plugins/twitch.py
bastimeyer: The first URL doesn't get matched by this group

bastimeyer:
The requests 2.0 dependency was bumped to `>=2` in this commit more than 2 years ago:
https://github.com/streamlink/streamlink/commit/334f763e72b7bcb3049bee182ec2341018cd403d#diff-2eeaed663bd0d25b7e608891384b7298L22

----

The only legitimate way for 3rd party applications of Twitch to let you sign into your account is using the OAuth flow. Everything else is just considered a hack. I'm not even sure if the cookie method works anymore. We should not support this kind of feature.

src/streamlink/plugins/twitch.py
beardypig: It doesn't need to if it redirects? It might break some old URLs people have around if they have the redirect option disabled...

PR #2388, opened 2019-03-31
[plugins.playtv] Fix for new stream data API

### LINKED ISSUE #1905: Playtv uk
playtv.fr working fine is there any possibility to run http://uk.play.tv/    looks like both sites france and uk owned by same  company. As streamlink support tvplay.fr. hope u get support for uk playtv.
http://uk.play.tv/live-tv/363/pick-tv/
http://uk.play.tv/live-tv/752/itv4-1/
http://uk.play.tv/live-tv/1106/itv3-1/
http://uk.play.tv/live-tv/1105/itv-1/

melmorabity:
This PR fixes the playtv plugin:
```
[cli][info] Found matching plugin playtv for URL http://playtv.fr/television/arte/
error: Unable to parse JSON: Expecting value: line 1 column 1 (char 0) ('eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1Ni ...)
```
Stream data API now returns a JWT-encoded URL. As a result, JWT support for Streamlink is required. This PR also add a new dependency [on the PyJWT library](https://pypi.org/project/pyjwt/).

This PR also:
* fixes #1905
* brings HTTPS support

### DIFF
--- a/src/streamlink/plugins/playtv.py
+++ b/src/streamlink/plugins/playtv.py
@@ -2,7 +2,6 @@ import base64
 import json
 import re
 
-from streamlink.exceptions import PluginError
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
 from streamlink.stream import HDSStream, HLSStream
--- a/src/streamlink/plugins/playtv.py
+++ b/src/streamlink/plugins/playtv.py
@@ -1,5 +1,7 @@
 import re
 
+import jwt
+
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
 from streamlink.stream import HDSStream, HLSStream
@@ -27,9 +29,12 @@ class PlayTV(Plugin):
             }
         )
     })
-    _api_schema = validate.Schema({
-        'url': validate.url()
-    })
+    _api_schema = validate.Schema(
+        validate.transform(lambda x: jwt.decode(x, algorithms=['HS256'], verify=False)),
+        {
+            'url': validate.url()
+        }
+    )
 
     @classmethod
     def can_handle_url(cls, url):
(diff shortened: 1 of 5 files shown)


beardypig:
Just to note that there probably need to be some changes for the windows installer too :)

src/streamlink/plugins/playtv.py
```diff
@@ -1,15 +1,17 @@
 import re
 
+import jwt
```
back-to: should probably be optional and not forced for every user.

bastimeyer:
Is there a way to install all optional dependencies via pip instead of listing them explicitly?

You currently have listed the plugin's optional dependency under a plugin-specific name in the `extras_require` field. If more plugins add optional dependencies in the future, then we have to explicitly list them as well in the `pip install` command(s). I don't think it's a good idea.

Btw, the optional dependencies also don't get installed in the CI environments and the only reason why it's working is that the changes are not being tested at the moment.

melmorabity:
> You currently have listed the plugin's optional dependency under a plugin-specific name in the `extras_require` field. If more plugins add optional dependencies in the future, then we have to explicitly list them as well in the `pip install` command(s). I don't think it's a good idea.
Maybe a single "generic" extras_require target (named "extras" or "optional" e.g.) may be appropriate for all plugins with optional deps, instead of the plugin's name.
BTW pyjwt is not an optional dependency for playtv. The plugin won't work without it. It requires a whole library just for it to work, which is a problem according to @back-to.

> Is there a way to install all optional dependencies via pip instead of listing them explicitly?
I don't think so, unfortunately. There's nothing in setuptools doc about it.

beardypig:
If you're not verifying the signature, and it's only for this website probably you can just roll your own method to decode it?

melmorabity:
> If you're not verifying the signature, and it's only for this website, probably you can just roll your own method to decode it?
> 
> Update: this should do it
> 
> ```python
> def jwt_decode(token):
>     info, payload, sig = token.split(".")
>     data = base64.urlsafe_b64decode(payload + '=' * (-len(payload) % 4))
>     return json.loads(data)
> ```

Sorry, I haven't seen your update before pushing my fixes. And I better like your implementation. Will use it instead

beardypig:
Restarting travis build for Python 3.7, failure was due to network error

PR #2393, opened 2019-03-31
Update supported wheels so the job doesn't bomb

### LINKED ISSUE #2392: 1.1.0: Windows specific wheels missing on PyPI (streamlinkw executable)
The PyPI deploy job of the 1.1.0 release has partially failed:
https://travis-ci.org/streamlink/streamlink/jobs/513828143#L3637
https://pypi.org/project/streamlink/#files

This means that there are no `win32` and `win-amd64` wheels on PyPI for the 1.1.0 release which include the `streamlinkw` executable (#2326).

The problem is the `cygwin` wheel, which was not accepted by PyPI. Fortunately, the `any` wheel (all platforms) was uploaded first, so the damage of the release failure is minimal. We should fix this in a `1.1.1` release soon, though.

The cygwin wheel can be removed from the next build, but now I'm not sure whether the other ones will be rejected on the next release as well, because there doesn't seem to be a list of platform names accepted by PyPI.

https://github.com/streamlink/streamlink/pull/2326#issuecomment-467825380

https://www.python.org/dev/peps/pep-0425/#platform-tag
https://www.python.org/dev/peps/pep-0425/#id1
https://www.python.org/dev/peps/pep-0427/#file-name-convention

----

The Windows specific wheels can be built and installed locally like this:
```bash
git clone https://github.com/streamlink/streamlink.git
cd streamlink
pip install -U setuptools wheel
pip install -U -r dev-requirements.txt
pip install pycountry
python setup.py bdist_wheel --plat-name=win32
pip install -U ./dist/streamlink-1.1.0-py2.py3-none-win32.whl
```

gravyboat:
fixes #2392.

### DIFF
--- a/script/sdistsign.sh
+++ b/script/sdistsign.sh
@@ -9,7 +9,7 @@ version=$(python setup.py --version)
 dist_dir=${STREAMLINK_DIST_DIR:-dist}
 temp_keyring=$(mktemp -d) && trap "rm -rf ${temp_keyring}" EXIT || exit 255
 
-wheel_platforms_windows=("win32" "win-amd64" "cygwin")
+wheel_platforms_windows=("win32" "win_amd64")
 
 if [[ -n "${TRAVIS}" ]]; then
       openssl aes-256-cbc -K ${encrypted_eeb8b970d3a3_key} -iv ${encrypted_eeb8b970d3a3_iv} -in signing.key.enc -out "${SDIST_KEY_FILE}" -d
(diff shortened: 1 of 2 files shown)


bastimeyer:
> and we need to change the format from win-amd64 to win_amd64

https://github.com/streamlink/streamlink/issues/2392#issuecomment-478392876

The dash is correct in `win-amd64`. It gets changed to an underscore in the final wheel file name:
https://www.python.org/dev/peps/pep-0425/#platform-tag

> The platform tag is simply distutils.util.get_platform() with all hyphens - and periods . replaced with underscore _.

I guess cygwin also doesn't need to be removed from setup.py. Then people can still build their own wheel for cygwin if they want.

PR #2401, opened 2019-04-03
plugins.mediaklikk: update plugin

kviktor:
Things were changed on the site, all the current livestreams work now.

### DIFF
--- a/src/streamlink/plugins/mediaklikk.py
+++ b/src/streamlink/plugins/mediaklikk.py
@@ -1,44 +1,45 @@
+import logging
 import re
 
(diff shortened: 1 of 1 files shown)


hreffero:
There is a working VLC lua script for Mediaklikk, maybe there is a way to fix it?

https://github.com/tiborbaksa/vlc-mediaklikk-video/blob/master/mediaklikk-video.lua

kviktor:
I think this PR haven't been released yet, no idea when it will but it should work (at least for streams).

image

hreffero:
My bad it actually works but my plugin was somehow older... Sorry for confusion

PR #2408, opened 2019-04-07
plugins.foxnews: new plugin for foxnews.com

vstavrinov:
USA TV channel official site foxnews.com.

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -84,6 +84,7 @@ euronews                euronews.com         Yes   No
 europaplus              europaplus.ru        Yes   --
 facebook                facebook.com         Yes   Yes
 filmon                  filmon.com           Yes   Yes   Only SD quality streams.
+foxnews                 foxnews.com          Yes   No
 foxtr                   fox.com.tr           Yes   No
 funimationnow           - funimation.com     --    Yes   :ref:`Requires session cookies <cli-funimationnow>`
                         - funimationnow.uk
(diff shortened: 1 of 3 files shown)


back-to:

> 4. Sites which require any sort of cable login or subscription

you are trying to bypass the login for a pay tv channel.

vstavrinov:
> you are trying to bypass the login for a pay tv channel.

Not really exact this way. It is free for limited amount of time: that is 10 minutes on the site and 3 hours with direct stream available even without streamlink. And other common true which is worth saying, that any site have some restrictions either explicit or implicit that streamlink bypasses. And exact list of categories  You provide doesn't changes this truth.  Motivation may differ, but often it is mere ads and almost always it is about money. Otherwise they would provide direct link to the stream or simply playlist, so streamlink would not be needed at all, but they don't do this. So streamlink always bypasses one or other restrictions  doing something unwanted by site owners. And common practice is that it is tolerant al long as they don't complain.

vstavrinov:
If You wish You could add password authenticating to this plugin, but in this case You should add correspondent authenticating code for every of 564 service providers listed on this site. You will just be redirected to one of those 564 providers. The site itself has no password authentication and provide temporary stream that is not password protected. It uses  complex  chain of cryptographic requests originating from browser fingerprint. But this is not authentication. I mere replaced this fingerprint with random number.

gravyboat:
I'm going to lock this as there's nothing left to discuss. We're not going to bend the rules for new plugins. It opens up too much potential for "but you did it for xyz plugin even after the rules were implemented!" which is enough of a headache as is on top of the maintenance work that is already being done.

PR #2415, opened 2019-04-10
Added Canal Sur plugin.

dgoiko:
Added Canal Sur plugin.

No integration testing, I dont know how to integrate it, I'm using it from %APPDATA%streamlink/plugins. If somoene could comment how the integration is done it would be great for future plugins.

I added the test build, but I dunno if I have to list the plugin anywhere else.

I recommend squeezing commits, as they're pretty useless separated.


dgoiko:
I'm new integrting plugins to this. I'm solving problems ad build proccesses fail, just added canalsur to plugin matrix, but I dunno if adding it to the rst is enough, but it looks like it is passing tests now.

back-to:
Not sure why you would want a new plugin here,
just use the youtube channel.

`streamlink "https://www.youtube.com/channel/UCqwjtgaJflHga2aPWphG5PQ"`

dgoiko:
> 
> 
> Not sure why you would want a new plugin here,
> just use the youtube channel.
> 
> `streamlink "https://www.youtube.com/channel/UCqwjtgaJflHga2aPWphG5PQ"`

Because youtube url changes everty couple hours, so you would need to update your site 12 times every day, while the covered url by the plugin changes only once in a year or so.

Maybe I should've explained it.

bastimeyer:
> https://www.youtube.com/channel/UCqwjtgaJflHga2aPWphG5PQ

This is a canonical channel URL on YouTube with a unique and static channel ID. You are confusing this with live stream IDs, which are unique for each broadcast and will therefore change.

Channels can only change their custom URL names, which are not to be confused with IDs.
https://support.google.com/youtube/answer/2657968?hl=en

dgoiko:
> 
> 
> > https://www.youtube.com/channel/UCqwjtgaJflHga2aPWphG5PQ
> 
> This is a canonical channel URL on YouTube with a unique and static channel ID. You are confusing this with live stream IDs, which are unique for each broadcast and will therefore change.
> 
> Channels can only change their custom URL names, which are not to be confused with IDs.
> https://support.google.com/youtube/answer/2657968?hl=en

Thanks for the clarification, but all I know is that using that url does not allways work, while using the plugin did not give problems so far. Maybe Canal Sur is doing things the wrong way and have multiple channels, I dont know.

back-to:
My presented URL still works, a new plugin for this is currently not needed.

dgoiko:
> 
> 
> My presented URL still works, a new plugin for this is currently not needed.

As I said, it failed 10 times in the last 11 days using the channel url, while no failures using CanalSur website. 

I removed the YouTube link and added the plugin in my application in order to be able to work with Canal Sur. If you consider it unnecesary, it is your program, not mine.

Maybe Canal Sur uses more than one channel? All I know is that I get around 2 hours more of video per day using my plugin than not using it. (Configured both as different channels just to check if you're right)

Anyway, thanks for your replies and suggestions.

PR #2419, opened 2019-04-12
Plugin: DLive

skulblakka:
This adds support for livestreams on DLive.

I know the request data for the livestream info is quite ridiculous but I didn't really find a cleaner way to get the name for the HLS variant playlist and I wasn't able to change the request data without the server rejecting it.


src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,390 @@
+"""
+Plugin for dlive.tv
+"""
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import validate
+from streamlink.stream import HLSStream
+from streamlink.utils import parse_json
+from streamlink import NoStreamsError
```
back-to: not needed

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+QUALITY_WEIGHTS = {
+   "src": 1080,
+}
+
+
+class dlive(Plugin):
```
back-to: use `Dlive` or `DLive`

tests/plugins/test_dlive.py
```diff
@@ -0,0 +1,9 @@
+import unittest
+
+from streamlink.plugins.dlive import dlive
```
back-to: `Dlive` or `DLive`

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,390 @@
...
+    def stream_weight(cls, key):
+        weight = QUALITY_WEIGHTS.get(key)
+        if weight:
+            return weight, "dlive"
+
+        return Plugin.stream_weight(key)
+
+    def _get_streams(self):
+        reqUrl = "https://graphigo.prd.dlive.tv/"
+        displayname = self.url[self.url.rfind('/') + 1:]
```
back-to: use `_url_re` for this

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,390 @@
...
+    @classmethod
+    def stream_weight(cls, key):
+        weight = QUALITY_WEIGHTS.get(key)
+        if weight:
+            return weight, "dlive"
+
+        return Plugin.stream_weight(key)
+
+    def _get_streams(self):
+        reqUrl = "https://graphigo.prd.dlive.tv/"
```
back-to: put this url into the class so you can use it with `self.API_URL` or something similar

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,390 @@
...
+                    }
+                __typename
+                }
+                """
+        }
+        res = self.session.http.post(reqUrl, json=jsonData)
+
+        jsonres = self.session.http.json(res, schema=self._INFO_SCHEMA)
+
+        hls_url = "https://live.prd.dlive.tv/hls/live/" \
```
back-to: use `.format` not ` + `

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,390 @@
...
+            + jsonres["data"]["userByDisplayName"]["username"] + ".m3u8"
+
+        try:
+            return HLSStream.parse_variant_playlist(self.session, hls_url)
+        except Exception:
```
back-to: don't use an empty `Exception` or an Exception at all here

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,390 @@
...
+        weight = QUALITY_WEIGHTS.get(key)
+        if weight:
+            return weight, "dlive"
+
+        return Plugin.stream_weight(key)
+
+    def _get_streams(self):
+        reqUrl = "https://graphigo.prd.dlive.tv/"
+        displayname = self.url[self.url.rfind('/') + 1:]
+        jsonData = {
```
back-to: this seems like it could break very fast,
is there no better way to get the username?

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,390 @@
...
+
+        jsonres = self.session.http.json(res, schema=self._INFO_SCHEMA)
+
+        hls_url = "https://live.prd.dlive.tv/hls/live/" \
+            + jsonres["data"]["userByDisplayName"]["username"] + ".m3u8"
+
+        try:
+            return HLSStream.parse_variant_playlist(self.session, hls_url)
+        except Exception:
+            raise NoStreamsError(self.url)
```
back-to: don't use `NoStreamsError` here

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,390 @@
+"""
+Plugin for dlive.tv
+"""
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import validate
+from streamlink.stream import HLSStream
+from streamlink.utils import parse_json
```
back-to: I think `parse_json` is not used, can be removed.

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,390 @@
+"""
+Plugin for dlive.tv
```
back-to: put the comment into the Plugin class or don't use it at all

tests/plugins/test_dlive.py
```diff
@@ -0,0 +1,9 @@
...
+
+
+class TestPluginDlive(unittest.TestCase):
+    def test_can_handle_url(self):
+        # should match
+        self.assertTrue(dlive.can_handle_url("https://dlive.tv/pewdiepie"))
```
back-to: add an `assertFalse` testcase

skulblakka:
Thanks for the review. I did some of the changes and added VOD support.

src/streamlink/plugins/dlive.py
skulblakka: The problem I had was that the user will see an exception if the server returns an error code if the playlist does not exist because the livestream is offline. I copied this one from another plugin... What would be the better solution here?

back-to: > What would be the better solution here?

getting the online/offline status from `__APOLLO_STATE__`

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,56 @@
...
+        weight = QUALITY_WEIGHTS.get(key)
+        if weight:
+            return weight, "dlive"
+
+        return Plugin.stream_weight(key)
+
+    def _get_streams(self):
+        res = self.session.http.get(self.url)
+
+        playback_url = self._playback_re.search(res.text)
```
back-to: 1. use `m =` or `match =` for `_re.search`, so it is similar to other plugins.

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,56 @@
...
+
+        if playback_url is not None:
+            hls_url = playback_url.group(0)
+            hls_url = bytes(unquote_plus(hls_url), "utf-8").decode(
+                "unicode_escape")
+        else:
+            if self._livestream_re.search(res.text) is not None:
+                return None
+
+            username = self._username_re.search(res.text)
```
back-to: see 1.

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,56 @@
...
+
+        playback_url = self._playback_re.search(res.text)
+
+        if playback_url is not None:
+            hls_url = playback_url.group(0)
+            hls_url = bytes(unquote_plus(hls_url), "utf-8").decode(
+                "unicode_escape")
+        else:
+            if self._livestream_re.search(res.text) is not None:
+                return None
```
back-to: add some log or PluginError

just `return` should be enough

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,56 @@
...
+        return Plugin.stream_weight(key)
+
+    def _get_streams(self):
+        res = self.session.http.get(self.url)
+
+        playback_url = self._playback_re.search(res.text)
+
+        if playback_url is not None:
+            hls_url = playback_url.group(0)
+            hls_url = bytes(unquote_plus(hls_url), "utf-8").decode(
```
back-to: does this work correctly with py2 and py3 ?

the facebook addon has something else there

https://github.com/streamlink/streamlink/blob/499301dd2720d9022c051c6004c3efebf217fc12/src/streamlink/plugins/facebook.py#L49-L52

src/streamlink/plugins/dlive.py
```diff
@@ -0,0 +1,56 @@
...
+                "unicode_escape")
+        else:
+            if self._livestream_re.search(res.text) is not None:
+                return None
+
+            username = self._username_re.search(res.text)
+            if username is not None:
```
back-to:
```py
if not m:
    log.error('...')
    return
```

or 

```py
if not m:
    raise PluginError('....')
```

src/streamlink/plugins/dlive.py
skulblakka: I updated and tested it with python2.7 and 3.7. However using ```string_escape``` for the py2 solution did not work. Had to use ```unicode_escape``` for both.

gravyboat:
Does this site actually work? When I went to it I just got a 502 error saying the lambda function couldn't satisfy the request:

dlive_error

skulblakka:
The site worked fine when I developed the plugin and while I did see the Error a few minutes ago it seems to work now. I guess they just had some server issues...

Issue #2353, opened 2019-03-11
Add option to force progress display


tonycpsu:
## Feature Request

- [x] This is a feature request and I have read the contribution guidelines.

### Description

Per #1770, `streamlink` will currently only output progress messages if stdout is a tty:

https://github.com/streamlink/streamlink/blob/71d550fedf7048288435b564c4c1b49373911f6c/src/streamlink_cli/main.py#L341

This is normally what a user wants, but I'm developing an application that runs `streamlink` as a subprocess, and I'd like to capture the progress messages to display the transfer rate in the app.  However, isatty() fails on pipes.  I can work around this by using `pty.openpty()`, but this causes the subprocesses to terminate when the app exits, which isn't what I want.

My proposed solution would be to add an option, e.g. `force-progress`, which would skip the isatty check if the option is supplied.  This way there is no change for users who count on the current behavior, but those who would like to monitor the progress on a pipe can do so.

I will submit a PR for this change if so desired.

### Expected / Actual behavior

Progress display can be forced even when stdout is not a TTY.

### Additional comments, screenshots, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

Quaraxkad:
I second this request.

beardypig:
@tonycpsu can you give an example of how you are invoking streamlink so that we can test it?

tonycpsu:
@beardypig, sure:

```python
import io
import asyncio

async def get_progress(proc):
    while True:
        line = await proc.stdout.readline()
        if line == b'':
            break

        print(line)

async def run(with_pty=False):

    proc = await asyncio.create_subprocess_exec(
        "streamlink", "https://www.youtube.com/watch?v=__dFiuEpW6c",
        "720p",
        "-o", "foo.mp4",
        stdout=asyncio.subprocess.PIPE
    )
    import os
    await asyncio.gather(get_progress(proc))


loop = asyncio.get_event_loop()
loop.run_until_complete(run())
```

back-to:
closing, it seems like it's too specific and not worth it.

rseanhall:
I don't understand why you think it's too specific. I think it's fairly common to write an application that spawns apps like streamlink as a subprocess. Is there some other way for those wrapper applications to get download progress when saving to a file?

fatalis:
[...]

PR #2438, opened 2019-04-20
cli: New command --force-progress

back-to:
show the download progress bar even if there is no terminal

closes https://github.com/streamlink/streamlink/issues/2353


gravyboat:
@back-to Looks good, thanks.

Issue #2416, opened 2019-04-10
Teamliquid's new domain tl.net usage broke the corresponding plugin


ptkovsky:
There's a major issue now where teamliquid's new domain tl.net usage has completely broken the corresponding plugin:
```
streamlink https://tl.net/video/streams/ioioiobb
error: No plugin can handle URL: https://tl.net/video/streams/ioioiobb
```

bastimeyer:
> major issue

This is just a simple regex change, unless they've changed their site's layout.
https://github.com/streamlink/streamlink/blob/1.1.1/src/streamlink/plugins/teamliquid.py#L6

The plugin also still uses the deprecated logger calls and needs to get updated.

----

Btw, what is the policy on these kinds of wrapper/proxy/alias plugins? I can remember that we had a discussion about this a few months ago, but I can't find the thread. Doesn't this technically violate the third point here?
https://github.com/streamlink/streamlink/blob/1.1.1/CONTRIBUTING.md#plugin-requests

> 3. Sites which are primarily rehosting content that is available from a legitimate source

My personal opinion is that these plugins are rather pointless, but at the same time they've existed for a long time now and there are people who are using them.

gravyboat:
@bastimeyer From what I remember the only reason we kept any of the teamliquid stuff was basically because it was grandfathered in. One of us (probably me since I hate the duplicate bloat) brought up getting rid of these specific types of plugins during some discussion and we came the the conclusion we would allow old plugins to stay, but wouldn't permit future restream plugins like this to exist.

If we ever want to bring down the hammer I'll be happy to put teamliquid and vaughnlive as number one on the chopping block.

Issue #2407, opened 2019-04-06
Teamliquid plugin best quality detector has an issue, Afreeca behaves properly under the same conditions


ptkovsky:
## Plugin Issue

Not sure if this is a plugin issue or not. Maybe I misunderstood something here. But guess it is.

### Description

Just wanted to check the stream real quick but was too lazy to copy the original afreeca stream link (in fact, this is why the appropriate plugin exists) so tried copying the teamliquid's one instead. It opened up but looked kind of wrong in terms of the quality. Later, tested the teamliquid plugin a little bit more, and noticed that it kept behaving kind of strange. It kept opening 540p stream as the `best` option, while 1080p 60fps was still available. My guess is that it happened because of the urls being embedded.

### Reproduction steps / Explicit stream URLs to test

1. try to open a stream with the specified commands in the log output, for example: `streamlink https://www.teamliquid.net/video/streams/byflash best`
2. observe the actual results

### Log output

Here is some of the actual output of `teamliquid` plugin that I was getting:

```
streamlink https://www.teamliquid.net/video/streams/byflash best
[cli][info] Found matching plugin teamliquid for URL https://www.teamliquid.net/video/streams/byflash
[plugin.teamliquid][info] Attempting to play streams from https://play.afreecatv.com/byflash
[cli][info] Available streams: original, sd (worst), hd (best)
[cli][info] Opening stream: hd (hls)
```

And here's the same stream output when playing it from the direct link, using `afreeca` plugin:

```
streamlink https://play.afreecatv.com/byflash best
[cli][info] Found matching plugin afreeca for URL https://play.afreecatv.com/byflash
[cli][info] Available streams: sd (worst), hd, original (best)
[cli][info] Opening stream: original (hls)
```

### Additional comments, screenshots, etc.
Environment: guess it's irrelevant here and not platform dependent either, but this case was tested using Windows 7 Ultimate 32 bit, and the Streamlink version is current, 1.1.1.

The expected result: both commands should equally open the `best`, highest quality stream, in the case, playing at 1080p 60fps.

The actual result: when trying to play video using teamliquid url and invoking `teamliquid` plugin, the `best` video is falsely detected as `hd`, which, in the case, appears to be a 540p stream at 31fps.

`afreeca` plugin, though, behaves adequately, as it detects the `best` output as the `original` one, and the correct 1080p 60fps video starts playing.

jacobmalmberg:
[...]

PR #2442, opened 2019-04-21
plugins.teamliquid: New domain, fix stream_weight

back-to:
closes https://github.com/streamlink/streamlink/issues/2407
closes https://github.com/streamlink/streamlink/issues/2416

---

not sure if there is a better way for fixing the `stream_weight`,
but this would work for now.

---

Plugin could also be removed, as the website is just reusing other Streams.


gravyboat:
Sorry for not getting this merged @back-to, I must have missed it.

PR #2453, opened 2019-04-30
Note that Totem requires --player-continuous-http

mavit:


back-to:
it also needs `gstreamer1.0-plugins-bad`

mavit:
Sorry, what response was outstanding?

PR #2466, opened 2019-05-14
Update lrt plugin to cover updated page

justasmalinauskas:
Hello, this piece of edit should resolve  streamlink#2440
Btw first commit to other's projects.

P.S streamlink should also cover VOD part or not?


back-to:
> [plugin.lrt][debug] No match for sources regex

does not fix it

---

using the video_id is probably a better way to fix this

PR #2471, opened 2019-05-20
plugins.dmax: New Plugin for dmax.de

TheBinaryLoop:


back-to:
VOD only plugin, probably not needed for Streamlink

gravyboat:
Since this is VODs only I'm going to close this. Thank you for your PR, unfortunately we cannot accept it due to our [plugin guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#plugin-requests).

PR #2478, opened 2019-05-26
Add a sponsor button

gravyboat:
This adds a sponsorship link from GitHub's newly added features.


gravyboat:
@bastimeyer Any concerns over this? I couldn't think of any potential issues since we already link here from elsewhere, but wanted to bring it up since I know you're very conscious of these things which I appreciate.

bastimeyer:
I wanted to post a PR with the funding config a few days ago, but wasn't sure whether to include a `custom` link to https://streamlink.github.io/donate.html or not, as it also lists individual members of streamlink. The current implementation on Github is a bit weird with "long" custom URLs though, so it's probably fine to leave it out for now, well, if it's even needed/desired to be included here in the future. Individual people should not be listed directly in the `github` field (which requires whitelisted accounts of their sponsor beta program). 

Regarding the changes to the config file, please add the missing line break at the end of the file and this PR should be good :)

gravyboat:
@bastimeyer Bah that's what I get for trying to do a quick edit on my Windows machine! Newline fixed. In terms of the custom URL I thought about including it then decided against it to keep confusion to a minimum. If people want they can enable the sponsorship link on their profile then update the associated page.

PR #2510, opened 2019-06-25
plugins.welt: fix plugin

bastimeyer:
Fixes the plugin after the recent site layout changes. Live TV stream and VODs are now working again.


gravyboat:

Issue #2402, opened 2019-04-05
plugins.twitch: Low Latency Test Plugin


Chibicon:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

Stream doesn't start if the low latency plugin is sideloaded.

Twitch low latency test plugin:
https://github.com/streamlink/streamlink/issues/1676#issuecomment-430304591

### Reproduction steps / Explicit stream URLs to test

1. Launch any stream with the sideloaded [twitch.py](https://gist.githubusercontent.com/back-to/8e9ed3c60e5932d8c7a67ccd43b906d0/raw/e8b4466ace87d59aa43799f2ee14fd7523ee031f/twitch.py) in the plugins folder.

### Log output

```
[session][debug] Plugin twitch is being overridden by C:\Users\User\AppData\Roaming\streamlink\plugins\twitch.py
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.6.6
[cli][debug] Streamlink: 1.1.1
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.56.0)
[cli][info] Found matching plugin twitch for URL twitch.tv/loltyler1
[plugin.twitch][warning] Test Plugin - Low Latency
[plugin.twitch][debug] Getting live HLS streams for loltyler1
[utils.l10n][debug] Language code: en_US
Traceback (most recent call last):
  File "runpy.py", line 193, in _run_module_as_main
  File "runpy.py", line 85, in _run_code
  File "C:\Program Files (x86)\Streamlink\bin\streamlink.exe\__main__.py", line 18, in <module>
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 1033, in main
    handle_url()
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 577, in handle_url
    streams = fetch_streams(plugin)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink_cli\main.py", line 457, in fetch_streams
    sorting_excludes=args.stream_sorting_excludes)
  File "C:\Program Files (x86)\Streamlink\pkgs\streamlink\plugin\plugin.py", line 317, in streams
    ostreams = self._get_streams()
  File "C:\Users\User\AppData\Roaming\streamlink\plugins\twitch.py", line 960, in _get_streams
    return self._get_hls_streams("live")
[...]

Issue #1676, opened 2018-05-21
Any way to have Twitch's reduced delay work on Streamlink?


KazukiMutou:
### Checklist

- [x] This is a feature request.
- [x] I have read the contribution guidelines.

### Description

Twitch now offers a "new" reduced delay option for broadcasters. But as of now, this only works when viewing on the website. Streamlink has an additional 5-8 second delay, regardless of what settings are used.

### Expected / Actual behavior

- 1-3 second stream delay for Twitch streams where the broadcaster has opted in for reduced delay.
- When viewing on the Twitch site, this works. However, with Streamlink it doesn't, even with hls-live-edge set to the minimal value.

### Reproduction steps / Explicit stream URLs to test

1. Open stream that has reduced delay on Twitch website.
2. Confirm it has reduced delay through the video stats.
3. Open same stream on Streamlink.
4. Streamlink stream has additional 5-8 second delay based on the settings picked.

### Comments, screenshots, etc.

I'm not sure if I'm just ill-informed and the ability is already there. I've tried changing a number of different settings from setting hls-live-edge to 1 and changing player-passthrough. None of the options I have tried reduce the delay from ~>7 seconds.

![reduceddelay](https://user-images.githubusercontent.com/7350221/40320292-74858d1a-5d2b-11e8-9606-3ce4f7f4445c.png)

beardypig:
If anyone knows of a broadcaster using the low latency option frequently, let me know and I'll check it out. If i'm online on Gitter and you see a stream that is running low latency ping me.

gravyboat:
@beardypig I checked a few streams this weekend and saw https://www.twitch.tv/lirik and https://www.twitch.tv/cohhcarnage using it.

KazukiMutou:
@beardypig @back-to https://www.twitch.tv/elajjaz is currently live for another 6-7 hours as of this message with low latency enabled.

beardypig:
Thanks @KazukiMutou, but it seems to be an HLS stream rather than a DASH stream. Maybe they just improved their infrastructure + player to provide lower latency... 

@bastimeyer where did you see they were using DASH?

There are a couple of improvements we could make to reduce the latency of the streams, we even tried before but it caused issues else where.

bastimeyer:
I was referring to https://github.com/streamlink/streamlink/pull/1637#issuecomment-389300549
I'm usually not using the Twitch website, but that is indeed a HLS stream. I should have checked it myself first, sorry about that.

[...]

PR #2513, opened 2019-06-27
plugins.twitch: low latency

### LINKED ISSUE #2180: HLS: New command to use stream=True from requests for segments
## Feature Request

### Description

Add an option to use `stream=True` for HLS Streams Stream

https://github.com/streamlink/streamlink/blob/cd6f94a5a4ee83cb5b61a295df9e13956bcfcb27/src/streamlink/stream/hls.py#L106-L110

Example `--hls-segment-stream-data` or another name
which will turn `stream=True` or `stream=False`

### **HLSStreams**

It was already added in https://github.com/streamlink/streamlink/pull/260
but removed in https://github.com/streamlink/streamlink/pull/522 because it had some issues.

it might be better with a command where you can turn it on and off

_there might be still some issues with it, but it can be ignored see below for an example_

basically required for Twitch low latency https://github.com/streamlink/streamlink/issues/1676
to get the same delay as the browser or better, you can't get the same time as the browser without _stream=True_

New command to use `stream=True` from requests for HLS segments as an optional option.

### Example Code

_ChunkedEncodingError_ might skip a segment but the latency should be more important for some livestreams than a missing frame,
also this issue is very rare.

this code is only used for normal hls streams

```diff
diff --git a/src/streamlink/stream/hls.py b/src/streamlink/stream/hls.py
index 8a67f564..fecc54f4 100644
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -4,6 +4,7 @@ import struct
 
 from collections import defaultdict, namedtuple
 from Crypto.Cipher import AES
+from requests.exceptions import ChunkedEncodingError
 
 from streamlink.exceptions import StreamError
 from streamlink.stream import hls_playlist
@@ -45,6 +46,8 @@ class HLSStreamWriter(SegmentedStreamWriter):
         self.byterange_offsets = defaultdict(int)
         self.key_data = None
         self.key_uri = None
+        self.stream_data = options.get("hls-segment-stream-data")
+
         if self.ignore_names:
             # creates a regex from a list of segment names,
             # this will be used to ignore segments.
@@ -104,6 +107,8 @@ class HLSStreamWriter(SegmentedStreamWriter):
                 return
 
             return self.session.http.get(sequence.segment.uri,
+                                         stream=(self.stream_data
+                                                 and not sequence.segment.key),
                                          timeout=self.timeout,
                                          exception=StreamError,
                                          retries=self.retries,
@@ -134,8 +139,12 @@ class HLSStreamWriter(SegmentedStreamWriter):
 
             self.reader.buffer.write(pkcs7_decode(decrypted_chunk))
         else:
-            for chunk in res.iter_content(chunk_size):
-                self.reader.buffer.write(chunk)
+            try:
+                for chunk in res.iter_content(chunk_size):
+                    self.reader.buffer.write(chunk)
+            except ChunkedEncodingError:
+                log.error("Download of segment {0} failed", sequence.num)
+                return
 
         log.debug("Download of segment {0} complete", sequence.num)
```

---

**EDIT**: Removed DASH, HLS is enough

bastimeyer:
Maybe this should be done in two separate pull requests, but here we go.
These changes are mainly based on the work of @back-to and his low latency test plugin for Twitch (see [here](https://github.com/streamlink/streamlink/issues/1676#issuecomment-430304591)).

**Description**
- Adds the `--hls-segment-stream-data` parameter, which makes Streamlink write the HLS segments to the output buffer while they are being downloaded.
- Adds HLS segment prefetching to the custom `TwitchM3U8Parser`.
- Lowers the playlist refresh time to the duration of the HLS segments (2 seconds).
- Adds the `--twitch-low-latency` shorthand plugin-parameter, which automatically sets `--hls-segment-stream-data` to true and limits the value of `--hls-live-edge` to a maximum of 2, as higher values (default is 3) are pointless for low latency streaming.

**Open questions**

- ~~Is reducing the playlist refresh time by this much necessary? It's also what Twitch's web player does nowadays for regular and low latency streams and it basically refreshes the playlist after each segment download. I'm not sure if this defeats the segment prefetching idea. It seems to be right though to ignore the `targetduration`. Maybe a refresh time of `number_of(prefetch_segments) * duration(segment)` would make sense...~~
- Regarding ad filtering and additional segment metadata, is it the right approach while prefetching segments to clone the last segment of the sequence and replace its URL?
- ~~Is introducing a shorthand parameter for setting other parameter values a good idea? I will have to reword the parameter description later, as segment prefetching is always being done.~~
- ~~Tests need to be added...~~

As for the results of the low latency changes, I'm able to beat Twitch's web player by 0-2 seconds while using MPV without additional player cache. Using a HLS live edge value of 1 sometimes results in micro-buffering though, but this can be fixed by pausing the playback for a short moment.

Resolves #2180, #1676, #2402

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -168,6 +168,13 @@ class TwitchM3U8Parser(M3U8Parser):
     def __init__(self, base_uri=None, stream=None, **kwargs):
         M3U8Parser.__init__(self, base_uri, **kwargs)
         self.stream = stream
+        self.has_prefetch_segments = False
+
+    def parse(self, *args):
+        m3u8 = super(TwitchM3U8Parser, self).parse(*args)
+        m3u8.has_prefetch_segments = self.has_prefetch_segments
+
+        return m3u8
 
     @parse_condition("disable_ads")
     def parse_tag_ext_x_scte35_out(self, value):
@@ -184,6 +191,7 @@ class TwitchM3U8Parser(M3U8Parser):
 
     @parse_condition("low_latency")
     def parse_tag_ext_x_twitch_prefetch(self, value):
+        self.has_prefetch_segments = True
         segments = self.m3u8.segments
         if segments:
             segments.append(segments[-1]._replace(uri=self.uri(value)))
@@ -219,6 +227,12 @@ class TwitchHLSStreamWorker(HLSStreamWorker):
             super(TwitchHLSStreamWorker, self)._set_playlist_reload_time(playlist, sequences)
         else:
             self.playlist_reload_time = sequences[-1].segment.duration
+
+    def process_sequences(self, playlist, sequences):
+        if self.playlist_sequence < 0 and self.stream.low_latency and not playlist.has_prefetch_segments:
+            log.info("This is not a low latency stream")
+
+        return super(TwitchHLSStreamWorker, self).process_sequences(playlist, sequences)
 
 
 class TwitchHLSStreamWriter(HLSStreamWriter):
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -244,8 +244,17 @@ class TwitchHLSStreamReader(HLSStreamReader):
 class TwitchHLSStream(HLSStream):
     def __init__(self, *args, **kwargs):
         HLSStream.__init__(self, *args, **kwargs)
-        self.disable_ads = self.session.get_plugin_option("twitch", "disable-ads")
-        self.low_latency = self.session.get_plugin_option("twitch", "low-latency")
+
+        disable_ads = self.session.get_plugin_option("twitch", "disable-ads")
+        low_latency = self.session.get_plugin_option("twitch", "low-latency")
+
+        if low_latency and disable_ads:
+            log.info("Low latency streaming with ad filtering is currently not supported")
+            self.session.set_plugin_option("twitch", "low-latency", False)
+            low_latency = False
+
+        self.disable_ads = disable_ads
+        self.low_latency = low_latency
 
     def open(self):
         if self.disable_ads:
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -152,17 +152,41 @@ _quality_options_schema = validate.Schema(
 
 Segment = namedtuple("Segment", "uri duration title key discontinuity scte35 byterange date map")
 
+LOW_LATENCY_MAX_LIVE_EDGE = 2
+
+
+def parse_condition(attr):
+    def wrapper(func):
+        def method(self, *args, **kwargs):
+            if hasattr(self.stream, attr) and getattr(self.stream, attr, False):
+                func(self, *args, **kwargs)
+        return method
+    return wrapper
+
 
 class TwitchM3U8Parser(M3U8Parser):
+    def __init__(self, base_uri=None, stream=None, **kwargs):
+        M3U8Parser.__init__(self, base_uri, **kwargs)
+        self.stream = stream
+
+    @parse_condition("disable_ads")
     def parse_tag_ext_x_scte35_out(self, value):
         self.state["scte35"] = True
 
     # unsure if this gets used by Twitch
+    @parse_condition("disable_ads")
     def parse_tag_ext_x_scte35_out_cont(self, value):
         self.state["scte35"] = True
 
+    @parse_condition("disable_ads")
     def parse_tag_ext_x_scte35_in(self, value):
         self.state["scte35"] = False
+
+    @parse_condition("low_latency")
+    def parse_tag_ext_x_twitch_prefetch(self, value):
+        segments = self.m3u8.segments
+        if segments:
+            segments.append(segments[-1]._replace(uri=self.uri(value)))
 
     def get_segment(self, uri):
         byterange = self.state.pop("byterange", None)
@@ -188,19 +212,18 @@ class TwitchM3U8Parser(M3U8Parser):
 
 class TwitchHLSStreamWorker(HLSStreamWorker):
     def _reload_playlist(self, text, url):
-        return load_hls_playlist(text, url, parser=TwitchM3U8Parser)
+        return load_hls_playlist(text, url, parser=TwitchM3U8Parser, stream=self.stream)
+
+    def _set_playlist_reload_time(self, playlist, sequences):
+        if not self.stream.low_latency:
+            super(TwitchHLSStreamWorker, self)._set_playlist_reload_time(playlist, sequences)
+        else:
+            self.playlist_reload_time = sequences[-1].segment.duration
 
 
 class TwitchHLSStreamWriter(HLSStreamWriter):
-    def __init__(self, *args, **kwargs):
-        HLSStreamWriter.__init__(self, *args, **kwargs)
-        options = self.session.plugins.get("twitch").options
-        self.disable_ads = options.get("disable-ads")
-        if self.disable_ads:
-            log.info("Will skip ad segments")
-
     def write(self, sequence, *args, **kwargs):
-        if self.disable_ads:
+        if self.stream.disable_ads:
             if sequence.segment.scte35 is not None:
                 self.reader.ads = sequence.segment.scte35
                 if self.reader.ads:
@@ -219,11 +242,28 @@ class TwitchHLSStreamReader(HLSStreamReader):
 
 
 class TwitchHLSStream(HLSStream):
+    def __init__(self, *args, **kwargs):
+        HLSStream.__init__(self, *args, **kwargs)
+        self.disable_ads = self.session.get_plugin_option("twitch", "disable-ads")
+        self.low_latency = self.session.get_plugin_option("twitch", "low-latency")
+
     def open(self):
+        if self.disable_ads:
+            log.info("Will skip ad segments")
+        if self.low_latency:
+            live_edge = max(1, min(LOW_LATENCY_MAX_LIVE_EDGE, self.session.options.get("hls-live-edge")))
+            self.session.options.set("hls-live-edge", live_edge)
+            self.session.options.set("hls-segment-stream-data", True)
+            log.info("Low latency streaming (HLS live edge: {0})".format(live_edge))
+
         reader = TwitchHLSStreamReader(self)
         reader.open()
 
         return reader
+
+    @classmethod
+    def _get_variant_playlist(cls, res):
+        return load_hls_playlist(res.text, base_uri=res.url)
 
 
 class UsherService(object):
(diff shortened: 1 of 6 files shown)


src/streamlink/plugins/twitch.py
```diff
@@ -175,12 +184,14 @@ class TwitchHLSStreamWorker(HLSStreamWorker):
     def _reload_playlist(self, text, url):
         return load_hls_playlist(text, url, parser=TwitchM3U8Parser)
 
+    def _set_playlist_reload_time(self, playlist, sequences):
+        self.playlist_reload_time = sequences[-1].segment.duration
```
back-to: seems a bit to much for normal Streams without `low-latency`
even if Twitch is doing it in the browser

bastimeyer:
- added `hls-segment-stream-data` to the Streamlink session options
- moved the `disable_ads` and `low_latency` properties to the TwitchHLSStream class, so that the other classes can access them via the `stream` property
- passed the `stream` instance to the TwitchM3U8Parser
- added HLS tag parsing conditions (as simple decorators)
- reverted the aggressive playlist reloading behavior when `low_latency` is not set
- improved tests and added tests for low latency logic
- changed / fixed plugin parameter description

bastimeyer:
Btw, while I was adding the tests, I noticed that the plugin parameters do not get initialized properly and I had to set them explicitly. Looks like they only get initialized by streamlink_cli here:
https://github.com/streamlink/streamlink/blob/1.1.1/src/streamlink_cli/main.py#L888-L914

src/streamlink/plugins/twitch.py
```diff
@@ -137,18 +137,45 @@
...
         self.state["scte35"] = False
 
+    @parse_condition("low_latency")
+    def parse_tag_ext_x_twitch_prefetch(self, value):
+        segments = self.m3u8.segments
+        if segments:
+            last_segment = segments[-1:][0]
+            new_segment = last_segment._replace(uri=self.uri(value))
+            segments.append(new_segment)
+            log.debug("Prefetching segment {0}".format(self.m3u8.media_sequence + len(segments) - 1))
```
back-to: to much spam and duplicate messages

it also doesn't prefetch anything here, it just queue it to a list so this is a bit misleading

```
[plugin.twitch][debug] Prefetching segment 3405
[plugin.twitch][debug] Prefetching segment 3406
[stream.hls][debug] Adding segment 3406 to queue
[stream.hls][debug] Download of segment 3405 complete
[plugin.twitch][debug] Prefetching segment 3406
[plugin.twitch][debug] Prefetching segment 3407
[stream.hls][debug] Adding segment 3407 to queue
[stream.hls][debug] Download of segment 3406 complete
[plugin.twitch][debug] Prefetching segment 3407
[plugin.twitch][debug] Prefetching segment 3408
[stream.hls][debug] Adding segment 3408 to queue
[stream.hls][debug] Download of segment 3407 complete
[plugin.twitch][debug] Prefetching segment 3408
[plugin.twitch][debug] Prefetching segment 3409
[stream.hls][debug] Adding segment 3409 to queue
[stream.hls][debug] Download of segment 3408 complete
[plugin.twitch][debug] Prefetching segment 3409
[plugin.twitch][debug] Prefetching segment 3410
[stream.hls][debug] Adding segment 3410 to queue
[stream.hls][debug] Download of segment 3409 complete
[plugin.twitch][debug] Prefetching segment 3410
[plugin.twitch][debug] Prefetching segment 3411
```

bastimeyer: True, I'll remove it again

src/streamlink/plugins/twitch.py
```diff
@@ -173,19 +199,18 @@ def get_segment(self, uri):
 
 class TwitchHLSStreamWorker(HLSStreamWorker):
     def _reload_playlist(self, text, url):
-        return load_hls_playlist(text, url, parser=TwitchM3U8Parser)
+        return load_hls_playlist(text, url, parser=TwitchM3U8Parser, stream=self.stream)
+
+    def _set_playlist_reload_time(self, playlist, sequences):
+        if not self.stream.low_latency:
+            HLSStreamWorker._set_playlist_reload_time(self, playlist, sequences)
```
beardypig: Really you should use `super` here, but whatever.

beardypig:
I think this will do for now :) I have, what _I_ think is, a better solution in the pipeline, but it is a little way off :)

beardypig:
One note: why not override `HLSStreamWorker.valid_sequence` instead of introducing the complexity of `parse_condition`? :)

bastimeyer:
> HLSStreamWorker.valid_sequence

Ideally, this should have been done in the disable-ads PR earlier this year, so that [`TwitchHLSStreamWriter.write`](https://github.com/streamlink/streamlink/pull/2513/files#diff-195bebeb65dac1f6a8e94f1957a95ec1L187) didn't need to be implemented. With these changes though, it doesn't make sense rewriting the entire thing again, unless you want it to be more complex.

Regarding the decorator, I've added the simple if statement to the prefetch tag method first, but it made sense to disable these advertisement tags too when the other parameter is not set, so adding a decorator made things look more simple, as all methods share this same logic. Btw, using `HLSStreamWorker.valid_sequence` to figure out whether the prefetch segments should be used or not would mean that they would always have to be created by the parser first and then discarded by the worker. A bit unnecessary.

----

I just wanted to get things finally done here with this PR. I know that most changes are not implemented the best way, but doing it right would require a much deeper refactoring of the entire thing and you already said that you had been working on something better a few months ago.

Let me fix the super class method call, though, before this gets merged.

beardypig:
Fair enough @bastimeyer, I think I was out of the loop on those changes, or not paying close enough attention.
I think the parser should parse all of the segments regardless, it should only have one job - parse, now it has a dependency on the stream. If the worker wants to discard a segment that should be its responsibility, I personally think it makes it more difficult to understand  (and less testable) to delegate to the parser, it might be slightly more efficient - but at what cost? Won’t someone think of the children!? 

Hopefully the work I’m doing on the stream classes will make this kind of thing easier.

src/streamlink/plugins/twitch.py
```diff
@@ -137,18 +137,44 @@
...
 
+    @parse_condition("disable_ads")
     def parse_tag_ext_x_scte35_in(self, value):
         self.state["scte35"] = False
 
+    @parse_condition("low_latency")
+    def parse_tag_ext_x_twitch_prefetch(self, value):
+        segments = self.m3u8.segments
+        if segments:
+            last_segment = segments[-1:][0]
```
beardypig: Did I miss something or is the same as `segments[-1]`?

bastimeyer: This line was copied from back-to's low latency test plugin here:
https://gist.github.com/back-to/8e9ed3c60e5932d8c7a67ccd43b906d0/#file-twitch-py-L444
I didn't change it when I created this branch, because I wasn't sure if there was something going on with `__getitem__` or so. Looks like it's a simple list object. Do you want me to change it or can we leave it?

By the way, I'm also still not sure how the cloned prefetch segments will work in combination with the ads filtering, because if hls-live-edge is set to 1 or 2, which is now a requirement here, then regular segments won't be used anymore at all and you're re-using the metadata of the last regular segment in the playlist for all the prefetched ones. So if there's an ad break or the end of an ad break, then you will only notice it if the playlist gets refreshed quickly enough and the last regular segment has the metadata set accordingly. It does get refreshed though after each segment now, as mentioned above, so it's *probably* fine.

beardypig: Shouldn't you just be able to use `get_segment` to create the "prefetch" segments?

beardypig: @back-to what's the deal with `segments[-1:][0]`? Am I missing something important?

back-to: > Am I missing something important?

nope

bastimeyer:
Any further concerns, comments or suggestions?

If refactoring the plugin and removing the stream dependency from the parser is that important, then tell me, and I will change it. Or just push the necessary changes to my branch yourself :)

beardypig:
As it's the twitch specific parser, I think it's OK. In my WIP branch, I have not used the `stream` (https://github.com/beardypig/streamlink/blob/stream-latency/src/streamlink/plugins/twitch.py). I haven't managed to get the latency down to faster than the browser though... What `mpv` cache settings did you use? And did you notice anything weird when reloading the playlist? I seem to get one prefetch segment in the first load, then in the next reload I get that previous prefetch segment as completed one, plus another completed one, and then more prefetch segments...

bastimeyer:
> What `mpv` cache settings did you use?

This depends on the value of `hls-live-edge`. If set to `2`, then [`--cache=no`](https://mpv.io/manual/master/#options-cache) is enough to not make it stutter (MPV's demuxer will still cache one additional second). If `hls-live-edge` is set to `1`, then you need to increase MPV's cache a bit by 0.5 secs or so via [`--demuxer-readahead-secs=1.5`](https://mpv.io/manual/master/#options-demuxer-readahead-secs) or [`--cache-secs`](https://mpv.io/manual/master/#options-cache-secs).

> did you notice anything weird when reloading the playlist?

Not really. Each playlist lists two prefetch segments. One of them is the previously completed one (live edge 1), and in the case that the reload/download takes too long, then the two prefetch segments are new ones.

I think using `get_segment` in your code is not the correct choice here, because the prefetch segments don't have a state which you can pop off the stack. That's why the last regular segment gets cloned for each prefetch segment in my PR.

beardypig:
Thanks for the `mpv` example. I’ll took again how `get_segment` and the state work. I need to look at more playlists :)

bastimeyer:
I've been thinking a little bit about the ads filtering, and I realized that we will probably have to make both parameters mutually exclusive.

As I've said, I don't get any ads on Twitch, so I can't test and verify this, but the way the low latency stuff is currently implemented is that it clones the last regular segment data, replaces the URL with the additional prefetch segment URLs and adds these new segments to the sequence as regular ones. This means that we're also shifting the live-edge here and since we're limiting the live-edge value to 1 or 2 when the parameter gets set, we don't have proper metadata to work with when it comes to ads filtering. This means that once an ad appears in the regular segments, we've already processed it as a prefetch segment without metadata and we're shifting the start/end of an ad-block by the amount of prefetch tags, which is two, or in other words 4 seconds. Since players like VLC don't handle discontinuities well, this breaks the whole ads-filtering implementation and is a problem.

Also, another thing I've realized is that we're also allowing low latency streaming for regular non-low-latency streams without prefetch segments. This isn't wrong per se, as it enables `--hls-segment-stream-data` and thus improves the latency by at most 2 seconds, but the log output implies that it's a low latency stream on Twitch's end, which isn't true. The log output needs to be changed, so that the user knows whether it's a true low latency stream or not. There's no information available on the Twitch API, so it has to be done by parsing the HLS playlist and checking whether it includes prefetch data.

thinkpad4:
Has this been added to the nightly build?  I tried a nightly from a week ago and couldn't get the --twitch-low-latency parameter to work.

bastimeyer:
This is an open pull request, so of course not. If you want to try it, install it from the branch of this PR. See https://github.com/streamlink/streamlink/issues/2402#issuecomment-480455258

back-to:
- should this be merged **or** do you want to make more changes?
- should we wait for the updated HLSStream version? (which will also take longer)

bastimeyer:
> should this be merged **or** do you want to make more changes?

I haven't had the time and motivation to change the things mentioned in my previous comment:
https://github.com/streamlink/streamlink/pull/2513#issuecomment-513464721

I think this should be done before this PR can be merged.

> should we wait for the updated HLSStream version?

No, not unless the features of this PR are in conflict with the promised improvements (`--hls-segment-stream-data` parameter and low latency streaming behind the `--twitch-low-latency` parameter).

beardypig:
My PR would make `--hls-segment-stream-data` obsolete, it would always apply, other than that there wouldn't be so much of a conflict.

kuon:
Do you think you might merge this in? I would love to have this feature built in.

bastimeyer:
As you can see since my last comment, no further changes have been added yet and the PR is still in the same WIP state.
https://github.com/streamlink/streamlink/pull/2513#issuecomment-518247628

kuon:
My view on both issue:

- Making this option mutually exclusive with ad filtering is fine. I see making them work together as a separate task as the problem seems non trivial.
- I think we should not over engineer the detection of low latency, adding a note in the option documentation about this option not working with non low latency twitch stream is fine.

thinkpad4:
I would love to see this finally added

axm2:
Would also like to see this merged

thinkpad4:
Any update on this getting pushed?

bastimeyer:
Locking this PR now.

I've already said multiple times that it's a WIP and not finished yet. This hasn't changed, so please stop asking, this is just unnecessary noise and not contributing anything. As you can see on my Github profile, I didn't have the time recently to work on any of my projects or pull requests.

I will see if I can add the necessary changes this weekend, but I'm not sure.

There's also @beardypig with his implementation, but I don't know anything about the status of that.

----

**What needs to be done?**

- [ ] Make ad filtering mutually exclusive with low latency streaming
- [ ] Disable low latency streaming if it's not a low latency stream
- [ ] Add proper documentation

src/streamlink/plugins/twitch.py
```diff
@@ -229,8 +229,17 @@ class TwitchHLSStreamReader(HLSStreamReader):
...
     def __init__(self, *args, **kwargs):
         HLSStream.__init__(self, *args, **kwargs)
-        self.disable_ads = self.session.get_plugin_option("twitch", "disable-ads")
-        self.low_latency = self.session.get_plugin_option("twitch", "low-latency")
+
+        disable_ads = self.session.get_plugin_option("twitch", "disable-ads")
+        low_latency = self.session.get_plugin_option("twitch", "low-latency")
+
+        if low_latency and disable_ads:
+            log.info("Low latency streaming with ad filtering is currently not supported")
```
back-to: it is not really clear in this message, which option is used and which is disabled.

bastimeyer: True, but both options print separate info log lines when set, so it's not important if this line doesn't tell which one gets disabled.

bastimeyer:
I've rebased the branch to 1.3.0. This doesn't mean though that there's anything new or that I'm confident getting this merged any time soon.

If you're testing the low latency branch, you should reinstall it. Btw, the next Twitch GUI version will require 1.3.0, so if you're using that, you will have to upgrade.

bastimeyer:
I've force-pushed a rebase from Streamlink's current master branch, so that the MPV fixes are now included and merge conflicts solved. Don't merge though, nothing has changed.

back-to:
> Regarding ad filtering and additional segment metadata, is it the right approach while prefetching segments to clone the last segment of the sequence and replace its URL?

if you could detect the Ad URLs they could be ignored

as you are using the prefetch URLs **there shouldn't be an issue**,
I don't think there are Ad URLs in the prefetch URLs.

but this is only testable with a larger user base

> Add proper documentation

nice to have, but not important for testing purposes

---

**i would merge this PR** and see what issues occupy on the fly.

bastimeyer:
> I don't think there are Ad URLs in the prefetch URLs.

Hm, that is actually true.

> Add proper documentation
> **i would merge this PR** and see what issues occupy on the fly.

Even if the `--twitch-low-latency` parameter gets set, there is still some tweaking to do by the user on the player side, otherwise the benefits are minimal. I think that should be mentioned at least. Also the live-edge value, which is all tied to it. Not sure if an example parameter string would be a good idea, because it depends on the player that's being used. I also haven't looked into VLC parameters yet, which is problematic, because it's Streamlink's default player. Let me improve the parameter description at least a bit though.

After I've rebased the branch and fixed the merge conflicts, then I think we can merge this. In another thread somewhere here I've said that I don't really feel comfortable with the changes, because there are still a couple of issues with occasional hiccups, but to be honest, I also don't want this PR to be open and stale anymore. Merging this will probably lead to lots of further questions, but so be it then.

bastimeyer:
After fiddling around a bit more with MPV's caching parameters (and also with python-sphinx), I've  decided against adding an example command line, because it's just too variable. It's probably also unnecessary when using the latest MPV version, because not adjusting anything at all didn't cause any major hiccups when I tested it and the latency was still good. This wasn't the case a few months ago.

I've also not changed the mutual exclusivity of the disable-ads parameter and also not mentioned it in the description. Since a message gets logged to `info`, I don't think it's necessary. We can change this stuff later if it causes issues.

Issue #2515, opened 2019-06-28
Can't login Pixiv account


hat3l0v3:
**Hello, Can't login in Pixiv account**

typing:
```
streamlink --pixiv-username "*******" --pixiv-password "*******" --player="D:\Program Files\vlc\vlc.exe" sketch.pixiv.net/@tuvai1031 best
```
result:
```
[plugin.pixiv][error] Failed to log in.
error: No playable streams found in this URL: sketch.pixiv.net/@tuvai1031
```
[Screenshot](https://i.imgur.com/rUhWRqZ.png)

**If I change artist that not requires you to login (not R18) - plugin works fine!**

typins the same with other stream link:
```
streamlink --pixiv-username "*******" --pixiv-password "*******" --player="D:\Program Files\vlc\vlc.exe" sketch.pixiv.net/@irodori-ace best
```
Result:
```
[plugin.pixiv][error] Failed to log in.
[cli][info] Avaliable streams: 720p (worst), 1080p (best)
[cli][info] Openins stream: 1080p (hls)
[cli][info] Starting player: D:\Program Files\vlc\vlc.exe
```
[Screenshot](https://i.imgur.com/lr2aco8.png)

**So problem is login/password I guess. I tried typing** `--pixiv-username=*******, --pixiv-username="*******"`, **adding** `;` **, double check my login/password to be sure. Nothing helps. Maybe im typing something wrong? I will be grateful for any help**

casmithyother:
I started having this issue the other day. It appears to be caused by a captcha system that may have been added recently.

I made modifications to my plugin script to use OAuth via https://oauth.secure.pixiv.net/auth/token. Here is where I took example from: https://github.com/Nandaka/PixivUtil2/blob/master/PixivOAuth.py

 I do not know if there is a more ideal solution.

brachna:
That might be the best solution for now. For my script I just use cookies from my browser.

hat3l0v3:
I

> That might be the best solution for now. For my script I just use cookies from my browser.

Is there some kind of guide how to fix problem like you? For people who know nothing about coding (like me).

brachna:
It does require a bit of coding, but if you don't have it and don't want to wait for a proper fix here's how to do it (I assume you're on Windows and using Firefox browser):
1. download browsercookie: 
[richardpenman-browsercookie-aab350c45967.zip](https://github.com/streamlink/streamlink/files/3342411/richardpenman-browsercookie-aab350c45967.zip)
[...]

PR #2521, opened 2019-06-30
plugins.pixiv: Use OAuth

casmithyother:
#2515


beardypig:
Thanks for the contribution! :) 

Where did you client secrets come from? Did you register an application with pixiv, or are they lifted from somewhere else?

casmithyother:
I've found the same secrets in a couple of places using their OAuth login which leads me to believe they are pulled from Pixiv's Android app. There also appears to be a pair for the IOS app as well.

back-to:
this is now not compatible, with https://github.com/streamlink/streamlink/pull/2529 merged

PR #2529, opened 2019-07-06
add login posibility for pixiv using sessionid and devicetoken

YuuichiMizuoka:
this adds the possibility to directly login to pixiv using the sessionid and devicetoken.

this can be useful for:
- automation scripts
- as a backup login method in case pixiv disables the oauth tokens that are used in the application


gravyboat:

PR #2532, opened 2019-07-08
plugins.ustreamtv: add support for proxying WebSocket connections

### LINKED ISSUE #2531: plugins.ustreamtv: Proxy and Socks do not work
Hi,
I noticed Proxy and Socks no longer work on Streamlink.
The channel I want to watch blocks all countries except the United States and I use US IP to see it.

```
streamlink --http-proxy xxx.xxx.xxx.xxx:xxxx https://www.ustream.tv/embed/23259921 best
```

And it doesn't work

```
[cli][info] Found matching plugin ustreamtv for URL https://www.ustream.tv/embed/23259921
[plugin.ustreamtv][error] This content is not available in your area
```

Meanwhile, I use the Proxy SwitchyOmega extension and the same IP, I can watch it on the browser.
Please help me, Thanks.

beardypig:
Supports establishing WebSocket connections using the same proxy as specified by `--http-proxy`. 
Fixes #2531.

### DIFF
--- a/src/streamlink/plugins/ustreamtv.py
+++ b/src/streamlink/plugins/ustreamtv.py
@@ -11,7 +11,7 @@ from socket import error as SocketError
 from threading import Thread, Event
 from time import sleep
 
-from streamlink.compat import range, urljoin, urlunparse
+from streamlink.compat import range, urljoin, urlunparse, urlparse, unquote_plus
 from streamlink.exceptions import PluginError, StreamError
 from streamlink.plugin import Plugin, PluginArguments, PluginArgument
 from streamlink.plugin.api import useragents, validate
@@ -57,6 +57,7 @@ class UHSClient(object):
         self._app_version = options.pop("app_version", self.APP_VERSION)
         self._cluster = options.pop("cluster", "live")
         self._password = options.pop("password")
+        self._proxy_url = options.pop("proxy")
         self._ws = None
 
     @property
@@ -79,11 +80,32 @@ class UHSClient(object):
         self._cluster = cluster
         self.reconnect()
 
+    @staticmethod
+    def parse_proxy_url(purl):
+        proxy_options = {}
+        if purl:
+            p = urlparse(purl)
+            proxy_options['proxy_type'] = p.scheme
+            proxy_options['http_proxy_host'] = p.hostname
+            if p.port:
+                proxy_options['http_proxy_port'] = p.port
+            if p.username:
+                proxy_options['http_proxy_auth'] = (unquote_plus(p.username), unquote_plus(p.password or ""))
+        return proxy_options
+
     def connect(self):
-        log.debug("Connecting to {0}".format(self.host))
+        proxy_options = self.parse_proxy_url(self._proxy_url)
+        if proxy_options.get('http_proxy_host'):
+            log.debug("Connecting to {0} via proxy ({1}://{2}:{3})".format(self.host,
+                                                                           proxy_options.get('proxy_type') or "http",
+                                                                           proxy_options.get('http_proxy_host'),
+                                                                           proxy_options.get('http_proxy_port') or 80))
+        else:
+            log.debug("Connecting to {0}".format(self.host))
         self._ws = websocket.create_connection(self.host,
                                                header=["User-Agent: {0}".format(useragents.CHROME)],
-                                               origin="https://www.ustream.tv")
+                                               origin="https://www.ustream.tv",
+                                               **proxy_options)
 
         args = dict(type="viewer",
                     appId=self._app_id,
@@ -304,6 +326,7 @@ class UHSStream(Stream):
             def _stopper(*args, **kwargs):
                 self.stop()
                 return f(*args, **kwargs)
+
             return _stopper
 
         def handle_module_info(self, args):
@@ -411,8 +434,8 @@ class UStreamTV(Plugin):
                     raise PluginError("Stream format is not supported: {0}".format(
                         ", ".join(data["streamFormats"].keys())))
             elif "stream" in arg and arg["stream"]["contentAvailable"] is False:
-                    log.error("This stream is currently offline")
-                    raise ModuleInfoNoStreams
+                log.error("This stream is currently offline")
+                raise ModuleInfoNoStreams
 
         return res
 
@@ -432,7 +455,11 @@ class UStreamTV(Plugin):
     def _get_streams(self):
         media_id, application = self._get_media_app()
         if media_id:
-            api = UHSClient(media_id, application, referrer=self.url, cluster="live", password=self.get_option("password"))
+            api = UHSClient(media_id, application,
+                            referrer=self.url,
+                            cluster="live",
+                            password=self.get_option("password"),
+                            proxy=self.session.get_option("http-proxy"))
             log.debug("Connecting to UStream API: media_id={0}, application={1}, referrer={2}, cluster={3}",
                       media_id, application, self.url, "live")
             api.connect()
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -1070,8 +1070,8 @@ def build_parser():
         "--http-proxy",
         metavar="HTTP_PROXY",
         help="""
-        A HTTP proxy to use for all HTTP requests, by default this proxy
-        will be used for all HTTPS requests too.
+        A HTTP proxy to use for all HTTP requests, including WebSocket connections. 
+        By default this proxy will be used for all HTTPS requests too.
 
         Example: "http://hostname:port/"
         """


beardypig:
We could also use the existing `http-proxy` argument. I think it is good to include the `websocket-proxy` option though, as it allows for more customisation. However, it might be nice if the default values for `https-proxy` and `websocket-proxy` are set to the value for `http-proxy` - but that's a different discussion.

src/streamlink/plugins/ustreamtv.py
```diff
@@ -79,11 +80,28 @@ def cluster(self, cluster):
         self._cluster = cluster
         self.reconnect()
 
+    @staticmethod
+    def parse_proxy_url(purl):
+        proxy_options = {}
+        if purl:
+            p = urlparse(purl)
+            if p.scheme != "http":
```
back-to: socks5 seems supported

https://github.com/websocket-client/websocket-client/pull/417

src/streamlink_cli/argparser.py
```diff
@@ -1084,6 +1084,15 @@ def build_parser():
         Example: "https://hostname:port/"
         """
     )
+    http.add_argument(
+        "--websocket-proxy",
```
back-to: this doesn't seems like an universal solution,
`--http-proxy` or `--https-proxy` might be easier.

src/streamlink/plugins/ustreamtv.py
beardypig: Ah, I didn’t see that - in the docs for the function it says http only.

src/streamlink_cli/argparser.py
beardypig: Depends - on the one I think it’s good for it to be more configurable. In case you want the WebSocket proxied but not the http connection, I didn’t try it but it might actually be better in the ustream case. What about `wss` does that use the `https-proxy` option?  

To be honest I’m quite undecided about it - I commented on the PR as soon as I made it suggesting the possibility of using `http-proxy`. Part of me thinks `http-proxy‘ should cover everything, http, https, and WebSocket. Life would be easier then, it would make it easier for regular users too. Then make it so you can override or unset the other options, of course this would be a breaking change...


Maybe it is better to just use `http-proxy` for this ... :|

bastimeyer: I think the distinction between `http-proxy` and `https-proxy` (and therefore also `websocket-proxy`) is unnecessary. I highly doubt that using different proxies for these parameters is a proper use case. Using different proxies is only useful if you know which requests Streamlink makes, so you can target certain URLs with different proxies. The distinction between the protocols alone doesn't make sense.

beardypig: Which is better? 

1. `http-proxy` for `ws://` and `https-proxy` for `wss://`.
2. Same as 1., but the default for `https-proxy` is the same as `http-proxy`.
3. Use `http-proxy` for both `ws://` and `wss://`.

I personally prefer 2. as it would kill two stones with one bird, there have been a few issues where people have needed to set `https-proxy` as well as `http-proxy`. As @bastimeyer mentioned, you have to know which connections streamlink will make - and sometimes we don’t even know which it will make, there could be redirects etc. 2. gives you a sensible default, but the ability to override it if needed, perhaps if your http proxy only supports https over https and http over http.

bastimeyer: The second point sounds like the best solution, as it's a non-breaking change. It should be done in a second PR, though.

beardypig: A separate PR indeed. There is a breaking change though, before `https-proxy` would be unset. I don’t think it’s a problem though :)

src/streamlink/plugins/ustreamtv.py
beardypig: Updated it so that it'll pass it through to the websocket library to handle.

beardypig:
If an unsupported proxy type is specified, it will fail like so
```
[plugin.ustreamtv][debug] Connecting to ws://r11721124-1-23259921-channel-ws-live.ums.ustream.tv:1935/1/ustream via proxy (https://localhost:666)
error: proxy_type must be 'http', 'socks4', 'socks5' or 'socks5h'
```

src/streamlink/session.py
```diff
@@ -291,8 +291,9 @@ def set_option(self, key, value):
             self.http.cert = value
         elif key == "http-timeout":
             self.http.timeout = value
-        else:
-            self.options.set(key, value)
+
+        # all options are available via `options.get`
+        self.options.set(key, value)
```
back-to: should probably be its own PR

---

`get_option` and `options.get` might be different _(not directly related to your changes)_ 

```
        print(self.session.get_option("http-proxy"))
http://127.0.0.1:80
        print(self.session.options.get("http-proxy"))
127.0.0.1:80
```

src/streamlink/plugins/ustreamtv.py
```diff
@@ -432,7 +455,11 @@ def handle_reject(self, api, args):
     def _get_streams(self):
         media_id, application = self._get_media_app()
         if media_id:
-            api = UHSClient(media_id, application, referrer=self.url, cluster="live", password=self.get_option("password"))
+            api = UHSClient(media_id, application,
+                            referrer=self.url,
+                            cluster="live",
+                            password=self.get_option("password"),
+                            proxy=self.session.options.get("http-proxy"))
```
back-to: `self.session.get_option("http-proxy")` should probably be used

src/streamlink_cli/argparser.py
```diff
@@ -1070,7 +1070,7 @@ def build_parser():
         "--http-proxy",
         metavar="HTTP_PROXY",
         help="""
-        A HTTP proxy to use for all HTTP requests.
+        A HTTP proxy to use for all HTTP requests, including WebSocket connections.
```
back-to: will be in conflict with the other PR

beardypig: Will have to rebase one or the other :)

src/streamlink/session.py
beardypig: True, it's not required when using `self.session.get_option` :)

beardypig:
Thanks for the review @back-to :)

src/streamlink/plugins/ustreamtv.py
```diff
@@ -79,11 +80,32 @@ def cluster(self, cluster):
...
-        log.debug("Connecting to {0}".format(self.host))
+        proxy_options = self.parse_proxy_url(self._proxy_url)
+        if proxy_options.get('http_proxy_host'):
+            log.debug("Connecting to {0} via proxy ({1}://{2}:{3})".format(self.host,
+                                                                           proxy_options.get('proxy_type') or "http",
+                                                                           proxy_options.get('http_proxy_host'),
+                                                                           proxy_options.get('http_proxy_port') or 80))
+        else:
+            log.debug("Connecting to {0}".format(self.host))
         self._ws = websocket.create_connection(self.host,
```
back-to: `websocket._exceptions.WebSocketProxyException:` should maybe be handled for invalid proxies

```
[plugin.ustreamtv][debug] Connecting to ws://xxx via proxy (http://xxx)
  File "streamlink/vv/bin/streamlink", line 11, in <module>
  File "streamlink/src/streamlink_cli/main.py", line 1037, in main
    handle_url()
  File "streamlink/src/streamlink_cli/main.py", line 577, in handle_url
  File "streamlink/src/streamlink_cli/main.py", line 457, in fetch_streams
  File "streamlink/src/streamlink/plugin/plugin.py", line 325, in streams
    ostreams = list(ostreams)
  File "streamlink/src/streamlink/plugins/ustreamtv.py", line 465, in _get_streams
    api.connect()
  File "streamlink/src/streamlink/plugins/ustreamtv.py", line 108, in connect
    **proxy_options)
  File "streamlink/vv/lib/python3.7/site-packages/websocket/_core.py", line 514, in create_connection
    websock.connect(url, **options)
  File "streamlink/vv/lib/python3.7/site-packages/websocket/_core.py", line 223, in connect
    options.pop('socket', None))
  File "streamlink/vv/lib/python3.7/site-packages/websocket/_http.py", line 122, in connect
    sock = _tunnel(sock, hostname, port, auth)
  File "streamlink/vv/lib/python3.7/site-packages/websocket/_http.py", line 293, in _tunnel
    "failed CONNECT via proxy status: %r" % status)
websocket._exceptions.WebSocketProxyException: failed CONNECT via proxy status: 403

  File "streamlink/vv/lib/python3.7/site-packages/websocket/_http.py", line 293, in _tunnel
    "failed CONNECT via proxy status: %r" % status)
websocket._exceptions.WebSocketProxyException: failed CONNECT via proxy status: 503

  File "streamlink/vv/lib/python3.7/site-packages/websocket/_http.py", line 293, in _tunnel
    "failed CONNECT via proxy status: %r" % status)
websocket._exceptions.WebSocketProxyException: failed CONNECT via proxy status: 400
```

beardypig: Yep - I’ll take a look :)

beardypig:
I rebased this one on to master.

src/streamlink/plugins/ustreamtv.py
back-to: this `exception` is **not important** when used with `--http-proxy` it will fail earlier

PR #2534, opened 2019-07-08
Add dependency for JS parser

vstavrinov:
Introducing esprima as JS parser. Inspired by discussions in https://github.com/streamlink/streamlink/pull/2425 and https://github.com/streamlink/streamlink/pull/2505
Source code:
https://github.com/Kronuz/esprima-python
Packages for nodejs  exists in Debian and other distro while python-esprima should be expected as well. 
Here is usage example. The key idea is to iterate list until find type / name pair:
```
import esprima
js = '''
function a(b) {
    b["c"] = "d"
    return b
}

class e {
    constructor() {
        this.f = {g: "h", c: "k"}
    }
    m(n) {
        var p = this.f
        p["g"] = "r"
        return n(p)
    }
}

s = new e
console.log(s.f)
console.log(s.m(a))

'''
data = ""
parse = esprima.parseScript(js, { "tolerant": True })
body = parse.body
for node in body:
    if node.type == "ClassDeclaration" and node.id.name == "e":
        for node in node.body.body:
            if node.type == "MethodDefinition" and node.key.name == "m":
                for node in node.value.body.body:
                    if node.type == "ExpressionStatement" and \
                        node.expression.left.object.name == "p" and \
                        node.expression.left.property.value == "g":
                        data = node.expression.right.value
                        break
                break
        break

print("data = " + data)

```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,8 @@ deps = [
     'win-inet-pton;python_version<"3.0" and platform_system=="Windows"',
     # shutil.get_terminal_size and which were added in Python 3.3
     'backports.shutil_which;python_version<"3.3"',
-    'backports.shutil_get_terminal_size;python_version<"3.3"'
+    'backports.shutil_get_terminal_size;python_version<"3.3"',
+    'esprima'
 ]
 
 # for encrypted streams


gravyboat:
Thanks for bringing this up for discussion @vstavrinov! I'm not super familiar with the various js parser options for Python, so I'd just like to make sure we do our due diligence in selecting one:

* Looking at the repo there hasn't been an update in almost a year. They also don't allow the creation of issues, is this still actively maintained?
* Are there any alternatives? What makes esprima the best? Is there anything with a higher adoption rate, better OS support, more active development, etc?
* How many distros support esprima with official packages for python-esprima?

vstavrinov:
@gravyboat:

They still accept pull-requests. If there are no maintenance we can take it over and make our own fork when it will makes sense. 

Other issues we have already discussed. You can add more details if consider it worth of that. 

And finally, feel free to propose an alternative and we will discuss Your proposal here. 

P.S. As for esprima itself, it looks like standard solution (I have already mentioned this), concern may touch it's python implementation only. I think it worth to take care about this issue.

back-to:
is there **any real use case** for a popular plugin?
(if possible already a finished plugin)

beardypig:
I suggested that a utility function for parsing things that are not quite JSON objects more reliable
```js
{
	test: "foo",
}
```

This is a simple case, but there are some more tricky cases where regex becomes quite fragile.

vstavrinov:
@back-to, It will come when You add the dependency. Who wants to waste time without perspective  to be accepted? I have already passed this way, You know for example ovvatv but not only that. And why "popular" only? It could be almost any one of Your choice as often as js comes into game.

vstavrinov:
Here is new plugin as example of esprima usage:
https://github.com/vstavrinov/streamlink/blob/maverick/src/streamlink/plugins/kanalukraina.py

vstavrinov:
One more example:

https://github.com/vstavrinov/streamlink/blob/master/src/streamlink/plugins/arconai.py

This is specific case, where regex based solution seems  impossible. Even using html/js parsers insufficient. The problem is that the request of the same html page return different js containing  the same type of chunks of target url  in different fields of the structure. Yes, it is possbile to create different patterns according to every structure, but  their exhaustive list is unknown, so there always are chance that next request fails. 

This is interesting but it turned out stream url could be assembled from the javascript structure by means of js beautifier independently of alternating structure. So it is more than only formatter while the parsers used only to locate approriate data. Ultimately workflow is as follow:  html parser  -> js beautifier -> js parser.

The story conclusion may differ, but there are no guarantee it is unique. Thus it may considered as yet another reason using  regular tools for js handling.

brouxco:
Also you should take a look at Kronuz/esprima-python#11

beardypig:
@brouxco yeah that is a bit weird, it could be used to host the malicious script on GitHub. The file referred to in that PR is not actually included in the distribution, it only appears to be used for testing. Weird none the less. 

Unfortunately esprima does seem to be the best JavaScript lexer for python and it would be a shame for that file to mess it up.

brouxco:
Yeah. The fact that the project does not look very active is concerning too as you mentioned.

I don't know if you had a look at https://github.com/ytdl-org/youtube-dl/issues/11292 but they are discussing this since November 2016.
The corresponding pull request is https://github.com/ytdl-org/youtube-dl/issues/11292 (no activity since June 2018)

Adding a dependency could help as long as it is maintained.
It does not look like any JS parser in python is actively maintained. (see other JS parser projects:  https://github.com/PiotrDabkowski/pyjsparser, https://github.com/rspivak/slimit)

Quoting: https://github.com/ytdl-org/youtube-dl/issues/11292#issuecomment-263053575
> Python wrappers for those existing engines [SpiderMonkey of Gecko, V8 of Blink, JavascriptCore of Webkit and ChakraCore of IE/Edge, and maybe other JS engines used in popular browsers] are much easier to write and maintain in comparison with a new Python implementation from scratch. 

Maybe this is what we should try to do if we deem that having a JS parser is really necessary.

beardypig:
The thing with esprima is that it’s a pure python implementation, so there are no dependencies to manage and it appears to be a pretty good le we for modern JavaScript. It should be sufficient for most of Streamlink’s requirements, and I don't think we should be adding anything that might be able to actually execute JavaScript.

setup.py
```diff
@@ -23,7 +23,8 @@
     'win-inet-pton;python_version<"3.0" and platform_system=="Windows"',
     # shutil.get_terminal_size and which were added in Python 3.3
     'backports.shutil_which;python_version<"3.3"',
-    'backports.shutil_get_terminal_size;python_version<"3.3"'
+    'backports.shutil_get_terminal_size;python_version<"3.3"',
+    'esprima'
```
beardypig, line 27: We should pin to this specific version, or possible a smaller range of versions. 

```suggestion
    'esprima==4.0.1'
```

gravyboat:
@beardypig Esprima is a dead project, why are we even discussing usage of this again?

beardypig:
@gravyboat it would help with some things, regex is pretty ugly for some things. It doesn't have to be a fully fledged all singing, all dancing JS parser, we don't want to execute JS - just extract useful bits of information slightly easier. I created a function for transforming JavaScript objects not in JSON to Python dicts (https://github.com/beardypig/streamlink/commit/edce2246d09c13a88531e0680d755d69bdaeff22) as an example use case. 

Do you mean that `python-esprima` is dead project?

gravyboat:
@beardypig Yes I am specifically referring to esprima. I have no issue with using some kind of (light and actively maintained/secure) js parser and I like the util you added.

beardypig:
As far as I understand `python-esprima` is following the release cycle of `esprima` which the latest version is 4.0.1. `python-esprima` is a pure python package, with no external dependencies - pretty light. 
It doesn't seem any less secure that any other dependency we have, the "shellcode" that was included in the tests (not in the package) was explained, as has since been removed. 
I dunno, I like it ...

vstavrinov:
And it is not dead. The author responds to pull requests. So python-esprima is managed.

vstavrinov:
One more new plugin using js parser:

https://github.com/vstavrinov/streamlink/blob/master/src/streamlink/plugins/ukraina24.py

Unfortunately, they all are out of upstream.

gravyboat:
Closing as there hasn't been any movement or a decision in months.

Issue #2550, opened 2019-07-14
Closing the command prompt is not handled on Windows


skulblakka:
## Bug Report

- [x ] This is a bug report and I have read the contribution guidelines.

### Description

When the user closes the command prompt (cmd/PowerShell) using the "X" button it is not handled by streamlink. This causes Windows to just kill it without giving streamlink the chance to terminate the player.

### Expected / Actual behavior

The player should be closed along streamlink when the command prompt window is closed.

### Reproduction steps / Explicit stream URLs to test

Using ```--player-passthrough``` or ```--player-args="--keep-open=yes {filename}"``` with mpv will keep the player open when closing the command prompt window. 

Additionally using CTRL-C does not work when using ```--player-passthrough```. Nothing happens until you close the player yourself or the stream has ended.

### Log output

I don't think log output is very useful here. If you disagree just tell me and I add some :)

### Additional comments, screenshots, etc.

I tried fixing this by [setting a Console Ctrl Handler](https://docs.microsoft.com/en-us/windows/console/setconsolectrlhandler) using win32api ([pypiwin32](https://pypi.org/project/pypiwin32)). 

I ran into an issue when using ```--player-passthrough```. The player is started using ```subprocess.call``` instead of ```subprocess.Popen```. As subprocess.call does not return until the subprocess is finished we can not kill the process when the handler is triggered*. I "fixed" this by not using call and waiting in the main thread instead. This also resolved the problem with streamlink not reacting to CTRL-C.

I also tested this on debian. I does not seem to break anything.

**I did not make a PR for this as I'm sure there is some room for improvement. I'm also not sure how to add the pypiwin32 dependency only for Windows.** My current solution is [here](https://gist.githubusercontent.com/skulblakka/7adacdd155b3bafedef31969cf932eee/raw/af545a222edcf6ff389c4b17fef4846d907fd5d9/win32_exit.diff). Hopefully someone can take a look at it.

_*This is just my observation on Windows. The same stuff works on Linux so I guess this is either Windows specific or my limited Know-How regarding python/this project._

My local setup:
- Windows 10 1809
- Python 3.7.3 x64
- streamlink commit 0f01f35

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

beardypig:
[...]

PR #1758, opened 2018-06-06
sdist and wheel release fixes

beardypig:
The script to build the sdist and wheel packages is run before the script that uploads them to PyPI. The sdist script doesn't clean up the environment, but extra prune options have been added to the MANIFEST.in file (thanks @back-to).

All the requirements have been changed to use PEP 345 environment markers, instead of deciding the packages to include based on the current interpreter. This means you can build a wheel on Python 3.6.3 and have it work on Python 2.7.14 :) It also tidies up the setup.py quite a bit. 

I'd like to get confirmation from @mmetak that this will work as expected on the streamlink-git AUR before merging :)

PR #2551, opened 2019-07-14
Feature: Use ConsoleCtrlHandler under Windows

skulblakka:
This PR is for #2550. @beardypig suggested me making a PR with my changes so we can better discuss them.

---

This attempts to correctly detect the signals received when the Windows command prompt is closed. Additionally I'm using ```subprocess.Popen``` instead of ```subprocess.call``` to correctly close the player when using ```--player-passthrough```.

There are still some issues here:
- I was getting a ```ConsoleCtrlHandler function failed``` message in the command prompt when using ```sys.exit(error_code)``` instead of ```return error_code``` at the end of ```win32_exit``` (L67 in main.py)
- When not using ```--player-passthrough``` and exiting streamlink using CTRL-C will result in the exception below
```
streamlink twitch.tv/lirik best --player-args="--volume=0 {filename}"
[cli][info] Found matching plugin twitch for URL twitch.tv/lirik
[cli][info] Starting player: "H:\Tresors\Dev\mpv\build\mpv.exe"
[cli][info] Stream ended
  File "F:\python-3.7.3 x64\App\Python\Scripts\streamlink-script.py", line 11, in <module>
    load_entry_point('streamlink==1.1.1+44.g51e7d58.dirty', 'console_scripts', 'streamlink')()
  File "F:\python-3.7.3 x64\App\Python\lib\site-packages\streamlink-1.1.1+44.g51e7d58.dirty-py3.7.egg\streamlink_cli\main.py", line 1058, in main
    handle_url()
  File "F:\python-3.7.3 x64\App\Python\lib\site-packages\streamlink-1.1.1+44.g51e7d58.dirty-py3.7.egg\streamlink_cli\main.py", line 612, in handle_url
    handle_stream(plugin, streams, stream_name)
  File "F:\python-3.7.3 x64\App\Python\lib\site-packages\streamlink-1.1.1+44.g51e7d58.dirty-py3.7.egg\streamlink_cli\main.py", line 465, in handle_stream
    success = output_stream(plugin, stream)
  File "F:\python-3.7.3 x64\App\Python\lib\site-packages\streamlink-1.1.1+44.g51e7d58.dirty-py3.7.egg\streamlink_cli\main.py", line 349, in output_stream
    read_stream(stream_fd, output, prebuffer)
  File "F:\python-3.7.3 x64\App\Python\lib\site-packages\streamlink-1.1.1+44.g51e7d58.dirty-py3.7.egg\streamlink_cli\main.py", line 386, in read_stream
    output.write(data)
  File "F:\python-3.7.3 x64\App\Python\lib\site-packages\streamlink-1.1.1+44.g51e7d58.dirty-py3.7.egg\streamlink_cli\output.py", line 37, in write
    return self._write(data)
  File "F:\python-3.7.3 x64\App\Python\lib\site-packages\streamlink-1.1.1+44.g51e7d58.dirty-py3.7.egg\streamlink_cli\output.py", line 296, in _write
    self.player.stdin.write(data)
ValueError: I/O operation on closed file
```
- With the added dependency the installation via ```python3 setup.py install``` fails under msys2 with python 3.7.3. It works fine with a "normal" installation of python3. I'm not sure if we would be able to use the windows signals in msys2 anyway but I was unable to make the dependency for non-msys2 python only.
```
Searching for pywin32>=223
Reading https://pypi.org/simple/pywin32/
No local packages or working download links found for pywin32>=223
error: Could not find suitable distribution for Requirement.parse('pywin32>=223')
```

The issues above only occur on Windows. Using Debian the issues do not appear and everything looks like it's working as expected. The player closes and only the expected output is generated when closing the player, using CTRL-C or closing the terminal window with and without ```--player-passthrough```.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ deps = [
 # support for win32api on Windows (win32 and win-amd64)
 # pypiwin32 is not available for python 3.4 and below
 if get_platform().startswith("win"):
-    deps.append('pypiwin32<=219;platform_system=="Windows" and python_version>"3.4"')
+    deps.append('pypiwin32>=223;platform_system=="Windows" and python_version>"3.4"')
 
 # for encrypted streams
 if environ.get("STREAMLINK_USE_PYCRYPTO"):
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,9 @@ deps = [
     'win-inet-pton;python_version<"3.0" and platform_system=="Windows"',
     # shutil.get_terminal_size and which were added in Python 3.3
     'backports.shutil_which;python_version<"3.3"',
-    'backports.shutil_get_terminal_size;python_version<"3.3"'
+    'backports.shutil_get_terminal_size;python_version<"3.3"',
+    # support for win32api on Windows
+    'pypiwin32<=223;platform_system=="Windows"'
 ]
 
 # for encrypted streams
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@
 import codecs
 from os import environ, path
 from sys import argv, path as sys_path
+from sysconfig import get_platform
 
 from setuptools import setup, find_packages
 
@@ -23,10 +24,12 @@ deps = [
     'win-inet-pton;python_version<"3.0" and platform_system=="Windows"',
     # shutil.get_terminal_size and which were added in Python 3.3
     'backports.shutil_which;python_version<"3.3"',
-    'backports.shutil_get_terminal_size;python_version<"3.3"',
-    # support for win32api on Windows
-    'pypiwin32<=223;platform_system=="Windows"'
+    'backports.shutil_get_terminal_size;python_version<"3.3"'
 ]
+
+# support for win32api on Windows (win32 and win-amd64)
+if get_platform().startswith("win"):
+    deps.append('pypiwin32<=223;platform_system=="Windows"')
 
 # for encrypted streams
 if environ.get("STREAMLINK_USE_PYCRYPTO"):
--- a/setup.py
+++ b/setup.py
@@ -28,8 +28,9 @@ deps = [
 ]
 
 # support for win32api on Windows (win32 and win-amd64)
+# pypiwin32 is not available for python 3.4 and below
 if get_platform().startswith("win"):
-    deps.append('pypiwin32<=223;platform_system=="Windows"')
+    deps.append('pypiwin32<=219;platform_system=="Windows" and python_version>"3.4"')
 
 # for encrypted streams
 if environ.get("STREAMLINK_USE_PYCRYPTO"):
--- a/setup.py
+++ b/setup.py
@@ -28,9 +28,9 @@ deps = [
 ]
 
 # support for win32api on Windows (win32 and win-amd64)
-# pypiwin32 is not available for python 3.4 and below
-if get_platform().startswith("win"):
-    deps.append('pypiwin32>=223;platform_system=="Windows" and python_version>"3.4"')
+# pywin32 is only available for python 2.7, 3.5 and above
+if not environ.get("STREAMLINK_NO_PYWIN32"):
+    deps.append('pywin32>=223;platform_system=="Windows" and (python_version=="2.7" or python_version>"3.4")')
 
 # for encrypted streams
 if environ.get("STREAMLINK_USE_PYCRYPTO"):
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -397,6 +397,13 @@ def read_stream(stream, output, prebuffer, chunk_size=8192):
                     console.exit("Error when writing to output: {0}, exiting", err)
 
                 break
+            except ValueError as err:
+                try:
+                    output.player.wait(10) # Wait a moment before we check the player
+                    if not output.opened:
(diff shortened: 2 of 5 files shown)


beardypig:
In principle I agree the suggested change, let’s see if you can iron out all the kinks :)

src/streamlink_cli/main.py
```diff
@@ -375,6 +397,13 @@ def read_stream(stream, output, prebuffer, chunk_size=8192):
                     console.exit("Error when writing to output: {0}, exiting", err)
 
                 break
+            except ValueError as err:
+                try:
+                    output.player.wait(10) # Wait a moment before we check the player
+                    if not output.opened:
+                        sys.exit(130)
+                except TimeoutError:
+                    console.exit("Error when writing to output: {0}, exiting", err)
```
skulblakka: For me it looks like the handler and the main thread are having some kind of race condition causing the issue with the exception occurring during the write I mentioned in the OP. This causes the player to be closed but the check for ```player.opened``` still returning ```True```. 

My thought process was that the easiest solution would be to catch the Exception and wait a few seconds to see if the player is closing. If yes it is assumed that the exception was caused by the race condition during the exit and not an actual error if not the usual error message is displayed. The first case should only occur if streamlink was interrupted with CTRL-C that's why I'm using the 130 exit code.

There is probably a better solution so let me know :)

skulblakka:
I'm just now realizing that the pip version of pywin32 does not support python3.4. I don't know what the plan regarding maintaining python3.4 support is considering it reached it's official end of life a few months ago.

skulblakka:
I think I resolved all the issues. Works fine for me on Windows and Debian. 

The only thing remaining is the race condition I mentioned above. While my workaround resolves the issue I'm not sure if that is the cleanest solution for the problem.

src/streamlink_cli/main.py
```diff
@@ -977,11 +1006,15 @@ def setup_logging(stream=sys.stdout, level="info"):
 
 
 def main():
+    global use_win32exit
```
beardypig: This isn't required is it? If you're just reading a `global` you don't need to declare it `global`, right?

skulblakka: You are right. Changed.

setup.py
```diff
@@ -26,6 +27,11 @@
     'backports.shutil_get_terminal_size;python_version<"3.3"'
 ]
 
+# support for win32api on Windows (win32 and win-amd64)
+# pypiwin32 is not available for python 3.4 and below
+if get_platform().startswith("win"):
```
back-to: the windows wheel might miss this dependency
similar to https://github.com/streamlink/streamlink/pull/1758/files#diff-2eeaed663bd0d25b7e608891384b7298

tests/test_cmdline.py
```diff
@@ -49,7 +49,7 @@ def fn(*args):
         if not passthrough:
             mock_popen.assert_called_with(commandline, stderr=ANY, stdout=ANY, bufsize=ANY, stdin=ANY)
         else:
-            mock_popen.assert_called_with(commandline, stderr=ANY, stdout=ANY)
+            mock_popen.assert_called_with(commandline, stderr=ANY, stdout=ANY, bufsize=ANY, stdin=ANY)
```
back-to: `if not` and `else` is the same now

skulblakka: Updated

setup.py
skulblakka: Sorry but I'm not really sure what I'm supposed to do to fix that.

back-to: add it directly into **deps**, without `get_platform().startswith("win")`

similar to

https://github.com/streamlink/streamlink/blob/56253563e92732e2f0ebffebabc566c3bc4a2124/setup.py#L22-L24

---

also add it into `makeinstaller.sh`

https://github.com/streamlink/streamlink/blob/master/script/makeinstaller.sh#L70-L100

skulblakka: > add it directly into deps, without get_platform().startswith("win")

But is it possible to check for the `sysconfig.get_platform()` with that method for the conditional dependency? The problem I mentioned in my original post was that the install would fail on msys2 because all the other methods for checking the platform have the same result as a "normal" Windows python installation. However it would then try to get pypiwin32 from pip for mingw which is not available.

gravyboat:
@back-to Any other feedback on this? It's been hanging around for a while.

skulblakka:
Ok. I haven't found a solution to the problem I described in the comments above regarding the use of `sysconfig.get_platform()` in the install script. Instead my new solution uses a environment variable to allow users to skip the installation of `pywin32`.

I also added the dependency to `makeinstaller.sh` and the docs.

bastimeyer:
Going to close this PR. It's been stale for 3 years, has unresolved issues, a ton of merge conflicts, and it hasn't been closed earlier because nobody knows what to do with it. And to be honest, I didn't pay any attention to it, sorry.

If you or anyone else wants to implement the handling of the termination of the windows terminal, then this should be done anew, ideally with a bit of refactoring of the CLI main module, as the solution proposed here relies on global vars and the bad design of the CLI package. I can of course offer my help then, as I'm far more familiar with the current CLI implementation and packaging requirements now compared to the time when the PR was opened.

PR #2553, opened 2019-07-18
plugins.beattv: partial fix for the be-at.tv streams

beardypig:
There is no easy way to determine which stream to pick in some cases, for example the series pages have multiple streams but nothing in the url to indicate which is playing.

No sure if there is much interest in the plugin, the current one is totally broken and I cannot find an issue about it.

### DIFF
--- a/src/streamlink/plugins/beattv.py
+++ b/src/streamlink/plugins/beattv.py
@@ -1,324 +1,48 @@
 import logging
+import base64
 import re
 
-from collections import namedtuple
-from io import BytesIO
-try:
-    from Crypto.Cipher import AES
-    CAN_DECRYPT = True
-except ImportError:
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/beattv.py
```diff
@@ -1,324 +1,48 @@
...
-
-        return reader
+log = logging.getLogger(__name__)
 
 
 class BeatTV(Plugin):
-    @classmethod
-    def can_handle_url(self, url):
-        return _url_re.match(url)
+    _url_re = re.compile(r"https?://(\w+\.)?be-at.tv/videos/.+")
```
back-to: there is also `/video/`

https://www.be-at.tv/video/adam_beyer_hyte_nye_germany_2018

PR #2557, opened 2019-07-20
plugins.youtube: fix url check

vstavrinov:
valid url example:
https://www.youtube.com/c/ТелеканалНАШ

### DIFF
--- a/src/streamlink/plugins/youtube.py
+++ b/src/streamlink/plugins/youtube.py
@@ -110,14 +110,14 @@ _url_re = re.compile(r"""(?x)https?://(?:\w+\.)?youtube\.com
         |
         (?:
             /(?:
-                (?:user|channel)/
+                (?:user|c(?:hannel)?)/
                 |
                 embed/live_stream\?channel=
             )(?P<user>[^/?&]+)
         )
         |
         (?:
-            /(?:c/)?(?P<liveChannel>[^/?]+)/
+            /(?:c/)?(?P<liveChannel>[^/?]+)/live/?$
         )
     )
 """)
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/youtube.py
```diff
@@ -117,7 +117,7 @@ def parse_fmt_list(formatsmap):
         )
         |
         (?:
-            /(?:c/)?(?P<liveChannel>[^/?]+)/live/?$
+            /(?:c/)?(?P<liveChannel>[^/?]+)/
```
back-to: this would now match `https://www.youtube.com/t/terms`

don't remove the `live` part it is for urls like `https://www.youtube.com/EXAMPLE1/live`

---

also your example url doesn't actually match the _liveChannel_ group because there is a `/` in the regex

`https://www.youtube.com/c/ТелеканалНАШ`

just add `/c/` to the user group some lines above

`(?:user|c(?:hannel)?)/`

---
and update the plugin tests also add an example url there

I know the `user` and `liveChannel` group is not used in the plugin right now,
but it still shouldn't match other youtube urls incorrectly.

vstavrinov: @back-to:

> I know the `user` and `liveChannel` group is not used in the plugin right now,
> but it still shouldn't match other youtube urls incorrectly.

In this case what for we retain those groups there?  To produce errors in the test? They should be removed from plugin as well from the test. Instead,  there are more urls patterns to add like this:

https://www.youtube.com/user/NewsmaxTV/live
https://www.youtube.com/channel/UCx6h-dWzJ5NpAlja1YsApdg/live

And slash after live should be removed because it only redirects to root of  the user or channel.

Issue #2488, opened 2019-06-04
Files of recorded of Facebook MPEG-DASH streams are broken


MaximVol:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

Streamlink produces broken files during recording of Facebook streams. Only small fragment of recorded file can be played, after this video freezes and only audio is playing.

### Reproduction steps / Explicit stream URLs to test

This happens with any Facebook live stream. This is example of recorded file:
[https://www.dropbox.com/s/04r53qctuoc85yb/test.ts?dl=1](url)

Duration of recorded media is 2 minute, but only first 12 seconds can be played.

### Log output

```
`streamlink -o test.ts "https://www.facebook.com/Fox4DFW/videos/460953291359115/" best
[cli][info] streamlink is running as root! Be careful!
[cli][info] Found matching plugin facebook for URL https://www.facebook.com/Fox4DFW/videos/460953291359115/
[cli][info] Available streams: 144p_alt (worst), 144p, 240p_alt, 240p, 360p_alt, 360p, 480p_alt, 480p, 720p_alt, 720p (best)
[cli][info] Opening stream: 720p (dash)
[download][test.ts] Written 5.7 MB (2m1s @ 15.9 KB/s)
^C
[stream.dash][warning] Download of segment: https://video-otp1-1.xx.fbcdn.net/hvideo-prn2-ash/v/rrejTXTVG8-z0ec49bal9/live-dash/live-hd1-v/10156192895032093_0-4036580.m4v aborted
[stream.ffmpegmux][error] Pipe copy aborted: /tmp/ffmpeg-27990-210
[cli][info] Stream ended
Interrupted! Exiting...
[cli][info] Closing currently open stream...`
```

FFMPEG says that duration of recorded media is unknown:

```
ffmpeg -hide_banner -i test.ts
Input #0, matroska,webm, from 'test.ts':
  Metadata:
    COMPATIBLE_BRANDS: isomiso6dash
    MAJOR_BRAND     : iso6
    MINOR_VERSION   : 1
    ENCODER         : Lavf57.71.100
  Duration: N/A, start: 3974.503000, bitrate: N/A
    Stream #0:0: Video: h264 (High), yuv420p(progressive), 1280x720, 29.97 fps, 29.97 tbr, 1k tbn, 2k tbc (default)
    Metadata:
      HANDLER_NAME    : VideoHandler
[...]

PR #2559, opened 2019-07-21
ffmpegmux: use -start_at_zero with -copyts

back-to:
- fixes invalid timestamp on recorded files
- fixes invalid timestamp on live player cache playback
- starts the timestamp always at 0 if -copyts was used
- removed wrong called log, it should have been `self.logger` when it was changed
  it was only used for one line instead of all at the moment.

Fixes https://github.com/streamlink/streamlink/issues/2488

### DIFF
--- a/src/streamlink/stream/ffmpegmux.py
+++ b/src/streamlink/stream/ffmpegmux.py
@@ -84,7 +84,6 @@ class FFMPEGMuxer(StreamIO):
 
         self.session = session
         self.process = None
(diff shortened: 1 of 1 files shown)


gravyboat:
Looks good @back-to. @beardypig Feel free to just merge after a review. I don't think we have any procedure around requiring more than one person.

beardypig:
Sometimes I like to wait in case op makes a change, or someone else wants to weigh in :)

PR #2563, opened 2019-07-26
CLI logging issues

ichramm:
* Fixed use of invalid member console.logger
* Fixed lint warning "Too many arguments for logging format string"


beardypig:
Looks good to me, should result in identical output.

PR #2565, opened 2019-07-27
plugins.youtube: fixed reason msg, updated _url_re

back-to:
closes #2557

### DIFF
--- a/src/streamlink/plugins/youtube.py
+++ b/src/streamlink/plugins/youtube.py
@@ -85,8 +85,13 @@ _config_schema = validate.Schema(
                 },
                 validate.optional("videoDetails"): {
                     validate.optional("isLive"): validate.transform(bool),
-                }
-            }
+                },
+                validate.optional("playabilityStatus"): {
+                    validate.optional("status"): validate.text,
+                    validate.optional("reason"): validate.all(validate.text,
+                                                              validate.transform(maybe_decode)),
+                },
+            },
         ),
         validate.optional("live_playback"): validate.transform(bool),
         validate.optional("reason"): validate.all(validate.text, validate.transform(maybe_decode)),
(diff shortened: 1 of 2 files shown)


vstavrinov:
Here is difference between Your and mine code from #2557:
```diff
         (?:
-            /(?:
-                (?:user|c(?:hannel)?)/
-                |
-                embed/live_stream\?channel=
-            )[^/?&]+
+            /(?:user|c(?:hannel)?)/[^/?&]+(/live)?
         )
         |
         (?:
-            /(?:c/)?[^/?]+/live/?$
+            /embed/live_stream\?channel=[^/?&]+
         )
```
Your variant is significantly longer, cumbersome, vague and ultimately it's wrong, because it matches what shouldn't, for example:

```
streamlink youtube.com/xxx/live
[cli][info] Found matching plugin youtube for URL youtube.com/xxx/live
error: Could not find a video on this page
 ```
Just for recap what all this issue is about: 

1. "live" suffix make sense with all and only /c/,  /user/ and /channel/ urls, but such urls matched part of Your expression that doesn't contains any of them (last one in quoted fragment).
2. There are should not be slash after this suffix. 
3.  In some  cases the same urls with and without the suffix could produce different live streams.

Two more examples:

```
streamlink youtube.com/embed/live_stream?channel=UCx6h-dWzJ5NpAlja1YsApdg/live
[cli][info] Found matching plugin youtube for URL youtube.com/embed/live_stream?channel=UCx6h-dWzJ5NpAlja1YsApdg/live
error: Could not find a video on this page

streamlink youtube.com/embed/live_stream?channel=UCx6h-dWzJ5NpAlja1YsApdg
[cli][info] Found matching plugin youtube for URL youtube.com/embed/live_stream?channel=UCx6h-dWzJ5NpAlja1YsApdg
[cli][info] Available streams: 144p (worst), 240p, 360p, 480p, 720p (best)

```
They both matches Your expression, but only last one should. It happens  because You combine embed url with with live suffix in the same part (first) of Your expression.  That's wrong too.

back-to:
> what shouldn't, for example:
streamlink youtube.com/xxx/live

this is a **valid** url https://www.youtube.com/rocketbeanstv/live
and your code doesn't match it.

> streamlink youtube.com/embed/live_stream?channel=UCx6h-dWzJ5NpAlja1YsApdg/live

this is an **invalid** url

> They both matches Your expression, but only last one should. 

sure, but it's an **user issue** if someone ever enters this invalid url
could be changed but not that important.

vstavrinov:
With such logic we could match any url for youtube.com.  But only part of expression containing group makes sense. Everything else could be collapsed.

PR #2570, opened 2019-07-31
session: use importlib to load plugins in Python 3

beardypig:
Using `importlib` to import the plugins in Python 3 speeds up the loading of plugins by 2x over the old `imp` method used in Python 2.

For me, with Python 3, the time to run `load_plugins` goes down from ~550ms to ~220ms (on average).

### DIFF
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -1,14 +1,13 @@
-import imp
 import logging
 import pkgutil
 import sys
 import traceback
+from collections import OrderedDict
 
 import requests
 
-from collections import OrderedDict
-
 from streamlink.logger import StreamlinkLogger, Logger
+from streamlink.plugin import Plugin
 from streamlink.utils import update_scheme, memoize
 from streamlink.utils.l10n import Localization
 from . import plugins, __version__
@@ -457,42 +456,19 @@ class Streamlink(object):
         """Attempt to load plugins from the path specified.
 
         :param path: full path to a directory where to look for plugins
-
         """
-        for loader, name, ispkg in pkgutil.iter_modules([path]):
-            file, pathname, desc = imp.find_module(name, [path])
-            # set the full plugin module name
-            module_name = "streamlink.plugin.{0}".format(name)
-
-            try:
-                self.load_plugin(module_name, file, pathname, desc)
-            except Exception:
-                sys.stderr.write("Failed to load plugin {0}:\n".format(name))
-                print_small_exception("load_plugin")
-
-                continue
-
-    def load_plugin(self, name, file, pathname, desc):
-        # Set the global http session for this plugin
         user_input_requester = self.get_option("user-input-requester")
-        api.http = self.http
-
-        module = imp.load_module(name, file, pathname, desc)
-
-        if hasattr(module, "__plugin__"):
-            module_name = getattr(module, "__name__")
-            plugin_name = module_name.split(".")[-1]  # get the plugin part of the module name
-
-            plugin = getattr(module, "__plugin__")
-            plugin.bind(self, plugin_name, user_input_requester)
-
-            if plugin.module in self.plugins:
-                log.debug("Plugin {0} is being overridden by {1}".format(plugin.module, pathname))
-
-            self.plugins[plugin.module] = plugin
-
-        if file:
-            file.close()
+        for (loader, name, _) in pkgutil.iter_modules([path], "streamlink.plugins."):
+            try:
+                mod = loader.find_module(name).load_module(name)
+            except ImportError:
+                log.exception("Failed to load plugin {0} from {1}\n".format(name, path))
+                continue
+            if hasattr(mod, "__plugin__") and issubclass(mod.__plugin__, Plugin):
+                mod.__plugin__.bind(self, name, user_input_requester)
+                if mod.__plugin__.module in self.plugins:
+                    log.debug("Plugin {0} is being overridden by {1}".format(mod.__plugin__.module, mod.__file__))
+                self.plugins[mod.__plugin__.module] = mod.__plugin__
 
     @property
     def version(self):
--- a/src/streamlink/plugin/api/support_plugin.py
+++ b/src/streamlink/plugin/api/support_plugin.py
@@ -1,5 +1,6 @@
 import os
 import inspect
+
 from streamlink.utils import load_module
 
 __all__ = ["load_support_plugin"]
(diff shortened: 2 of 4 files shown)


beardypig:
Would be nice to get some people to try it out, especially those on lower spec machines with Python 3 to see if it improves things at all. I'll see if I can dig out a Raspberry Pi to test it out :)

skulblakka:
I just tested this on my Raspberry Pi 3 (Raspbian 9.9 and Python 3.5.3). 

Average time for ```load_plugins``` goes down from 2236ms to 1747ms (about 22%).

beardypig:
Thanks @skulblakka! If you have time, could you test out this version of [session.py](https://raw.githubusercontent.com/beardypig/streamlink/plugin-import-tweak-cache/src/streamlink/session.py), it implements some caching in Python 3 that might improve performance even more on lower spec machines :)

skulblakka:
I actually reinstalled my Pi last night updating it to Raspian 10 (including updating to Python 3.7.3). I repeated all the tests and got a few different results: The average time for ```load_plugins``` goes down from 1162ms to 1006ms (about 13%) using the version from this PR and down to 973ms (additional 3%) using the version linked above.

I also repeated the test using Python 3.5.4 (still running on Debian 10 tho) and got from 1354ms to 1192ms (about 12%) for the version from this PR and down to 1177ms (additional 1.5%) using the version linked above. 

_Additional Note: 
When I was using the version linked above I had to run streamlink as root as I got_ 
```
PermissionError:   [Errno 13] Permission denied:   '/usr/local/lib/python3.7/dist-packages/streamlink-1.1.1+57.g1aaab4b.dirty-py3.7.egg/streamlink/plugins/plugin.cache'
```
_otherwise._

beardypig:
Ah yeah, I need to change the cache location - I was really wondering if it was worth it, but if it only saves a few ms it's not worth the added complexity :)

beardypig:
@skulblakka if you get time, I updated the original PR to simplify the whole thing (turns out `pkgutils` wraps the different loading methods), can you see if you still get a speed up?

src/streamlink/session.py
```diff
@@ -453,42 +452,20 @@ def load_plugins(self, path):
...
-                log.debug("Plugin {0} is being overridden by {1}".format(plugin.module, pathname))
-
-            self.plugins[plugin.module] = plugin
-
-        if file:
-            file.close()
+            if hasattr(mod, "__plugin__") and issubclass(mod.__plugin__, Plugin):
+                mod.__plugin__.bind(self, name, user_input_requester)
+                if mod.__plugin__.module in self.plugins:
+                    log.debug("Plugin {0} is being overridden".format(mod.__plugin__.module))
```
back-to: are there any reason for removing the path?

beardypig: It wasn’t easily available in the previous iteration, I will add it back :)

src/streamlink/session.py
```diff
@@ -453,42 +452,20 @@ def load_plugins(self, path):
...
-            # set the full plugin module name
-            module_name = "streamlink.plugin.{0}".format(name)
-
+        user_input_requester = self.get_option("user-input-requester")
+        for (loader, name, _) in pkgutil.iter_modules([path]):
             try:
-                self.load_plugin(module_name, file, pathname, desc)
-            except Exception:
+                mod = loader.find_module(name).load_module(name)
+            except:
```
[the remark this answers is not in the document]
beardypig: I will add the correct exceptions, I just wanna see if it works first :)

skulblakka:
@beardypig With the new version I get from 1153ms to 1044ms (about 9%)

beardypig:
Seems about the same as the first version then.

back-to:
looks like a cleanup, does it still speed up without the cache changes?

also is it ready?

beardypig:
There were some improvements over the current version, but I need to update this branch because some of the method uses are now deprecated.

beardypig:
Seems to cause more problems than its worth.

bastimeyer:
Why did you close this, @beardypig? And what about #2582?

I haven't followed these two PRs, but it looks like that you tried out a different implementation first in the other PR and then simplified it via pkgutil. What's the problem with the latest commit here? Was it because of the failing tests?
https://travis-ci.org/github/streamlink/streamlink/jobs/670619725#L600-L601
With the py2 removal, we can/should drop deprecated stuff.

beardypig:
@bastimeyer I can/will review in light of dropping Python 2. That was the main issue with it, the compatibility between the two versions - it didn't offer enough of a gain to be worth while. But, with just Python 3 it is worth cleaning up.

bastimeyer:
@beardypig Well, let's go then. It's almost September, py2 is dead since 8 months and we've been pushing it back and ignoring it again and again. No reason to further delay this. If I'm not mistaken, all we need to do is remove py2 from the CI matrix and setup.py script, which we have a PR for - hopefully not outdated. Everything else should be optional for now and can come afterwards, like this change for example. No more py2 restrictions then. Those who still rely on py2 can keep using the older releases.

----

In regards to the plugin imports though, would it make sense exploring different plugin registration systems? The only reason why all (default / non-user) plugins need to be imported at once is because of the `can_handle_url` method. Since all it does on basically every (?) plugin is matching only a single regex, one way of improving the initial load time could be making the regexes monolithic with a lazy plugin import on match, so that no unused plugins need to be imported. Another way of improving it could also be splitting up each plugin into a separate plugin and regex module, so that the initial imports are much lighter, with a lazy plugin import on match as well. Not sure though if this would be beneficial when there's already a warm bytecode cache.

beardypig:
The real gains are for the lower-spec'd machines, RPi etc. I don't know that there is a better way to do the plugins, the way we have it at the moment means that every `can_handle_url` method (including from side-loaded plugins) always called and ranked which gives us lots of flexibility.

bastimeyer:
> But, with just Python 3 it is worth cleaning up.

@beardypig Do you want to reopen this and rebase once you get the time? The import changes should be fine now on py3.6+. I'd like to get rid of the deprecated `imp` module in `2.0.0`.

bastimeyer:
Just tinkered around a bit with both import interfaces, the module returned by `pkgutil.iter_modules(...) -> loader.find_module(...).load_module(...)` as well as `importlib.machinery.PathFinder.find_spec(...).loader.exec_module(...)`, and for some reason, modules dynamically imported by this don't share the same logging config, which means every plugin's log calls get ignored in the output. Not sure what this is about.

Issue #2607, opened 2019-08-29
BBC iplayer keyerror: goto


werid:
### Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

When trying to connect to bbc iplayer, live stream of bbc channel, it crashes with KeyError: 'goto'

### Reproduction steps / Explicit stream URLs to test

1. i updated streamlink to latest from github: pip-3.6 install --user git+https://github.com/streamlink/streamlink.git
2. i ran the following cmd: streamlink --loglevel debug --http-no-ssl-verify --bbciplayer-username "username" --bbciplayer-password password --player-external-http --player-external-http-port 3280 https://www.bbc.co.uk/iplayer/live/bbctwo best

### Log output

```
[cli][debug] OS:         FreeBSD-12.0-RELEASE-p3-amd64-64bit-ELF
[cli][debug] Python:     3.6.8
[cli][debug] Streamlink: 1.2.0+12.g1fb9391
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.50.0)
[cli][info] Found matching plugin bbciplayer for URL https://www.bbc.co.uk/iplayer/live/bbctwo
[cli][debug] Plugin specific arguments:
[cli][debug]  --bbciplayer-username=***@*** (username)
[cli][debug]  --bbciplayer-password=******** (password)
[plugin.bbciplayer][info] A TV License is required to watch BBC iPlayer streams, see the BBC website for more information: https://www.bbc.co.uk/iplayer/help/tvlicence
Traceback (most recent call last):
  File "/home/username/bin/streamlink", line 11, in <module>
    load_entry_point('streamlink==1.2.0+12.g1fb9391', 'console_scripts', 'streamlink')()
  File "/home/username/.local/lib/python3.6/site-packages/streamlink_cli/main.py", line 1038, in main
    handle_url()
  File "/home/username/.local/lib/python3.6/site-packages/streamlink_cli/main.py", line 578, in handle_url
    streams = fetch_streams(plugin)
  File "/home/username/.local/lib/python3.6/site-packages/streamlink_cli/main.py", line 458, in fetch_streams
    sorting_excludes=args.stream_sorting_excludes)
  File "/home/username/.local/lib/python3.6/site-packages/streamlink/plugin/plugin.py", line 323, in streams
    ostreams = list(ostreams)
[...]

PR #2614, opened 2019-09-03
Fixes streamlink issue #2607.

mkbloke:
Changed code for nonce extraction, plus small typo fixes.


mkbloke:
Breaks current testing code...

Issue #2590, opened 2019-08-17
Bilibili playurl api changed


aliofwind:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

Bilibili was changed get playurl api, source stream must using "quality=4" in playurl api, "quality=0" is same with "quality=3" get 1500kbps/720p stream

### Reproduction steps / Explicit stream URLs to test

1. https://live.bilibili.com/13

back-to:
> [cli][debug] Streamlink: 1.0.0

your Streamlink version is way to old.

aliofwind:
this issue is also in version 1.1.1, and i reinstalled py37 yesterday so debug log is 1.0.0 on py36

aliofwind:
Bilibili replaces the original quality parameter with the qn parameter, and the qn parameter can be refined to more than 3 resolutions.
eg: 
liveroom: https://live.bilibili.com/6 
playurlapi: https://api.live.bilibili.com/room/v1/Room/playUrl?cid=7734200&platform=web
request:
`"data":{"current_quality":4,"accept_quality":["4","3","2"],"current_qn":10000,"quality_description":[{"qn":10000,"desc":"原画"},{"qn":400,"desc":"蓝光"},{"qn":250,"desc":"超清"},{"qn":150,"desc":"高清"},{"qn":80,"desc":"流畅"}],`

PR #2625, opened 2019-09-10
plugins.bilibili: fix resolution issue

Troogle:
The bilibili API has changed and quality options are now in range of [4,3,2]. Using quality=0 could only fetch 720p stream
#2590


gravyboat:
Good stuff, thanks.

PR #2628, opened 2019-09-15
plugins.foxtr: Extended support

ozankaraali:
With a small url regex change, this plugin could play both of the media sites of Fox TR.


gravyboat:
Looks good, thanks.

PR #2630, opened 2019-09-17
plugins.twitcasting: new plugin for TwitCasting.tv

printempw:
Add support for https://twitcasting.tv

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -240,6 +240,7 @@ tvplayer                tvplayer.com         Yes   No    Streams may be geo-rest
 tvrby                   tvr.by               Yes   No    Streams may be geo-restricted to Belarus.
 tvrplus                 tvrplus.ro           Yes   No    Streams may be geo-restricted to Romania.
 tvtoya                  tvtoya.pl            Yes   --
+twitcasting             twitcasting.tv       Yes   No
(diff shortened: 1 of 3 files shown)


gravyboat:
I'm hesitant to add this site because it has had almost no one streaming and very few viewers every time I've checked. There have also been a few instances where it seems people are just using the site to restream restricted content. I'll wait and see what others say.

printempw:
It's a quite famous live streaming platform in Japan though. Wait for other opinions.

gravyboat:
@printempw I'll try to check it out when more Japanese people are awake I guess (so far I've checked late night and early morning JPN time). For such a famous streaming platform it seems like no one is using it.

MaserIsNotALaser:
Please add this site, its popular in Japan. (You can see much more streams if you change site language to Japanese)

gravyboat:
Thanks for the contribution @printempw, changing the language to Japanese made actual streams with viewers show up. I didn't realize they categorized them in such an obtuse way.

printempw:
That's nice. Thank you ;)

i-cant-git:
Sometimes streamer has bad signal and connection is lost for few seconds, because of that streamlink closes websocket connection and you have to start streamlink again and again and again. Is it possible to add "reconnect to websocket if connection is lost" feature?

Issue #2636, opened 2019-09-29
Support for wasd.tv


Uinden:
There is 2 type of links. For example:
https://wasd.tv/channel/86692/videos/269793
https://wasd.tv/channel/86692
Both of them can be live stream or replay of past stream.

What I will say next is not confirmed by any documentation or anything its just how I think this website working.

For example user call streamlink for URL https://wasd.tv/channel/86692
Then we request
https://wasd.tv/api/media-containers?media_container_status=RUNNING&limit=1&offset=0&channel_id=86692&media_container_type=SINGLE,COOP
Response is in JSON format.
*I replaced `media_container_status=STOPPED,RUNNING` to just `RUNNING`. I think if user calls streamlink without /videos/ part in URL he wants live stream. Not past record.
If there is no live stream then response will be `{"result":[]}`
From response we need `json["result"][0]["user_id"]`. For example this channel's `user_id=179599`
Then we look into array `json["result"][0]["media_container_streams"]` and search for element where `json["result"][0]["media_container_streams"][i]["user_id"]` is equals to channel's `user_id=179599`
Then inside that element of array we get another element `json["result"][0]["media_container_streams"][i]["stream_media"]` which has different streams with different quality. For example:
`json["result"][0]["media_container_streams"][0]["stream_media"][0]["media_type"]="HLS"`
`json["result"][0]["media_container_streams"][0]["stream_media"][0]["media_status"]="RUNNING"`
`json["result"][0]["media_container_streams"][0]["stream_media"][0]["media_meta"]["media_current_resolution"]="1280x720"`
`json["result"][0]["media_container_streams"][0]["stream_media"][0]["media_meta"]["media_url"]="https://cdn.wasd.tv/live/179599/index.m3u8"`

Now if user call streamlink for URL https://wasd.tv/channel/86692/videos/269793
We request https://wasd.tv/api/media-containers/269793
And its basically same response and we need to do same job here. 
If this is live stream then in `json["result"][0]["media_container_streams"][0]["stream_media"]` array will be element with `media_status="RUNNING"`
If this is replay of past stream then in `json["result"][0]["media_container_streams"][0]["stream_media"]` array will be element with `media_status="STOPPED"`
And instead of `media_url` we need to get `media_archive_url` which will be `.m3u8` file.

I have checked other plugins code and its too hard for me to make it myself. 
I hope somebody will be able to do it with results of my "investigation".

Uinden:
[...]

PR #2638, opened 2019-09-29
plugins.wasd: new plugin for WASD.TV streaming platform

Uinden:
Check #2636


Uinden:
Please add
```
wasd                    wasd.tv              Yes   Yes   Requires session cookies for AUTH API token
```
To `/docs/plugin_matrix.rst`. Looks like I can't do it in browser. 
As I understand it should solve tests problem.

PR #2640, opened 2019-09-30
plugins.wasd: new plugin for WASD.TV

### LINKED ISSUE #2636: Support for wasd.tv

Uinden:
Closes #2636

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -264,6 +264,7 @@ vinhlongtv              thvli.vn             Yes   No    Streams are geo-restric
 vk                      vk.com               Yes   Yes
 vrtbe                   vrt.be/vrtnu         Yes   Yes
 vtvgo                   vtvgo.vn             Yes   No
+wasd                    wasd.tv              Yes   Yes   Requires session cookies for AUTH API token
 webcast_india_gov       webcast.gov.in       Yes   No    You can use #Channel to indicate CH number.
 webtv                   web.tv               Yes   --
 welt                    welt.de              Yes   Yes   Streams may be geo-restricted to Germany.
--- a/src/streamlink/plugins/wasd.py
+++ b/src/streamlink/plugins/wasd.py
@@ -1,0 +1,147 @@
+import logging
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.stream import HLSStream
+
+log = logging.getLogger(__name__)
+
+class WASD(Plugin):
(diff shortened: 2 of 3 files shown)


Uinden:
I'm sorry for spamming pull requests but I'm trying my best

PR #2641, opened 2019-09-30
plugins.wasd: new plugin for WASD.TV

### LINKED ISSUE #2636: Support for wasd.tv

Uinden:
Closes #2636

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -274,7 +274,7 @@ vk                      vk.com               Yes   Yes
 vlive                   vlive.tv             Yes   No    Embedded Naver VODs are not supported.
 vrtbe                   vrt.be/vrtnu         Yes   Yes
 vtvgo                   vtvgo.vn             Yes   No
-wasd                    wasd.tv              Yes   Yes   Requires session cookies for AUTH API token
+wasd                    wasd.tv              Yes   Yes
 webcast_india_gov       webcast.gov.in       Yes   No    You can use #Channel to indicate CH number.
 webtv                   web.tv               Yes   --
 welt                    welt.de              Yes   Yes   Streams may be geo-restricted to Germany.
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -274,6 +274,7 @@ vk                      vk.com               Yes   Yes
 vlive                   vlive.tv             Yes   No    Embedded Naver VODs are not supported.
 vrtbe                   vrt.be/vrtnu         Yes   Yes
 vtvgo                   vtvgo.vn             Yes   No
+wasd                    wasd.tv              Yes   Yes   Requires session cookies for AUTH API token
 webcast_india_gov       webcast.gov.in       Yes   No    You can use #Channel to indicate CH number.
 webtv                   web.tv               Yes   --
 welt                    welt.de              Yes   Yes   Streams may be geo-restricted to Germany.
--- a/src/streamlink/plugins/wasd.py
+++ b/src/streamlink/plugins/wasd.py
@@ -1,0 +1,147 @@
+import logging
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.stream import HLSStream
+
+log = logging.getLogger(__name__)
+
+class WASD(Plugin):
+	_url_re = re.compile(r"https?://(?:www\.)?wasd\.tv/")
+	_url_channel = re.compile(r"https?://(?:www\.)?wasd\.tv/channel/(\d+)")
+	_url_video = re.compile(r"https?://(?:www\.)?wasd\.tv/channel/\d+/videos/(\d+)")
+
+	@classmethod
+	def can_handle_url(cls, url):
+		return WASD._url_re.match(url)
+
+	def _get_streams(self):
+		re_video_id = self._url_video.match(self.url)
+		re_channel_id = self._url_channel.match(self.url)
+
+		if not re_video_id and not re_channel_id:
+			return log.error("Expected to see url like https://wasd.tv/channel/12345 or https://wasd.tv/channel/12345/videos/67890")
+
+		# Check if we have AUTH token stored in session cache and request it
+		cookies = self.session.http.cookies
+		if not cookies or not cookies['cronos-auth-token'] or not cookies['cronos-auth-token-signature']:
+			log.debug("There is no AUTH token in cache. Attempting to request one")
+			self.getAuthToken()
+		if not cookies or not cookies['cronos-auth-token'] or not cookies['cronos-auth-token-signature']:
+			return log.error("Failed to get AUTH token")
+
+		if re_video_id:
+			streams = self.getStreamsFromURL("https://wasd.tv/api/media-containers/" + re_video_id.group(1), 'video')
+		else:
+			streams = self.getStreamsFromURL("https://wasd.tv/api/media-containers?media_container_status=RUNNING&limit=1&offset=0&channel_id=" + re_channel_id.group(1) + "&media_container_type=SINGLE,COOP", 'channel')
+
+		if streams:
+			for video in streams:
+				# TODO: This is how I planned it initially. video[0] is quality and video[1] is URL for .m3u8 file.
+				# But it doesn't work and I can't find channels with more than 1 quality to debug it. So for now it will be copy-pasted code from other plugins
+				# yield video[0], HLSStream(self.session, video[1])
+
+				# This one is working. But I feel its not how it should be
+				return HLSStream.parse_variant_playlist(self.session, video[1])
+
+	# WASD.TV requires token to make API requests. We will get it here
+	def getAuthToken(self):
+		log.debug("Requesting AUTH token for future API requests")
+		res = self.session.http.post('https://wasd.tv/api/auth/anon-token')
+
+		if not res:
+			return None
+		if not res.cookies:
+			return None
+		if not res.cookies['cronos-auth-token']:
+			return None
+		if not res.cookies['cronos-auth-token-signature']:
+			return None
+
+		self.save_cookies()
+		log.debug("Succesfully got AUTH token")
+
+	# type = video / channel. Type of API link
+	def getStreamsFromURL(self, URL, type):
+		log.debug("Requesting stream URL: " + URL)
+
+		res = self.session.http.get(URL)
+		json_res = self.session.http.json(res)
+
+		if not json_res:
+			return log.error("Server sent empty response for URL: " + URL)
+		if not json_res["result"]:
+			log.debug("There is no results so we assume no streams available")
+			return None
+
+		if type == "video":
+			return self.getStreamsFromJSON(URL, json_res["result"], type)
+		else:
+			if not json_res["result"][0]:
+				return log.error("Missing result[0] in response from server for URL: " + URL)
+			return self.getStreamsFromJSON(URL, json_res["result"][0], type)
+
+	def getStreamsFromJSON(self, URL, json, type):
+		if not json["user_id"]:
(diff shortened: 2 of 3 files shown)


docs/plugin_matrix.rst
```diff
@@ -264,6 +264,7 @@ vinhlongtv              thvli.vn             Yes   No    Streams are geo-restric
 vk                      vk.com               Yes   Yes
 vrtbe                   vrt.be/vrtnu         Yes   Yes
 vtvgo                   vtvgo.vn             Yes   No
+wasd                    wasd.tv              Yes   Yes   Requires session cookies for AUTH API token
```
back-to: > Requires session cookies for AUTH API token

you don't need to mention it if it is automatically

src/streamlink/plugins/wasd.py
```diff
@@ -0,0 +1,147 @@
...
+from streamlink.stream import HLSStream
+
+log = logging.getLogger(__name__)
+
+class WASD(Plugin):
+	_url_re = re.compile(r"https?://(?:www\.)?wasd\.tv/")
+	_url_channel = re.compile(r"https?://(?:www\.)?wasd\.tv/channel/(\d+)")
+	_url_video = re.compile(r"https?://(?:www\.)?wasd\.tv/channel/\d+/videos/(\d+)")
+	
+	@classmethod
```
back-to: you are using tabs everywhere, use 4 spaces instead.

src/streamlink/plugins/wasd.py
```diff
@@ -0,0 +1,147 @@
...
+	def can_handle_url(cls, url):
+		return WASD._url_re.match(url)
+		
+	def _get_streams(self):
+		re_video_id = self._url_video.match(self.url)
+		re_channel_id = self._url_channel.match(self.url)
+		
+		if not re_video_id and not re_channel_id:
+			return log.error("Expected to see url like https://wasd.tv/channel/12345 or https://wasd.tv/channel/12345/videos/67890")
```
back-to: you don't need this, update `_url_re` instead of using `_url_channel` and `_url_video`

example

```py
>>> f = r'https?://(?:www\.)?wasd\.tv/channel/(?P<channel_id>\d+)(?:/videos/(?P<video_id>\d+))?'
>>> import re
>>> r = re.compile(f)
>>> url1 = 'https://wasd.tv/channel/12345/videos/67890'
>>> url2 = 'https://wasd.tv/channel/12345'
>>> r.match(url1).groups()
('12345', '67890')
>>> r.match(url2).groups()
('12345', None)
>>> r.match(url2).group('video_id')
>>> r.match(url2).group('channel_id')
'12345'
```

src/streamlink/plugins/wasd.py
```diff
@@ -0,0 +1,147 @@
...
+		if not res.cookies['cronos-auth-token']:
+			return None
+		if not res.cookies['cronos-auth-token-signature']:
+			return None
+			
+		self.save_cookies()
+		log.debug("Succesfully got AUTH token")
+	
+	# type = video / channel. Type of API link 
+	def getStreamsFromURL(self, URL, type):
```
back-to: don't use the name `type`

src/streamlink/plugins/wasd.py
```diff
@@ -0,0 +1,147 @@
...
+	def getUserContainerFromArray(self, user_id, arr):
+		for stream in arr:
+			if not stream["user_id"]:
+				return log.error("Failed to check user_id. WASD.TV changed JSON format?")
+			if stream["user_id"] == user_id:
+				return stream
+		
+		return log.error("Failed to find media_container for user_id " + stream["user_id"])
+		
+	def getStreamInstancesArray(self, media):
```
back-to: looks to complicated, could be done without it.

src/streamlink/plugins/wasd.py
```diff
@@ -0,0 +1,147 @@
...
+			return log.error("Failed to get AUTH token")
+		
+		if re_video_id:
+			streams = self.getStreamsFromURL("https://wasd.tv/api/media-containers/" + re_video_id.group(1), 'video')
+		else:
+			streams = self.getStreamsFromURL("https://wasd.tv/api/media-containers?media_container_status=RUNNING&limit=1&offset=0&channel_id=" + re_channel_id.group(1) + "&media_container_type=SINGLE,COOP", 'channel')
+		
+		if streams:
+			for video in streams:
+				# TODO: This is how I planned it initially. video[0] is quality and video[1] is URL for .m3u8 file.
```
back-to: no need for such long comments and `video[0]` is not needed if you use 
`HLSStream.parse_variant_playlist`

src/streamlink/plugins/wasd.py
```diff
@@ -0,0 +1,147 @@
...
+		if not cookies or not cookies['cronos-auth-token'] or not cookies['cronos-auth-token-signature']:
+			log.debug("There is no AUTH token in cache. Attempting to request one")
+			self.getAuthToken()
+		if not cookies or not cookies['cronos-auth-token'] or not cookies['cronos-auth-token-signature']:
+			return log.error("Failed to get AUTH token")
+		
+		if re_video_id:
+			streams = self.getStreamsFromURL("https://wasd.tv/api/media-containers/" + re_video_id.group(1), 'video')
+		else:
+			streams = self.getStreamsFromURL("https://wasd.tv/api/media-containers?media_container_status=RUNNING&limit=1&offset=0&channel_id=" + re_channel_id.group(1) + "&media_container_type=SINGLE,COOP", 'channel')
```
back-to: a bit long, you could split the params from the url

src/streamlink/plugins/wasd.py
```diff
@@ -0,0 +1,147 @@
...
+	def _get_streams(self):
+		re_video_id = self._url_video.match(self.url)
+		re_channel_id = self._url_channel.match(self.url)
+		
+		if not re_video_id and not re_channel_id:
+			return log.error("Expected to see url like https://wasd.tv/channel/12345 or https://wasd.tv/channel/12345/videos/67890")
+		
+		# Check if we have AUTH token stored in session cache and request it
+		cookies = self.session.http.cookies
+		if not cookies or not cookies['cronos-auth-token'] or not cookies['cronos-auth-token-signature']:
```
back-to:
```py
if not (cookies.get('cronos-auth-token') and cookies.get('cronos-auth-token-signature')):
```

src/streamlink/plugins/wasd.py
```diff
@@ -0,0 +1,147 @@
...
+			log.debug("There is no AUTH token in cache. Attempting to request one")
+			self.getAuthToken()
```
back-to: > self.getAuthToken()

just open the website here instead of `self.getAuthToken()` and let it fail here if you don't have both cookies

src/streamlink/plugins/wasd.py
```diff
@@ -0,0 +1,147 @@
...
+			for video in streams:
+				# TODO: This is how I planned it initially. video[0] is quality and video[1] is URL for .m3u8 file. 
+				# But it doesn't work and I can't find channels with more than 1 quality to debug it. So for now it will be copy-pasted code from other plugins
+				# yield video[0], HLSStream(self.session, video[1])
+				
+				# This one is working. But I feel its not how it should be
+				return HLSStream.parse_variant_playlist(self.session, video[1])
+	
+	# WASD.TV requires token to make API requests. We will get it here
+	def getAuthToken(self):
```
back-to: looks to complicated, could be done without it.

src/streamlink/plugins/wasd.py
```diff
@@ -0,0 +1,147 @@
...
+			return log.error("Missing media_container_streams[0] in response from server for URL: " + URL)
+		
+		user_id = json["user_id"]
+		user_container = self.getUserContainerFromArray(user_id, json["media_container_streams"])
+		
+		if type == "channel" and user_container["stream_status"] != "RUNNING":
+			return log.debug("Stream on channel page is not live (recording of last stream?)")
+			
+		if not user_container["stream_media"]:
+			return log.error("stream_media is missing inside container. WASD.TV changed JSON format?")
```
back-to: you don't need the questions in the log errors

src/streamlink/plugins/wasd.py
```diff
@@ -0,0 +1,147 @@
...
+log = logging.getLogger(__name__)
+
+class WASD(Plugin):
+	_url_re = re.compile(r"https?://(?:www\.)?wasd\.tv/")
+	_url_channel = re.compile(r"https?://(?:www\.)?wasd\.tv/channel/(\d+)")
+	_url_video = re.compile(r"https?://(?:www\.)?wasd\.tv/channel/\d+/videos/(\d+)")
+	
+	@classmethod
+	def can_handle_url(cls, url):
+		return WASD._url_re.match(url)
```
back-to: `cls._url_re.match(url) is not None`

src/streamlink/plugins/wasd.py
Uinden: Explain this please.
Cookies `cronos-auth-token` and `cronos-auth-token-signature` are required to make API request.
To get those cookies I need to request `https://wasd.tv/api/auth/anon-token` and it will response with `set-cookie` headers with life time of 7 days.
What you mean by "just open the website"?

back-to: you are not reusing `self.getAuthToken` , so you don't need to create a new method for it

without it, it will be easier to read the code in the future in something breaks

---

you also don't need all the `if not res`
just use it with `res.cookies.get('XXX')`

---

some of the other methods are also not required

PR #2642, opened 2019-10-01
Remove direct installation instructions, link to docs

slhck:
Change default installation instructions to remove use of "sudo", as it is highly discouraged.

References:

- https://pages.charlesreid1.com/dont-sudo-pip/
- https://askubuntu.com/questions/802544/is-sudo-pip-install-still-a-broken-practice
- https://stackoverflow.com/questions/21055859/what-are-the-risks-of-running-sudo-pip

Other very popular packages also refraining from using "sudo" in their documentation:

- https://pyyaml.org/wiki/PyYAMLDocumentation
- https://github.com/psf/requests
- https://docs.scipy.org/doc/numpy/user/building.html

---

**Edit:** PR changed according to feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,27 +16,12 @@ services when they are available for the content you consume and enjoy.
 
 # [Installation][streamlink-installation]
 
-#### Installation via Python pip
-
-```bash
-sudo pip install streamlink
-```
-
-#### Manual installation via Python
-
-```bash
-git clone https://github.com/streamlink/streamlink
-cd streamlink
-sudo python setup.py install
-```
-
-#### Windows, MacOS, Linux and BSD specific installation methods
+Please refer to our documentation for different ways to install Streamlink:
 
 - [Windows][streamlink-installation-windows]
-- [Windows portable version][streamlink-installation-windows-portable]
-- [MacOS][streamlink-installation-others]
-- [Linux and BSD][streamlink-installation-linux]
-
+- [macOS][streamlink-installation-macos]
+- [Linux and BSD][streamlink-installation-linux-and-bsd]
+- [Source code][streamlink-installation-source]
 
 # Features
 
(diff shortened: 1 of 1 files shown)


Uinden:
Isn't it `pip install streamlink --user`?
Without `--user` it will fail. At least on my installation. 
Like ```error: could not create '/usr/lib/python3.7/site-packages/%package%': Permission denied```

slhck:
It depends!™️

If you're using a user-environment Python like from `pyenv` or Homebrew, or Pipenv, or a Virtualenv, then you don't need `--user`. On a “standard” Linux with system Python, you need `--user`.

If there's a danger that people might just do `sudo` if they see a “permission denied” error, then perhaps we should add `--user`. It doesn't hurt.

bastimeyer:
The install instructions on the repo's README are a remnant of the Livestreamer fork, I believe. Streamlink's documentation has a dedicated install page, where the `pip install --user` situation is explained, although kind of mixed with installing it from git, see here:
https://streamlink.github.io/install.html#source-code

Not sure if everyone else agrees with me here, but I think it would be better to remove the install commands from the README as a whole and just refer to the install documentation instead. Ideally, this should have been done when the the install page and Github release template got rewritten earlier this year, but I forgot about it.

I won't close any PRs this month due to Hacktoberfest, but if we agree on removing the commands, I would be happy if you could change this PR real quick. If not, then just removing the sudo part is also fine.

gravyboat:
@bastimeyer I agree. I think a link to the actual install docs would be better, then we don't have to maintain them in two locations.

slhck:
Thanks for your feedback everyone. I've changed the README to just link to the documentation for installation instructions. PR title changed accordingly.

slhck:
You're right. I fixed those links, removed the explicit mention of Windows portable (since it's linked from the first link anyways), added macOS and the source code installation.

gravyboat:
Looks good, thanks @slhck!

PR #2649, opened 2019-10-04
Added Canvas plugin

Mz49:
Recently, Canvas has grown in popularity as more schools and colleges pick it up. Many teachers and students use it share instructional content and other media. This adds a plugin that supports Canvas media_object URLs. For example:

`https://canvas.instructure.com/media_objects/m-Zeo1NE689yHQMYUXzRX4yIEtoNOIymxP/`

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -36,7 +36,7 @@ brightcove              players.brig... [6]_ Yes   Yes
 btsports                sport.bt.com         Yes   Yes   Requires subscription account
 btv                     btv.bg               Yes   No    Requires login, and geo-restricted to Bulgaria.
 canalplus               mycanal.fr           No    Yes   Streams may be geo-restricted to France.
-canvas                  instructure.com      No    Yes
+canvas                  instructure.com      --    Yes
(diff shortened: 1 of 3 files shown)


Mz49:
Fixing issues arising from the check.

beardypig:
Never heard of that site, but appears to be VOD only no live streams?

Mz49:
Yes, unfortunately that is the case...

gravyboat:
@beardypig I'm not sure what stance we should take here, on the one hand this is beneficial for students, but on the other we've taken a pretty hard stance against plugins that are VOD only. It seems hypocritical to allow this one when we haven't allowed others and have closed requests/PRs that add them. What do you think?

gravyboat:
I'm going to have to close this, sorry @Mz49. I've just closed a request where it was for VOD only content and it would be unfair and hypocritical of me not to reject this as well since we've taken the stance of focusing on live streams only.

Mz49:
@gravyboat Oh my! I apologize for my very delayed response. I understand and that's fine. I feel in the long run support VOD-only sites would be good for the users in general rather than having to depend on a fork for VOD.  Thanks though! Have a good day.

Issue #2450, opened 2019-04-27
Niconico Live


panhartstuff:
## Plugin Request

- [x] This is a plugin request and I have read the contribution guidelines and plugin request requirements.

### Description

https://live.nicovideo.jp/

Niconico Live is a very popular Japanese livestreaming service created by Nobuo Kawakami, the founder and CEO of Niconico. The total quantity of aired programs had reached 1 million within its first half year since user programs started, and it had reached 100 million on July 11th 2012. Nowadays, hundreds of thousands of user broadcasting programs are aired in a day.

It is a cornerstone of the Japanese internet culture; covering anime, computer games and pop music and more. By integrating this site as a plugin, it will hopefully open up opportunities for the preservation of important Niconico livestreams in the best quality possible.

### Example stream URLs

1. https://ch.nicovideo.jp/weathernews/live
2. https://ch.nicovideo.jp/ch2525
3. https://ch.nicovideo.jp/ch2640322
4. https://live.nicovideo.jp/watch/co2172061
5. https://live.nicovideo.jp/watch/co2267262
6. https://live.nicovideo.jp/watch/co3375579 

### Additional comments, screenshots, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

PR #2651, opened 2019-10-04
NicoLive: new plugin for Niconico Live

k9yyy:
A new plugin for Niconico Live streams. 

* This plugin handles URL like `https://live2.nicovideo.jp/watch/lv#########`.
* Both live and the Timeshift (VOD) feature are supported. 
* For content that requires login or premium access, user can provide login credentials or session token via command-line arguments.

https://github.com/streamlink/streamlink/issues/2450


beardypig:
Nice, thanks for the contribution. Is there any way you can decouple the `HLSStream` from the `Plugin`? Would be nice not to have to subclass the `HLSStream`, I know it's been done before (I've done it before) - just wondering if it can be avoided :)

k9yyy:
> Nice, thanks for the contribution. Is there any way you can decouple the `HLSStream` from the `Plugin`? Would be nice not to have to subclass the `HLSStream`, I know it's been done before (I've done it before) - just wondering if it can be avoided :)

Thanks for your response!

I subclassed `HLSStream` because I'd like to have access to `HLSStreamReader` from the `Plugin`, so that I can close the stream cleanly by calling `HLSStreamReader.close()`.

The stream URL will return 403 after the stream ends, so If I don't explicitly call `close()`, `HLSStreamReader` will keep running and attempt to retry, and finally quits after numerous retries.

src/streamlink/plugins/nicolive.py
```diff
@@ -0,0 +1,330 @@
...
+    """
+    proxy_options = {}
+    if purl:
+        p = urlparse(purl)
+        proxy_options['proxy_type'] = p.scheme
+        proxy_options['http_proxy_host'] = p.hostname
+        if p.port:
+            proxy_options['http_proxy_port'] = p.port
+        if p.username:
+            proxy_options['http_proxy_auth'] = (unquote_plus(
```
back-to: > undefined name 'unquote_plus'flake8(F821)

src/streamlink/plugins/nicolive.py
```diff
@@ -0,0 +1,330 @@
...
+from streamlink.compat import urlparse
+
+
+_log = logging.getLogger(__name__)
+
+_url_re = re.compile(
+    r"^http(|s)://(?P<domain>live[0-9]*\.nicovideo\.jp)/watch/lv[0-9]*")
+
+_login_url = "https://account.nicovideo.jp/login/redirector?" \
+             "show_button_twitter=1&" \
```
back-to: you could use this as **params** with `self.session.http.get(... params={'...': ...})` instead of making a long url with `\`

src/streamlink/plugins/nicolive.py
```diff
@@ -0,0 +1,330 @@
...
+from streamlink.plugin import Plugin, PluginArguments, PluginArgument
+from streamlink.plugin.api import useragents
+from streamlink.stream import HLSStream
+from streamlink.compat import urlparse
+
+
+_log = logging.getLogger(__name__)
+
+_url_re = re.compile(
+    r"^http(|s)://(?P<domain>live[0-9]*\.nicovideo\.jp)/watch/lv[0-9]*")
```
back-to: `https?`

k9yyy:
Sorry for the delay in response. Totally forgot about this...

I have improved the code and it's ready for review again.

k9yyy:
> there might be `[stream.hls][error] Failed to open segment` for people outside of Japan
> 
> increasing `--hls-segment-thread` fixes it

I think it's fine to just leave it, since `--hls-segment-thread` is a generic option, right? Or should I address this issue in the plugin if possible?

Although not the cause of this, there is no handling for unexpected disconnecting of the websocket API, which may also cause problems under bad network condition. If the API connection is broken, the session will expire. So I should make it auto-retry on unexpected disconnect events.

back-to:
> I think it's fine to just leave it, since --hls-segment-thread is a generic option, right?

Yes, it should be fine.

> Although not the cause of this, there is no handling for unexpected disconnecting of the websocket API, which may also cause problems under bad network condition. If the API connection is broken, the session will expire (and the stream url will be inaccessible). So I should make it auto-retry on unexpected disconnect events.

if it’s not too time consuming to change, but could also be added without it.

k9yyy:
> if it’s not too time consuming to change, but could also be added without it.

It shouldn't take long. I will add this feature.

k9yyy:
> > if it’s not too time consuming to change, but could also be added without it.
> 
> It shouldn't take long. I will add this feature.

Unexpected websocket disconnects are properly handled now. It will reconnect automatically on such events.

If there aren't other issues I think it should be ready to be merged.

PR #2652, opened 2019-10-04
stream.streamprocess: check that a process is usable before using it

beardypig:
StreamProcess provided an is_usable method, use that method to check if the process can be launch and give a more useful error message if not.

~I renamed RTMPStream to `rtmpdump`, because it uses that. I don't think it will be a problem as it's just a pretty name.~ @bastimeyer pointed out this was bad :)

### DIFF
--- a/src/streamlink/stream/rtmpdump.py
+++ b/src/streamlink/stream/rtmpdump.py
@@ -27,7 +27,6 @@ class RTMPStream(StreamProcess):
     def __init__(self, session, params, redirect=False, **kwargs):
         StreamProcess.__init__(self, session, params=params, **kwargs)
 
-        self.cmd = self.session.options.get("rtmp-rtmpdump")
         self.timeout = self.session.options.get("rtmp-timeout")
         self.redirect = redirect
 
@@ -42,6 +41,10 @@ class RTMPStream(StreamProcess):
                 self.parameters["debug"] = True
             else:
                 self.parameters["verbose"] = True
+
+    @property
+    def cmd(self):
+        return self.session.options.get("rtmp-rtmpdump")
 
     def __repr__(self):
         return "<RTMPStream({0!r}, redirect={1!r}>".format(self.parameters,
--- a/src/streamlink/stream/streamprocess.py
+++ b/src/streamlink/stream/streamprocess.py
@@ -1,7 +1,8 @@
 import logging
 import subprocess
-from operator import itemgetter
+import os.path
 
+from operator import itemgetter
 from streamlink.stream import Stream
 from streamlink.stream.wrappers import StreamIOThreadWrapper
 from streamlink.compat import devnull, which
@@ -39,7 +40,6 @@ class StreamProcess(Stream):
         """
         super(StreamProcess, self).__init__(session)
 
-        self.cmd = None
         self.parameters = params or {}
         self.arguments = args or []
         self.timeout = timeout
@@ -54,6 +54,10 @@ class StreamProcess(Stream):
             self.stderr = devnull()
 
     @property
+    def cmd(self):
+        raise NotImplementedError
+
+    @property
     def params(self):
         return self.parameters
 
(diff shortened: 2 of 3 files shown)


src/streamlink/stream/rtmpdump.py
```diff
@@ -21,7 +21,7 @@ class RTMPStream(StreamProcess):
     - :attr:`params` A :class:`dict` containing parameters passed to rtmpdump
     """
 
-    __shortname__ = "rtmp"
+    __shortname__ = "rtmpdump"
```
bastimeyer: Changing the Stream class's shortname is a breaking change, as it changes the expected/documented values of the player-passthrough and related parameters:
https://streamlink.github.io/cli.html#cmdoption-player-passthrough

src/streamlink/stream/streamprocess.py
```diff
@@ -62,19 +62,22 @@ def is_usable(cls, session):
...
+                if hasattr(self.stderr, "name"):
+                    raise StreamError(("Error while executing subprocess, "
+                                       "error output logged to: {0}").format(self.stderr.name))
+                else:
+                    raise StreamError("Error while executing subprocess")
 
-        return StreamProcessIO(self.session, process, process.stdout, timeout=self.timeout)
+            return StreamProcessIO(self.session, process, process.stdout, timeout=self.timeout)
+        else:
+            raise StreamError("{0} is not installed or not supported on your system".format(self.shortname()))
```
bastimeyer: You should probably use the value of the `rtmp-rtmpdump` option here instead of the class's shortname.

src/streamlink/stream/rtmpdump.py
beardypig: True that

src/streamlink/stream/streamprocess.py
beardypig: I would have done, but this is `StreamProcess`.

bastimeyer: You're right, I was looking at the wrong file locally. You should still be able to use `self.cmd` though if I'm not mistaken, since the implementing classes need to set `self.cmd`.

beardypig: Yep, it's not so well exposed, so I'll refactor it a bit...

src/streamlink/stream/streamprocess.py
```diff
@@ -62,19 +66,22 @@ def is_usable(cls, session):
...
+                if hasattr(self.stderr, "name"):
+                    raise StreamError(("Error while executing subprocess, "
+                                       "error output logged to: {0}").format(self.stderr.name))
+                else:
+                    raise StreamError("Error while executing subprocess")
 
-        return StreamProcessIO(self.session, process, process.stdout, timeout=self.timeout)
+            return StreamProcessIO(self.session, process, process.stdout, timeout=self.timeout)
+        else:
+            raise StreamError("{0} is not installed or not supported on your system".format(os.path.basename(self.cmd)))
```
back-to: `_check_cmd` might need an update or a cleanup,
it seems like it won't be triggered anymore at least with a rtmp example

https://github.com/streamlink/streamlink/blob/178d386f2f35a9c8d934690cc68f813a5cd77d9c/src/streamlink/stream/streamprocess.py#L163-L172

old

```
$ streamlink --rtmp-rtmpdump vlcc rtmp://localhost:33333
[cli][error] Try 1/1: Could not open stream <RTMPStream({'rtmp': 'rtmp://localhost:33333', 'flv': '-'}, redirect=False> (Could not open stream: Unable to find `vlcc' command)
error: Could not open stream <RTMPStream({'rtmp': 'rtmp://localhost:33333', 'flv': '-'}, redirect=False>, tried 1 times, exiting

```

new

```
$ streamlink --rtmp-rtmpdump vlcc rtmp://localhost:33333
[cli][error] Try 1/1: Could not open stream <RTMPStream({'rtmp': 'rtmp://localhost:33333', 'flv': '-'}, redirect=False> (Could not open stream: vlcc is not installed or not supported on your system)
error: Could not open stream <RTMPStream({'rtmp': 'rtmp://localhost:33333', 'flv': '-'}, redirect=False>, tried 1 times, exiting
```

PR #2655, opened 2019-10-06
ci/travis: install pynsist 2.4

bastimeyer:
This "bumps" the `pynsist` (Windows installer) dependency to `2.4` and downloads it from pypi instead of cloning the git repo from github and checking out a pre-release commit ID.

The old commit ID was added because pynsist's author didn't publish a new release when we needed the fixes earlier this year, see here:
https://github.com/takluyver/pynsist/commits/master
This has been resolved now since July.

This PR basically doesn't do much other than making the CI config more readable. Since we're going to add Github actions as a secondary CI service soon, I don't want to have mixed or weird dependency definitions, as it's confusing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
       pip install doctr;
     fi
   - if [[ $BUILD_INSTALLER == 'yes' ]]; then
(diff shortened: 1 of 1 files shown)


beardypig:
Cool. Looks good :)

gravyboat:

PR #2678, opened 2019-11-03
Add basic Twitch VOD support

legionxvx:
VOD support has been a requested feature before, I'm not sure if the developers are aware but The Twitch API supports this functionality easily, and only requires another Twitch API method to be made.

~I also threw in (basic) flatpak VLC support for fun because VLC on Fedora is... hard to acquire.~

See https://github.com/streamlink/streamlink/pull/2678#issuecomment-549158170

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,8 @@ nbproject
 *.sublime-workspace
 .idea
 
+.vscode
+
 *.mo
 *.egg-info
 *.egg
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -359,6 +359,12 @@ class Twitch(Plugin):
         simpler way of doing it now.
         """
                        ),
+        PluginArgument("check-vods",
+                       action="store_true",
+                       help="""
+        Interactive VOD check for the channel.
+        """
+        ),
         PluginArgument("disable-hosting",
                        action="store_true",
                        help="""
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -293,6 +293,9 @@ class TwitchAPI(object):
     def videos(self, video_id, **params):
         return self.call("/kraken/videos/{0}".format(video_id), **params)
 
+    def all_videos(self, channel_id, **params):
+        return self.call("/kraken/channels/{0}/videos".format(channel_id), **params)
+
     def channel_info(self, channel, **params):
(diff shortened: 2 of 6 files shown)


bastimeyer:
Thanks for taking the time to submit a pull request. There are a couple of issues though:

First, you are mixing three different changes into one pull request. Please don't do that. Merging such pull requests makes it unnecessarily difficult to properly find and understand the origin of the changes in the future.

Secord, regarding the changes to Twitch.tv VODs, you seem to have implemented a VOD browsing feature here. Streamlink is not a stream/vod browser and should only accept input URLs for a specific resource beforehand, either a live stream, or a VOD. It also already supports Twitch.tv VODs if you give it a proper Twitch.tv VOD input URL.
Redirecting a user to a VOD in case no live stream could be found is also problematic because of how Streamlink works. Watching VODs almost always requires the user to use the player-passthrough method, so that they can skip to a different position in a VOD from the player. You can't do this while using Streamlink regularly in its default configuration, as it pipes the fetched and concatenated HLS stream into the player's stdin channel.
Your plugin specific implementation also doesn't belong into streamlink_cli.
A better solution for this whole VOD fallback idea would be writing a custom wrapper module/script/application.

And third, regarding VLC and flatpak, it should already be supported by setting the required `run flatpak-app-id` parameters manually via `--player-args`. By adding `flatpak` to the list of VLC executable names, Streamlink would always assume VLC if the user had chosen flatpak as the player executable, even if they were using a different flatpak player application.

legionxvx:
Thanks for the feedback. As you can tell I am new to this.

Duly noted on the PR. I will endeavor to make it 1:1 (one feature to one PR) in the future. 

You are definitely correct in that streamlink _does_ support VODs, however they are not easily discoverable in the Twitch webpage, and not discoverable at all through streamlink's CLI. The crux of this particular PR was to provide a way for people to get to a channel's VODs easily without having to go to the actual webpage, track down the ID and then re-run the command. This still fulfills that need.

Also you are totally correct on the flatpak handling. It was short sighted and I didn't see the --player.args argument which is exactly what people should use.

gravyboat:
Yep I agree with @back-to and @bastimeyer. If you'd like to use this functionality feel free to fork Streamlink if you would like or make a wrapper application of some sort that meets your needs. Streamlink is not intended for this usage and we don't want to enhance VOD functionality further currently. I'm going to close this out since it won't get merged, sorry if you feel as though your time was wasted @legionxvx.

PR #2684, opened 2019-11-08
plugins.onetv:  adjust for akamai returned 403 "Forbidden" other way

vstavrinov:
streamlink 1tv.ru/live
[cli][info] Found matching plugin onetv for URL 1tv.ru/live
error: Unable to open URL: http://live1tvakamai.akamaized.net/dash-live11/streams/1tv/1tvdash.mpd?e=1573218827 (403 Client Error: Forbidden for url: http://live1tvakamai.akamaized.net/dash-live11/streams/1tv/1tvdash.mpd?e=1573218827)

### DIFF
--- a/src/streamlink/plugins/onetv.py
+++ b/src/streamlink/plugins/onetv.py
@@ -72,7 +72,7 @@ class OneTV(Plugin):
         if self.is_live:
             self.logger.debug("Loading live stream for {0}...", self.channel)
 
-            self.session.http.headers = {"origin": "http://1tv.ru"}
+            self.session.http.headers.update({"origin": "http://1tv.ru"})
             res = self.session.http.get(self.live_api_url, data={"r": random.randint(1, 100000)})
             live_data = self.session.http.json(res)
 
--- a/src/streamlink/plugins/onetv.py
+++ b/src/streamlink/plugins/onetv.py
@@ -72,6 +72,7 @@ class OneTV(Plugin):
         if self.is_live:
             self.logger.debug("Loading live stream for {0}...", self.channel)
 
+            self.session.http.headers = {"origin": "http://1tv.ru"}
             res = self.session.http.get(self.live_api_url, data={"r": random.randint(1, 100000)})
             live_data = self.session.http.json(res)
 


vstavrinov:
This is strange, but it happens occasionally for a few minutes or hours, that is it is not persistent problem. Fortunately adding the header provides stable operation.

src/streamlink/plugins/onetv.py
```diff
@@ -72,6 +72,7 @@ def _get_streams(self):
         if self.is_live:
             self.logger.debug("Loading live stream for {0}...", self.channel)
 
+            self.session.http.headers = {"origin": "http://1tv.ru"}
```
beardypig: This will overwrite the existing headers, including any set on the command line. 
```python
# before
{'User-Agent': 'python-requests/2.22.0', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive'}
# after
{'origin': 'http://1tv.ru'}
```

You might need to change/remove the `User-Agent` header though as that might be affecting it too. 

```suggestion
            self.session.http.headers.update({"origin": "http://1tv.ru"})
```

vstavrinov:
The header doesn't help. The issue is that api json returns  from time to time different cdn one of which works and other doesn't. The problem seen on the sites too.

back-to:
> The header doesn't help. The issue is that api json returns from time to time different cdn one of which works and other doesn't. The problem seen on the sites too.

the plugin seems to work and this is a random issue on there side as you said,
maybe there is not much we can do about it and that we need to fix.

**closing**, as this issue is open for a long time.

---

for people with issues, there is a dash stream which don't need a session and which is not geo blocked.

```
streamlink "https://edge2.1internet.tv/dash-live11/streams/1tv/1tvdash.mpd"
```

vstavrinov:
Eventually "origin" header required. Please, merge this PR. I hope it becomes the final state of the issue on the site side. In any case, it does not hurt.

vstavrinov:
Here is an example:

* **With** "origin" header:
```
streamlink --stream-url 1tv.ru/live
http://mobdrm.1tv.ru/hls-live8/streams/1tv/1tv5.m3u8?cdn=http://live1tvakamai.akamaized.net&netprav=/hls-vod11/streams/1tv/netprav.m3u8&e=1584715292&s=9akh7%2FBOIABBiIYcJoP%2Byrqg2s3cK9aYAm1bZhlIIC36Autg0qT%2FSa5i2xM%2FPiOkcGfEA0I%2FS492AGwwj4bvIT7sL0m6qH3OyDRg7ictlKWFscr7ip4fOehUB3a%2FYh3CaMrf0slHcD0qDBf4mRuazb8oNMHakqNtlGRyU%2FMNU9vSzqX3kxzIdyvBKIIqA1Bg
```
*  **Without** "origin" header:
```
streamlink --stream-url 1tv.ru/live
error: Unable to open URL: http://live1tvakamai.akamaized.net/dash-live11/streams/1tv/1tvdash.mpd?e=1584715724 (403 Client Error: Forbidden for url: http://live1tvakamai.akamaized.net/dash-live11/streams/1tv/1tvdash.mpd?e=1584715724)
```

I've tried this few times - it is reproducible.
As for the direct links, they all are not permanent. And there are no geo-restrictions, they just want "origin" header. I think it is a good idea always to provide it so as it is common to check if a request originating from the site.

Issue #2680, opened 2019-11-06
plugins.twitch: 410 Client Error (Gone for url)


gimli520:
This just started happening today sadly.

 Unable to open URL: https://api.twitch.tv/api/channels/sjow/access_token.json (410 Client Error: Gone for url: https://api.twitch.tv/api/channels/sjow/access_token.json?as3=t&platform=_)

This happens for any stream I try to start. Not sure what is wrong.

ThienJP:
same

cifvts:
It was working and suddenly this happened to me as well.

gimli520:
Well, guess I am not alone at the very least. Hopefully some light can be shed on how to fix it.

vertex101:
they have removed that from the api it looks like as when i visit the link it gives it shows this
`
{"error":"Gone","status":410,"message":"this API has been removed."}
`

gimli520:
How do we fix it?

bastimeyer:
Looks like Twitch has changed their private API endpoints which Streamlink needs to acquire the access token.

Please try to keep the number of posts low in this thread and only post informations which are necessary to fix this issue, so that this thread doesn't end in a big mess. Thanks!

bastimeyer:
Ok, this is weird.

If you `curl` the `access_token.json` request with the same headers and parameters which your browser makes when you're watching a stream on Twitch's website, it's returning 200 with the necessary JSON data.

However, if you're using an oauth token generated by Streamlink's or Streamlink Twitch GUI's API client_id, then it's returning a 410 error with `{"error":"Gone","status":410,"message":"this API has been removed."}`.

Providing no or in invalid oauth token results in an error 400 with `{"error":"Bad Request","status":400,"message":"No client id specified"}`.

This doesn't necessarily mean that Twitch is actively blocking Streamlink, but it's weird that their website is still using the same API endpoints when they remove them for 3rd party applications (without deprecating them).

I currently don't have access to my computer. This makes debugging this a bit painful. If somebody has more informations, please share.

KynikossDragonn:
Can we still make OAuth tokens for Streamlink? Because I had deleted mine when I noticed using a OAuth token caused the SurePlay advertisements to get injected into the m3u8 playlists where as using without a OAuth token made everything "ad free".

skulblakka:
The issue just resolved itself on my end while trying to find what's different between the site and streamlinks requests...

KynikossDragonn:
[...]

PR #2685, opened 2019-11-08
plugins.twitch: fix twitch plugin

161134050:
New client ID supplied by @ekimekim fixes twitch plugin


bastimeyer:
As I've said in https://github.com/streamlink/streamlink/issues/2680#issuecomment-551981368, this is the client-id of Twitch's website and Streamlink should not use that, as it's not intended for 3rd party applications to be used.

PR #2686, opened 2019-11-08
Added twitch plugin option --twitch-client-id to use custom API ID

Grumm:
#2680


skulblakka:
As far as I know this can already be achieved by using `--http-header`

bastimeyer:
@skulblakka I thought so too, but apparently the `Client-ID` header set here
https://github.com/streamlink/streamlink/blob/1.2.0/src/streamlink/plugins/twitch.py#L271
overrides the custom `Client-ID` header.

A custom client ID shouldn't be necessary though, since individual users can simply set an OAuth token, which takes precedence over the client-id anyway. OAuth tokens are tied to the client-id they've been generated with.

back-to:
#2692

PR #2692, opened 2019-11-14
plugins.twitch: fix API issue - 410 gone error

bastimeyer:
Since it should be pretty clear now that #2680 won't get fixed any time soon on Twitch's end or that it's probably even an intentional change (which "accidentally gets fixed" from time to time), we have to make a decision now.

As said in #2680, we don't really want to use Twitch's own client-id in Streamlink, but if we want to fix this issue, we don't have any other choice.

Twitch's developer ToS can be read here:
https://www.twitch.tv/p/legal/developer-agreement/

----

The changes of this PR make all private API calls use Twitch's own client-id and also remove authorization data from these API calls, since OAuth tokens are tied to the client-id which they have been granted from and take precedence when set. All other API calls are still using Streamlink's own client-id.

This means that OAuth tokens, unless granted from Twitch's client-id on their website, are entirely pointless in Streamlink now. Their only purpose was getting a stream access token with authorization data (meant for receiving item drops or being able to access private streams, etc), but this is not possible anymore, because there is no way of distinguishing between Streamlink's and Twitch's OAuth tokens, so they have to be disabled as a whole on the private API. And this further means that aquiring an OAuth token from Streamlink is pointless as well.

I haven't removed the entire authorization feature in this PR though. I hope this doesn't cause any confusion when people are trying to use --twitch-oauth-token (which still shows the login message, but is pointless now) or --twitch-oauth-authenticate. If this feature should get removed, we will also have to change the documentation in this regard.

----

- adds private=False parameter to TwitchAPI.call method
- uses Twitch's own client-id for all private API calls
- disables OAuth tokens on private API calls
- sets Authorization header instead of oauth_token parameter
- removes unnecessary platform=_ parameter from access_token call

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -39,7 +39,10 @@ QUALITY_WEIGHTS = {
     "mobile": 120,
 }
 
+# Streamlink's client-id used for public API calls (don't steal this and register your own application on Twitch)
 TWITCH_CLIENT_ID = "pwkzresl8kj2rdj6g7bvxl9ys1wly3j"
+# Twitch's client-id used for private API calls (see issue #2680 for why we are doing this)
+TWITCH_CLIENT_ID_PRIVATE = "kimne78kx3ncx6brgo4mv6wki5h1ko"
 
 _url_re = re.compile(r"""
     http(s)?://
@@ -254,11 +257,8 @@ class TwitchAPI(object):
     def add_cookies(self, cookies):
         self.session.http.parse_cookies(cookies, domain="twitch.tv")
 
-    def call(self, path, format="json", schema=None, **extra_params):
+    def call(self, path, format="json", schema=None, private=False, **extra_params):
         params = dict(as3="t", **extra_params)
-
-        if self.oauth_token:
-            params["oauth_token"] = self.oauth_token
 
         if len(format) > 0:
             url = "https://{0}.twitch.tv{1}.{2}".format(self.subdomain, path, format)
@@ -266,7 +266,12 @@ class TwitchAPI(object):
             url = "https://{0}.twitch.tv{1}".format(self.subdomain, path)
 
         headers = {'Accept': 'application/vnd.twitchtv.v{0}+json'.format(self.version),
-                   'Client-ID': TWITCH_CLIENT_ID}
+                   'Client-ID': TWITCH_CLIENT_ID if not private else TWITCH_CLIENT_ID_PRIVATE}
+
+        # OAuth tokens created from Streamlink's own client-id can't be used anymore on the private API (#2680)
+        # Since we don't know the origin of the provided OAuth token, we unfortunately need to disable all
+        if self.oauth_token and not private:
+            headers["Authorization"] = "OAuth {}".format(self.oauth_token)
 
         res = self.session.http.get(url, params=params, headers=headers)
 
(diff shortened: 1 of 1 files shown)


bastimeyer:
Btw, we still haven't cleaned up the old API stuff. I can't seem to find @back-to's PR with some of the cleaning work.

If we're removing features, we should also remove the stuff which has been unsupported / unused for a while now.

back-to:
should probably make a new version _1.3.0_ with this added

> Btw, we still haven't cleaned up the old API stuff.

can be done, but it is not important

src/streamlink/plugins/twitch.py
```diff
@@ -254,19 +257,21 @@ def __init__(self, session, beta=False, version=3):
...
             url = "https://{0}.twitch.tv{1}.{2}".format(self.subdomain, path, format)
         else:
             url = "https://{0}.twitch.tv{1}".format(self.subdomain, path)
 
         headers = {'Accept': 'application/vnd.twitchtv.v{0}+json'.format(self.version),
-                   'Client-ID': TWITCH_CLIENT_ID}
+                   'Client-ID': TWITCH_CLIENT_ID if not private else TWITCH_CLIENT_ID_PRIVATE}
+
+        # OAuth tokens created from Streamlink's own client-id can't be used anymore on the private API (#2680)
+        # Since we don't know the origin of the provided OAuth token, we unfortunately need to disable all
```
bastimeyer: Regarding what I said here, it \*is\* possible to check whether it's an OAuth token from Twitch's website or not. That would mean that we don't have to disable the `Authorization` header as a whole on the private API endpoints and people could still log in via an OAuth token from Twitch's website. We would have to make another API request for the auth permissions on the kraken root endpoint first though, which would delay launching streams by a bit when an OAuth token is provided. This does only work because certain permission scopes can only be granted by Twitch and thus are not available on 3rd party OAuth tokens, but I'm not sure if that's a proper and solid way, as those permissions can and probably will change over time. Thoughts?

back-to: > but I'm not sure if that's a proper and solid way, as those permissions can and probably will change over time. Thoughts?

don't think to complicated, working Streams should be enough for now.

---

yt-dl also uses the api key since 8-2018

https://github.com/ytdl-org/youtube-dl/commit/cb1c3a3c07b2afe854ecd8c97f77d8f62720d7a4

gravyboat:
@bastimeyer I still really don't like the idea of using Twitch's client ID as our official stance. I'm pretty sure doing so puts us in violation of the ToS where as before we've been able to avoid doing so since we use our own key. Using their client ID is a clear case of reverse engineering the site to pull that information. I'm wondering instead if we should just modify how Streamlink works to require the user to configure this from their own account. I know it will suck but it's the only legitimate alternative I can think of.

bastimeyer:
I don't like using it either, but the sad reality is that we have to at some point if these issues keep continuing, which they clearly do. Other big applications like youtube-dl are using it as well. It doesn't justify it of course, but asking every user of Streamlink who wants to watch Twitch streams to manually change Streamlink's client-id or asking them to open their web browser's dev tools and extracting a certain header is complete nonsense.

The OAuth tokens from the website are by the way "not legitimate" either, because they are generated from Twitch's own client-id and thus are "reverse engineered" too, if you're applying the same logic here. Streamlink has always used the same private API endpoint for aquiring the stream's access token, which means that we were "reverse engineering" the website the whole time, but just telling Twitch that we're not a user of the website by sending our own client-id instead.

So what would happen if this is indeed a ToS violation? If Streamlink's cliend-id should get revoked by Twitch if they decide to go against all third party applications which are accessing streams, then regular public API requests (kraken v5) would not be possible anymore. Those API requests are only needed for optional stuff though, like retrieving the account name after authentication, which is off the table anyway, or for getting stream/video metadata, like the title, streamer name, etc. Whether they would apply legal measures, I'm not sure.

Btw, the API rejects all third party client-ids / OAuth tokens, not just Streamlink. And it's also been down during the week, not just the weekend. We simply don't know why it is happening, which makes having a decision here so difficult. All we know is that it's usually happening around 9pm CET (whatever timezone it is for them) and that it lasts for several hours or days on the weekends (second weekend now).

I'd say that we should calm down and don't rush things. It's a stupid situation, but putting us into an even more stupid position would be a bad idea.

gravyboat:
@bastimeyer I agree that it's nonsense to ask that of users, but Twitch's lack of communication and clarity on this issue for an obvious failing in their public APIs is also nonsense. I know the OAuth tokens aren't legit either but at least in that scenario it's still being generated. The ID that we're sending as the Streamlink user was generated by their system from their dev tools and then used by us for our application so I feel like it at least has some more legitimacy even if it uses a non-public API. I didn't realize however that other apps like youtube-dl were using Twitch's key. I thought they had their own that they requested from the dev platform. That actually annoys me much more because we're basically being punished for trying to do things as properly as we can.

I know that others have been unable to use their creds as well, as you said we simply don't know why it is happening and they won't give us an answer because it's a "private" API that they make publicly available. I'm not in any rush either and am perfectly calm about it I just think we should discuss what options we have (if any to be seen that we are trying to be good "Twitch citizens" since I know they monitor this project). I'm not sure what you mean by putting us in an even more stupid position however, are you talking about pushing people to approve via the oauth? I agree that the whole situation is stupid.

bastimeyer:
I meant that by unnecessarily risking to break ToS and removing features, because it's still possible that this issue does get fixed eventually by Twitch.

gravyboat:
Ahh okay that makes sense and I agree.

h1z1:
To be clear, I can understand why Twitch would want to kill Streamlink and anything based on it due in part to the way it uses their API (there is no connection pooling for example).  However to understand why Twitch reacts the way it does, you have to think like a complete sociopathic moron.

Read their replies very carefully   and you'll start to understand..  though anyone with half a brain knows the API is used by many applications, in their tiny heads they didn't break anything because it was considered 'private'.   Randomly breaking something with increasing occurrences is literally part of the plan many developers / tech operations consider acceptable..  

They claim to be run by a community of volunteers (BarryCarlyon's responses to you are sraight ignorant)  -- On a site literally called "Twitch developers" none the less where supposed twitch staff post and despite being given a dubious name as a subdomain under twitch.tv.

tldr You will never "fix" this.  Create an app on Twitch, any app.  Use that application's client_id/whatever they require today(tm).    Rather ironic thing is how much ytdl and streamlink are used internally especially before Twitch even had vods.

gravyboat:
@bastimeyer Seems like this is going to stick around permanently now. It hasn't magically recovered for quite a few days now unlike previously. How much longer do you want to wait before we merge this?

bastimeyer:
Yes, it's pretty clear now that Twitch has made a decision and that it was not a mistake on their end (the dev forum answers were just useless non-staff answers, so waiting was justified). Let's not wait any longer and get it fixed here.

bastimeyer:
As I've said, the twitch oauth parameters will not be removed by this PR and will still be available as "placebo" parameters (except for public API calls where auth data is optional anyways).

Removing them would require a major version bump I think. We should add a note to the docs though. I'll see if I can do this later today so that we can have a new release. I'll have to update the Twitch GUI anyway today because of this.

gravyboat:
Okay. I'm going to merge this in then, I'm sure if @beardypig had any concerns they would have weighed in with concerns.

Issue #2703, opened 2019-11-18
stream/hls.py: name prefix is ignored when checking for alternate stream qualities


pmrowla:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description

https://github.com/streamlink/streamlink/blob/074c270d85c8da1581146671dd652704fc922069/src/streamlink/stream/hls.py#L442-L450

This code block will not be executed properly if `name_prefix` was passed into `parse_variant_playlist`.

### Expected / Actual behavior

The `if` conditional should be checking if `streams` contains `name_prefix + stream_name`

### Reproduction steps / Explicit stream URLs to test

Pass any `name_prefix` into `parse_variant_playlist()` for an HLS playlist containing multiple url's for a single stream name (quality)

### Log output

```
N/A
```

### Additional comments, screenshots, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

back-to:
> stream/hls.py: name prefix is ignored when checking for alternate stream qualities

it checks only the current `parse_variant_playlist` not other `parse_variant_playlist`

https://github.com/streamlink/streamlink/blob/074c270d85c8da1581146671dd652704fc922069/src/streamlink/stream/hls.py#L387

if you set `name_prefix` it will be `Available streams: test720p, test432p, test288p, test224p`

> The if conditional should be checking if streams contains name_prefix + stream_name

wouldn't do anything different as it only uses the current playlist and just adds the prefix at the end,
it would be the same if you add it beforehand.

---

not sure what you really what with this issue, but `name_prefix` doesn't do that much in HLSStream

maybe you got an example how you would use it.

pmrowla:
I noticed it when I was submitting PR #2702 

The m3u8 playlists for that source include two separate entries for 720p (one for 30fps and one for 60). That plugin needs to parse up to 5 separate m3u8 playlists (as one NHL game will have multiple TV broadcasts available) and each one of those m3u8 playlists will have the two 720p entries. I was using `name_prefix` to prepend `home_`, `away_`, etc. prefixes to differentiate the broadcasts.

The expected behavior would be that `parse_variant_playlist` would return a dict containing `home_720p` and `home_720p_alt` (where the `_alt` stream is the second 720p 60fps entry from the m3u8 playlist).

[...]

Issue #2701, opened 2019-11-17
Hls variant: ascii' codec can't encode characters


campones:
## Bug Report

- [ x] This is a bug report and I have read the contribution guidelines.

Reproduction steps / Explicit stream URLs to test
(using swiss ip)

```
streamlink "hlsvariant://https://streams.wilmaa.com/m3u8/get?channelId=m_6_hd"
[cli][info] streamlink is running as root! Be careful!
[cli][info] Found matching plugin hls for URL hlsvariant://https://streams.wilmaa.com/m3u8/get?channelId=m_6_hd
error: 'ascii' codec can't encode characters in position 4-5: ordinal not in range(128)
```

I am using streamlink 1.0.2 with Python 2.7.12 on ubuntu 16.04 

I m able to retrieve the video only with livestreamer, and to grab everything with ffmpeg.
It works fine with streamlink 1.0.2 on windows 64

Output of the master playlist:

```
#EXTM3U
#EXT-X-VERSION:5
#EXT-X-INDEPENDENT-SEGMENTS

#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="audio-mp4a.40.2",NAME="Français",DEFAULT=YES,AUTOSELECT=YES,LANGUAGE="fr",URI="t_track_audio_bw_128_num_0_p_10_l_fr.m3u8?z32=MNZWSZB5GE2UIN2FG4ZTEQKFHA2DIN2CIIWTONJYGFAUGNRSGAZTSMSFGAZEIJTNMF4HEYLUM$
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="audio-mp4a.40.2",NAME="Other (qad)",DEFAULT=NO,AUTOSELECT=YES,LANGUAGE="qad",URI="t_track_audio_bw_128_num_2_p_10_l_qad.m3u8?z32=MNZWSZB5GE2UIN2FG4ZTEQKFHA2DIN2CIIWTONJYGFAUGNRSGAZTSMSFGAZEIJTNMF4HE$


#EXT-X-STREAM-INF:BANDWIDTH=5000000,CODECS="avc1.4d4020,mp4a.40.2",RESOLUTION=1280x720,FRAME-RATE=50,AUDIO="audio-mp4a.40.2",CLOSED-CAPTIONS=NONE
[...]

PR #2705, opened 2019-11-19
stream.hls: Fix UnicodeDecodeError for log msg and name_prefix for stream_name

back-to:
Fixes https://github.com/streamlink/streamlink/issues/2701
Fixes https://github.com/streamlink/streamlink/issues/2703

not worth to create two PRs as it uses the same log line.


gravyboat:
Yeah I'm fine with this. Thanks @back-to.

PR #2722, opened 2019-12-01
Allow to disable twitch reruns

### LINKED ISSUE #2721: Twitch disable reruns
## Feature Request

### Description

There are disabling hosts (--twitch-disable-hosting), can you also add disabling reruns?

skulblakka:
This adds the option to disable twitch reruns similar to disabling hosted streams.

~I know the check for the `stream_type` could also be directly done in `_get_hls_streams`. If you like that better tell me and I can change it :)~

Edit: I decided to change it. This should have less changes and I think my previous solution broke the tests for the skip-ads option.

Closes #2721

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -703,21 +703,18 @@ class Twitch(Plugin):
             self.logger.info("{0} is hosting {1}".format(self.channel, host_info["target_login"]))
             return host_info["target_login"]
 
-    def _get_stream_type(self):
-        return self.api.streams(self.channel_id)["stream"]["stream_type"]
+    def _check_for_rerun(self):
+        return self.api.streams(self.channel_id)["stream"]["stream_type"] == "rerun"
 
     def _get_hls_streams(self, stream_type="live"):
         self.logger.debug("Getting {0} HLS streams for {1}".format(stream_type, self.channel))
         self._authenticate()
         self._hosted_chain.append(self.channel)
 
-        if stream_type == "live" or stream_type == "rerun":
-            if stream_type == "rerun" and self.options.get("disable_reruns"):
+        if stream_type == "live":
+            if self._check_for_rerun() and self.options.get("disable_reruns"):
                 self.logger.info("Reruns were disabled by command line option")
                 return {}
-            elif stream_type == "rerun":
-                self.logger.info("Starting rerun")
-                stream_type = "live"
 
             hosted_channel = self._check_for_host()
             if hosted_channel and self.options.get("disable_hosting"):
@@ -794,8 +791,7 @@ class Twitch(Plugin):
         elif self.clip_name:
             return self._get_clips()
         elif self._channel:
-            stream_type = self._get_stream_type()
-            return self._get_hls_streams(stream_type)
+            return self._get_hls_streams("live")
 
 
 __plugin__ = Twitch
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -375,6 +375,12 @@ class Twitch(Plugin):
         Skip embedded advertisement segments at the beginning or during a stream.
         Will cause these segments to be missing from the stream.
         """
+                       ),
+        PluginArgument("disable-reruns",
+                       action="store_true",
+                       help="""
+        Do not open the stream if the target channel is currently broadcasting a rerun.
+        """
                        ))
 
     @classmethod
@@ -697,12 +703,22 @@ class Twitch(Plugin):
             self.logger.info("{0} is hosting {1}".format(self.channel, host_info["target_login"]))
             return host_info["target_login"]
 
+    def _get_stream_type(self):
+        return self.api.streams(self.channel_id)["stream"]["stream_type"]
+
     def _get_hls_streams(self, stream_type="live"):
         self.logger.debug("Getting {0} HLS streams for {1}".format(stream_type, self.channel))
         self._authenticate()
         self._hosted_chain.append(self.channel)
 
-        if stream_type == "live":
+        if stream_type == "live" or stream_type == "rerun":
+            if stream_type == "rerun" and self.options.get("disable_reruns"):
+                self.logger.info("Reruns were disabled by command line option")
+                return {}
+            elif stream_type == "rerun":
+                self.logger.info("Starting rerun")
+                stream_type = "live"
+
             hosted_channel = self._check_for_host()
             if hosted_channel and self.options.get("disable_hosting"):
                 self.logger.info("hosting was disabled by command line option")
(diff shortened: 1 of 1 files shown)


gravyboat:
Thanks @skulblakka, I didn't realize the reruns were flagged by the stream type.

EnterGin:
Thank you!

src/streamlink/plugins/twitch.py
```diff
@@ -697,12 +703,19 @@ def _check_for_host(self):
...
+    def _check_for_rerun(self):
+        return self.api.streams(self.channel_id)["stream"]["stream_type"] == "rerun"
+
     def _get_hls_streams(self, stream_type="live"):
         self.logger.debug("Getting {0} HLS streams for {1}".format(stream_type, self.channel))
         self._authenticate()
         self._hosted_chain.append(self.channel)
 
         if stream_type == "live":
+            if self._check_for_rerun() and self.options.get("disable_reruns"):
```
bastimeyer: This will make an API call even if reruns are not disabled, which is unnecessary and will delay launching the stream by a bit.

src/streamlink/plugins/twitch.py
```diff
@@ -697,12 +703,19 @@ def _check_for_host(self):
             self.logger.info("{0} is hosting {1}".format(self.channel, host_info["target_login"]))
             return host_info["target_login"]
 
+    def _check_for_rerun(self):
+        return self.api.streams(self.channel_id)["stream"]["stream_type"] == "rerun"
```
bastimeyer: Btw, according to the [kraken v5 documentation](https://dev.twitch.tv/docs/v5/reference/streams#get-live-streams), supported values are `live`, `playlist` and `all`, but that's not true. There's also a second attribute, `broadcast_platform`, which isn't documented, but is related to reruns and IIRC, it was used by Twitch before they added the `stream_type` attribute. Both attributes have changed quite a bit over the past years after reruns were implemented by Twitch. For example, there also was the `watch_party` value which was used to indicate reruns.

Even if `stream_type` is the documented property, I am not sure if it's the right property to check. Most attribute names can be used for filtering streams via a GET parameter, like `game` for example, but `stream_type` can not be used for filtering whereas `broadcast_platform` can, which is weird. See here:
```bash
curl -s \
  -H "Accept: application/vnd.twitchtv.v5+json" \
  -H "Client-ID: phiay4sq36lfv9zu7cbqwz2ndnesfd8" \
  "https://api.twitch.tv/kraken/streams?stream_type=rerun&limit=1" | \
  jq '.streams[0] | {stream_type: .stream_type, broadcast_platform: .broadcast_platform}'
```
```json
{
  "stream_type": null,
  "broadcast_platform": null
}
```
```bash
curl -s \
  -H "Accept: application/vnd.twitchtv.v5+json" \
  -H "Client-ID: phiay4sq36lfv9zu7cbqwz2ndnesfd8" \
  "https://api.twitch.tv/kraken/streams?broadcast_platform=rerun&limit=1" | \
  jq '.streams[0] | {stream_type: .stream_type, broadcast_platform: .broadcast_platform}'
```
```json
{
  "stream_type": "rerun",
  "broadcast_platform": "rerun"
}
```

Then there's also the problem of broadcasters not setting the rerun flag while streaming a rerun. The Twitch GUI does some additional regex checks of the channel's status message (stream title) in addition to checking both `stream_type` and `broadcast_platform` attributes.

skulblakka: >Btw, according to the kraken v5 documentation, supported values are live, playlist and all

To be honest I totally forgot to check the kraken documentation. I was basically going by what the website was doing. Wasn't the playlist feature replaced by the reruns feature anyway? Maybe they are just no longer updating the docs?

>stream_type can not be used for filtering whereas broadcast_platform can, which is weird

Filtering by `stream_type` was removed on their Helix API a while ago. Maybe the same issues / reasons existed on Kraken too and they decided to remove it there as well?

>Then there's also the problem of broadcasters not setting the rerun flag while streaming a rerun.

My intention was to only disable reruns done via Twitch's rerun feature. Are you sure broadcasters have to flag their stream as a rerun when using it?

bastimeyer: > I totally forgot to check the kraken documentation

The documentation is complete garbage, so forget about that. Checking for `rerun` as value should be correct, but what I meant is that I'm not sure if it's enough since there's also the `broadcast_platform` attribute. The entire API is a total mess and you can't be 100% sure unless they document it properly, which they don't. Not sure about the playlist streams and whether they were removed, but there's also the `is_playlist` attribute. I think they added this after the playlist stream announcement and the documentation is just old and therefore still includes the `playlist` value.

> Filtering by `stream_type` was removed on their Helix API a while ago.

I haven't really followed the past developments on the helix API, but if you say that `stream_type` has been there and was removed (it's currently not there), then it's probably the right attribute to check on kraken and they've simply forgot to remove it there as well. The fact that it's been removed doesn't sound good though.

> Are you sure broadcasters have to flag their stream as a rerun when using it?

What I meant is that there are many channels which don't set the rerun flag while re-streaming a locally recorded stream and they just set a rerun stream title. See the various ESL channels for example (Dota2, CSGO, etc).

EnterGin:
There also a problem occurs when stream is offline

src/streamlink/plugins/twitch.py
gravyboat: Bah my bad @bastimeyer I totally forgot that it would cause an API call.

gravyboat:
@skulblakka Do you have time to try and fix this at some point? Otherwise maybe we should revert since this is kind of hard to test. I only know of a few streamers like followgrubby that regularly have reruns. Seems like the items bastimeyer noted might be biting us in the butt a bit here, might be worth adding some tests as well? I didn't think about how poorly patched together the API is when I merged this.

skulblakka:
I'm sorry for the mess I caused.

@bastimeyer Like I mentioned I assumed the OP of the issue was talking about reruns made via the Rerun-Feature. That's why I didn't consider "live"-streams that broadcasted a video. Getting streams by type was [removed March 2018](https://discuss.dev.twitch.tv/t/new-twitch-api-deprecating-get-streams-by-type/14554) on Helix (although the field there is only called `type`) and other values than `live` were removed from the documentation. Do you think filtering by `stream_type` is enough or should I also add `broadcast_platform`?

@gravyboat I have to leave for work now but I will try fixing this in the evening (and will also see about some tests).

gravyboat:
@skulblakka There is no need to apologize or any reason to stress about it. There's no release currently planned so we're not in a rush, I just wanted to clarify whether you wanted to work on it or if you were going to be too busy to do so and we should back it out. Don't pressure yourself or inconvenience your personal time.

bastimeyer:
I'll submit a PR with a fix later. Currently adding some tests... No pressure...

PR #2733, opened 2019-12-10
plugins.nhkworld: the site migrates from xml to json stream data

vstavrinov:
adapt to the site changes


vstavrinov:
may be py34 should be removed

gravyboat:
Py3.4 was removed in another PR so I'm going to merge this in.

Issue #2726, opened 2019-12-04
SHOWROOM: Questions about missing segments in Showroom-live.com


CplSgt82:
## Issue

- [x ] This is not a bug report, feature request, or plugin issue/request.
- [x ] I have read the contribution guidelines.

### Description

Problem: Streamlink will "skip" segments when downloading live streams (m3u8 playlist) from https://www.showroom-live.com/ 

There is no error msg, the log displayed "adding segment 500 to the queue" and then it will jump to "adding segment 502 to the queue" but i know that a segment is skipped because the video will freeze for a few seconds during playback. 

Addition info: Each segment is very short (less than 4s in playback) , and i have a very high ping to the server.
This "missing segment problem" occurs very frequently only on all hls playlist I download from this website, about 40 times average in a 2hr live stream. I have no such issue from other sites like youtube and bilibili.

Options tried: 
streamlink --loglevel debug --hls-segment-threads 10 --hls-segment-timeout 60.0 --hls-segment-attempts 60 --hls-live-edge 3 --ringbuffer-size 100M --hls-playlist-reload-attempts 3 https://hls-cdn17-showroom.cdn-dena.com/liveedge/7e41091a3cd59397262028d1c90bce30c6824df276d4e91b425fe0b72bfd0803/chunklist_v2.m3u8 best  -o file.mp4

I tried setting --hls-live-edge from 1 to 9, but the missing segments problem is still there. 

### Expected / Actual behavior

Questions: 
1. Any solutions to solve this missing segments problem? for eg, Is there any options in streamlink to force it to download segments in order without skipping? 
2. Is there any options to add more than 2 segments to the queue?
3. Any options in streamlink to keep hammering the server for 1 particular segment until it fetches it? 

### Reproduction steps / Explicit stream URLs to test

1. ...go to https://www.showroom-live.com/
2. ...Enter anyone of the live stream channels and fetch the m3u8 playlist. (i use stream detector FF plugin)
3. ...start streamlink and download to file 

### Logs

```
streamlink --loglevel debug --hls-segment-threads 10 --hls-segment-timeout 60.0 --hls-segment-attempts 60 --hls-live-edge 3 --ringbuffer-size 100M --hls-playlist-reload-attempts 3 https://hls-cdn17-showroom.cdn-dena.com/liveedge/7e41091a3cd59397262028d1c90bce30c6824df276d4e91b425fe0b72bfd0803/chunklist_v2.m3u8 best  -o file.mp4
.
.
.
.
[...]

PR #2735, opened 2019-12-11
Showroom: Detect Reload Duration

wlerin:
This addresses #2726 , though perhaps not in the most elegant fashion. 

A method of HLSStreamWorker is monkeypatched, and thus any future improvements to that method will not propagate to this plugin, however it hasn't been meaningfully changed in 3 years so I think that is an acceptable risk.

### DIFF
--- a/src/streamlink/plugins/showroom.py
+++ b/src/streamlink/plugins/showroom.py
@@ -79,7 +79,8 @@ _info_pages = set((
 ))
 
 
-# force shorter reload_time than TARGETDURATION
+# base reload time on segment duration rather than TARGETDURATION,
+# as the latter is consistently wrong
 class ShowroomHLSStreamWorker(hls.HLSStreamWorker):
     def process_sequences(self, playlist, sequences):
         first_sequence, last_sequence = sequences[0], sequences[-1]
@@ -89,8 +90,10 @@ class ShowroomHLSStreamWorker(hls.HLSStreamWorker):
 
         self.playlist_changed = ([s.num for s in self.playlist_sequences] !=
                                  [s.num for s in sequences])
-        self.playlist_reload_time = min(playlist.target_duration or 5,
-                                     last_sequence.segment.duration)
+        self.playlist_reload_time = max(
+            first_sequence.segment.duration,
+            last_sequence.segment.duration
+        )
         self.playlist_sequences = sequences
 
         if not self.playlist_changed:
--- a/src/streamlink/plugins/showroom.py
+++ b/src/streamlink/plugins/showroom.py
@@ -3,7 +3,8 @@ import re
 
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate, useragents
-from streamlink.stream import HLSStream, RTMPStream
+from streamlink.stream import RTMPStream
+from streamlink.stream import hls
 
 _url_re = re.compile(r'''^https?://
     (?:\w*.)?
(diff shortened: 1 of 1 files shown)


beardypig:
FYI for others, the diff of the function is 

```diff
**diff --git a/src/streamlink/stream/hls.py b/src/streamlink/stream/hls.py
index 1328d1ae..79600506 100644
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -221,8 +221,10 @@ class HLSStreamWorker(SegmentedStreamWorker):

         self.playlist_changed = ([s.num for s in self.playlist_sequences] !=
                                  [s.num for s in sequences])
-        self.playlist_reload_time = (playlist.target_duration or
-                                     last_sequence.segment.duration)
+        self.playlist_reload_time = max(
+            first_sequence.segment.duration,
+            last_sequence.segment.duration
+        )
         self.playlist_sequences = sequences

         if not self.playlist_changed:
```

beardypig:
I'd actually be tempted to change it to something like

```python
self.playlist_reload_time = min([s.segment.duration for s in sequences])
```
Perhaps the average duration rounded down would also be a good value. Why did you choose the `max` duration of the two @wlerin ?

src/streamlink/plugins/showroom.py
```diff
@@ -78,6 +79,46 @@
 ))
 
 
+# base reload time on segment duration rather than TARGETDURATION,
+# as the latter is consistently wrong
+class ShowroomHLSStreamWorker(hls.HLSStreamWorker):
+    def process_sequences(self, playlist, sequences):
```
bastimeyer: You should not rewrite the whole `process_sequences` method here. Once the parent class gets changed, your custom one can break. The `HLSStreamWorker` class is not that flexible unfortunately and needs a bit of refactoring, but for now you could move the reload time caluclation into a separate method and then just override that in your custom class, like it's been done in my Twitch low latency PR:
https://github.com/streamlink/streamlink/pull/2513/files#diff-7ddabf15fb3a46558218929a060133e0R225-R237

wlerin:
> I'd actually be tempted to change it to something like
> 
> ```python
> self.playlist_reload_time = min([s.segment.duration for s in sequences])
> ```
> 
> Perhaps the average duration rounded down would also be a good value. Why did you choose the `max` duration of the two @wlerin ?

I chose the max duration because that's closer to what the `TARGETDURATION` is *supposed* to be. Similarly I only used the first and last segment because accessing those segments is just a local lookup (probably inconsequential difference), and because at least right now, those are only two segments available at any given time anyway.

src/streamlink/plugins/showroom.py
wlerin: Right... but is that going to be merged soon? I didn't want to touch hls.py

bastimeyer: > but is that going to be merged soon?

The linked PR probably not, but that's why I meant that you should add the same change in your PR.

It's a reasonable change and there's no reason why your PR shouldn't get merged if you improve the basic HLS class with that necessary change. If you want to keep this PR as simple as possible, open another one with the HLS class change alone and rebase this one after the other has been merged.

back-to:
```py
self.playlist_reload_time = max(
            first_sequence.segment.duration,
            last_sequence.segment.duration
        )

self.playlist_reload_time = 2
```

won't work

```py
self.playlist_reload_time = 1.5
```

might work

PR #2738, opened 2019-12-14
docs/tests: remove python 3.4, use 3.8 and nightly for travis-ci

back-to:
Python 3.4 has reached end-of-life.
https://www.python.org/downloads/release/python-3410/


gravyboat:
Looks good to me. Thanks @back-to!

Issue #2712, opened 2019-11-24
[Bilibili Live Error] Unable to run on Linux server


Dreammer12138:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

There are always errors running on the Linux server, but there is nothing wrong with running on the PC side and on my raspberrypi 3B. 
I also checked the network connection and the python version and updated to the latest version. 
I also tried to install streamlink on different servers, but I got the same result.

### Reproduction steps / Explicit stream URLs to test

1. https://live.bilibili.com/13946381
2. https://live.bilibili.com/631
3. https://live.bilibili.com/4895312

### Log output

```
streamlink https://live.bilibili.com/13946381 -o test.mp4
[cli][info] streamlink is running as root! Be careful!
[cli][info] Found matching plugin bilibili for URL live.bilibili.com/13946381
error: Unable to validate JSON: Unable to validate key 'data': {'current_quality': 0, 'accept_quality': None, 'current_qn': 0} does not equal None or Key 'durl' not found in {'current_quality': 0, 'accept_quality': None, 'current_qn': 0}
```

### Additional comments, screenshots, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

bastimeyer:
> Use the `--loglevel debug` parameter and avoid using parameters which suppress log output.
> https://streamlink.github.io/cli.html#cmdoption-l

Dreammer12138:
```
streamlink --loglevel debug https://live.bilibili.com/13946381 -o test.mp4
[cli][info] streamlink is running as root! Be careful!
[cli][debug] OS:         Linux-4.15.0-70-generic-x86_64-with-Ubuntu-18.04-bionic
[cli][debug] Python:     3.6.8
[cli][debug] Streamlink: 1.3.0
[cli][debug] Requests(2.22.0), Socks(1.7.1), Websocket(0.56.0)
[cli][info] Found matching plugin bilibili for URL https://live.bilibili.com/13946381
error: No playable streams found on this URL: https://live.bilibili.com/13946381
```
I tried.

Dreammer12138:
I tested the get request part of the plug-in and found that the server-side Response is different from the client-side Response when get requests API_URL. 
I tried to add headers, but the result hasn't changed.
```
#plugins/bilibili.py
res = self.session.http.get(API_URL.format(room_id), headers=_headers_API)       <-here
[...]

Issue #2676, opened 2019-10-31
bilibili error


eller54:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

keep getting error message few days ago

### Reproduction steps / Explicit stream URLs to test

1. streamlink http://live.bilibili.com/544767 best
2. ...
3. ...

### Log output

```
[cli][info] Found matching plugin bilibili for URL http://live.bilibili.com/544767
[cli][info] Available streams: source_alt, source_alt2, source (worst, best)
[cli][info] Opening stream: source (http)
[cli][error] Try 1/1: Could not open stream <HTTPStream('https://ws.live-play.acgvideo.com
/live-ws/706128/live_307431_5209301.flv?wsSecret=95c1ddaca4779fbc825afff9199f87f6&wsTime=1
572565769&trid=ed3f36c02cb14e1380941a9e9209a26b&pt=web&oi=619134137&order=1&sig=no')> (Cou
ld not open stream: Unable to open URL: https://ws.live-play.acgvideo.com/live-ws/706128/l
ive_307431_5209301.flv?wsSecret=95c1ddaca4779fbc825afff9199f87f6&wsTime=1572565769&trid=ed
3f36c02cb14e1380941a9e9209a26b&pt=web&oi=619134137&order=1&sig=no (459 Client Error: Unkno
wn for url: https://3geauymtsgrzdrcbzfahug.ourdvsss.com/ws.live-play.acgvideo.com/live-ws/
706128/live_307431_5209301.flv?wsSecret=95c1ddaca4779fbc825afff9199f87f6&wsTime=1572565769
&trid=ed3f36c02cb14e1380941a9e9209a26b&pt=web&oi=619134137&order=1&sig=no&wshc_tag=0&wsts_
tag=5dbb64f9&wsid_tag=24e73cb9&wsiphost=ipdbm))
error: Could not open stream <HTTPStream('https://ws.live-play.acgvideo.com/live-ws/706128
/live_307431_5209301.flv?wsSecret=95c1ddaca4779fbc825afff9199f87f6&wsTime=1572565769&trid=
ed3f36c02cb14e1380941a9e9209a26b&pt=web&oi=619134137&order=1&sig=no')>, tried 1 times, exi
ting
```

### Additional comments, screenshots, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

Afanyiyu:
The same question. 
[...]

PR #2740, opened 2019-12-15
plugins.bilibili: fix Livestreams with status 1 (set Referer)

back-to:
closes https://github.com/streamlink/streamlink/issues/2676
closes https://github.com/streamlink/streamlink/issues/2712


gravyboat:
Thanks @back-to.

Issue #2694, opened 2019-11-15
[YouTube] 404 Client Error: Not Found for url


jack-michaud:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description

When trying to stream a certain video (not live, already uploaded), I am getting an unusual error. After streamlink selects a stream and dispatches it to ffmpegmux, it says it's unable to open the stream URL (404 Client Error: Not Found for url).

### Expected / Actual behavior

I expected streamlink to stream the video.

### Reproduction steps / Explicit stream URLs to test

`streamlink -l debug https://www.youtube.com/watch\?v\=9lb0gZRO2Go best`
 

### Log output

```
[cli][debug] OS:         Linux-4.15.0-58-generic-x86_64-with-Ubuntu-18.04-bionic
[cli][debug] Python:     2.7.15+
[cli][debug] Streamlink: 1.2.0
[cli][debug] Requests(2.22.0), Socks(1.7.0), Websocket(0.56.0)
[cli][info] Found matching plugin youtube for URL https://www.youtube.com/watch?v=9lb0gZRO2Go
[plugin.youtube][debug] Video ID from URL
[plugin.youtube][debug] Using video ID: 9lb0gZRO2Go
[plugin.youtube][debug] get_video_info - 1: Found data
[plugin.youtube][debug] MuxedStream: v 137 a 251 = 1080p
[plugin.youtube][debug] MuxedStream: v 299 a 251 = 1080p60
[plugin.youtube][debug] MuxedStream: v 302 a 251 = 720p60
[plugin.youtube][debug] MuxedStream: v 303 a 251 = 1080p60
[cli][info] Available streams: audio_webm, audio_mp4, 360p (worst), 720p, 720p60, 1080p, 1080p60 (best)
[cli][info] Opening stream: 1080p60 (muxed-stream)
[stream.ffmpegmux][debug] Opening http substream
[...]

PR #2741, opened 2019-12-15
plugins.youtube: Remove itag 303

back-to:
1080p60 video/webm format that doesn't work

example video: https://www.youtube.com/watch?v=9lb0gZRO2Go

closes https://github.com/streamlink/streamlink/issues/2694


gravyboat:
Good find, thanks @back-to.

PR #2742, opened 2019-12-15
plugins.ustream: Added support for video.ibm.com

back-to:


gravyboat:
@back-to Do we really need to support this? Seems like the view count on most of these streams is basically 0. What's the benefit of adding this?

beardypig:
UStream _is_ IBM Video - I think it should be supported :p

PR #2745, opened 2019-12-17
Fox

calculon-jr:
Added support for fox.com live streaming channels


src/streamlink/plugins/fox.py
```diff
@@ -0,0 +1,93 @@
...
+        headers = {
+            'x-api-key': x_api_key,
+            'Authorization': 'Bearer %s' % access_token,
+            'x-dcg-udid': segment_device_id,
+        }
+
+        return headers
+
+    def _get_streams(self):
+        slug = self.url_re.match(self.url)['slug'].upper()
```
beardypig: Python 2 compatibility again :)

src/streamlink/plugins/fox.py
```diff
@@ -0,0 +1,93 @@
...
+        segment_device_id = None
+
+        if 'mvpd-auth' in self.session.http.cookies:
+            mvpd_auth = self.session.http.cookies['mvpd-auth']
+
+        if 'segment-device-id' in self.session.http.cookies:
+            segment_device_id = self.session.http.cookies['segment-device-id']
+
+        if not mvpd_auth or not segment_device_id:
+            self.logger.error("Login cookie(s) missing. Provide 'mvpd-auth' and 'segment-device-id' cookies with --http-cookie.")
```
beardypig: Python logging API again.

src/streamlink/plugins/fox.py
```diff
@@ -0,0 +1,93 @@
...
+        # keep above zero in case current show has ended
+        program_index = 1
+
+        epg_listing_url = channels[channel_index]['items']['member'][program_index]['@id']
+        epg_listing_res = self.session.http.get(epg_listing_url, headers=headers)
+
+        live_player_url = epg_listing_res.json()['livePlayerScreenUrl']
+
+        try:
+            live_player_res = self.session.http.get(live_player_url, headers=headers) # 403 error here if channel is not included in cable subscription
```
beardypig: If you prefer, instead of catching the Exception you can allow 403 errors to pass and then check the `status_code`. You can pass `acceptable_status=(200, 403)` as an argument to `get` and then check `live_player_res.status_code` to see if it was 403. Other status codes are not expected and will be treated as exceptions, you can just `return` in the case of 403 and no streams will be found.

src/streamlink/plugins/fox.py
```diff
@@ -0,0 +1,93 @@
...
+
+        except PluginError as ex:
+            if ex.err.response.status_code == 403:  # pylint: disable=no-member
+                self.logger.error("Error 403: Forbidden. This is most likely because your cable subscription does not have access to this channel.")
+            raise ex
+
+        the_platform_url = live_player_res.json()['url']
+        the_platform_res = self.session.http.get(the_platform_url)
+
+        stream_url = the_platform_res.json()['playURL']
```
beardypig: As a general rue, `self.session.http.json(the_platform_res)` would be preferred as it automatically handles different (out of spec) character encodings.  

Same for the other line above :)

PR #2746, opened 2019-12-17
Discovery

calculon-jr:
Added support for Discovery and related channel live streams


src/streamlink/plugins/discovery.py
```diff
@@ -0,0 +1,72 @@
...
+    ''', re.VERBOSE)
+
+    api_url = 'https://api.discovery.com/v1/content/livestreams?slug='
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls.url_re.match(url) is not None
+
+    def _get_streams(self):
+        slug = self.url_re.match(self.url)['slug'].lower()
```
beardypig: We need to maintain Python 2 compatibility, you need to use `groups()` as Python 2 regex match items don't support `__getitem__`.

src/streamlink/plugins/discovery.py
```diff
@@ -0,0 +1,72 @@
...
+        else:
+            stream_url = link_data[0]['links'][0]['href']
+
+        try:
+            self.session.http.get(stream_url)  # check for 401 error
+
+        except PluginError as ex:
+            if ex.err.response.status_code == 401:  # pylint: disable=no-member
+                # some Discovery sites have a separate eosAf cookie that does not work with other sites
+                self.logger.error("Error 401: Unauthorized. This is most likely because you used an eosAf cookie from another Discovery site or one that has expired.")
```
beardypig: Please use the standard Python logging API.

src/streamlink/plugins/discovery.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls.url_re.match(url) is not None
+
+    def _get_streams(self):
+        slug = self.url_re.match(self.url)['slug'].lower()
+        eosAf = None
+
+        if 'eosAf' in self.session.http.cookies:
```
beardypig: You could also do this with a `PluginArgument` to make it slightly more user-friendly.

PR #2747, opened 2019-12-17
plugins.pluto: new plugin for https://pluto.tv/live-tv/

calculon7:
Added support for pluto.tv


src/streamlink/plugins/pluto.py
```diff
@@ -0,0 +1,36 @@
...
+    url_re = re.compile(r'https?://(?:www\.)?pluto\.tv/live-tv/(?P<slug>[a-zA-Z0-9-]+)')
+
+    api_url = 'https://api.pluto.tv/v2/channels'
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls.url_re.match(url) is not None
+
+    def _get_streams(self):
+        slug = self.url_re.match(self.url)['slug'].lower()
```
beardypig: This won't work in Python 2.7, which we are still supporting.

src/streamlink/plugins/pluto.py
```diff
@@ -0,0 +1,36 @@
...
+
+        channels_res = self.session.http.get(self.api_url)
+        channels_data = channels_res.json()
+        channel_match = next(filter(lambda x: x['slug'] == slug, channels_data), None)
+
+        if not channel_match:
+            self.logger.error('Channel %s not found. It may have been removed.' % slug)
```
beardypig: You should use the standard python logging api, the `plugin.logger` api has been deprecated.

```
import logging
log = loggering.getLogger(__name__)

...

log.error(....)
```

src/streamlink/plugins/pluto.py
```diff
@@ -0,0 +1,113 @@
...
+            movies_data = self.session.http.json(movies_res, schema=self.movies_schema)
+
+            for category in movies_data['categories']:
+                for movie in category['items']:
+                    if movie['slug'] == slug:
+                        match = movie
+
+        return self.handle_media_match(match)
+
+    def handle_media_match(self, match):
```
back-to: no point in creating this method just use `_get_streams`

src/streamlink/plugins/pluto.py
```diff
@@ -0,0 +1,113 @@
...
+            re.match(cls.url_re_live, url)
+            or re.match(cls.url_re_tv, url)
+            or re.match(cls.url_re_movie, url)
+        )
+
+    def _get_streams(self):
+        match = None
+
+        # live TV
+        if re.match(self.url_re_live, self.url):
```
back-to: you don't need `re.`

`self.url_re_live.match(self.url)`

src/streamlink/plugins/pluto.py
```diff
@@ -0,0 +1,113 @@
...
+        if not match:
+            return
+
+        if 'stitched' not in match:
+            log.error('No streams found. This video may use an unsupported format.')
+            return
+
+        stream_link_no_sid = match['stitched']['urls'][0]['url']
+        sid = str(uuid4())
+        stream_link = stream_link_no_sid.replace('&sid=', '&sid=' + sid)
```
back-to: you could use `update_qsd`

https://github.com/streamlink/streamlink/blob/ad276c4eefae7aea23428ca927952ece7c619f34/src/streamlink/utils/url.py#L66
https://github.com/streamlink/streamlink/blob/master/tests/test_utils_url.py#L45-L48

calculon-jr: Whenever I try to use `update_qsd`, it removes all the url params with no value, which breaks this plugin. 

So `update_qsd('http://url.com?param1=&param2=', {'param1': '', 'param2': 'value'})` becomes `http://url.com?param2=value`

`{'param1': None}` also gives the same result

beardypig: Sounds like a bug in `update_qsd`...

gravyboat:
No movement for 2 months, closing.

gravyboat:
@back-to Do you want to open this back up and merge it now that #2892 is in?

back-to:
@gravyboat this PR is outdated, I will open a different PR soon.

PR #2760, opened 2020-01-06
plugins.galatasaraytv: Add support for GALATASARAY SK TV live streams 

yavuzKomecoglu:
Added free live stream [GALATASARAY SK](https://galatasaray.com/) support.
Tested. 360p (worst), 480p, 720p (best) quality is also broadcast continuously.

### DIFF
--- a/src/streamlink/plugins/canalplus.py
+++ b/src/streamlink/plugins/canalplus.py
@@ -55,9 +55,9 @@ class CanalPlus(Plugin):
         parsed = []
         headers = {'User-Agent': self._user_agent}
 
-        # Some videos may be also available on Dailymotion (especially on CNews)
+        # Some videos may be also available on  (especially on CNews)
         if videos['ID_DM'] != '':
-            for stream in self.session.streams('https://www.dailymotion.com/video/' + videos['ID_DM']).items():
+            for stream in self.session.streams('https://www..com/video/' + videos['ID_DM']).items():
                 yield stream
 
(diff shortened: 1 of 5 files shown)


src/streamlink/plugins/canalplus.py
```diff
@@ -55,9 +55,9 @@ def _get_streams(self):
         parsed = []
         headers = {'User-Agent': self._user_agent}
 
-        # Some videos may be also available on Dailymotion (especially on CNews)
+        # Some videos may be also available on  (especially on CNews)
         if videos['ID_DM'] != '':
-            for stream in self.session.streams('https://www.dailymotion.com/video/' + videos['ID_DM']).items():
+            for stream in self.session.streams('https://www..com/video/' + videos['ID_DM']).items():
```
bastimeyer: This looks like you've accidentially replaced the string "DailyMotion" with "" in the entire project.

yavuzKomecoglu: Yes, a single file was accidentally modified. Sorry. Thanks for the fix.

yavuzKomecoglu:
@bastimeyer What are my obligations to be a merge? will an action be taken?

src/streamlink/plugins/canalplus.py
bastimeyer: Which fix? I've requested a change.

yavuzKomecoglu: I undoed the change made by mistake.

PR #2765, opened 2020-01-12
cli.output: fix mpv player parameter format

### LINKED ISSUE #2764: Change in mpv options for 2 dashes options breaks it connection to Streamlink
lately mpv changed the way they handle [2 dashes options](https://github.com/mpv-player/mpv/commit/d3cef97ad38fb027262a905bd82e1d3d2549aec7)

```
 --- mpv 0.31.1 ---
    - change behavior when using legacy option syntax with options that start
      with two dashes (``--`` instead of a ``-``). Now, using the recommended
      syntax is required for options starting with ``--``, which means an option
      value must be strictly passed after a ``=``, instead of as separate
      argument. For example, ``--log-file f.txt`` was previously accepted and
      behaved like ``--log-file=f.txt``, but now causes an error. Use of legacy
      syntax that is still supported now prints a deprecation warning.
```

So now mpv just closes with no errors when used with streamlink as streamlink creates an mpv command with --title "title" instead of --title="title"

```
[15:57] zouhair@box <5402:63> [~] -> :)
┖╴$ streamlink --verbose-player --loglevel debug https://www.twitch.tv/videos/533706108 best
[cli][debug] OS:         CYGWIN_NT-10.0-18362-3.1.2-340.x86_64-x86_64-64bit-WindowsPE
[cli][debug] Python:     3.7.4
[cli][debug] Streamlink: 1.3.0
[cli][debug] Requests(2.22.0), Socks(1.7.0), Websocket(0.56.0)
[cli][info] Found matching plugin twitch for URL https://www.twitch.tv/videos/533706108
[plugin.twitch][debug] Getting video HLS streams for gamesdonequick
[cli][info] Available streams: audio, 160p (worst), 360p, 480p, 720p, 720p60, 1080p60 (best)
[cli][info] Starting player: /home/zouhair/mpv.exe --cache 2048
[cli.output][debug] Calling: /home/zouhair/mpv.exe --cache 2048 --title https://www.twitch.tv/videos/533706108 https://vod-secure.twitch.tv/7dc8f3a102cc1793f7bc_gamesdonequick_36629634272_1358097847/chunked/index-dvr.m3u8
[15:57] zouhair@box <5403:64> [~] -> :)
┖╴$ echo $?
0
[15:57] zouhair@box <5404:65> [~] -> :)
┖╴$
```

bastimeyer:
Fixes #2764

See https://github.com/mpv-player/mpv/commit/d3cef97ad38fb027262a905bd82e1d3d2549aec7

The MPV change is supposed to print out a deprecation warning, but if it causes the player to immediately terminate as described in #2764, this would mean that we should publish a patch release soon. As I've said on Gitter, such a breaking change with the parameter parsing should actually require a major version bump on the side of MPV. Weird.

### DIFF
--- a/src/streamlink_cli/output.py
+++ b/src/streamlink_cli/output.py
@@ -190,7 +190,7 @@ class PlayerOutput(Output):
             if self.player_name == "mpv":
                 # see https://mpv.io/manual/stable/#property-expansion, allow escaping with \$, respect mpv's $>
                 self.title = self._mpv_title_escape(self.title)
-                extra_args.extend(["--title", self.title])
+                extra_args.append("--title={}".format(self.title))
 
             # potplayer
             if self.player_name == "potplayer":
@@ -202,7 +202,7 @@ class PlayerOutput(Output):
 
         args = self.args.format(filename=filename)
         cmd = self.cmd
-        
+
         # player command
         if is_win32:
             eargs = maybe_decode(subprocess.list2cmdline(extra_args))
--- a/tests/test_cmdline_title.py
+++ b/tests/test_cmdline_title.py
@@ -26,7 +26,7 @@ class TestCommandLineWithTitlePOSIX(CommandLineTestCase):
 
     def test_open_player_with_title_mpv(self):
         self._test_args(["streamlink", "-p", "/usr/bin/mpv", "--title", "{title}", "http://test.se", "test"],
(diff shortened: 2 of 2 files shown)


gravyboat:
That really is odd they would make this change without a major version bump.  Either way, looks good to me. Thanks @bastimeyer.

Issue #2769, opened 2020-01-16
Cannot play with mpv on Linux and cannot set format from text config


Digitalone1:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description
Using streamlink on Arch Linux and I'm not able to play Twitch streams with mpv.
I created a text config inside `~/config/streamlink/config` with only `player=mpv`. It's parsed by command line, but streamlink is not able to start mpv:

```
[cli][info] Available streams: audio_only, 160p (worst), 360p, 480p, 720p, 720p60, 1080p60 (best)
[cli][info] Opening stream: 1080p60 (hls)
[cli][info] Starting player: mpv
error: Failed to start player: mpv (Process exited prematurely)
[cli][info] Closing currently open stream...

```

Tried also with `player=/usr/bin/mpv` but nothing changed.

Besides, I can't find a way to set a preferred format from the text configuration. Is this intended or am I missing something?
`STREAM=1080p60` is not parsed, `stream-sorting-excludes=">1080p"` gives me error, but reading documentation should be correct. Is there a way to specify a format from text config? I can do it with youtube-dl. Something like `-f best[height<=1080]`

In addition, the parameter reported [here](https://streamlink.github.io/issues.html#issues-player-caching) regarding mpv is wrong. As stated [here](https://mpv.io/manual/master/#options-cache), --cache parameter only accept _yes_, _no_ and _auto_.

Digitalone1:
Okay, I solved compiling streamlink-git from AUR. So old version is not working anymore with mpv.

Please, release a new version, so Arch upstream repository will be updated.

Found out `default-stream` option to select a specified format from text config.

I'd like to ask something different: `ringbuffer-size` option will cache the stream to an amount of size before sending it to mpv or mpv will read instantly without waiting? I know I can set a specific cache on mpv, but it's a little tricky and I'd like to set something on streamlink if it's possible, but I'm not sure if it's working as I intended.

I specified `ringbuffer-size=20M`, but mpv is starting very fast, so I assume it is not waiting.

Thanks

bastimeyer:
> Cannot play with mpv on Linux

The issue has just been fixed in #2765 (for the MPV 0.30.1 release). There will be a Streamlink patch release soon. We are aware of this issue. The MPV devs are breaking the semantic versioning scheme with this change btw, as it would usually require a major version bump on their side, even if they called the old parameter format "legacy".

[...]

PR #2771, opened 2020-01-16
docs: fix MPV parameters on common issues page

bastimeyer:
and add link to --player-args parameter

See #2769


bastimeyer:
Not sure if we should also add a note regarding MPV's legacy parameter syntax.

gravyboat:

PR #2792, opened 2020-02-08
Added support for ATV Mas, peruvian news-channel

malvinas2:


back-to:
isn't this covered by https://github.com/streamlink/streamlink/pull/2791 ?

malvinas2:
That is a very good question. In fact the channel ATV and ATV más belong to the same group Albavision. But in contrast to other websites of this group, the site of ATV contains two live-streams: The normal channel called ATV and a news channel called ATV mas. There is no difference in the url to call the streams, instead of this the user chooses the desired channel by clicking on a button. 
The albavision plugin does a search for the expression "Math.floor", which in case of ATV appears two times. The same rules for the expression "rsk", used in the method _get_token_url to detect the page which is generating the token: It will be found two times. But the albavision plugin assumes there is just one hit for Math.floor and rsk. 
I didn't know how to handle this case. I imagine one possibility would have been to introduce some kind of additional command line parameter like "streamlink http://www.atv.pe/envivo _atvmas_ best", but that seemed to complicated to me (and I wouldn't have known how to realize this....) 
Instead of that I added a second plugin, called by "streamlink http://www.atvmas.pe/envivo best", although the used url "atvmas.pe" in reality does not exist. 
Let me know if there is a more elegant solution.

back-to:
>  There is no difference in the url to call the streams, instead of this the user chooses the desired channel by clicking on a button.

- http://www.atv.pe/envivo/ATV
- http://www.atv.pe/envivo/ATVMas

> The albavision plugin does a search for the expression "Math.floor", which in case of ATV appears two times. The same rules for the expression "rsk", used in the method _get_token_url to detect the page which is generating the token: It will be found two times. But the albavision plugin assumes there is just one hit for Math.floor and rsk.

just use the 2nd rsk/token for ATVMas and the first for ATV

---

closing, use one PR https://github.com/streamlink/streamlink/pull/2791

malvinas2:
> * http://www.atv.pe/envivo/ATV
> * http://www.atv.pe/envivo/ATVMas

Oooops.....

PR #2795, opened 2020-02-08
plugin.api: use Firefox as default User-Agent instead of python-requests

back-to:
- User-Agent update

---

less robot detection for http calls without a User-Agent


gravyboat:
Thanks @back-to.

Issue #2775, opened 2020-01-18
filmon plugin 504 Server Error: Gateway Time-out


johnthecracker:
## Bug Report

filmon plugin, 504 Server Error: Gateway Time-out

### Description

Here is filmon plugin bug for streamlink's team attention
in case somebody would be willing to fix the plugin.

Error is intermittent and occurs about once in every half an hour.
To replicate this error would require watching the stream for up to 30 minutes
and player would freeze at some point in time.
Player must be killed as a process, otherwise it hangs on desktop.
It happens with vlc, mpv and ffplay.

### Reproduction steps / Explicit stream URLs to test

https://www.filmon.tv/channel/bbc-world-news

```
$url = 'https://www.filmon.tv/channel/bbc-world-news'
run('streamlink.exe --http-no-ssl-verify --player=ffplay.exe ' & $url & '  best','',@SW_MINIMIZE)
```

### Log output

```
[cli][info] Found matching plugin filmon for URL https://www.filmon.tv/channel/bbc-world-news
[cli][info] Available streams: low (worst), high (best)
[cli][info] Opening stream: high (hls-filmon)
[cli][info] Starting player: ffplay.exe
Exception in thread Thread-FilmOnHLSStreamWorker:
Traceback (most recent call last):
  File "C:\P\STREAMLINK\pkgs\streamlink\plugin\api\http_session.py", line 166, in request
    res.raise_for_status()
  File "C:\P\STREAMLINK\pkgs\requests\models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 504 Server Error: Gateway Time-out for url: http://www.filmon.com/api-v2/channel/3268?protocol=hls

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "threading.py", line 916, in _bootstrap_inner
  File "C:\P\STREAMLINK\pkgs\streamlink\stream\segmented.py", line 59, in run
    for segment in self.iter_segments():
  File "C:\P\STREAMLINK\pkgs\streamlink\stream\hls.py", line 280, in iter_segments
    self.reload_playlist()
  File "C:\P\STREAMLINK\pkgs\streamlink\plugins\filmon.py", line 24, in reload_playlist
    parsed = urlparse(self.stream.url)
  File "C:\P\STREAMLINK\pkgs\streamlink\plugins\filmon.py", line 104, in url
    for stream in self._get_stream_data():
[...]

PR #2796, opened 2020-02-08
plugins.filmon: retry for 50X errors

back-to:
closes https://github.com/streamlink/streamlink/issues/2775

### DIFF
--- a/src/streamlink/plugins/filmon.py
+++ b/src/streamlink/plugins/filmon.py
@@ -149,7 +149,15 @@ class FilmOnAPI(object):
     )
 
     def channel(self, channel):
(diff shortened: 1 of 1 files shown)


gravyboat:
Looks good, thanks @back-to.

mkbloke:
I was looking at the idea of using try/except blocks here, but in the process I seemed to find (not thoroughly tested yet), that simply adding ```retries=3``` to ```res = self.session.http.get(self.channel_url.format(channel))``` seemed like it also solved the issue.  I wanted to wait until I had a successful FilmOn recording and had watched it to see the effect before suggesting it as a fix though.  Thanks for fixing it @back-to.

PR #2581, opened 2019-08-07
setup.cfg: added flake8 settings

back-to:

PR #2804, opened 2020-02-19
Flake8

bastimeyer:
This is an attempt of cleaning up Streamlink's codebase a bit. ~~I've tried to fix as many errors and warnings as possible, but there's still a lot to do, as I didn't touch those which would require altering the code or reformatting it by a lot (unfortunately Python programmers love silly indentation and code styles). This additional work can be done later in another PR or I'll add some more commits if I feel like.~~

A flake8 config was already added in #2581, but the linting tool was never actually used in the CI runs, so I've enabled that as well (only on TravisCI). It will always succeed though, so if there are any build errors, then it was due to me making mistakes while cleaning up. Some errors/warnings are suppressed via inline comments or in the flake8 config (`setup.cfg`).

To anyone reviewing the changes, I think it'll be better to go through each commit individually. I've tried to only fix the errors/warnings mentioned in the commit message, but there are also some other minor changes here and there, like deleting an unnecessary empty line between class and method definitions. 

Remaining errors: none

There's still more to do than just fixing the reported linting errors/warnings:
- Remove unnecessary file encoding comments
  - Add an editorconfig file which sets the encoding value and other configs globally (or per file / ext)
- Remove unnecessary python shebangs
- Remove unnecessary `unittest.main()` calls from test modules
- Replace `self.logger` calls with native `logging` module

Other things I've noticed:
- Some tests don't inherit from `unittest.TestCase`, but `unittest` still gets imported (unused import error needs to be suppressed)


bastimeyer:
I've decided to fix W504 (line break after binary operator), E741 (ambiguous variable name '...') and most importantly E501 (line too long (... > 128 characters)).

These are a lot of further changes and I had to rewrite lots of url matching tests and also a couple of regular expressions. This needs to be reviewed before it can be merged.

bastimeyer:
All errors are fixed now. This required fixing some plugins which I initially didn't want to do. In case of F821, these errors were all caused by #1925, which looks like it was a simple replace-string-in-directory job and thus broke lots of stuff. That's kinda bad. It got unnoticed here, as the broken code was unreachable or behind a double geo-block check and thus not executed.

Since the current flake8 config doesn't report any problems now with these changes, are we fine with making the lint-check a build success requirement? Open PRs would all have to be rebased once this gets merged. Other PRs which will get merged in the mean time will most likely require me to resolve some merge conflicts then, but I'm fine with that.

Also please remember to review these changes before merging (I know it's a lot :weary:). Any further fixes/improvements/cleanups will be done in another PR. I'm done here for now unless there are some issues.

gravyboat:
@bastimeyer  Took me a while but I went through this tonight and it looks good. Maybe we can try for one more review just in case I missed something? I'm also fine with making the lint check a build success requirement, I don't want to review another PR like this in the future. Thanks for doing this.

bastimeyer:
Fixed the class attributes.
Also removed the `--exit-zero` parameter so that builds can fail if there are linting errors (and reordered the commits).

bastimeyer:
Interesting... It uses a different set of linting rules when run on Python 2. Didn't know that. Added another commit and reordered again. Should be good now.

PR #2809, opened 2020-02-25
Further cleanup

bastimeyer:
These are the additional (simple) changes I've mentioned in #2804. 

Regarding the editorconfig, please check if this config is good enough, as there might be a couple of file types/paths which I've left out.

And regarding the removed `unittest.main()` calls from the test modules, people should instead use the [`pytest`](https://docs.pytest.org/en/latest/usage.html) or [`unittest`](https://docs.python.org/3/library/unittest.html) command line interfaces to run test modules or single tests. Or even better, from within their IDEs. For the sake of completeness though, CLI examples here:
```sh
# pytest CLI
python -m pytest
python -m pytest tests/path/to/file.py
python -m pytest tests/path/to/file.py::ClassName
python -m pytest tests/path/to/file.py::ClassName::test_name
# unittest CLI
python -m unittest --verbose
python -m unittest --verbose tests.path.to.file
python -m unittest --verbose tests.path.to.file.ClassName
python -m unittest --verbose tests.path.to.file.ClassName.test_name
```

That should be it for now with my cleanup changes (I think :sweat_smile:).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,25 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+max_line_length = 128
+trim_trailing_whitespace = true
+
+[*.{cfg,svg,yml}]
+indent_size = 2
+
+[*.{md,markdown}]
+trim_trailing_whitespace = false
+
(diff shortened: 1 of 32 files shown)


gravyboat:
Looks good,  thanks @bastimeyer. Now please no more.! 😆

Issue #2750, opened 2019-12-19
Youtube plugin not working with some vods


Maqken:
## Plugin Issue

- [x ] This is a plugin issue and I have read the contribution guidelines.

### Description

On some vods i get no playable streams found and cant seem to identify why.

### Reproduction steps / Explicit stream URLs to test

1. Try to load streamlink https://www.youtube.com/watch?v=FmDUUGOTFCM

### Log output

```
streamlink https://www.youtube.com/watch?v=jXg93FFZXWk -l debug
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.6.6
[cli][debug] Streamlink: 1.3.0
[cli][debug] Requests(2.22.0), Socks(1.7.1), Websocket(0.56.0)
[cli][info] Found matching plugin youtube for URL https://www.youtube.com/watch?v=jXg93FFZXWk
[plugin.youtube][debug] Video ID from URL
[plugin.youtube][debug] Using video ID: jXg93FFZXWk
[plugin.youtube][debug] get_video_info - 1: Found data
[plugin.youtube][debug] MuxedStream: v 137 a 251 = 1080p
Available streams: audio_mp4, audio_webm, 360p (worst), 720p, 1080p (best)

streamlink https://www.youtube.com/watch?v=FmDUUGOTFCM -l debug
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.6.6
[cli][debug] Streamlink: 1.3.0
[cli][debug] Requests(2.22.0), Socks(1.7.1), Websocket(0.56.0)
[cli][info] Found matching plugin youtube for URL https://www.youtube.com/watch?v=FmDUUGOTFCM
[plugin.youtube][debug] Video ID from URL
[plugin.youtube][debug] Using video ID: FmDUUGOTFCM
[plugin.youtube][debug] get_video_info - 1: Found data
error: No playable streams found on this URL: https://www.youtube.com/watch?v=FmDUUGOTFCM
```

### Additional comments, screenshots, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

Infernion:
I've faced with similar issue with following link: https://www.youtube.com/watch?v=UhkgggX_-QE&feature=youtu.be

phytoporg:
+1, repro here as well using 1.3.0+19.gad276c4 and video link https://www.youtube.com/watch?v=fCZsk0E-Y5A

[...]

PR #2822, opened 2020-03-10
Youtube: no playable streams found - fix attempt

fgietzen:
Hello everyone,

since the youtube plugin does seem not work correctly lately (#2750) i have spend some time to investigate this issue.

This might be no final solution but this might help to understand how the structure of the stream info page has changed.

So far I am able to open streams which are not seperated in video and audio (the 360p and 720p one). 
I am currently not sure why `MuxedStream.is_usable(self.session)` in line 294 returns false and how i can fix that.


skhalymon:
@fgietzen do you have any progress on this?

PR #2826, opened 2020-03-12
plugins.radiko: Add support for radiko.jp

iucario:
New plugin for http://radiko.jp

* The site is geo-restricted and a Japan IP is needed.
* Supports both live and timefree.


src/streamlink/plugins/radiko.py
```diff
@@ -0,0 +1,113 @@
...
+            'User-Agent': useragents.FIREFOX,
+            'X-Radiko-AuthToken': token
+        }
+        self.session.http.headers = headers
+        r = self.session.http.get(url)
+        url = r.text.split('\n')[-2]
+        yield 'source', HLSStream(self.session, url)
+
+    def _live(self, station_id):
+        live_url = 'https://rpaa.smartstream.ne.jp/so/playlist.m3u8'
```
back-to: > 504 Server Error: Gateway Timeout for url: https://rpaa.smartstream.ne.jp/medialist

won't work with a hard coded URL

---

some streams need different URLs

example `http://radiko.jp/#!/live/RN1` needs
`http://f-radiko.smartstream.ne.jp/RN1/_definst_/simul-stream.stream/playlist.m3u8`

src/streamlink/plugins/radiko.py
```diff
@@ -0,0 +1,113 @@
+import re
+import random
+import hashlib
+import base64
+import datetime
+import xml.etree.ElementTree as ET
```
back-to: sort ABC and use a new line between `import` and `from` lines

src/streamlink/plugins/radiko.py
```diff
@@ -0,0 +1,113 @@
...
+from streamlink.compat import urlencode
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import useragents
+from streamlink.stream import HLSStream
+
+
+class Radiko(Plugin):
+    _url_re = re.compile(r'http://radiko\.jp/(#!/)?(?P<state>live|ts)/(?P<station_id>\w+)/?(?P<start_at>\d+)?')
+    _api_auth_1 = "https://radiko.jp/v2/api/auth1"
```
back-to: use `'` not `"` , so it is the same in the whole file.

src/streamlink/plugins/radiko.py
```diff
@@ -0,0 +1,113 @@
...
+            url, token = self._live(station_id)
+        else:
+            start_at = match.group('start_at')
+            url, token = self._timefree(station_id, start_at)
+        headers = {
+            'User-Agent': useragents.FIREFOX,
+            'X-Radiko-AuthToken': token
+        }
+        self.session.http.headers = headers
+        r = self.session.http.get(url)
```
back-to: you don't need this, just use

```py
       for s in HLSStream.parse_variant_playlist(self.session, url).items():
            yield s
```

docs/plugin_matrix.rst
```diff
@@ -158,6 +158,7 @@ pluzz                   - france.tv          Yes   Yes   Streams may be geo-rest
                         - francetvinfo.fr
 powerapp                powerapp.com.tr      Yes   No
 qq                      live.qq.com          Yes   No
+radiko                  radiko.jp             Yes   Yes   Streams are geo-restricted to Japan.
```
back-to:
```suggestion
radiko                  radiko.jp            Yes   Yes   Streams are geo-restricted to Japan.
```

Issue #2752, opened 2019-12-23
Twitch clips fail with 410 Gone


achurch:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

Streamlink is no longer able to play clips from Twitch, giving a "410 Gone" error for `https://clips.twitch.tv/api/v2/clips/[clip_name]/status`. This is separate from the now-fixed 410 error for live streams and VODs, though the root cause is probably the same (Twitch removing old private APIs).

### Reproduction steps / Explicit stream URLs to test

Any Twitch clip, e.g. https://clips.twitch.tv/LuckyExquisiteHamburgerAsianGlow

### Log output

(using master at ad276c4)
```
[cli][debug] Python:     3.6.9
[cli][debug] Streamlink: 0+unknown
[cli][debug] Requests(2.22.0), Socks(1.6.7), Websocket(0.48.0)
[cli][info] Found matching plugin twitch for URL https://clips.twitch.tv/LuckyExquisiteHamburgerAsianGlow
error: Unable to open URL: https://clips.twitch.tv/api/v2/clips/LuckyExquisiteHamburgerAsianGlow/status (410 Client Error: Gone for url: https://clips.twitch.tv/api/v2/clips/LuckyExquisiteHamburgerAsianGlow/status?as3=t)
```

### Additional comments, screenshots, etc.

youtube-dl had the same problem and fixed it with: https://github.com/ytdl-org/youtube-dl/commit/232ed8e6e0ec8b86156e68002e496a8bc89e6346

hristiyanvasilev:
was looking into fixing this for a first issue, however I noticed that youtube-dl are using the undocumented GraphQL endpoint, which should not be used by third party apps.
https://discuss.dev.twitch.tv/t/is-it-legal-to-use-https-gql-twitch-tv-gql-endpoint/21811/5

jamezrin:
@hristiyanvasilev this has also been the case for the /api endpoint, which is used right now, so it isn't really any different.

gootecks:
I just wanted to bump this up because I'm having the same issue and thought it was just me.  

I built a really cool (IMO) Twitch clip downloader and would really like to see this working again.  

@hristiyanvasilev please work some magic! <3

mjevans:
I too am experiencing the twitch api v2 issue (I have a shell wrapper that calls with my personal oauth-token as well as some other flags I always use).

[...]

PR #2843, opened 2020-03-22
plugins.twitch: Fixed clips

back-to:
closes https://github.com/streamlink/streamlink/issues/2752


gravyboat:
Good stuff, thanks @back-to.

PR #2845, opened 2020-03-24
plugins.invintus: Add support for Invintus Media live streams and VOD

thefinn93:
This adds a plugin for videos from player.invintus.com. This is my first contribution to this project, please let me know if I've missed anything.


src/streamlink/plugins/invintus.py
```diff
@@ -0,0 +1,50 @@
+import re
+import json
+try:
+    from urllib.parse import urlparse, parse_qs
+except ImportError:
+    from urlparse import urlparse, parse_qs
```
bastimeyer: `from streamlink.compat import ...`
https://github.com/streamlink/streamlink/blob/24c4318f29c31ad572e26c1ef94be3bf5722f899/src/streamlink/compat.py#L35-L43

src/streamlink/plugins/invintus.py
```diff
@@ -0,0 +1,50 @@
...
+    url_re = re.compile(r"https?://player\.invintus\.com/\?clientID=\d+&eventID=\d+")
+    WSC_API_KEY = "7WhiEBzijpritypp8bqcU7pfU9uicDR"  # hard coded in the middle of https://player.invintus.com/app.js
+    API_URL = "https://api.v3.invintusmedia.com/v2/Event/getDetailed"
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls.url_re.match(url) is not None
+
+    def _get_streams(self):
+        query = parse_qs(urlparse(self.url).query)
```
bastimeyer: Why urlparse if you're defining a strict regex anyway? Just set regex capture groups

src/streamlink/plugins/invintus.py
```diff
@@ -0,0 +1,50 @@
...
+        }
+        api_response = self.session.http.post(self.API_URL, data=json.dumps(postdata), headers=headers).json()
+        if "data" not in api_response:
+            return
+        if "streamingURIs" not in api_response["data"]:
+            return
+        if not isinstance(api_response["data"]["streamingURIs"], dict):
+            return
+        if "main" not in api_response["data"]["streamingURIs"]:
+            return
```
bastimeyer: Please use Streamlink's validation API instead of concatenating lots of ifs for checking a JSON response. See other plugins or test cases for a reference of that validation API.

src/streamlink/plugins/invintus.py
```diff
@@ -0,0 +1,50 @@
...
+        return HLSStream.parse_variant_playlist(self.session, "https:{}".format(api_response["data"]["streamingURIs"]["main"]))
```
bastimeyer: `update_scheme` from `streamlink.utils.url` is a bit more safe here:
https://github.com/streamlink/streamlink/blob/24c4318f29c31ad572e26c1ef94be3bf5722f899/src/streamlink/utils/url.py#L6-L22

tests/plugins/test_invintus.py
```diff
@@ -0,0 +1,16 @@
+import unittest
+
+from streamlink.plugins.invintus import InvintusMedia
+
+
+class TestPluginInvintusMedia(unittest.TestCase):
+    def test_can_handle_url(self):
```
bastimeyer: Please also add a simple non-matching test

thefinn93:
Oops, sorry! here's a brief description (based on what I've been able to cobble together):

[Invintus Media](https://www.invintus.com/) is company that provides live and VOD video streaming solutions. According to [the Invintus Media "about" page](https://www.invintus.com/about), they were originally part of [TVW](https://en.wikipedia.org/wiki/TVW_(Washington)), Washington state's public affairs network, but has since spun off and also serves other clients, which seem to primarily be other state government agencies.

I believe this plugin fits into streamlink, as this site does not seem to do any of the things listed in in `CONTRIBUTING.md` and it seems generally in line with other sites supported by streamlink.

Working to resolve the review comments now.

PR #2846, opened 2020-03-26
plugins.twitch: remove auth

bastimeyer:
This removes or rather disables (non-breaking change) both authentication methods of the Twitch plugin:

- **Cookies** (`--twitch-cookie`)
  This has been deprecated and obsolete since many years and is non-functional anyway ever since they've upgraded their website. The API endpoint for aquiring a token with cookies does not exist anymore.
- **OAuth token** (`--twitch-oauth-token`)
  This was already "disabled" by #2680/#2692 when we had to switch to Twitch's own API client-id for all private API calls. Since the streaming access token API endpoint is private and we can't use our own client-id here, all user's OAuth tokens can't be used as well, as they are tied to our client-id. When we added the changes, we kept the auth methods and the dummy login message, as we didn't want to delay the release any further (https://github.com/streamlink/streamlink/pull/2692#issuecomment-555840502). This PR finally removes these remnants. The only effect is therefore the lack of the login log message if the parameter still gets used.
  It also removes/disables the `--twitch-oauth-authenticate` parameter, which is used for logging in and getting an OAuth token. The OAuth return page in the docs has also been removed, as it's not needed anymore.


bastimeyer:
PR is split into two commits, in case the private API endpoint stuff changes in the future, so we can simply revert the oauth token commit.

gravyboat:
Changes look good, thanks @bastimeyer.

milanserfaty:
There is something I don’t get: is it still possible to watch or download twitch streams with restricted access or not anymore?

milanserfaty:
There is something I don’t get: is it still possible to watch or download twitch streams with restricted access or not anymore?

gravyboat:
@milanserfaty It is not.

milanserfaty:
Do you think this feature will be available again ?

bastimeyer:
This has nothing to do with Streamlink, but with Twitch.
Read https://github.com/streamlink/streamlink/issues/2680#issuecomment-557605851 or the entire thread if you don't know why this has been removed. It was actually already removed in 1.3.1, and the only difference in 1.4.0 is that it doesn't log your username anymore and that the auth parameters are dummy parameters now.

If Twitch decides to make a public API endpoint available for retrieving the stream access tokens, then the Twitch auth feature could be added back to Streamlink. But since Twitch won't do this and has instead been actively blocking the access tokens for 3rd party API client IDs, this means that they'll probably never do this.

There's no need to keep asking or begging. This is nothing which we have control over and the goal for us here is to make watching streams possible without going through loopholes like reading OAuth tokens from the web browser's dev tools.

PR #2849, opened 2020-03-27
docs: fix multiple options on the same line

bastimeyer:
Multiple double-dash parameters are currently ignored in the docs. Example:
https://streamlink.github.io/latest/cli.html#cmdoption-twitch-low-latency

Parameter descriptions are now parsed line by line and lines starting with [two spaces](https://streamlink.github.io/latest/cli.html#cmdoption-hls-segment-key-uri) (rendered as code blocks) or with [`Example: `](https://streamlink.github.io/latest/cli.html#cmdoption-hls-segment-ignore-names) get ignored. Then on each matching line, all double-dash params at the start of the line or with a leading space get transformed into anchors.

image


gravyboat:
Nice catch @bastimeyer.

PR #2857, opened 2020-04-01
docs: switch to sphinx_rtd_theme

### LINKED ISSUE #2851: Suggestion to replace custom sphinx theme with official sphinx rtd theme
## Feature Request

### Description

Currently, Streamlink documentation uses a custom sphinx read the docs theme originally made by chrippa.

As it ages, [stuff goes deprecated](https://github.com/readthedocs/sphinx_rtd_theme/pull/728/files) and the official sphinx read the docs theme is updated accordingly by their maintainers, but not the custom sphinx theme currently used.

So I'm suggesting to switch to the official sphinx_rtd_theme instead as an external theme.
This should mean:
- Remove the theme folder in Streamlink's git
- Add relevant configuration in docs/conf.py according to [readthedocs installation instructions](https://github.com/readthedocs/sphinx_rtd_theme#installation)

I've made a build of the Streamlink's documentation with the official readthedocs theme so you can see what it looks like, here: [streamlink-docs-with-sphinx_rtd_theme.zip](https://github.com/streamlink/streamlink/files/4396373/streamlink-docs-with-sphinx_rtd_theme.zip)

I'm making this suggestion primarily as the distribution packager to ease maintenance over Streamlink's documentation when other packages are updated (like sphinx).

Thanks for your consideration !

bastimeyer:
Resolves #2851 

Spent a bit more time trying to fix my annoyances with the theme and am now in a state where I think it's kind of usable. As mentioned in https://github.com/streamlink/streamlink/issues/2851#issuecomment-605673167, I'm not the biggest fan of it, especially because of the left-aligned menu and the color theme plus overall styles, but after fixing the alignment of the page contents (while not trying to break other stuff defined in the theme's unfortunately static stylesheets), it's much better now (although not perfect), which is the reason why I'm posting this PR to get further feedback.

I've also added a couple of other changes and decided to rewrite and reformat some sections of the docs. This includes the donation page, so if anyone has some more changes here, feel free to tell me, or add them to the PR branch.

What I'm not 100% sure about is whether not specifying a strict version(-range) in the `docs-requirements.txt` is a good idea. This opens the door for problems, should the sphinx and/or rtd-theme maintainers introduce breaking changes.

-----

```bash
mkdir streamlink-rtd-theme
cd streamlink-rtd-theme

virtualenv ./venv
source ./venv/bin/activate

git clone --branch docs/sphinx-rtd-theme https://github.com/bastimeyer/streamlink.git
cd streamlink

pip install -e .
pip install -r docs-requirements.txt

$BROWSER ./docs/_built/html/index.html
```

image


amurzeau:
Nice work !
To have the sidebar alignment centered, it seems "left: initial;" works when manually adding a style in firefox.
Debian is planning the move to sphinx 2.4 from 1.8, there are changes required in themes and that's ongoing in sphinx_rtd_theme (https://github.com/readthedocs/sphinx_rtd_theme/pull/784).
This help about that kind of maintenance.

bastimeyer:
The reason why I left the menu aligned to the left is its background color. It would just look weird having it centered as well. When aligning it to the center, it has to be re-styled completely first, including the TOC (nested links).

Personally, I prefer having it centered, but re-styling a static/pre-built theme and manually overriding CSS selectors/properties all while using the latest version which will change from time to time is a bit of work and annoying.

bastimeyer:
@back-to As said, I'm not a fan of it either. The menu can of course be moved to the center, but that requires restyling it, which is a lot of effort when using a prebuilt theme as a dependency.

Regarding sphinx 3, this requires a fix in the parsers definition, I believe. Looked it up earlier today, but haven't tested it yet. And yeah, the version should be set in the requirements file. Btw, the theme devs are still working on the sphinx 2 migration from what it looks like.

PR #2859, opened 2020-04-04
ci: remove old CI configs

bastimeyer:
Remove old CI configs and the now unused bintray deploy scripts


bastimeyer:
Looks like we'll have to remove the old and deprecated TravisCI app integration as well. Fingers crossed this won't delete its build history, which would be a shame.

bastimeyer:
@gravyboat All the old stuff has been removed, but it still expects a successful build from travis. Do I have to reopen the PR? What do you think? Or did I miss something?

PR #2860, opened 2020-04-04
ci: remove old CI configs and fix codecov report

bastimeyer:
Remove old CI configs and the now unused bintray deploy scripts.

Opening a new PR, because the old TravisCI integration was still active when I opened the other one. Hopefully this is working now. Sorry for the spamming. :weary:


bastimeyer:
Found it, it was the master branch protection rule (hence the required flag). Updated it, will force-push now.

bastimeyer:
Btw, I think it's listing 34 checks here, because I've updated the rules while the PR was open and it doubled the 17 (3*5+2) build job checks.

----

Looks like CodeCov is not working as it should be

> Error: Missing repository upload token

https://github.com/streamlink/streamlink/runs/559962509?check_suite_focus=true#step:7:24

Secret env vars are unavailable in PRs. Let me see.

bastimeyer:
Codecov upload tokens are not required for public repos. This wasn't working recently, which was the reason I added this. I didn't think about the env var being missing in PRs though. It's possible that the codecov upload script is picking up an empty env var and therefore refuses to upload the report.

According to the recent codecov docs, tokenless uploads should be working now.
https://community.codecov.io/t/whitelist-github-action-servers-to-upload-without-a-token/491/17

It's also possible that the codecov python package we're using has not been updated yet. Let me remove the token definition from the workflow config and see if it's working.

bastimeyer:
https://github.com/streamlink/streamlink/pull/2860/checks?check_run_id=559983179#step:7:23

We'll have to use the codecov github action for that instead of the codecov python module.

**Let me fix this first before merging.**

bastimeyer:
That's not right...
Why does it only find the coverage results on Windows?

**Ubuntu:**
https://github.com/streamlink/streamlink/pull/2860/checks?check_run_id=559988929#step:7:27
**macOS:**
https://github.com/streamlink/streamlink/pull/2860/checks?check_run_id=559988942#step:7:27
**Windows:**
https://github.com/streamlink/streamlink/pull/2860/checks?check_run_id=559988954#step:7:28

Is it because `coveragepy` is for some reason automatically installed on Windows?

bastimeyer:
We're fine now. Please merge. Otherwise I will merge myself in a few.

back-to:
https://github.com/streamlink/streamlink/blob/master/build.cmd can be removed

also some Travis links on https://github.com/streamlink/streamlink/blob/master/README.md

bastimeyer:
True that, @back-to 
Do you want to fix this real quick or do you want me to do this? Anything else that was missed? In the docs maybe?

back-to:
> Do you want to fix this real quick or do you want me to do this?

you can just add a 2nd commit with your other PR, might be easier and faster.

> Anything else that was missed? In the docs maybe?

maybe some of your travis compatibility envs in `script/github_releases.py` `script/makeinstaller.sh`

https://github.com/streamlink/streamlink/search?q=travis

PR #2862, opened 2020-04-04
ci: cleanup

bastimeyer:
https://docs.codecov.io/docs/pull-request-comments#branches

> Generally, you should set after_n_builds equal to the number of reports you upload per commit.

Set this to 15. If we remove macOS, this needs to be changed. It's annotated though.


bastimeyer:
This is getting annoying. Why is it not working? We're uploading 15 reports.

No idea why this doesn't appear in the logs anymore after the build succeeded, but some coverage reports fail to upload and it's stuck in an upload loop with a 30s timeout. Scroll to the bottom here:
https://pipelines.actions.githubusercontent.com/eFmcCyrUFDy9z9353ozAyQf6s8IEKyozIRvlSLnsHPO1JgSIpe/_apis/pipelines/1/runs/11/signedlogcontent/20?urlExpires=2020-04-04T09%3A08%3A42.9997602Z&urlSigningMethod=HMACV1&urlSignature=47YxmPNNGWv8qWmVmmrdtG8Y%2F%2FymuQlP%2BHDzclCbh58%3D

Could it be that it's an anti-spam filter on Codecov because lots of jobs finish in parallel? Doesn't explain though why there's no comment after all builds have uploaded a report.

bastimeyer:
https://codecov.io/gh/streamlink/streamlink/commit/ac534a2313a0b6d305bf69520a06943f5eb00900

> Waiting for CI to complete.

back-to:
one failed, there are only 14

https://github.com/streamlink/streamlink/pull/2862/checks?check_run_id=560110976


```
Run codecov/codecov-action@v1
##[warning]Codecov warning: connect ETIMEDOUT 35.199.43.247:443
bash codecov.sh -n  -F 
codecov.sh: line 1: undefined: command not found
##[warning]Codecov warning: The process 'bash' failed with exit code 127
```

bastimeyer:
So how do we fix this?
My suggestion would be removing macOS, since it's unnecessary and spams the whole thing, and then setting the value to 6, which means that at least one ubuntu and one windows build must have passed.

back-to:
> removing macOS

could be done

>  and then setting the value to 6,

as the message is not required, maybe 10 would be fine anyways.

bastimeyer:
If I set it to 10, then it's the same situation where one failure will not trigger a comment or even full report on codecov's website.

Since there's no difference in coverage between linux and macOS, removing macOS and setting the number of required build to 6 should be good. If all builds succeed and there's a coverage change in the remaining runs, then the comment gets updated accordingly.

The situation before when not requiring a specific number of build successes also works, but the comment that gets then edited multiple times by the codecov bot does not always get re-rendered while the PR is opened in the browser and can therefore cause confusion.

Let's remove macOS from the tests, we don't need it.

bastimeyer:
Looks like it's working. Had to delete the empty comment that was made immediately after force pushing.

PR #2866, opened 2020-04-06
plugins.tv360: updated URL and HLS stream extraction method

beardypig:


gravyboat:

PR #1999, opened 2018-08-07
tests: use unittest instead of pytest for itertags error

back-to:
The tests won't pass with `python setup.py test`,
use `unittest` instead of `pytest`.

```
======================================================================
FAIL: test_itertags_multi_attrs (tests.test_plugin_utils.TestPluginUtil)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/test_plugin_utils.py", line 59, in 
test_itertags_multi_attrs
    self.assertEqual(metas[0].text, None)
AssertionError: 'Title' != None

======================================================================
FAIL: test_no_end_tag (tests.test_plugin_utils.TestPluginUtil)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/test_plugin_utils.py", line 80, in test_no_end_tag
    self.assertEqual(links[0].text, None)
AssertionError: '' != None

```

as an example https://aur.archlinux.org/packages/streamlink-git/ uses it.
https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=streamlink-git#n57

---

Ref https://github.com/streamlink/streamlink/pull/1675
Ref https://github.com/streamlink/streamlink/issues/1979

PR #2869, opened 2020-04-07
utils: update_qsd to leave blank values unchanged

beardypig:
Updates the `update_qsd` to support blank param values.

As pointed out by @calculon-jr in [#2747](https://github.com/streamlink/streamlink/pull/2747/files/b90362964e5c582eb9d960520386f4f12f8e3f54..b0a5f75146b1e0ca0bca6ecd8405878735c01aa7﻿).

I also refactored the tests for the url utils to be more pytesty.

### DIFF
--- a/tests/test_utils_url.py
+++ b/tests/test_utils_url.py
@@ -1,48 +1,47 @@
-import unittest
+from collections import OrderedDict
 
 from streamlink.utils.url import (
     update_scheme, url_equal, url_concat, update_qsd
 )
 
 
-class TestUtilURL(unittest.TestCase):
-    def test_update_scheme(self):
-        self.assertEqual(
-            "https://example.com/foo",  # becomes https
-            update_scheme("https://other.com/bar", "//example.com/foo")
-        )
-        self.assertEqual(
-            "http://example.com/foo",  # becomes http
-            update_scheme("http://other.com/bar", "//example.com/foo")
-        )
-        self.assertEqual(
-            "http://example.com/foo",  # remains unchanged
-            update_scheme("https://other.com/bar", "http://example.com/foo")
-        )
-        self.assertEqual(
-            "https://example.com/foo",  # becomes https
-            update_scheme("https://other.com/bar", "example.com/foo")
-        )
+def test_update_scheme():
+    assert update_scheme("https://other.com/bar", "//example.com/foo") == "https://example.com/foo", "should become https"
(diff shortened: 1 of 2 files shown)


tests/test_utils_url.py
```diff
@@ -1,48 +1,46 @@
...
+    assert url_concat("http://test.se/one", "../two", "../three") == "http://test.se/three"
+
+
+def test_update_qsd():
+    assert update_qsd("http://test.se?one=1&two=3", {"two": 2}) == "http://test.se?one=1&two=2"
+    assert update_qsd("http://test.se?one=1&two=3", remove=["two"]) == "http://test.se?one=1"
+    assert update_qsd("http://test.se?one=1&two=3", {"one": None}, remove="*") == "http://test.se?one=1"
+    assert update_qsd("http://test.se", OrderedDict([("one", ""), ("two", "")])) == "http://test.se?one=&two=", \
+        "should add empty params"
+    assert update_qsd("http://test.se?one=", {"one": None}) == "http://test.se?one=", "should leave empty params unchanged"
```
bastimeyer, line 41: Maybe also add this for the sake of completeness:
```python
assert update_qsd("http://test.se?one=", {"two": 2}) == "http://test.se?one=&two=2"
```
(blank value in input, no key in qsd param and keep_blank_values is True)

beardypig: In PyCharm you just have to change the default test runner to `pytest` to get it to work nicely. We also already have some non-`unittest` tests that use `pytest`, some that mix `unittest.TestCase` and plain `test_` methods :)

back-to:
> We also already have some non-unittest tests that use pytest, some that mix unittest.TestCase and plain test_ methods :)

None of these tests actually work with `python setup.py test`

I changed the tests, so **they must fail**.

```
----------------------------------------------------------------------
Ran 890 tests in 28.962s

OK (skipped=34)
```

with pytest

```
====== 3 failed, 888 passed, 38 skipped, 1 warnings in 32.24s
```

---

similar to https://github.com/streamlink/streamlink/pull/1999

beardypig:
As mentioned in #1999 we could set up the tests requires etc. We can create a separate issue to fix that if you feel it is important.
I believe that the `test` command in `setuptools` is being deprecated.

back-to:
>  I believe that the test command in setuptools is being deprecated.

https://setuptools.readthedocs.io/en/latest/setuptools.html#test-build-package-and-run-a-unittest-suite

well then it is unimportant and can be ignored.

PR #2774, opened 2020-01-18
Add support for specifying output file format and audio sync option

lcapriotti:
The change is introducing two new optional cmdline options:

--ffmpeg-fout: The output file format when muxing with ffmpeg
--ffmpeg-start-at-zero: When used with ffmpeg and copyts, shift input timestamps so they start at zero

Both options defaults are keeping the predefined values, hence backwards compatibility is enforced.

PR #2870, opened 2020-04-07
ci.github: fix continue on error

bastimeyer:
This fixes the irrelevant but potentially confusing error message if a test has failed on one of the required test jobs. Github expects a boolean value on the `continue-on-error` config, but `matrix.continue` is only set on additional allowed-to-fail jobs, which we currently don't have, so the value is an empty string instead, which it doesn't like.

Example:
https://github.com/streamlink/streamlink/runs/567342748?check_suite_focus=true#step:6:15

Fixed:
https://github.com/bastimeyer-test/streamlink-test/runs/567542014?check_suite_focus=true#step:5:271

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,16 +32,16 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
       - name: Install dependencies
-        continue-on-error: ${{ matrix.continue }}
+        continue-on-error: ${{ matrix.continue || false }}
         run: bash ./script/install-dependencies.sh
       - name: Test
-        continue-on-error: ${{ matrix.continue }}
+        continue-on-error: ${{ matrix.continue || false }}
         run: pytest -r a --cov --cov-report=xml
       - name: Lint
-        continue-on-error: ${{ matrix.continue }}
+        continue-on-error: ${{ matrix.continue || false }}
         run: flake8
       - name: Upload coverage data
-        continue-on-error: ${{ matrix.continue }}
+        continue-on-error: ${{ matrix.continue || false }}
         uses: codecov/codecov-action@v1
 
   documentation:


beardypig:
flake8 is busting my balls...

bastimeyer:
I wasn't trying to mock you, @beardypig, I just took the most recent workflow failure :sweat_smile: 

Btw, there's another CI issue which I'd like to fix, but I'm not sure if it's worth it:
Currently, Github doesn't support auto-cancelling runs which are in progress if another commit gets pushed to a PR (or a regular repo branch - not required here). The only workaround right now is adding a custom community cancel-action which cancels any other in-progress runs of the same PR via the available API endpoints. Problem though is that it bloats the workflow a bit and needs to run in a separate job before everything else can start running. Other CI providers all have an option for this implemented, but Github still does not. If this doesn't get added by them soon and if it's causing issues here, we should add the custom cancel-action.

- https://github.com/Khan/pull-request-workflow-cancel
- https://github.community/t5/GitHub-Actions/Github-actions-Cancel-redundant-builds/td-p/29549
- https://github.community/t5/GitHub-Actions/Is-it-possible-to-cancel-previous-jobs-automatically-when-the/td-p/37139

As an example, in #2774, 5 code suggestions from a review were merged via GH's website within one minute, which for some reason spawned 9 workflow runs (with 17 jobs each), causing other PRs that were submitted at the same time to be unnecessarily queued up for a while. Not to mention that it all adds up to the available monthly runner time.

https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits

beardypig:
I know @bastimeyer, but flake8 was out to get me :D 

That `cancel-action` might be worth it to speed things like that up. 

Where does it say about the monthly runner limits? The way I read those limits it was per run?

bastimeyer:
Ok, I guess I have misread the monthly limitations when I read the documentation some time ago, or I'm just not remembering it correctly. The limitation only affects private repos and it's 2000 mins / month for free accounts. It's completely free for public repos. Only the concurrent runner limitation exists, which is 20.
https://help.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions

Let me check out cancel action on my test repo and see if it's working correctly and reliably... Btw, I've removed the deploy keys you've set up there, as I don't need them anymore.

bastimeyer:
@beardypig I can't get the cancel workflow action to work. Tried two different actions and I think it's not working because of an API restriction (which is weird because that's why these actions exist):
https://github.com/bastimeyer-test/streamlink-test/pull/4/checks?check_run_id=568628887#step:3:34
Looks like a custom access token with the `repo` scope is required (now - maybe this was changed recently?!). Wouldn't work on PRs though.

----

Enough off-topic now. This PR here can be safely merged if there's no further feedback.

beardypig:
We can set up cancelling redundant jobs if they add first party support for it, I don't think we need to worry about it too much :)

PR #2877, opened 2020-04-10
plugin.api.useragents: update User-Agent

back-to:


gravyboat:
Thanks for staying on top of these @back-to.

Issue #2616, opened 2019-09-05
plugin.younow is broken


TaoziDB:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Reproduction steps / Explicit stream URLs to test

1. ...
2. ...
3. ...

### Log output

```
streamlink https://www.younow.com/JustboBShow --loglevel debug
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.7.3
[cli][debug] Streamlink: 1.2.0
[cli][debug] Requests(2.22.0), Socks(1.7.0), Websocket(0.56.0)
[cli][info] Found matching plugin younow for URL https://www.younow.com/JustboBShow
error: Unable to validate JSON: Unable to validate result: Key 'media' not found in {<deleted long json data here>} 
```

### Additional comments, screenshots, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

PR #2879, opened 2020-04-10
plugins.younow: remove plugin

back-to:
seems like you need a **third party account** for a normal login

not really worth to fix this

closes https://github.com/streamlink/streamlink/issues/2616


gravyboat:
Agreed, not worth it.

Issue #2175, opened 2018-11-20
AUTHORS Sorting


frenchja:
Came across a tiny observation to discuss.  Currently, `./script/authors.sh` uses an alphabetical, case-sensitive sorting mechanism. I thought this behavior might be a discrepancy between the `sort` bundled with macOS and GNU, but version of `sort` provided by `coretools` in Homebrew also seems to have a `-f` option to ignore case.  This is a bit weird since the [current AUTHORS file](https://github.com/streamlink/streamlink/blob/master/AUTHORS) appears to be case insensitive.  I might be missing something... 

Does it makes sense to change this to case-insensitive?

```bash
git log --format='%aN' | sort -uf
```

Alternatively, the file could present authors by their commit amounts: 

```bash
git shortlog -s -n | awk '{$1="";print}'
```

back-to:
**the script is not used, it was only used once.**

there are a lot of **duplicates** if you want to do it automatically

it might be easier to update it **manually** when a `First-time contributor` PR gets merged

of course the current list could be updated with your first command,
but the duplicates should be removed from it.

```
git log --format='%aN' | sort -uf
```

---

also this part `__credits__` and/or `__author__`

https://github.com/streamlink/streamlink/blob/cd6f94a5a4ee83cb5b61a295df9e13956bcfcb27/src/streamlink/__init__.py#L17-L71

could be changed with a link to https://raw.githubusercontent.com/streamlink/streamlink/master/AUTHORS

PR #2889, opened 2020-04-12
AUTHORS: removed unused script and removed outdated list

back-to:
Removed the list because it is **totally outdated**,
if someone wants to search the Authors list for whatever reason just use the git command.

closes https://github.com/streamlink/streamlink/issues/2175


AUTHORS
```diff
@@ -1,123 +1,5 @@
-This AUTHORS file is generated by ./script/authors.sh
 Thank you to everyone whom has contributed to streamlink / livestreamer.
```
bastimeyer, line 1: Not your edit, I know, but isn't "whom" grammatically incorrect here? "Who" is the subject of that clause.

back-to: so it should be who instead of whom ?

gravyboat: I agree `who` sounds better grammatically.

PR #2892, opened 2020-04-13
Add support for specifying output file format and audio sync option

beardypig:
The change is introducing two new optional command line options:

- `--ffmpeg-fout`: The output file format when muxing with ffmpeg
- `--ffmpeg-no-start-at-zero`: When used with ffmpeg and copyts, disables automatically shifting input timestamps so they start at zero

Both options defaults are keeping the predefined values, hence backwards compatibility is enforced.


beardypig:
I added some extensive changes that couldn't be done with suggestions. I added some tests and fixed some bugs that were exposed by the tests...

beardypig:
Replaces #2774

src/streamlink/stream/ffmpegmux.py
```diff
@@ -91,13 +91,14 @@ def __init__(self, session, *streams, **options):
                              for stream, np in
                              zip(self.streams, self.pipes)]
 
-        ofmt = options.pop("format", "matroska")
+        ofmt = options.pop("format", session.options.get("ffmpeg-fout"))
```
back-to, line 94: the user command **can't work** because `format` is not None

```
$ streamlink https://www.zdf.de/live-tv --ffmpeg-fout "matroska"
[stream.ffmpegmux][debug]

ffmpeg command: /usr/bin/ffmpeg -nostats -y
-i /tmp/ffmpeg-8930-729 -i /tmp/ffmpeg-8930-644
-c:v copy -c:a copy -map 0:v? -map 0:a?
-map 1:a -f mpegts pipe:1
```

same for vcodec acodec start_at_zero

beardypig: Yep, you're right. I have it backwards :/

It should go `ffmpeg-fout` -> `options.format` -> `default`

beardypig: I added some more tests, and defaults in the class - it should behave as expected now.

back-to:
setting the value might work, but is there an actual working use case

not sure which formats are valid https://ffmpeg.org/ffmpeg-formats.html
but most of the time it will just fail.

@lcapriotti what was your intention with this feature?

```
[cli][error] Try 1/1: Could not open stream <Stream()> (No data returned from stream)
error: Could not open stream <Stream()>, tried 1 times, exiting

...

[stream.ffmpegmux][error] Pipe copy aborted: /tmp/ffmpeg-23548-515
[stream.ffmpegmux][debug] Closing ffmpeg thread
[stream.ffmpegmux][debug] Closed all the substreams
[cli][info] Stream ended
[stream.ffmpegmux][debug] Closing ffmpeg thread
[stream.ffmpegmux][debug] Closed all the substreams
[stream.ffmpegmux][debug] Closing ffmpeg thread
[stream.ffmpegmux][debug] Closed all the substreams
```

lcapriotti:
> @lcapriotti what was your intention with this feature?
I am using my forked version to stream from a Swiss site (Wilmaa), because master was not giving proper results for these MPEG-DASH streams (audio delay); I also am in need to have TS streams as output hence the ffmpeg options.

back-to (approved):
I would merge it like this if there are no concerns

---

PR is required for https://github.com/streamlink/streamlink/pull/2747

gravyboat:
@back-to It needs a rebase. I closed the other PR for the time being.

src/streamlink/stream/ffmpegmux.py
campones: sorry but I see you are using "matroska" for mpegts 
what would be the value to fill in for "flv" ?

gravyboat: This issues is ancient, why are you bumping it?

PR #2372, opened 2019-03-23
 plugins.twitch: implement disable-ads parameter

bastimeyer:
Resolves #2368, #2357 (deleted, for whatever reason, which is really sad - I've opened a support ticket to get it hopefully restored) 

Please see #2369 and my comment here first:
https://github.com/streamlink/streamlink/pull/2369#issuecomment-475528275

Adds a custom `TwitchHLSStream` implementation and the `--twitch-disable-ads` CLI parameter, which will skip HLS segments between `#EXT-X-SCTE35-OUT` and `#EXT-X-SCTE35-IN` tags.

If this PR gets merged, **please don't squash the commits**.

PR #2895, opened 2020-04-14
plugins.twitch: rewrite disable ads logic

### LINKED ISSUE #2894: Twitch streams showing Commercial Break In Progress video

### Description

When launching a Twitch stream you're shown a Commercial Break video before the actual stream starts playing. This behavior started tonight.

The commercial video is expected to be ignored.

Launch a Twitch stream.

### Log output

```
C:\Users\User>streamlink https://www.twitch.tv/esl_csgo best --loglevel debug
[cli][debug] Streamlink: 1.3.1
[cli][debug] Requests(2.22.0), Socks(1.7.1), Websocket(0.57.0)
[cli][info] Found matching plugin twitch for URL https://www.twitch.tv/esl_csgo
[plugin.twitch][debug] Getting live HLS streams for esl_csgo
[utils.l10n][debug] Language code: sv_SE
[stream.hls][debug] First Sequence: 0; Last Sequence: 2
[stream.hls][debug] Start offset: 0; Duration: None; Start Sequence: 0; End Sequence: None
[stream.hls][debug] Adding segment 1 to queue
[stream.hls][debug] Download of segment 0 complete
[cli.output][debug] Opening subprocess: "C:\Program Files\VideoLAN\VLC\vlc.exe" --input-title-format https://www.twitch.tv/esl_csgo -
[stream.hls][debug] Download of segment 1 complete
[stream.hls][debug] Download of segment 2 complete
[stream.hls][debug] Download of segment 3 complete
[stream.hls][debug] Download of segment 4 complete
[stream.hls][debug] Download of segment 5 complete
[stream.hls][debug] Download of segment 6 complete
[stream.hls][debug] Download of segment 7 complete
[stream.hls][debug] Download of segment 8 complete
[stream.hls][debug] Adding segment 11 to queue
[stream.hls][debug] Download of segment 9 complete
[stream.hls][debug] Download of segment 10 complete
[stream.hls][debug] Download of segment 11 complete
[stream.hls][debug] Adding segment 12 to queue
[stream.hls][debug] Adding segment 13 to queue
[stream.hls][debug] Download of segment 12 complete
[stream.hls][debug] Download of segment 13 complete
[stream.hls][debug] Adding segment 14 to queue
[stream.hls][debug] Adding segment 15 to queue
[stream.hls][debug] Adding segment 16 to queue
[stream.hls][debug] Download of segment 14 complete
[stream.hls][debug] Download of segment 15 complete
[stream.hls][debug] Download of segment 16 complete
[stream.hls][debug] Adding segment 17 to queue
[stream.hls][debug] Adding segment 18 to queue
[stream.hls][debug] Download of segment 17 complete
[stream.hls][debug] Download of segment 18 complete
[cli][info] Player closed
[cli][info] Stream ended
```

[Screenshot](https://i.imgur.com/BbljzTw.png)

bastimeyer:
Resolves #2894

Simply filters out segments that are annotated with `#EXTINF:2.002,Amazon[...]`.
`--twitch-disable-ads` needs to be set, just like before.

edit:
Well, it doesn't "resolve" the issue, just replaces the `--twitch-disable-ads` logic with new one. As I've mentioned in the linked thread, Twitch's webplayer is not showing any ads to me and I haven't figured out yet why. If someone knows how to "trick" Twitch into providing the real stream data, then that would be considered the real "fix" for the issue. Nevertheless, the parameter needs new filtering logic.

----

The stream discontinuities when the "ad" switches to the regular stream data seems to be causing a lot of issues in various players, especially VLC (no idea about the used version though). Maybe we should re-evaluate switching the default player to MPV (#2144), as it's working just fine there.

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -232,21 +232,24 @@ class HLSStreamWorker(SegmentedStreamWorker):
         sequences = [Sequence(media_sequence + i, s)
                      for i, s in enumerate(playlist.segments)]
 
-        if sequences:
-            self.process_sequences(playlist, sequences)
+        self.process_sequences(playlist, sequences)
 
     def _set_playlist_reload_time(self, playlist, sequences):
         self.playlist_reload_time = (playlist.target_duration
-                                     or sequences[-1].segment.duration)
+                                     or len(sequences) > 0 and sequences[-1].segment.duration)
 
     def process_sequences(self, playlist, sequences):
+        self._set_playlist_reload_time(playlist, sequences)
+
+        if not sequences:
+            return
+
         first_sequence, last_sequence = sequences[0], sequences[-1]
 
         if first_sequence.segment.key and first_sequence.segment.key.method != "NONE":
             log.debug("Segments in this playlist are encrypted")
 
         self.playlist_changed = ([s.num for s in self.playlist_sequences] != [s.num for s in sequences])
-        self._set_playlist_reload_time(playlist, sequences)
         self.playlist_sequences = sequences
 
         if not self.playlist_changed:
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -119,25 +119,16 @@ _video_schema = validate.Schema(
     }
 )
 
-
-Segment = namedtuple("Segment", "uri duration title key discontinuity scte35 byterange date map")
+Segment = namedtuple("Segment", "uri duration title key discontinuity ad byterange date map")
 
 LOW_LATENCY_MAX_LIVE_EDGE = 2
 
 
-def parse_condition(attr):
-    def wrapper(func):
-        def method(self, *args, **kwargs):
-            if hasattr(self.stream, attr) and getattr(self.stream, attr, False):
-                func(self, *args, **kwargs)
-        return method
-    return wrapper
-
-
 class TwitchM3U8Parser(M3U8Parser):
-    def __init__(self, base_uri=None, stream=None, **kwargs):
-        M3U8Parser.__init__(self, base_uri, **kwargs)
-        self.stream = stream
+    def __init__(self, base_uri=None, disable_ads=False, low_latency=False, **kwargs):
+        super(TwitchM3U8Parser, self).__init__(base_uri, **kwargs)
+        self.disable_ads = disable_ads
+        self.low_latency = low_latency
         self.has_prefetch_segments = False
 
     def parse(self, *args):
@@ -146,21 +137,16 @@ class TwitchM3U8Parser(M3U8Parser):
 
         return m3u8
 
-    @parse_condition("disable_ads")
-    def parse_tag_ext_x_scte35_out(self, value):
-        self.state["scte35"] = True
+    def parse_extinf(self, value):
+        duration, title = super(TwitchM3U8Parser, self).parse_extinf(value)
+        if title and str(title).startswith("Amazon") and self.disable_ads:
+            self.state["ad"] = True
 
-    # unsure if this gets used by Twitch
-    @parse_condition("disable_ads")
-    def parse_tag_ext_x_scte35_out_cont(self, value):
-        self.state["scte35"] = True
+        return duration, title
 
-    @parse_condition("disable_ads")
-    def parse_tag_ext_x_scte35_in(self, value):
-        self.state["scte35"] = False
-
-    @parse_condition("low_latency")
     def parse_tag_ext_x_twitch_prefetch(self, value):
+        if not self.low_latency:
+            return
         self.has_prefetch_segments = True
         segments = self.m3u8.segments
         if segments:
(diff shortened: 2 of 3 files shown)


gravyboat:
I'm able to get ads every time in a Chrome Incognito window (logged out) but I don't see anything particularly useful past what you're seeing here. What data do you want? I don't know if there is any way to trick Twitch in this sense as it looks to me like they either present and ad, or they present the commercial break notification, there's no stream data to pull.

It seems to me like the big issue with players is being caused by buffering. They're not sending any stream data prior to the commercial break ending, so when it does end that's when you finally get stream data and it causes problems for the player because nothing is cached. It behaves similarly to when I was experimenting with extreme low latency live edge settings. I am fine with making MPV the default suggested player, but the issues brought up by people still have not been addressed in any way by the MPV team. The lack of official builds being by far my biggest concern.

bastimeyer:
@gravyboat It's possible that ads are not shown to logged in users. Doesn't help Streamlink though, as we can't use any auth data when acquiring the access token (users can still set a custom auth HTTP header I believe).

Regarding MPV, I don't have an answer here. My points from the linked issue still stand, I guess.

----

Pushed another clean up commit. This should be in a better shape overall now.
If there are preroll ads, another info message will be logged. I think this is better than just saying "Will ignore ad segments".

src/streamlink/stream/hls.py
```diff
@@ -234,10 +234,12 @@ def reload_playlist(self):
 
         if sequences:
             self.process_sequences(playlist, sequences)
+        else:
+            self._set_playlist_reload_time(playlist, sequences)
```
bastimeyer, line 238: I'm not sure if this could potentially break other HLS streams. @back-to @beardypig ?

The reason I've added this here is that the default playlist reload time is 15 and if there are no segments in the first playlist "refresh" (first download), then the `_set_playlist_reload_time` method does not get called. This makes the Twitch preroll ad test wait unnecessarily long for 15 seconds.

gravyboat:
I like the updated comment @bastimeyer, makes it more clear regarding what is going on.

src/streamlink/stream/hls.py
```diff
@@ -232,21 +232,24 @@ def reload_playlist(self):
...
         self.playlist_reload_time = (playlist.target_duration
-                                     or sequences[-1].segment.duration)
+                                     or len(sequences) > 0 and sequences[-1].segment.duration)
 
     def process_sequences(self, playlist, sequences):
+        self._set_playlist_reload_time(playlist, sequences)
+
+        if not sequences:
+            return
+
```
bastimeyer, line 246: I'm not sure if this could potentially break other HLS streams. @back-to @beardypig ?

The reason I've changed this here is that the default playlist reload time is 15 and if there are no segments in the first playlist "refresh" (first download), then the _set_playlist_reload_time method does not get called. This makes the Twitch preroll ad test wait unnecessarily long for 15 seconds.

(sorry about posting the same comment twice)

beardypig: seems reasonable, I don't think it would break anything

bastimeyer: Ok, good.

What about the rest of the changes? There's a bit of refactoring here which I haven't really mentioned yet. Eg. I've removed the stream dependency on the M3U8Parser, removed the now unnecessary method decorator and also removed the custom StreamWriter class, as the ad filtering is now done by the parser which simply ignores the next line if the tag above is identified as an ad. This further means that there are no log messages anymore which indicate the start or end of an ad break, but that couldn't be done anyway due to the lack of proper annotations in the playlist (previously we had the start and end tags which made that possible). Basically the same issue with just relying on the discontinuity tags, which you've tried last year when they implemented the embedded ads system the first time.

bastimeyer:
There might be a problem with the current implementation. I didn't run into non-preroll ads yesterday (pre-roll ads were filtered out just fine), but it's possible that Twitch requires HLS clients to still download the segments, like mentioned last year here:
https://github.com/streamlink/streamlink/pull/2372#issuecomment-476438732

That would mean that the filtering can't be done in the parser anymore and the removal of the custom StreamWriter has to be reverted.

Unfortunately, this can't be tested right now as the embedded ads have stopped.

bastimeyer:
Reverted the custom StreamWriter removal. Ad segments will now be always downloaded and then discarded afterwards.

----

Even if Twitch currently doesn't embed ads, it's safe to assume that this new system will be used in the future. Since this PR doesn't only implement the new ad filtering logic, but also removes the unnecessary restriction that low latency streaming and ads filter can't be used at the same time, I'd like to see this being included in the next release.

Being able to always set the low latency streaming parameter, even for non-LL streams, is a good thing, because it overrides the default playlist refresh timer, which is stupid IMO. The default refresh time is 6 seconds and causes 3 whole segments to be queued each time, meaning that with a default live edge value of 3, a short hickup while refreshing the playlist may cause a hickup in the playback if the player's own cache isn't large enough. And short delays while reloading the playlist happen a lot.

This makes me wonder whether a parameter for manually overriding the refresh time would make sense, maybe with support for both static time values and keywords like min-segment-length or so.

gravyboat:
@bastimeyer I agree. There's no reason not to include it and it will save us time going forward from answering repetitive questions. It has no negative impact. I really like the idea of adding an option to manually override the refresh timer, I've run in to this myself where I git a hiccup just because it's waiting for the playlist to update when I initially open a stream.

bastimeyer:
Noticed some "incorrect" values in the tests which are now fixed and also patched the HLSStreamWorker's `wait` method, so that the tests run without unnecessarily waiting between playlist refreshes.

PR #2912, opened 2020-04-17
plugins.cdnbg: update plugin to support new sites, and remove old sites

### LINKED ISSUE #2902: [bug] CDNBG - multiple issues
## Bug Report
1. BITelevision should be removed from the plugin and/or wiki/info pages, as it no longer exists.
2. Inlife.bg shouldn't be listed as supported under CDNbg.
3. Tvbulgare.bg should be listed as supported in inlife.bg's place (latter shares the former's stream)
4. Mu-vi.tv gives an error.
5. CDNBG should cover VTK - the national military channel
6. Kanal3's livestream is not found.
7. CDNBG should cover Cherno More - the regional channel for Varna, Bulgaria.

1. https://bitelevision.com/ is not a thing anymore.
2. Inlife.bg can't be opened and shouldn't be listed - it is a 'media partner' that restreams https://tvbulgare.bg/, which could be put in as a replacement for it.
3. https://tvbulgare.bg/ - No playable streams found.
4. http://mu-vi.tv/LiveStreams/pages/Live.aspx - Error: Unable to open URL.
5. https://www.armymedia.bg/
6. https://kanal3.bg/live
7. https://www.chernomore.bg/

beardypig:
- added support for armymedia.bg and chernomore.bg
- removed inlife.bg as it has migrated to cdn.netbadgers.com
- removed bitelevision.com as it no longer exists
- added support for i.cdn.bg urls
- removed support for video.bstv.tv

fixes #2902


roshavagarga:
@beardypig Not sure if important, but I noticed a few links that might need updating in the test plugin, namely:

- Old BNT links should be replaced with tv.bnt.bg/bnt1, tv.bnt.bg/bnt2, tv.bnt.bg/bnt3, tv.bnt.bg/bnt4. Only the last one isn't geoblocked..

- http://videochanel.bstv.bg/ should be changed to http://video.bstv.bg/

beardypig:
http://video.bstv.bg needs some extra work to make it work again. It uses JavaScript to load the iframe from somewhere.

roshavagarga:
@beardypig If it's too much work I'd drop that one, it seems like they added the stream there as an afterthought.

back-to:
> http://video.bstv.bg needs some extra work to make it work again. It uses JavaScript to load the iframe from somewhere.

the iframe url might be static `https://i.cdn.bg/live/zjq1oO6KW7` just use it

beardypig:
@back-to that's why I added support for `i.cdn.bg` URLs :) I wanted to see if it was static, or if the request was a simple look up - but if it's static then problem solved.

http://mu-vi.tv/LiveStreams/pages/Live.aspx is the only remaining one that doesn't work, need to work out why it doesn't work properly. There is a cookie that gets passed as a key, but I didn't work out where the cookie gets set yet.

roshavagarga:
@beardypig I think that one can arguably be dropped if it's too much of a hassle - it's the TV station for one of the two main medical universities in the country and it shouldn't be of that high interest.

PR #441, opened 2017-01-17
Rewritten StreamProcess class

beardypig:
### Rational

The current `StreamProcess` (and `RTMPStream` subclass) make use of a deprecated package `pbs` that is vendored with Streamlink and includes a few patches made back in Livestreamer days. The `pbs` package has been replaced by a new package, `sh`, however it no longer has support for Windows - which is a necessity for this project. The current `StreamProcess` class also has limitations, for example multiple parameters with the same name are not possible (eg. `--conn B:0 --conn S:Hello`) and positional arguments are not possible (not that rtmpdump uses positional arguments). 

### Proposed Solution

This PR proposes the solution. The `pbs` package has been removed and the Python `subprocess` module is used directly. New `StreamProcess` and `RTMPStream` classes that use `subprocess.Popen` to spawn an instance of the process (`rtmpdump`) have been implemented. 

The new classes have additional features and a backwards compatible API with the existing `StreamProcess` and `RTMPStream` classes. Positional arguments are supported (however, `rtmpdump` takes no positional arguments) and specifying multiple parameters with the same name is now possible (fixing #321). 

Multiple arguments of the same name use a list with one entry for each repetition (the ordering is maintained):

```python
RTMPStream(session, 
           params={
            "rtmp": "rtmp://foo.com/bar",
            "conn": ["B:0", "S:Hello"]
           })
```

Hope you don't mind @xkbd that I included the changes in #438 in this PR :)

PR #2916, opened 2020-04-17
installer: external assets

### LINKED ISSUE #2872: Update ffmpeg included in Windows installer
The version of `ffmpeg` included with the Windows installer is pretty old (3.2.1). We should consider updating it to a newer version, like 4.2.2 from the Zeranoe builds. 

We could also take this as an opportunity to move the `ffmpeg` and `rtmpdump` executables out of this repo and in to the [streamlink/streamlink-assets](https://github.com/streamlink/streamlink-assets) repo.

bastimeyer:
Resolves #2872
and upgrades FFMpeg to 4.2.2 (Zeranoe)

### Changes

1. Rewrite makeinstaller.sh
   - Move removed plugins list into a text file and read from it
   - Add log and err functions
   - Define build dependencies
   - Set repo root directory instead of assuming it
   - Clean up build directory before building the installer
   - Set dist dir to same STREAMLINK_DIST_DIR
   - Suppress warning/error messages from inkscape and imagemagick
   - Replace some strings in pynsist config with variables
   - Update python license text according to the currently used version
2. Download assets from streamlink/streamlink-assets
   - Download metadata from latest assets release
   - Download assets into cache dir (useful for re-building locally)
   - Compare checksums
   - Extract specific files following the metadata
3. Finally remove binary files from the repo


script/makeinstaller.sh
```diff
@@ -1,62 +1,81 @@
...
+cache_dir="${build_dir}/cache"
 nsis_dir="${build_dir}/nsis"
 files_dir="${build_dir}/files"
 icons_dir="${files_dir}/icons"
-# get the dist directory from an environment variable, but default to the build/nsis directory
-dist_dir="${STREAMLINK_DIST_DIR:-${nsis_dir}}"
-mkdir -p "${build_dir}" "${dist_dir}" "${nsis_dir}" "${files_dir}" "${icons_dir}"
 
-echo "Building streamlink-${STREAMLINK_VERSION} package..." 1>&2
+removed_plugins_file="${ROOT}/src/streamlink/plugins/_removed.txt"
```
beardypig, line 53: I personally prefer `.removed.txt` as the name for that file, or even just `.removed`

script/makeinstaller.sh
```diff
@@ -1,62 +1,81 @@
...
-  "vgtv" "weeb" "oldlivestream" "ok_live" "rte" "tvcatchup" "npo" "ovvatv"
-  "mlgtv" "kanal7" "younow" "eurocom" "tv1channel"
-)
-for i in "${old_files[@]}"
-do
-    touch "${build_dir_plugins}/$i.py"
-done
+while read -r pluginname; do
+    touch "${build_dir_plugins}/${pluginname}.py"
+done < <(sed -e 's/[[:space:]]*[#;].*//; /^[[:space:]]*$/d' "${removed_plugins_file}")
```
beardypig, line 68: I think it's a bit more readable to use `readarray`, it's a Bash 4 command but that shouldn't be a problem. 

```suggestion
readarray -t removed_plugins < "${removed_plugins_file}"
for pluginname in "${removed_plugins[@]}"; do
    touch "${build_dir_plugins}/${pluginname}.py"
done
```

script/makeinstaller.sh
bastimeyer: The idea was to have the file sorted at the top. A leading dot would hide it in most *nix file managers by default.

script/makeinstaller.sh
```diff
@@ -1,62 +1,81 @@
...
+    [unzip]=unzip
+)
+
+for dep in "${!DEPS[@]}"; do
+    command -v "${dep}" 2>&1 >/dev/null || err "${DEPS["${dep}"]} is required to build the installer. Aborting."
+done
 
-set -e # stop on error
+ROOT=$(git rev-parse --show-toplevel 2>/dev/null || realpath "$(dirname "$(readlink -f "${0}")")/..")
+cd "${ROOT}"
```
beardypig, line 29:
```suggestion
pushd "${ROOT}"
```

And `popd` on exit :)

script/makeinstaller.sh
bastimeyer: Does readarray remove comments with a leading `#` or `;`?

script/makeinstaller.sh
bastimeyer: The entire script is executed in a subshell, so it shouldn't matter whether `cd` or `pushd/popd` is used, right?

script/makeinstaller.sh
```diff
@@ -262,17 +280,51 @@ StrCmp \$0 \${ffmpeg} "" +2
...
+log "Extracting assets"
+TEMP=$(mktemp -d) && trap "rm -rf ${TEMP}" EXIT || exit 255
+for ((i=$(jq length <<< "${assets_data}") - 1; i >= 0; --i)); do
+    IFS=" " read -r filename sourcedir targetdir \
+        < <(jq -r ".[$i] | \"\(.filename) \(.sourcedir) \(.targetdir)\"" <<< "${assets_data}")
+    unzip -qq "${cache_dir}/${filename}" -d "${TEMP}"
+    while IFS=" " read -r from to; do
+        log "Copying ${sourcedir}/${from} to ${targetdir}/${to}"
+        install -D -T "${TEMP}/${sourcedir}/${from}" "${files_dir}/${targetdir}/${to}"
+    done < <(jq -r ".[$i].files[] | \"\(.from) \(.to)\"" <<< "${assets_data}")
```
beardypig, line 323: I find it a bit nicer to pipe into while, it appears to work the same for me. 

```suggestion
    jq -r ".[$i].files[] | \"\(.from) \(.to)\"" <<< "${assets_data}" | while read -r from to; do
        log "Copying ${sourcedir}/${from} to ${targetdir}/${to}"
        install -D -T "${TEMP}/${sourcedir}/${from}" "${files_dir}/${targetdir}/${to}"
    done
```

script/makeinstaller.sh
```diff
@@ -262,17 +280,51 @@ StrCmp \$0 \${ffmpeg} "" +2
...
+        curl -s -L --output "${cache_dir}/${filename}" "${url}"
+    fi
+    checksum=$(jq -r "[.[] | select(.filename == \"${filename}\")] | first | .checksum" <<< "${assets_data}")
+    echo "${checksum} ${cache_dir}/${filename}" | sha256sum --check -
+done < <(jq -r '.assets[] | "\(.name) \(.size) \(.browser_download_url)"' <<< "${assets_release_data}")
+
+log "Extracting assets"
+TEMP=$(mktemp -d) && trap "rm -rf ${TEMP}" EXIT || exit 255
+for ((i=$(jq length <<< "${assets_data}") - 1; i >= 0; --i)); do
+    IFS=" " read -r filename sourcedir targetdir \
```
beardypig, line 317: Do you need to set `$IFS` here?

script/makeinstaller.sh
beardypig: Hidden is the best place for it, imo.

script/makeinstaller.sh
beardypig: True, you got comments in there uglying it up :p 
I made a similar change on another branch, but didn't put comments in the plugin list - I wanted to make a script to check to make sure a removed plugin was added to that file.

script/makeinstaller.sh
beardypig: True, force of habit

script/makeinstaller.sh
bastimeyer: Piping into while makes it run in a sub-shell, which is what I usually try to avoid.
https://stackoverflow.com/questions/13726764/while-loop-subshell-dilemma-in-bash

Not relevant in this particular case, but I tried to keep the same style throughout the assets and makeinstaller scripts. In the assets script, I ran into an issue because of piping into while.

beardypig: I like it because it isolates the variables, keeps them in their scope. It's easier to read too.

script/makeinstaller.sh
bastimeyer: I've explicitly set it on all read commands in case the user has set it. I don't think it's really needed, but better safe than sorry. We can export it at the start of the script though, if you prefer it that way.

script/makeinstaller.sh
bastimeyer: Sure, let's change it. Will update after some more feedback and issue resolving.

script/makeinstaller.sh
beardypig: Doesn't bash set it as `" \t\n"` at the start of the script automatically?

script/makeinstaller.sh
beardypig: Leave it how it is then, it's mostly a style issue for me anyway - I prefer to have the input to at the start of the loop rather than at the end :)
Bash script are hard to read as it is, probably this script would be better off in Python ;D

script/makeinstaller.sh
bastimeyer: I don't have anything against removing the comments. I just included them for anyone randomly reading the file.

> I wanted to make a script to check to make sure a removed plugin was added to that file

That script/test could also remove the comments while readin from the file. :smiley: 

But ok, let me change this, too.

script/makeinstaller.sh
bastimeyer: Yep, not necessary and can be removed as it doesn't inherit the value. The bash man page is only talking about default values, so wasn't 100% sure.

script/makeinstaller.sh
bastimeyer: I'll leave it. Otherwise the line gets too long and moving the while statement with the pipe symbol into the next line with added indentation makes it look even more weird.

script/makeinstaller.sh
beardypig: Nah, leave the comments they are useful. `grep -x` fixes everything.
```
while read -r rplugin; do echo grep -Fxq "${rplugin}" src/streamlink/plugins/_removed.txt; done < <(git diff --cached --name-only --diff-filter=D master | grep -oP '^src/streamlink/plugins/(\K.*)(?=.py)');
```

bastimeyer: Ok, restored again. For some reason Github doesn't update the review comments and I have to refresh the page.

bastimeyer:
The log output on Github actions is so useless. It's scrambling the lines for some reason, even in the raw log.

- https://github.com/streamlink/streamlink/pull/2916/checks?check_run_id=595990158#step:7:10
- https://pipelines.actions.githubusercontent.com/eFmcCyrUFDy9z9353ozAyQf6s8IEKyozIRvlSLnsHPO1JgSIpe/_apis/pipelines/1/runs/115/signedlogcontent/16?urlExpires=2020-04-17T16%3A40%3A41.9663179Z&urlSigningMethod=HMACV1&urlSignature=OMJc2qdc9T5spL2ug1vb4%2FUkgedJNRT2ga8Ij3%2BVBE8%3D

bastimeyer:
What about the python license file? Do we need to include it? I feel like this is pynsist's job.

beardypig:
It is pynsist's job. But it probably should be included, depends on what the licence says :)

bastimeyer:
https://github.com/streamlink/streamlink/pull/2916/files#diff-6871e707e71cec9606ed3e5534cc4516

Ok, that should probably not be deleted, but instead fixed up.

bastimeyer:
Restored `win32/LICENSE.txt`, fixed typo, fixed python version (would probably be wise turning it into a template in the future when the python version gets bumped again) and removed the redundant RTMPDump license text.

Still not sure how to feel about the removal of the license files of the libs bundled by FFmpeg. The zeranoe build archive should include these, but it doesn't. Maybe include this?
https://ffmpeg.zeranoe.com/builds/readme/win64/static/ffmpeg-4.2.2-win64-static-readme.txt
Would require a quick update of the assets repo and a fix here, as it always expects zip archives.

win32/LICENSE.txt
```diff
@@ -1,662 +1,320 @@
...
+7. Nothing in this License Agreement shall be deemed to create any relationship
+   of agency, partnership, or joint venture between PSF and Licensee.  This License
+   Agreement does not grant permission to use PSF trademarks or trade name in a
+   trademark sense to endorse or promote products or services of Licensee, or any
+   third party.
+
+8. By copying, installing or otherwise using Python 3.6.6, Licensee agrees
+   to be bound by the terms and conditions of this License Agreement.
+
+pbs License
```
back-to, line 47: >  pbs License

can be removed

Ref https://github.com/streamlink/streamlink/pull/441

bastimeyer: True, will remove that.

Does anyone know why some dependencies have their licenses included here and some not?

bastimeyer:
I've submitted streamlink/streamlink-assets#2 earlier today which should resolve the ffmpeg license issue. This needs to be merged first before I can push my changes here.

Btw, also did a bit of research regarding the Python license and from what it looks like, it's simply missing from the `python-3.6.6-embed-win32.zip` file `pynsist` downloads + extracts. The `3.8.2` zip archive includes a license file, so once the installer's python version gets bumped (not in this PR), the custom license text can be removed.

bastimeyer:
The ffmpeg build info is now included and the assets related log output a bit more verbose:
https://github.com/streamlink/streamlink/runs/599289704?check_suite_focus=true#step:7:308

That should be it regarding the external assets and their licenses.

Btw, `pynsist` 2.5 - we're currently using 2.4 - has some optimizations regarding the assembly of the executable launchers. Those are now done at build time and the result is that the install directory is much cleaner. Also includes high-DPI scaling fixes. Could be explored in another PR.

beardypig:
Have you made all the changes you want to make for this PR @bastimeyer?

bastimeyer:
Unless you've spotted another issue, I think it's ready to be merged.

beardypig:
Looks good to me, let's see if @back-to can spot anything :D

bastimeyer:
Ehm, do we even include Streamlink's license in the installer? I don't think so...

> Redistributions in binary form must reproduce the above copyright notice,
   this list of conditions and the following disclaimer in the documentation
   and/or other materials provided with the distribution.

script/makeinstaller.sh
```diff
@@ -1,72 +1,91 @@
...
-echo "Building ${STREAMLINK_INSTALLER} installer..." 1>&2
+log "Configuring installer"
 
 cat > "${build_dir}/streamlink.cfg" <<EOF
 [Application]
 name=Streamlink
 version=${STREAMLINK_VERSION}
 entry_point=streamlink_cli.main:main
 icon=${icons_dir}/icon.ico
+license_file=${ROOT}/LICENSE
```
back-to, line 85: would be better as a `.txt` file on windows

gravyboat:
Changes look good to me. 👍

Issue #2896, opened 2020-04-15
Python 2.7 support


brouxco:
## Issue

- [x] This is not a bug report, feature request, or plugin issue/request.
- [x] I have read the contribution guidelines.

### Description

Hi ! 
This is more of a question and the beginning of a discussion than an issue.

Python 2.7 is not maintained anymore (see: https://github.com/python/devguide/pull/344)

It has been mentioned a few times in the project's issues that Python 2.7 is sometimes problematic.
Last time was only a few days ago (see: https://github.com/streamlink/streamlink/pull/2871#issuecomment-611939149)

I'm really new in this project and I understand that there are historical reasons for a Python 2.7 version of StreamLink but I would like to know if there is a plan on Python 2.7 support.

Maybe some users are still relying on Python 2.7, should they expect an end of support soon?
How many users would be impacted? Is there an easy way to know?

gravyboat:
It really comes down to resources. If I could wave a magic wand I'd remove 2.7 support today. We're not going to support it, make code changes for it, or do any work to make Streamlink work with 2.7. The only plan around 2.7 is to stop supporting 2.7. If people are still relying on 2.7 they should expect it to go away at any time as 2.7 was already EoL, we just haven't gotten to it yet.

There is no way to know how many users would be impacted, we don't track stats like that at all within the application but even if it impacted 95% of people I'd still push for removing it.

brouxco:
I understand. 
Should we stop testing Python 2.7 then?
see: https://github.com/streamlink/streamlink/blob/master/.github/workflows/main.yml#L23

> we don't track stats like that at all within the application

And I'm grateful for that.

beardypig:
I imagine 99% of the Windows installs are from the installer or a portable version, which use Python 3.5+. 
I would imagine a lot of the Linux installs are Python 2.7, because older distros use Python 2.7 by default, same goes for macOS I guess.

gravyboat:
@brouxco When someone makes the PR to drop 2.7 support they'll stop testing 2.7, until then it needs to be tested. @beardypig I'm running it on 3.6+ on macOS since it's easy to run both with Brew. I don't think it's something worth worrying about as long as we announce it in the release notes.

bastimeyer:
Fedora and Ubuntu will be switching to python3 as default python interpreter in a couple of days. Apart from that, all available `streamlink` "system-packages" are based on python3.

[...]

PR #2919, opened 2020-04-18
Drop support for EOL Python 2

hugovk:
Fixes https://github.com/streamlink/streamlink/issues/2896.

Also upgrade syntax for Python 3 using https://github.com/asottile/pyupgrade.


bastimeyer:
Thanks for taking the time doing this and submitting the pull request!
Oh boy, that are a ton of changes (+1,859 −2,728)...

Just a couple of early thoughts from my side after a first quick review:

- It would probably be better to split this up into more commits, for reviewing reasons. One commit for each operation `pyupgrade` supports, for example, if it's possible. This would also be better for the commit history, git blame, etc...
- The string format related changes are probably unneeded or rather too much, as they only make explicit string templates implicit. Also, there's the 3.5 support drop in September, which enables the usage of f-strings once we go >=3.6, making these changes therefore a bit redundant.
- The included flashmedia package still has a compat file with py2 stuff, even though some of its files were modified.
- The order of imports in some modules has changed after the removal of the compat module. There's no linting rule for that here and not really important now, just wanted to mention it though.
- By removing one python version from the build matrix in the GH workflow file, the CodeCov comment threshold also needs to be changed in its config file. (not important now)

beardypig:
I would say the first change should be to update the docs and disallow Python 2 installs by updating `setup.py` and `setup.cfg` to remove the Python 2 specifics.

I believe that `setup.cfg` needs to be updated to remove this:
```
[wheel]
universal = 1
```

Perhaps a guard in `setup.py` needs to be added, something like
```
if sys.version_info < (3, 5):
    sys.exit('streamlink is only supported on Python 3.5 and later')
```

Once it is only possible to install `streamlink` on Python 3, then we can remove the Python 2 cruft. 

It would be nice to target Python 3.6+, but I don't know if that will be popular :)

hugovk:
Thanks for the comments! Yeah, it's a big one!

PR https://github.com/streamlink/streamlink/pull/2920 is a minimal version of this, with the first commit from here, plus a few bits from the last commit here, plus @beardypig's suggestions.

Then we can deal with fine-grained pyupgrade and the compat replacements separately.

> The order of imports in some modules has changed after the removal of the compat module. There's no linting rule for that here and not really important now, just wanted to mention it though.

I've found https://github.com/timothycrosley/isort/ great for taking the hassle out this.

bastimeyer:
#3270 already has all compat imports removed.

PR #2921, opened 2020-04-19
Upgrade Python syntax with pyupgrade

hugovk:
Split out from https://github.com/streamlink/streamlink/pull/2919.

https://github.com/asottile/pyupgrade has few options:

```console
$ pyupgrade --help
usage: pyupgrade [-h] [--exit-zero-even-if-changed] [--keep-percent-format] [--py3-plus] [--py36-plus] [--py37-plus]
                 [filenames [filenames ...]]

positional arguments:
  filenames

optional arguments:
  -h, --help            show this help message and exit
  --exit-zero-even-if-changed
  --keep-percent-format
  --py3-plus, --py3-only
  --py36-plus
  --py37-plus
```

This runs it with the minimal changes, and retains Python 2.7 support.

A follow-up can run with `--py3-plus`.

---

From https://github.com/streamlink/streamlink/pull/2919:

> * It would probably be better to split this up into more commits, for reviewing reasons. One commit for each operation `pyupgrade` supports, for example, if it's possible, similar to #2804. This would also be better for the commit history, git blame, etc...
>
> * The string format related changes are probably unneeded or rather too much, as they only make explicit string templates implicit. Also, there's the 3.5 support drop in September, which enables the usage of f-strings once we go >=3.6, making these changes therefore a bit redundant.

Unfortunately there's no option to skip explicit string template things. If you prefer, I can `git add -p` to skip those from the PR. But if they were be acceptable, it would make future `--py3-plus`, `--py36-plus` etc. runs easier.

### DIFF
--- a/src/streamlink/logger.py
+++ b/src/streamlink/logger.py
@@ -10,7 +10,7 @@ from streamlink.utils.encoding import maybe_encode
 TRACE = 5
 _levelToName = dict([(CRITICAL, "critical"), (ERROR, "error"), (WARN, "warning"), (INFO, "info"), (DEBUG, "debug"),
                      (TRACE, "trace"), (NOTSET, "none")])
-_nameToLevel = dict([(name, level) for level, name in _levelToName.items()])
+_nameToLevel = {name: level for level, name in _levelToName.items()}
 
 for level, name in _levelToName.items():
     logging.addLevelName(level, name)
--- a/src/streamlink/options.py
+++ b/src/streamlink/options.py
@@ -131,7 +131,7 @@ class Arguments(object):
 
         :return: list of dependant arguments
         """
-        results = set([name])
+        results = {name}
         argument = self.get(name)
         for reqname in argument.requires:
             required = self.get(reqname)
--- a/src/streamlink/plugin/api/utils.py
+++ b/src/streamlink/plugin/api/utils.py
@@ -24,5 +24,5 @@ def itertags(html, tag):
     """
     for match in tag_re.finditer(html):
         if match.group("tag") == tag:
-            attrs = dict((a.group("key").lower(), a.group("value")) for a in attr_re.finditer(match.group("attr")))
+            attrs = {a.group("key").lower(): a.group("value") for a in attr_re.finditer(match.group("attr"))}
             yield Tag(match.group("tag"), attrs, match.group("inner"))
--- a/src/streamlink/plugins/bbciplayer.py
+++ b/src/streamlink/plugins/bbciplayer.py
@@ -1,4 +1,3 @@
-
 import base64
 import logging
 import re
--- a/src/streamlink/plugins/idf1.py
+++ b/src/streamlink/plugins/idf1.py
@@ -1,4 +1,3 @@
-
 import re
 
 from streamlink.plugin import Plugin
--- a/src/streamlink/plugins/onetv.py
+++ b/src/streamlink/plugins/onetv.py
@@ -48,7 +48,7 @@ class OneTV(Plugin):
         res = self.session.http.get(update_scheme(self.url, self._session_api))
         data = self.session.http.json(res)
         # the values are already quoted, we don't want them quoted
-        return dict((k, unquote(v)) for k, v in data.items())
+        return {k: unquote(v) for k, v in data.items()}
 
     @property
     def is_live(self):
--- a/src/streamlink/plugins/showroom.py
+++ b/src/streamlink/plugins/showroom.py
@@ -59,7 +59,7 @@ _quality_weights = {
     "low": 160
 }
 # pages that definitely aren't rooms
-_info_pages = set((
+_info_pages = {
     "onlive",
     "campaign",
     "timetable",
@@ -76,7 +76,7 @@ _info_pages = set((
     "s",
     "organizer_registration",
     "lottery"
-))
+}
 
 
 class ShowroomHLSStreamWorker(HLSStreamWorker):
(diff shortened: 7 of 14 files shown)


hugovk:
Also, `--keep-percent-format` doesn't make much difference.

Omitting, it makes changes to `src/streamlink/packages/flashmedia/ordereddict.py` and `versioneer.py`, both external, so ignoring those. And this one small change:

```diff
diff --git a/src/streamlink/_version.py b/src/streamlink/_version.py
index d7a2069..75a107b 100644
--- a/src/streamlink/_version.py
+++ b/src/streamlink/_version.py
@@ -90,7 +90,7 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False,
             return None, None
     else:
         if verbose:
-            print("unable to find command, tried %s" % (commands,))
+            print("unable to find command, tried {}".format(commands))
         return None, None
     stdout = p.communicate()[0].strip()
     if sys.version_info[0] >= 3:
```

beardypig:
`_version.py` is automatically generated, or it was at some point. We should leave that file as is, and not lint it perhaps?

hugovk:
I've rebased out the changes to `./src/streamlink/_version.py`

bastimeyer:
> asottile/pyupgrade has few options:
> Unfortunately there's no option to skip explicit string template things.

Yeah, that's unfortunate, already had this feeling while reading the readme.

Individual AST transformation calls could be commented out though:
- https://github.com/asottile/pyupgrade/blob/f58732779c5ffe08fe1317ba3623a5ca84ef2a60/pyupgrade.py#L2448-L2455
- https://github.com/asottile/pyupgrade/blob/f58732779c5ffe08fe1317ba3623a5ca84ef2a60/pyupgrade.py#L758-L791

Maybe it's worth trying, if it's not too much of a hassle.

hugovk:
https://github.com/hugovk/streamlink/commit/3c48326f2f6e8aaea53c28f89d161fe65cd9101b is with this commented out from pyupgrade:

```python
        elif token.src == 'format' and i > 0 and tokens[i - 1].src == '.':
            _fix_format_literal(tokens, i - 2)
```

Is that better?

bastimeyer:
Yes, much better. From +635 −637 (160 files) to +74 -76 (14 files). That is acceptable in one commit.

A couple of changes seem to be missing though, eg:
https://github.com/hugovk/streamlink/blob/3c48326f2f6e8aaea53c28f89d161fe65cd9101b/src/streamlink/plugins/abematv.py#L40

hugovk:
> Yes, much better. From +635 −637 (160 files) to +74 -76 (14 files). That is acceptable in one commit.

Great, I've reset this PR to that commit.

> A couple of changes seem to be missing though, eg:
> [hugovk/streamlink:src/streamlink/plugins/abematv.py@`3c48326`#L40](https://github.com/hugovk/streamlink/blob/3c48326f2f6e8aaea53c28f89d161fe65cd9101b/src/streamlink/plugins/abematv.py#L40)

Yes, this is `pyupgrade` with no `--py3-plus`, as Python 2 is still supported until it's dropped in the minimal PR https://github.com/streamlink/streamlink/pull/2920.

beardypig:
@bastimeyer as there are going to be a lot of changes that could be made to remove python 2 cruft, what do you think about creating a separate branch for the python 2 -> 3 migration? We could merge all the smaller PRs in to that separate branch, and then merge that branch back in to master. It might make managing the changes easier.

bastimeyer:
The question is, when do we want to cut off py2? I think we should prepare a 1.4.0 release first (as soon as possible - we're overdue by a lot) and then deal with the py2 removal, as it'll require a major version bump anyway. That'll also be a good opportunity to remove some deprecated features, like dummy parameters, etc.

Regarding the organization of this migration, I guess adding a secondary branch to the repo here makes sense instead of keeping it on a personal branch in a PR. There will be merge conflicts either way, because it'll probably take a bit of time to finish this properly while regular changes and fixes will be done in parallel on the master branch.

Let's add a py2 removal branch after the next release, add a milestone on the issue tracker and assign issues/PRs to it. Unfortunately, this means that the PRs right now will need to be re-done. I hope that's okay.

hugovk:
Yeah, I don't mind redoing things.

I'd suggest doing the 1.4.0 release first, and then this stuff can be re-done. It might make sense to merge things in stages, so it doesn't go too much out of sync and cause more merge conflicts.

The minimal PR is easy, then of course `pyupgrade` and `pyupgrade --py3-plus` are automated. The last commit to replace compat stuff is more manual work, and should come after the `pyupgrade --py3-plus` changes.

beardypig:
I agree regarding a 1.4.0 release first. We should add a warning for deprecating Python 2.7 support in 1.4.0. Maybe something like this
```diff
diff --git a/src/streamlink/session.py b/src/streamlink/session.py
index ea554f6c..4813311e 100644
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -3,6 +3,7 @@ import logging
 import pkgutil
 import sys
 import traceback
+import warnings

 import requests

@@ -42,11 +43,21 @@ def print_small_exception(start_after):
     sys.stderr.write("\n")


+class PythonDeprecatedWarning(UserWarning):
+    pass
+
+
 class Streamlink(object):
     """A Streamlink session is used to keep track of plugins,
        options and log settings."""

     def __init__(self, options=None):
+        if sys.version_info[0] == 2:
+            warnings.warn("Python 2.7 has reached the end of its life.  A future version of streamlink will drop "
+                          "support for Python 2.7. Please upgrade your Python to at least 3.5.",
+                          category=PythonDeprecatedWarning,
+                          stacklevel=2)
+
         self.http = api.HTTPSession()
         self.options = Options({
             "hds-live-edge": 10.0,
```

gravyboat:
Definitely agree with a 1.4.0 release first as well as @beardypig's idea for a notice.

PR #2656, opened 2019-10-06
CI/CD: Github actions

bastimeyer:
## Why

We're currently using two CI providers, Appveyor for running tests on Windows and TravisCI for everything else. While Appveyor has caused issues in the past and is usually much slower than TravisCI, TravisCI's Windows VMs are still in beta with lack of support for secret env vars. See my previous PR #2342.

Github now supports CI/CD workflows via Github actions on Linux, Windows and macOS, which allows us to have all CI tasks being run in one place with direct integration into Github.

Docs can be found here:
https://help.github.com/en/categories/automating-your-workflow-with-github-actions

## Checklist

- [ ] **Re-enable actions**
  I've disabled all actions in the repo settings for now. These changes should be tested in a secondary repo first. See below...
- [x] **`test` job**
  - [x] Run in parallel on ubuntu-latest, macOS-latest and windows-latest, using Python 2.7 and 3.5..3.8
  - [x] Test (pytest)
  - [x] Lint (flake8)
  - [x] Report code coverage (codecov)
    - [x] Add `CODECOV_TOKEN` to secret env vars
- [x] **`documentation` job** (runs in parallel)
- [x] **`windows-installer` job** (runs in parallel)
  - [x] Deploys build artifact on tag with current version
  - [x] Deploys build artifact on scheduled run with current version
- [x] **`deploy-documentation` job**
  Depends on `test`, `documentation` and `windows-installer`.
  Runs when pushed to master or a tag gets pushed. Forks are disabled.
  - [x] Rewrite as bash script with similar/same behavior as doctr
  - [x] Add secret env vars to repo settings
    - [x] `DOCS_DEPLOY_PASSPHRASE`
  - [x] Replace deploy key
    - [x] Add public key to @streamlink-bot
  - [x] Give @streamlink-bot write access to docs repo
  - [x] Add 2FA to @streamlink-bot 
- [x] **`release` job**
  Depends on `deploy-documentation`.
  Runs if tag was pushed. Forks are disabled.
  - [x] Add secret env vars to repo settings
    - [x] `RELEASES_API_KEY`
    - [x] `RELEASE_SIGN_PASSPHRASE`
    - [x] `PYPI_USER`
    - [x] `PYPI_PASSWORD`
  - [x] Replace signing key (same key, different encryption)
  - [x] Deploy to PyPI
    - [x] Split sdist/wheel script into build and deploy script
  - [x] Deploy to Github releases
    - [x] Cache Windows installer artifact
[...]

PR #2924, opened 2020-04-21
actions: no need to use a secret for the PyPI username

beardypig:
See https://github.com/streamlink/streamlink/pull/2656#discussion_r411630309.

I took the liberty of doing it, in case @back-to didn't get round to it :)

We'll need to remove the `PYPI_USER` secret after this is merged too.


bastimeyer:
Looks like I didn't delete the env var fast enough, because it's still censored in the merge commit. Fixed on the latest commit though.

beardypig:
Perfect.

PR #2925, opened 2020-04-21
stream.hls: customizable playlist reload times

bastimeyer:
Adds support for customizable HLS playlist reload times via `--hls-playlist-reload-time`.

Values can be
- numbers (float) greater than 2 - self-explanatory
- `"segment"` - sets refresh time to the duration of the last segment
- `"live-edge"` - sets refresh time to the sum of durations of the `live-edge - 1`  last segments

For all other values, falls back to the playlist's `target_duration` if it is set, otherwise does the same as `live-edge` (this logic got changed, was `segment` before), and if no target duration and no segments are present, then the default value of 15 is used.

Also renames `HLSStreamWorker._set_playlist_reload_time` to `HLSStreamWorker._playlist_reload_time` and sets the reload time from its return value instead.

----

The changes are rather trivial, but I'd like to improve the integration tests a bit. It currently only tests one playlist refresh. Ideally, multiple playlists should be tested, but I couldn't set up a simple method spy with unittest.mock. Help and feedback is appreciated.

### DIFF
--- a/src/streamlink/plugins/showroom.py
+++ b/src/streamlink/plugins/showroom.py
@@ -80,8 +80,8 @@ _info_pages = set((
 
 
 class ShowroomHLSStreamWorker(HLSStreamWorker):
-    def _set_playlist_reload_time(self, playlist, sequences):
-        self.playlist_reload_time = 1.5
+    def _playlist_reload_time(self, playlist, sequences):
+        return 1.5
 
 
 class ShowroomHLSStreamReader(HLSStreamReader):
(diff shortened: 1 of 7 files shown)


gravyboat:
Do we care this doesn't work on 2.7? I'd rather just note that it doesn't. We already have PRs open to remove it.

bastimeyer:
To be honest, I didn't properly check my changes regarding Python 2 (lack of experience) before submitting and also didn't run local tests on Python 2.

I wasn't expecting this to be included in 1.4.0 anyway, but if somebody feels like it's important, then this can be fixed of course. Just wanted to get this out for review before I forget about it.

gravyboat:
Now that we're past that pycountry hurdle I'm going to merge this since the next release will drop 2.7 support. Thanks @bastimeyer.

PR #2926, opened 2020-04-21
Release 1.4.0

bastimeyer:
Let's prepare the 1.4.0 release...

Since there are so many changes, I had to add a list of notable changes instead of writing the usual text. It's better anyway, but "breaks" the "old format". The order is arbitrarily chosen and I've probably missed a few changes which should go on that list (only picked those with lots of comments and thumbs up reactions, if that makes sense).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,171 @@
 # Changelog
 
+## streamlink 1.4.0 (2020-04-22)
+
+This will be the last release with support for Python 2, as it has finally reached its EOL at the beginning of this year.
+
+Streamlink 1.4.0 comes with lots of plugin fixes/improvements, as well as some new features and plugins, and also a few plugin removals.
+
+Notable changes:
+
+- New: low latency streaming on Twitch via `--twitch-low-latency` ([#2513](https://github.com/streamlink/streamlink/pull/2513))
+- New: output HLS segment data immediately via `--hls-segment-stream-data` ([#2513](https://github.com/streamlink/streamlink/pull/2513))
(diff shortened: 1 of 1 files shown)


bastimeyer:
#2922 needs to be merged first

CHANGELOG.md
```diff
@@ -1,5 +1,171 @@
...
+- Fixed: Youtube plugin ([#2858](https://github.com/streamlink/streamlink/pull/2858))
+- Fixed: Crunchyroll plugin ([#2788](https://github.com/streamlink/streamlink/pull/2788))
+- Fixed: Pixiv plugin ([#2840](https://github.com/streamlink/streamlink/pull/2840))
+- Fixed: TVplayer plugin ([#2802](https://github.com/streamlink/streamlink/pull/2802))
+- Fixed: Zattoo plugin ([#2887](https://github.com/streamlink/streamlink/pull/2887))
+- Changed: set Firefox User-Agent HTTP header by default ([#2795](https://github.com/streamlink/streamlink/pull/2795))
+- Changed: upgraded bundled FFmpeg to `4.2.2` in Windows installer ([#2916](https://github.com/streamlink/streamlink/pull/2916))
+
+
+```
```
back-to, line 25: not sure if it changes anything, but all the other ones are 
```suggestion
```text
```

bastimeyer:
Are we good here?

bastimeyer:
I was just writing a comment here regarding the docs, which didn't update correctly on the stable directory, even though everything was built correctly. Github must be caching the GH-pages builds when building multiple commits in a fast succession. It's working now, so ignore this comment. Just wanted to mention it, in case anybody is (or will be) wondering (on future releases).

So yeah, everything worked fine with the new deploy config :tada:

PR #2927, opened 2020-04-21
plugins.nbcnews: new plugin for http://nbcnews.com/now

vstavrinov:
NBC News NOW is a streaming network from NBC News. Get all the latest on the day’s biggest stories and all aspects of the coronavirus crisis from the global resources of NBC News.


bastimeyer:
Thanks for the plugin submission!

Not sure how safe it is to use a static API ID here (the `220018` value), but extracting it from the script files is problematic, too, as they're minified. This is where the value is coming from:

`main` js script loaded by the site's script tag (with appended hash):
`https://ndassets.s-nbcnews.com/main-35a7208c5c9dd9c46a24.js`
```js
var g = a("j9g7")
          , m = n(g)
          , N = a("y3TV")
          , T = n(N)
          , L = t.NEWS_NOW_PID = "220018"
          , h = t.MSNBC_TVE_PID = "2000762"
          , w = 0;
        t.playmakerConfiguration = {
            news: {
                leapConfigUrl: "https://api.leap.nbcsports.com/pid/NBCNews/{PID}/v4/{PLATFORM}"
            },
            msnbc: {
                leapConfigUrl: "https://api.leap.nbcsports.com/pid/MSNBC/{PID}/v4/{PLATFORM}"
            }
        }
```

There's also `https://stream.nbcsports.com/data/event_config_220018.json` which includes a link to `https://stream.nbcnews.com/data/live_sources_220018.json` which has all the m3u8 URLs as well.

Also, regarding the implementation, it'd be better to use Streamlink's validation API for getting the HLS URLs from the JSON.

vstavrinov:
@bastimeyer :
> extracting it from the script files is problematic, too, as they're minified. 

Here is a solution: #2534

bastimeyer:
Parsing a minimized JS file and getting the value from an AST is just a little bit better than reading it from a regex. It's still unstable and can break very easily. They just have to change a variable name for example.

I personally don't have anything against a JS parser, but such a dependency should be maintained, which esprima doesn't seem to be, especially in the JS world which changes so rapidly.

src/streamlink/plugins/nbcnews.py
```diff
@@ -0,0 +1,34 @@
...
+    }, validate.get("videoSources"), validate.get(0), validate.get("sourceUrl"))
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls.url_re.match(url) is not None
+
+    def _get_streams(self):
+        html = self.session.http.get(self.url).text
+        match = self.js_re.search(html)
+        js = self.session.http.get(match[0]).text
```
back-to, line 27: depends if the next release should work with python2 or not

```
TypeError: '_sre.SRE_Match' object has no attribute '__getitem__'
```

back-to:
the channel is not live, but the plugin will play a 4h long VOD

this should probably ignored, instead of playing it.

vstavrinov:
> the channel is not live, but the plugin will play a 4h long VOD

Sad surprise.  There is no live stream. The live stream is here: 

https://pluto.tv/live-tv/nbc-news-now

Waiting for  #2747

vstavrinov:
Weird: there was no live stream on weekend but it has emerged now. Does it mean the live stream is not permanent? But nevertheless, for now, the plugin produces the same live stream (really live, not VOD) as the site does. So You can safely merge it.

vstavrinov:
> check if the stream is live with `type` or `status`

What for? For debug info and don't play if VOD? It looks like this VOD is the most recent stream that was live before terminated (or maybe digest or the last fragment)

back-to:
> What for? ... don't play if VOD?

Yes don't open the Stream if it is a VOD only play it as a Livestream.

vstavrinov:
The interesting thing: when you get stream URL and play it directly You can see it is VOD, but streamlink itself has no such indicator. So when You play VOD with streamlink You could think it is live.  Maybe it is worth to add such feature.

PR #2930, opened 2020-04-24
build: include correct signing key: 0xE3DB9E282E390FA0

beardypig:
I didn't check the signing key last time, this is the correct key with the correct ID. 

Resolves #2929.

PR #2932, opened 2020-04-25
actions: fixed incorrect versions and use names for codecov

back-to:
such as `streamlink==0+untagged`
https://github.com/streamlink/streamlink/runs/616367673?check_suite_focus=true#step:4:89

also `depth=1` won't work see
https://github.com/streamlink/streamlink.github.io/blob/e066aa52513b89304bf3396e52615c8b3109f1b4/latest/api.html#L10

use better upload names for codecov
> os:ubuntu-latest py:3.7

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,6 +43,8 @@ jobs:
       - name: Upload coverage data
         continue-on-error: ${{ matrix.continue || false }}
         uses: codecov/codecov-action@v1
+        with:
+          name: os:${{ matrix.os }} py:${{ matrix.python }}
 
   documentation:
     name: Test docs
@@ -52,7 +54,7 @@ jobs:
         with:
           fetch-depth: 300
       - name: Fetch tags
-        run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+        run: git fetch --depth=300 origin +refs/tags/*:refs/tags/*
       - uses: actions/setup-python@v1
         with:
           python-version: 3.8
@@ -71,7 +73,7 @@ jobs:
         with:
           fetch-depth: 300
       - name: Fetch tags
-        run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+        run: git fetch --depth=300 origin +refs/tags/*:refs/tags/*
       - uses: actions/setup-python@v1
         with:
           python-version: 3.8
@@ -106,7 +108,7 @@ jobs:
         with:
           fetch-depth: 300
       - name: Fetch tags
-        run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+        run: git fetch --depth=300 origin +refs/tags/*:refs/tags/*
       - uses: actions/setup-python@v1
         with:
           python-version: 3.8
@@ -132,7 +134,7 @@ jobs:
         with:
           fetch-depth: 300
       - name: Fetch tags
-        run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+        run: git fetch --depth=300 origin +refs/tags/*:refs/tags/*
       - uses: actions/setup-python@v1
         with:
           python-version: 3.8


back-to:
also seems to be wrong

https://github.com/streamlink/streamlink/runs/617922804?check_suite_focus=true#step:5:87

it uses a merge commit, so the version is wrong

https://github.com/streamlink/streamlink/commits/3bb25c7

it should be 1 but it is 2

.github/workflows/main.yml
```diff
@@ -28,6 +28,10 @@ jobs:
     timeout-minutes: 60
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 300
+      - name: Fetch tags
+        run: git fetch --depth=300 origin +refs/tags/*:refs/tags/*
```
bastimeyer, line 34: I've left the default parameters of `actions/checkout@v2` (which is `--depth=1 --no-tags`) on the test runners for performance reasons, so that it doesn't fetch unnecessary files from the git history, like the ffmpeg binary file (~30MiB) for example that we've now finally removed. Fetching all tags also means fetching this file.

I am aware that versioneer (or `git describe --tags`) doesn't work if there is not a single tag in the cloned repo, but is there a reason why the test runners need a correct version string? All that's important is the commit ID.

The depth value of `300` on all the additional "Fetch tags" commands should be set back to `1`, as it's either unnecessary or redundant. It is there to limit fetching more commits than the fetch depth value defines in the `actions/checkout@v2` action, but it'll still fetch all tags, regardless the depth value. Unfortunately, you can't fetch just the latest tag with the current git version, which would be the perferred thing to do here. If we **really** need only the latest tag, then it could be read from the Github API and the additional command be put into a local action inside the `.github` directory for DRY reasons.

.github/workflows/main.yml
```diff
@@ -43,6 +47,8 @@ jobs:
       - name: Upload coverage data
         continue-on-error: ${{ matrix.continue || false }}
         uses: codecov/codecov-action@v1
+        with:
+          name: os:${{ matrix.os }} py:${{ matrix.python }}
```
bastimeyer, line 51: Where can I see the different report names on codecov?

.github/workflows/main.yml
```diff
@@ -52,7 +58,7 @@ jobs:
         with:
           fetch-depth: 300
       - name: Fetch tags
-        run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+        run: git fetch --depth=300 origin +refs/tags/*:refs/tags/*
```
bastimeyer, line 61: 300 is not needed on the test docs runner. It should just verify the docs config and files. The final build is irrelevant.

.github/workflows/main.yml
```diff
@@ -71,7 +77,7 @@ jobs:
...
```
bastimeyer, line 80: Not needed

.github/workflows/main.yml
```diff
@@ -106,7 +112,7 @@ jobs:
...
```
bastimeyer, line 115: The latest docs surprise me a bit though.

> 1.4.0+123.gaf86d38

https://github.com/streamlink/streamlink/commit/af86d38
https://github.com/streamlink/streamlink/runs/616381381?check_suite_focus=true#step:3:1

Why does it use `1.3.1` as the base (`+123`), but still refers to `1.4.0` in its output string? It should be `1.4.0+1`.
By the way, `1.4.0` is correct here, because the `1.4.1` tag wasn't pushed to the repo yet on the `1.4.1` release commit when the latest docs were built.

Is this a bug in versioneer? Does `git describe --tags` work?

.github/workflows/main.yml
```diff
@@ -132,7 +138,7 @@ jobs:
...
```
bastimeyer, line 141: Not needed

.github/workflows/main.yml
back-to: https://codecov.io/gh/streamlink/streamlink/commit/6ef3bbc99eee5abbf3e049c5e8f01326b456a7c4/build

bastimeyer:
Sorry for the confusion. I needed a bit to properly figure this out first.

The reason why the version string of the latest docs build is based on `1.3.1` is that git incorrectly follows the commit history and therefore sees a different `master` branch. That's because of PR #2930, which was merged into `master` after the `1.4.0` release got tagged and was based on 1e8df7a8, which was committed before `1.4.0` got tagged. It basically skipped the `1.4.0` tag in-between the merge and PR base. That is a bit weird, because it's a merge **onto** `master` and not the other way around, so it shouldn't skip the real `master` branch. The reason why it's still using `1.4.0` in the version string is because all tags get fetched afterwards, but are not always connected to the followed commit branch.

This is happening because of the difference in fetch logic of both fetch commands. The fetch command which the checkout action performs always sets the `--no-tags` parameter in addition to its depth, which means that it will simply follow the current branch's commit history (possibly an invalid path) to that depth while ignoring the tags. The custom command however fetches all tags first and can then fill in all the missing commits based on that information until a certain depth is reached, regardless "which branch" the tags are on, but this depends on the depth value used here.

This means that setting `--depth=300` on all custom fetch-tag commands "assures" that we get the correct commit history.

----

From my local repo with the 1.4.1 tag temporarily removed:

**Incorrect**

```
$ cd $(mktemp -d)
$ git clone --depth=300 --no-tags file:///home/basti/repos/streamlink .
$ git fetch --depth=1 origin '+refs/tags/*:refs/tags/*'
$ git describe --tags
1.4.0-123-gaf86d38
$ git log --graph --oneline 1e8df7a8~1..
* af86d38 (HEAD -> master, origin/master, origin/HEAD) Release 1.4.1 (#2931)
*   b56d294 Merge pull request #2930 from beardypig/correct-sdist-key
|\  
| * 462165d build: include correct signing key: 0xE3DB9E282E390FA0
| * 1e8df7a plugins.twitch: rewrite disable ads logic
* 9e05355 (grafted, tag: 1.4.0) Merge pull request #2926 from bastimeyer/releases/1.4.0
```

**Correct**

```
$ cd $(mktemp -d)
$ git clone --depth=300 --no-tags file:///home/basti/repos/streamlink .
$ git fetch --depth=300 origin '+refs/tags/*:refs/tags/*'
$ git describe --tags
1.4.0-3-gaf86d38a
$ git log --graph --oneline 1e8df7a8~1..
* af86d38a (HEAD -> master, origin/master, origin/HEAD) Release 1.4.1 (#2931)
*   b56d294f Merge pull request #2930 from beardypig/correct-sdist-key
|\  
| * 462165da build: include correct signing key: 0xE3DB9E282E390FA0
* |   9e053551 (tag: 1.4.0) Merge pull request #2926 from bastimeyer/releases/1.4.0
|\ \  
| * | 5da7e3ae Release 1.4.0
|/ /  
* | 39f64c09 add python 2.7 deprecation warning
* |   151abce3 Merge pull request #2924 from beardypig/actions-secrets-fix
|\ \  
| |/  
|/|   
| * 4269aae9 actions: no need to use a secret for the PyPI username
|/  
* 1e8df7a8 plugins.twitch: rewrite disable ads logic
```

.github/workflows/main.yml
bastimeyer: @back-to What's your opinion here? Do you want to keep this or can this part be reverted?

I'd really like to merge the other changes, but if you think that we need to fetch the tags on all test runners, then I don't want to block this. I'm just saying that it's unnecessary, as I don't see any negative impact when fetching only one commit other than having a couple of "untagged" lines in the parts of the build log that are not relevant to the test runs and coverage reports.

back-to:
so without this
https://github.com/streamlink/streamlink/compare/6ef3bbc99eee5abbf3e049c5e8f01326b456a7c4..2dfb86cd986f26ceb0bafb4df51a5b1ea2849237
?

Issue #2918, opened 2020-04-18
Showroom: "No playable streams found on this URL"


davidyu0728:
## Bug Report

- [x ] This is a bug report and I have read the contribution guidelines.

### Description

I have checked some live link in https://www.showroom-live.com/

It seems cannot get the room_id by using regular expression

### Log output

```
[cli][info] streamlink is running as root! Be careful!
[cli][debug] OS:         Linux-4.4.0-18362-Microsoft-x86_64-with-Ubuntu-18.04-bionic
[cli][debug] Python:     3.6.9
[cli][debug] Streamlink: 1.3.1+88.g222ab71
[cli][debug] Requests(2.23.0), Socks(1.7.1), Websocket(0.57.0)
[cli][info] Found matching plugin showroom for URL https://www.showroom-live.com/48_YABUSHITA_FU
[plugin.showroom][debug] Failed to find room_id for 48_YABUSHITA_FU using primary regex
[utils.l10n][debug] Language code: en_US
[utils.l10n][debug] Language code: en_US
error: No playable streams found on this URL: https://www.showroom-live.com/48_YABUSHITA_FU
```

littmus:
I looked at this problem a bit, and room_id is found normally by _room_id_alt_re. it's not a problem.
Problem is at hls_playlist.py. Showroom stream's m3u8 seems that there is no stream information, so it's playlist is not created when parsing m3u8.
I don't know much about m3u8 parsing, so hope someone who knows well solve this problem.

PR #2939, opened 2020-04-26
plugins.showroom: use normal HLSStreams

back-to:
closes https://github.com/streamlink/streamlink/issues/2918

```
$ streamlink https://www.showroom-live.com/--- low --hls-segment-threads 10 -l info
[cli][info] Found matching plugin showroom for URL https://www.showroom-live.com/---
[cli][info] Available streams: low (worst), high (best)
[cli][info] Opening stream: low (hls)
```


gravyboat:
Looks good, thanks @back-to.

Issue #2943, opened 2020-05-01
plugin(DouyuTV): 401 Client Error: Unauthorized


hxzhao527:
## Plugin Issue
- [x] This is a plugin issue and I have read the contribution guidelines.

### Description
As the log-output show, plugin DouyuTV doesn't work.

### Reproduction steps / Explicit stream URLs to test
see log output

### Log output

```
streamlink --loglevel debug douyu.com/9999
[cli][debug] OS:         Linux-4.19.118-1-MANJARO-x86_64-with-glibc2.2.5
[cli][debug] Python:     3.8.2
[cli][debug] Streamlink: 1.4.1
[cli][debug] Requests(2.23.0), Socks(1.7.1), Websocket(0.57.0)
[cli][info] Found matching plugin douyutv for URL douyu.com/9999
error: Unable to open URL: https://capi.douyucdn.cn/api/v1/room/9999?aid=wp&cdn=ws&client_sys=wp&time=1588307909&auth=ab4f747bfebca8b6e58e710b1bea2b88 (401 Client Error: Unauthorized for url: https://capi.douyucdn.cn/api/v1/room/9999?aid=wp&cdn=ws&client_sys=wp&time=1588307909&auth=ab4f747bfebca8b6e58e710b1bea2b88)
```

### History
#2328 have been locked, but the bug of plugin DouyuTV still exists.

There are some codes shown below the old issue, but there is no any pr to fix it.

hxzhao527:
The new api to get rtmp-address is  https://www.douyu.com/lapi/live/getH5Play.
![image](https://user-images.githubusercontent.com/8402502/80798056-01f02900-8bd6-11ea-81bc-68029cc008bc.png)

The major problem is the parameter sign generation. Using python to impl the generation function which is js originally does't work well.

[Js2py](https://github.com/PiotrDabkowski/Js2Py) works, but this patch need to import some deps.

If importing new deps is acceptable, I'm glad to make a pr.

PR #2945, opened 2020-05-03
fix(plugin:douyutv): fix issue 2943

hxzhao527:
fix issue #2943 

import new dep [js2py](https://github.com/PiotrDabkowski/Js2Py)

NO py2 support

### DIFF
--- a/src/streamlink/plugins/douyutv.py
+++ b/src/streamlink/plugins/douyutv.py
@@ -1,25 +1,26 @@
+import hashlib
 import re
 import time
-import hashlib
 
-from requests.adapters import HTTPAdapter
+import js2py
 
 from streamlink.plugin import Plugin
-from streamlink.plugin.api import validate, useragents
-from streamlink.stream import HTTPStream, HLSStream, RTMPStream
+from streamlink.plugin.api import validate
+from streamlink.stream import HLSStream, HTTPStream
 
-API_URL = "https://capi.douyucdn.cn/api/v1/{0}&auth={1}"
-VAPI_URL = "https://vmobile.douyu.com/video/getInfo?vid={0}"
-API_SECRET = "zNzMV1y4EMxOHS6I5WKm"
-SHOW_STATUS_ONLINE = 1
-SHOW_STATUS_OFFLINE = 2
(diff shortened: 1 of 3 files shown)


src/streamlink/plugins/douyutv.py
```diff
@@ -1,25 +1,26 @@
+import hashlib
 import re
 import time
-import hashlib
 
-from requests.adapters import HTTPAdapter
+import js2py
```
back-to, line 5: adding a new dependency is bad

Erk-:
I think that we could look into is if the js is static in any way, Last time I looked at it they sent what seemed to be a lot different js. we could probably dump it and see if there is a difference between the code from channel to channel or connection to connection.

hxzhao527:
Old: https://web.archive.org/web/20200502075004/https://www.douyu.com/9999  
Now: https://www.douyu.com/9999  
we can see the sign function `ub98484234` have different content.  
and the value returned is not same

Erk-:
If you try a bunch of streams? Will the sign function be similar or seemingly random?

hxzhao527:
The sign function seems updated daily. But it is same between  different rooms.

If importing new dep is not allowed, we have to make a new release when douyu update.  
Or implement a transformer to translate the js code by ourself.

Erk-:
Could you try and capture a few days of sign code so we can diff them and see if there is any pattern we can use. 

I do agree that the easiest way would be to use something like js2py or similar though.

steven7851:
I think you can create a JS engine for plugin:douyutv. Like [ykdl](https://github.com/zhangn1985/ykdl/blob/master/ykdl/util/jsengine.py) have their own one.

Erk-:
That would add a external depedency, which may or may not be preffered.

hxzhao527:
I have captured three days, and upload the result to https://github.com/hxzhao527/4Issue/tree/master/streamlink-plugin-douyu

here is the diff result with vscode
image
image

Erk-:
I cannot really see if there is a way to make a generic implementation, it looks like it is doing some kind of hash and/or encryption. So I think it will be up to the maintainers how to continue from here.

Issue #2948, opened 2020-05-05
Film on fails channel id not found


jamiejacksoncumbria:
## Plugin Issue

- [ ] This is a plugin issue and I have read the contribution guidelines.

### Description

<Since version 1.4.1 filmon does not work. Found channel ID: None>

### Reproduction steps / Explicit stream URLs to test

```
/usr/local/bin/streamlink http://www.filmon.com/channel/cbs-action high --player-external-http --player-external-http-port 23530 --player-continuous-http -l debug
```

### Log output
```
/usr/local/lib/python2.7/dist-packages/streamlink_cli/main.py:760: PythonDeprecatedWarning: Python 2.7 has reached the end of its life.  A future version of streamlink will drop support for Python 2.7. Please upgrade your Python to at least 3.5.
  streamlink = Streamlink({"user-input-requester": ConsoleUserInputRequester(console)})
[cli][debug] OS:         Linux-4.15.0-99-generic-x86_64-with-Ubuntu-18.04-bionic
[cli][debug] Python:     2.7.17
[cli][debug] Streamlink: 1.4.1
[cli][debug] Requests(2.23.0), Socks(1.7.1), Websocket(0.57.0)
[cli][info] Found matching plugin filmon for URL http://www.filmon.com/channel/cbs-action
[plugin.filmon][debug] Found channel ID: None
[plugin.filmon][debug] channel sleep 0
[plugin.filmon][debug] channel sleep 1
[plugin.filmon][debug] channel sleep 2
[plugin.filmon][debug] channel sleep 3
[plugin.filmon][debug] channel sleep 4
[plugin.filmon][debug] Reset cached channel: cbs-action
Traceback (most recent call last):
  File "/usr/local/bin/streamlink", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/streamlink_cli/main.py", line 1024, in main
    handle_url()
  File "/usr/local/lib/python2.7/dist-packages/streamlink_cli/main.py", line 585, in handle_url
    streams = fetch_streams(plugin)
  File "/usr/local/lib/python2.7/dist-packages/streamlink_cli/main.py", line 465, in fetch_streams
    sorting_excludes=args.stream_sorting_excludes)
  File "/usr/local/lib/python2.7/dist-packages/streamlink/plugin/plugin.py", line 323, in streams
    ostreams = list(ostreams)
[...]

PR #2962, opened 2020-05-10
plugins.filmon: use /tv/ url and raise PluginError for invalid channels

back-to:
closes https://github.com/streamlink/streamlink/issues/2948


gravyboat:
Thanks @back-to.

PR #2973, opened 2020-05-12
docs: update debian install instructions

amurzeau:
Users should now use Debian backports when using Debian stable instead of using bintray.


gravyboat:
Good stuff, thanks.

PR #2991, opened 2020-05-26
plugins.ceskatelevize: add new api for some links

### LINKED ISSUE #2984: CT Sport with ceskatelevize.py doesn't work
## Bug Report

CT Sport with ceskatelevize.py doesn't work.
CT1, CT2, CT24, CT Decko and CT Art are okay, but CT Sport see Description Logfiles

c:\Streamlink\1_3_1\bin\streamlink.exe https://www.ceskatelevize.cz/sport/zive-vysilani/ best -l trace
or
c:\Streamlink\1_3_1\bin\streamlink.exe https://sport.ceskatelevize.cz/#live best -l trace

1. latest Streamlink Version
2. latest ceskatelevize.py but this Version is two years old
3. ...

### Log output

```
c:\Streamlink\1_3_1\bin\streamlink.exe https://www.ceskatelevize.cz/sport/zive-vysilani/ best -l trace
[09:22:36,953][cli][debug] OS:         Windows 10
[09:22:36,954][cli][debug] Python:     3.6.6
[09:22:36,956][cli][debug] Streamlink: 1.3.1
[09:22:36,957][cli][debug] Requests(2.22.0), Socks(1.7.1), Websocket(0.57.0)
[09:22:36,958][cli][info] Found matching plugin ceskatelevize for URL https://www.ceskatelevize.cz/sport/zive-vysilani/
[09:22:36,958][plugin.ceskatelevize][warning] SSL certificate verification is disabled.
error: Cannot find playlist info in the player url!
```

But that is not correct.
At the CT Sport Homepage Chrome Network Tools finds a mpd Link.
The mpd Link is not a DRM Link
This Links are at CT1, CT2, CT24 ... too

Billy2011:
fixes #2984

### DIFF
--- a/src/streamlink/plugins/ceskatelevize.py
+++ b/src/streamlink/plugins/ceskatelevize.py
@@ -13,89 +13,52 @@ Additionally, videos from iVysilani archive should work as well.
 """
 import logging
 import re
+import json
 
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import useragents, validate
-from streamlink.stream import HLSStream
+from streamlink.stream import HLSStream, DASHStream
 from streamlink.exceptions import PluginError
-
-_url_re = re.compile(
-    r'http(s)?://([^.]*.)?ceskatelevize.cz'
-)
-_player_re = re.compile(
-    r'ivysilani/embed/iFramePlayer[^"]+'
-)
-_hash_re = re.compile(
-    r'hash:"([0-9a-z]+)"'
-)
-_playlist_info_re = re.compile(
-    r'{"type":"([a-z]+)","id":"([0-9]+)"'
-)
-_playlist_url_schema = validate.Schema({
-    validate.optional("streamingProtocol"): validate.text,
-    "url": validate.any(
-        validate.url(),
-        "Error",
-        "error_region"
-    )
-})
-_playlist_schema = validate.Schema({
-    "playlist": [{
-        validate.optional("type"): validate.text,
-        "streamUrls": {
-            "main": validate.url(),
-        }
-    }]
-})
+from streamlink.compat import html_unescape, quote
 
 log = logging.getLogger(__name__)
-
-
-def _find_playlist_info(response):
-    """
-    Finds playlist info (type, id) in HTTP response.
-
-    :param response: Response object.
-    :returns: Dictionary with type and id.
-    """
-    values = {}
-    matches = _playlist_info_re.search(response.text)
-    if matches:
-        values['type'] = matches.group(1)
-        values['id'] = matches.group(2)
-
-    return values
-
-
-def _find_player_url(response):
-    """
-    Finds embedded player url in HTTP response.
-
-    :param response: Response object.
-    :returns: Player url (str).
-    """
-    url = ''
-    matches = _player_re.search(response.text)
-    if matches:
-        tmp_url = matches.group(0).replace('&amp;', '&')
-        if 'hash' not in tmp_url:
-            # there's no hash in the URL, try to find it
-            matches = _hash_re.search(response.text)
-            if matches:
-                url = tmp_url + '&hash=' + matches.group(1)
-        else:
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/ceskatelevize.py
```diff
@@ -145,10 +115,158 @@ def _get_streams(self):
...
+                        "quality": pl["quality"],
+                        "region": pl["region"]
+                    }]
+                }
+
+        headers = {
+            "Content-Type": "application/x-www-form-urlencoded; charset=UTF-8",
+            'Accept': 'application/json, text/javascript, */*; q=0.01',
+            'Accept-Language': 'de,en-US;q=0.7,en;q=0.3',
+            'Origin': 'https://sport.ceskatelevize.cz',
```
back-to, line 255: so this would only work for `sport` ?

src/streamlink/plugins/ceskatelevize.py
```diff
@@ -145,10 +115,158 @@ def _get_streams(self):
...
+                        "canBePlay": pl["canBePlay"],
+                        "quality": pl["quality"],
+                        "region": pl["region"]
+                    }]
+                }
+
+        headers = {
+            "Content-Type": "application/x-www-form-urlencoded; charset=UTF-8",
+            'Accept': 'application/json, text/javascript, */*; q=0.01',
+            'Accept-Language': 'de,en-US;q=0.7,en;q=0.3',
```
back-to, line 254: seems wrong or not needed

src/streamlink/plugins/ceskatelevize.py
Billy2011: I've not found any other link.

back-to: if `https://sport.ceskatelevize.cz` only uses the new api,
it might be more readable if you use a new plugin file for this subdomain instead of two plugins in one file.

src/streamlink/plugins/ceskatelevize.py
Billy2011: This is probably the minimum required header, but I do not know if it is always like this:
`{"Content-Type": "application/x-www-form-urlencoded; charset=UTF-8"}`

If you have nothing to complain about, then I will adapt the code accordingly?

src/streamlink/plugins/ceskatelevize.py
Billy2011: However, this is not always so clearly recognizable:
`https://www.ceskatelevize.cz/sport/zive-vysilani/`

back-to: `https://www.ceskatelevize.cz/sport/zive-vysilani/` as an URL is invalid,
it redirects to `https://sport.ceskatelevize.cz/zive-vysilani#live` so it should work if you would use the correct URL.

Billy2011: This is not my concern, but the unrecognizable / existing sport subdomain in the link.

A new plugin is not enough, because it also requires changes to the original plugin etc. and the associated complication of the url regex. It is also not excluded that other links switch to the new API.

I find my solution best because the links that cannot be resolved are automatically redirected to the new API and it also requires the least code changes.

Billy2011: Is a new plugin still the best solution and if so, should it only resolve URLs of the sport subdomain?

src/streamlink/plugins/ceskatelevize.py
back-to: use the minimum required header,
using `'Accept-Language': 'de,en-US;` on a **Czech** website as default seems wrong

src/streamlink/plugins/ceskatelevize.py
back-to: just let it like this for now,
it is not that important

PR #3000, opened 2020-05-31
plugins.reshet: plugin removal

### LINKED ISSUE #2996: Plugin 'reshet' changed URL
## Plugin Issue

### Description

Hi,
"reshet.tv" has changed to "13tv.co.il", making the plugin unuseable.
I suspect there are also changes to the API.

mkbloke:
Closes #2996


mkbloke:
It looks like the original plugin supported VODs as well as live.  I have not looked into the VODs on the site.  If VOD support is wanted, I could look into it and close this PR, although I believe streamlink is generally moving away from supporting VODs in the plugins, from the comments made by various streamlink devs in the past...

PR #3003, opened 2020-06-03
Streamlink was added to Windows Package Manager, adds info on how to install it to the docs

superusercode:
This PR adds info on how to install it from there, with the same format at the chocolatey link.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -24,11 +24,21 @@ Portable                             See the `Windows portable version`_ section
                                         choco install streamlink
 
                                      `Installing Chocolatey packages`_
+                                     
+`Windows Package Manager`_           .. code-block:: bat
+
+                                       winget install streamlink
+                                        
+                                     `Installing Winget packages`_
 ==================================== ===========================================
 
 .. _Chocolatey: https://chocolatey.org/packages/streamlink
+
+.. _Windows Package Manager: https://github.com/microsoft/winget-pkgs/tree/master/manifests/Streamlink/Streamlink
+
 .. _Installing Chocolatey packages: https://chocolatey.org
 
+.. _Installing Winget packages: https://devblogs.microsoft.com/commandline/windows-package-manager-preview/
 
 macOS
 -----


gravyboat:
Looks good to me, probably worth updating https://github.com/microsoft/winget-pkgs/blob/master/manifests/Streamlink/Streamlink/1.4.1.yaml at some point with our newer description from the README that cuts out the flash mention, `Streamlink is a CLI utility which pipes video streams from various services into a video player, such as VLC.`. Or are we still using that old description somewhere?

docs/install.rst
```diff
@@ -24,11 +24,21 @@ Portable                             See the `Windows portable version`_ section
...
+                                     `Installing Winget packages`_
 ==================================== ===========================================
 
 .. _Chocolatey: https://chocolatey.org/packages/streamlink
+
+.. _Windows Package Manager: https://github.com/microsoft/winget-pkgs/tree/master/manifests/Streamlink/Streamlink
+
 .. _Installing Chocolatey packages: https://chocolatey.org
 
+.. _Installing Winget packages: https://devblogs.microsoft.com/commandline/windows-package-manager-preview/
```
bastimeyer, line 41: The link should point to the official docs instead of a blog article.
https://docs.microsoft.com/en-us/windows/package-manager/

superusercode:
@gravyboat It's the first line of the chocolatey package description https://chocolatey.org/packages/streamlink

docs/install.rst
```diff
@@ -24,11 +24,21 @@ Portable                             See the `Windows portable version`_ section
                                         choco install streamlink
 
                                      `Installing Chocolatey packages`_
+                                     
+`Windows Package Manager`_           .. code-block:: bat
+
+                                       winget install streamlink
```
bastimeyer, line 30: Wrong indentation. The block content should be aligned with the first character of the block label, which means one space is missing.

gravyboat: Missed this on my review, sorry @bastimeyer, shouldn't review PRs when it's almost time to sleep!

PR #3012, opened 2020-06-05
ci: no test/documentation jobs on scheduled run

bastimeyer:
These jobs are unnecessary on scheduled builds, since only commits get merged into master which have passed the tests previously.
Also avoids recurring code coverage uploads if no commits get merged into master between two scheduled builds.

----

Ideally we should cancel the whole workflow run if there were no new commits since the last scheduled build, but that would require adding an additional job that runs before the test, documentation and installer jobs.

----

Duplicate code coverage upload:
- https://github.com/streamlink/streamlink/runs/740545127?check_suite_focus=true#step:7:40
- https://codecov.io/gh/streamlink/streamlink/commit/90e634ec17cd7b90f062f25f2b620e50285dd9c0/build

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,6 +15,7 @@ env:
 jobs:
   test:
     name: Test
+    if: github.event_name != 'schedule'
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


back-to:
Might result in not seeing other dependencies issues as fast, but should not be important.

PR #3017, opened 2020-06-07
Fixed typo (synonmys => synonyms)

resloved:
Only saw it while reading through the source code. Really doesn't matter.


gravyboat:
A fix is a fix, thanks!

PR #3018, opened 2020-06-09
fix radionet plugin so it works with new radio.net page format

lanroth:
I think the page layout for radio.net changed about a month ago, at least that's when the radionet plugin stopped working for me.

I hacked the code so it works for me.

Note - the plugin is hardcoded to return `live` for bitrate as I don't think this is available any more.

Here's a trimmed example of the JSON we scrape from the page:
```
{
  'id': 'bbcradio4lw',
  'name': 'BBC Radio 4 Long Wave',
  'streams': [
    {
      'url': 'http://bbcmedia.ic.llnwd.net/stream/bbcmedia_radio4lw_mf_q',
      'contentFormat': 'audio/mpeg'
    }
  ],
  'type': 'STATION'
}
```

### DIFF
--- a/src/streamlink/plugins/radionet.py
+++ b/src/streamlink/plugins/radionet.py
@@ -1,9 +1,13 @@
+import logging
 import re
 
+from streamlink.compat import urlparse, urlunparse
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
-from streamlink.stream import HTTPStream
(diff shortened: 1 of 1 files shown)


lanroth:
Thanks for the suggestions, @bastimeyer 

I've tidied the code but haven't had time to look at `.m3u8` streams.

PR #3045, opened 2020-06-26
Support bilibili custom API host

NyaMisty:
Bilibili live stream's API will block requests from cloud services like GCP/AWS. However they don't block the cloud service from downloading the stream from CDN.
So we can replace the API host with custom reverse proxy to bypass that restriction.

### DIFF
--- a/src/streamlink/plugins/bilibili.py
+++ b/src/streamlink/plugins/bilibili.py
@@ -2,7 +2,7 @@ import logging
 import re
 
 from streamlink.compat import urlparse
-from streamlink.plugin import Plugin
+from streamlink.plugin import Plugin, PluginArguments, PluginArgument
 from streamlink.plugin.api import validate, useragents
 from streamlink.stream import HTTPStream
 
--- a/src/streamlink/plugins/bilibili.py
+++ b/src/streamlink/plugins/bilibili.py
@@ -8,8 +8,9 @@ from streamlink.stream import HTTPStream
 
 log = logging.getLogger(__name__)
 
-API_URL = "https://api.live.bilibili.com/room/v1/Room/playUrl"
-ROOM_API = "https://api.live.bilibili.com/room/v1/Room/room_init?id={}"
+API_HOST = "https://api.live.bilibili.com"
+API_URL = "/room/v1/Room/playUrl"
+ROOM_API = "/room/v1/Room/room_init?id={}"
 SHOW_STATUS_OFFLINE = 0
 SHOW_STATUS_ONLINE = 1
 SHOW_STATUS_ROUND = 2
@@ -43,6 +44,15 @@ _room_stream_list_schema = validate.Schema(
 
 
 class Bilibili(Plugin):
+    arguments = PluginArguments(
+        PluginArgument(
+            "apihost",
+            metavar="APIHOST",
+            default=API_HOST,
+            help="Use custom api host url to bypass bilibili's cloud blocking"
+        )
+    )
+    
     @classmethod
     def can_handle_url(self, url):
         return _url_re.match(url)
@@ -60,7 +70,7 @@ class Bilibili(Plugin):
             'Referer': self.url})
         match = _url_re.match(self.url)
         channel = match.group("channel")
-        res_room_id = self.session.http.get(ROOM_API.format(channel))
+        res_room_id = self.session.http.get(self.options.get("apihost") + ROOM_API.format(channel))
         room_id_json = self.session.http.json(res_room_id, schema=_room_id_schema)
         room_id = room_id_json['room_id']
         if room_id_json['live_status'] != SHOW_STATUS_ONLINE:
@@ -71,7 +81,7 @@ class Bilibili(Plugin):
             'quality': '4',
             'platform': 'web',
         }
-        res = self.session.http.get(API_URL, params=params)
+        res = self.session.http.get(self.options.get("apihost") + API_URL, params=params)
         room = self.session.http.json(res, schema=_room_stream_list_schema)
         if not room:
             return
--- a/src/streamlink/plugins/bilibili.py
+++ b/src/streamlink/plugins/bilibili.py
@@ -52,7 +52,7 @@ class Bilibili(Plugin):
             help="Use custom api host url to bypass bilibili's cloud blocking"
         )
     )
-    
+
     @classmethod
     def can_handle_url(self, url):
         return _url_re.match(url)


bastimeyer:
> will block requests from cloud services like GCP/AWS
> So we can replace the API host with custom reverse proxy

This argument can be made for every streaming service once they start blocking you or your entire network. I don't think it makes sense adding a plugin specific parameter for replacing certain HTTP(S) requests. @gravyboat, thoughts?

gravyboat:
@bastimeyer Agreed. This is simply an exploit so people can either re-host or download the content via a server instead of their own system which Streamlink has purposefully pivoted away from. It's intended for people to be able to watch livestreams live, not so you can abuse some service provider from your cloud server. I'm closing this.

NyaMisty:
I don't think you understood the scenario, so before all, I'll re-explain the idea. Bilibili's live stream works by retriving a stream video url with api, and then downloading it. Only the first step got blocked, so neither using global proxy, which will double you traffic cost, nor pivoting to other service, which almost all cloud service got blocked, makes any sense in this scenario.
Then first, adding this option won't affect other plugins, which means it's managable & maintainable.
Second, these tweaks must lives in their own plugin, as each service will block different requests, and can be bypassed using different techniques. And it OFC make sense to have a plugin option to tweak the plugin's own behaviour, even if it's only a HTTP request. So @bastimeyer I really can't agree with you.
Third, you think it's "intended" for people be able to watch, however why don't we expand its scenario with only a single option? Being a OSS we can be inclusive, and allow people with different usage scenarios to make use of it together and contributing to it together.
Last, personally I'm feeling offended having a normal, non-spam PR closed without any conversation. :(

bastimeyer:
> I don't think you understood the scenario

I do understand what you've said, but I don't think that what you're trying to achieve here fits in the scope of the project.

> first, adding this option won't affect other plugins, which means it's managable & maintainable.

It does affect other plugins, because it would open the door for more plugins to add their own parameters for stuff like that. And once your workaround becomes obsolete or stops working, then we'll have to deal with a breaking change, as we'd have to remove the parameter again, which is the opposite of maintainable.

> Second, these tweaks must lives in their own plugin, as each service will block different requests
> And it OFC make sense to have a plugin option to tweak the plugin's own behaviour, even if it's only a HTTP request.

Just like I've said, having 100s of plugins with individual request-modification parameters and workarounds of that kind isn't something that's well designed or makes sense. If you want to modify a plugin's logic for your personal needs, then write your own plugin, but it doesn't necessarily mean that it'll be useful for everyone else and that it should be included in Streamlink's master branch. This is especially true for reasons you've described, like your IP address being blocked by the streaming service, even if it's just the API for aquiring the actual stream URL.

> Third, you think it's "intended" for people be able to watch, however why don't we expand its scenario with only a single option? Being a OSS we can be inclusive, and allow people with different usage scenarios to make use of it together and contributing to it together.

This isn't meant in a rude way, but Streamlink uses a free license, so if you don't like what the maintainers think is the best for the project or if your ideas don't align with the project, then please go ahead and create a fork. When we forked the Livestreamer project and created Streamlink, we added a list of goals and plugin request restrictions in order to shift the project into a better direction. This also includes plugin/feature improvements or support requests, which are meant for abusing a certain streaming services.

> Last, personally I'm feeling offended having a normal, non-spam PR closed without any conversation. :(

That's understandable, but if you look at the contributing guidelines, you'll see that it mentions asking first before submitting a pull request with significant changes, or you'll risk wasting your time if your work won't get merged.
https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-requests

Closing an issue or pull request, unless it also gets locked, also doesn't mean the end of a discussion or conversation, hence my reply.

NyaMisty:
> > first, adding this option won't affect other plugins, which means it's managable & maintainable.
> 
> It does affect other plugins, because it would open the door for more plugins to add their own parameters for stuff like that. And once your workaround becomes obsolete or stops working, then we'll have to deal with a breaking change, as we'd have to remove the parameter again, which is the opposite of maintainable.
I don't think leaving an obsolete/not-working option is a bad idea. This option is a general option, which meant to proxy through some requests, and in fact various tools have these options to customize the API endpoint.
It's true that you have to be care to add options, however that doesn't mean to add no options.

> > Second, these tweaks must lives in their own plugin, as each service will block different requests
> > And it OFC make sense to have a plugin option to tweak the plugin's own behaviour, even if it's only a HTTP request.
> 
> Just like I've said, having 100s of plugins with individual request-modification parameters and workarounds of that kind isn't something that's well designed or makes sense. If you want to modify a plugin's logic for your personal needs, then write your own plugin, but it doesn't necessarily mean that it'll be useful for everyone else and that it should be included in Streamlink's master branch. This is especially true for reasons you've described, like your IP address being blocked by the streaming service, even if it's just the API for acquiring the actual stream URL.

I don't think having 100s of plugins with different request modification parameters isn't a well design. At least it's better than *nothing*. You can make these these options only shown with additional help command like -hh -hhh. Big softwares like clang / gcc all has various obsolete, un-documented switches, but no one says it "isn't a well design", as they are meant for advanced users.

> > Third, you think it's "intended" for people be able to watch, however why don't we expand its scenario with only a single option? Being a OSS we can be inclusive, and allow people with different usage scenarios to make use of it together and contributing to it together.
> 
> This isn't meant in a rude way, but Streamlink uses a free license, so if you don't like what the maintainers think is the best for the project or if your ideas don't align with the project, then please go ahead and create a fork. When we forked the Livestreamer project and created Streamlink, we added a list of goals and plugin request restrictions in order to shift the project into a better direction. This also includes plugin/feature improvements or support requests, which are meant for abusing certain streaming services.
> https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#plugin-requests
For this one, it's netither a breaking change, nor a DRM-related problem (as that platform only restricts cloud service to avoid server overload). If you say abusing, then in fact the intercepting undocumented js code & http requests, making the user leave the original site, which at last make the page rank of that site drops, is also kind of "abusing".

> > Last, personally I'm feeling offended having a normal, non-spam PR closed without any conversation. :(
> 
> That's understandable, but if you look at the contributing guidelines, you'll see that it mentions asking first before submitting a pull request with significant changes, or you'll risk wasting your time if your work won't get merged.
> https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-requests
> 
> Closing an issue or pull request, unless it also gets locked, also doesn't mean the end of a discussion or conversation, hence my reply.
Thanks for your clarification though.

Issue #3042, opened 2020-06-24
Picarto.tv VODs Not Working


Gameplayer9198:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

Picarto VODs Aren't working 

### Reproduction steps / Explicit stream URLs to test

1. Find a Premium Member streamer with VODs enabled (NSFW [https://picarto.tv/LiveForTheFunk](https://picarto.tv/LiveForTheFunk))
2. Enter there videos tab & copy link address (right click menu) a VOD
3. Pop it in too Streamlink to be disappointed 

### Log output

```
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.6.6
[cli][debug] Streamlink: 1.4.1
[cli][debug] Requests(2.23.0), Socks(1.7.1), Websocket(0.57.0)
[cli][info] Found matching plugin picarto for URL https://picarto.tv/videopopout/LiveForTheFunk_2020.06.20.01.17.30_nsfw.mkv
[plugin.picarto][error] The channel LiveForTheFunk_2020.06.20.01.17.30_nsfw.mkv does not exist
error: No playable streams found on this URL: https://picarto.tv/videopopout/LiveForTheFunk_2020.06.20.01.17.30_nsfw.mkv
```

### Additional comments, screenshots, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

PR #3046, opened 2020-06-27
plugins.picarto: fixed vod url detection

back-to:
closes https://github.com/streamlink/streamlink/issues/3042


gravyboat:
Looks good, thanks.

PR #3049, opened 2020-07-01
plugins.ard_mediathek: fix / update

Billy2011:


src/streamlink/plugins/ard_mediathek.py
```diff
@@ -100,24 +105,27 @@ def _get_streams(self):
...
                     parser_name = "SMIL"
-                elif stream_.endswith(".m3u8"):
+                elif ".m3u8" in stream_:
                     parser = self._get_hls_streams
                     parser_name = "HLS"
                 elif stream_.startswith("http"):
                     parser = self._get_http_streams
                     parser_name = "HTTP"
+                else:
+                    raise PluginError("Unexpected stream type: '{0}'".format(stream_))
```
back-to, line 122: might result in an invalid raise

src/streamlink/plugins/ard_mediathek.py
```diff
@@ -100,24 +105,27 @@ def _get_streams(self):
...
 
                 try:
                     for s in parser(stream):
                         yield s
                 except IOError as err:
-                    self.logger.error("Failed to extract {0} streams: {1}",
+                    raise PluginError("Failed to extract {0} streams: {1}",
```
back-to, line 128: same here

src/streamlink/plugins/ard_mediathek.py
```diff
@@ -15,8 +18,8 @@
     0: "144p"
 }
 
-_url_re = re.compile(r"http(s)?://(?:(\w+\.)?ardmediathek.de/tv|mediathek.daserste.de/)")
-_media_id_re = re.compile(r"/play/(?:media|config)/(\d+)")
+_url_re = re.compile(r"http(s)?://(?:(\w+\.)?ardmediathek.de/|mediathek.daserste.de/)")
```
back-to, line 21: `mediathek.daserste.de` can be removed

src/streamlink/plugins/ard_mediathek.py
Billy2011: Why an invalid raise?

src/streamlink/plugins/ard_mediathek.py
Billy2011: If there is only a log message, you might not recognize the error.

src/streamlink/plugins/ard_mediathek.py
Billy2011: Is this solution not better?:
`_url_re = re.compile(r"http(s)?://(?:(\w+\.)?ardmediathek.de/((?!daserste).)*/|mediathek.daserste.de/)")`

back-to: I don't think the domain is needed, it just redirects to the new domain.

Billy2011: This Link can't be resolved with this plugin, this is better done with **ard_live** or if we rename it to **daserste** .
https://www.ardmediathek.de/daserste/sendung/babylon-berlin/Y3JpZDovL2Rhc2Vyc3RlLmRlL2JhYnlsb24tYmVybGlu/

Billy2011: Sorry, I've test it with a wrong link.

Billy2011: > mediathek.daserste.de can be removed

It fails with: [http://mediathek.daserste.de/live](http://mediathek.daserste.de/live)

PR #3052, opened 2020-07-02
plugins.ard_live: fix / rewrite

Billy2011:


src/streamlink/plugins/ard_live.py
```diff
@@ -1,56 +1,68 @@
...
+    _player_re = re.compile(r'''data-ctrl-player\s*=\s*"(?P<jsondata>.*?)"''')
     _player_url_schema = validate.Schema(
         validate.transform(_player_re.search),
         validate.any(
             None,
-            validate.all(validate.get("url"), validate.text)
-        )
+            validate.all(validate.get("jsondata"), validate.text)
+        ),
+        validate.transform(lambda v: json.loads(v.replace("'", '"'))),
```
back-to, line 23: https://www.daserste.de/information/reportage-dokumentation/unser-wetter/sendung/wo-unser-wetter-entsteht-die-alpen-folge-1-100.html

`AttributeError: 'NoneType' object has no attribute 'replace'`

might need to check for `videos` in the __url_re_

https://www.daserste.de/information/reportage-dokumentation/unser-wetter/videos/wo-unser-wetter-entsteht-die-alpen-folge-1-video-100.html

src/streamlink/plugins/ard_live.py
```diff
@@ -1,56 +1,68 @@
...
-                        yield "{0}k".format(stream["bitrate"]), HTTPStream(self.session, url)
-                except IOError as err:
-                    self.logger.warning("Error parsing stream: {0}", err)
+        res = self.session.http.get(self.url)
+        data_url = self._player_url_schema.validate(res.text)
+        data_url = urljoin(res.url, data_url)
+        log.debug("Player URL: '{0}'", data_url)
+        res = self.session.http.get(data_url)
+        mediainfo = self._mediainfo_schema.validate(json.loads(res.content))
+        log.debug("Mediainfo: {0}", json.dumps(mediainfo))
```
back-to, line 51:
```suggestion
        mediainfo = self.session.http.json(res, schema=self._mediainfo_schema)
        log.trace("Mediainfo: {0!r}".format(mediainfo))
```

src/streamlink/plugins/ard_live.py
```diff
@@ -1,56 +1,68 @@
...
+                if isinstance(stream_, list):
+                    if not stream_:
+                        continue
+                    stream_ = stream_[0]
+
+                if ".m3u8" in stream_:
+                    for s in HLSStream.parse_variant_playlist(self.session, stream_).items():
+                        yield s
+                else:
+                    raise PluginError("Unexpected stream type: '{0}'".format(stream_))
```
back-to, line 65: you can't **raise** something in a loop which might have valid streams

https://www.daserste.de/unterhaltung/film/filme-im-ersten/videos/mankells-wallander-eifersucht-video-102.html

src/streamlink/plugins/ard_live.py
```diff
@@ -1,56 +1,68 @@
 import re
+import json
```
back-to, line 2: there is also `parse_json`

https://github.com/streamlink/streamlink/blob/1b436da7942e492cfbb40f5088deec4ce5d4a3f9/src/streamlink/utils/__init__.py#L48

src/streamlink/plugins/ard_live.py
Billy2011: > AttributeError: 'NoneType' object has no attribute 'replace'

With me there is a Pluginerror:
`error: Unexpected stream type: 'https://pdvideosdaserste-a.akamaihd.net/int/2020/05/18/0b016d84-eec7-425e-9a1a-74631ccea882/320-1_678758.mp4'`

> might need to check for videos in the _url_re

I just got the test streams looked at and thought that the plugin is only for live streams, as the plugin name also says.

If also the videos are to be considered, then it would be better to rename the plugin in **daserste**?

src/streamlink/plugins/ard_live.py
Billy2011: No, that’s right, because I had assumed that the plugin is only for live streams and there are no **mp4** streams, for example.

back-to: if you use **raise** here, just **one invalid stream** can fail the whole plugin,
if you use a log message the plugin would still work for the **2nd valid stream**

**as an example,** when mp4 was not supported the plugin was broken even if there was a valid m3u8 url

Billy2011: > as an example, when mp4 was not supported the plugin was broken even if there was a valid m3u8 url

It is important to me to eliminate mistakes and not to ignore them.

PR #3054, opened 2020-07-02
ci: fix pycountry issue temporarily with a fixed version

back-to:


gravyboat:
Pretty annoying, but if it works it works.

bastimeyer:
I've spent a bit of time debugging this and it looks like a bug in the latest pycountry version.

Streamlink's `util/l10n` module is calling pycountry's `languages.lookup(language)` method for making unknown language lookups:
https://github.com/streamlink/streamlink/blob/1.4.1/src/streamlink/utils/l10n.py#L75-L76

This has to be done if you don't know which type of code or format you're dealing with. If you know which code it is, you could call `languages.get(alpha_2=language)` for example instead.

So what `languages.lookup(language)` does is checking for matches in each type of code in its `indices` dict. During the iteration of the `indices` dict, if the input is `en` for example, `alpha_3` gets checked first and fails as the input is only 2 characters long, and then it proceeds with `name`. This incorrectly matches and it returns a different result which doesn't have all the attributes needed in the next call, which then leads to an `AttributeError`.
https://github.com/flyingcircusio/pycountry/blob/20.7.3/src/pycountry/db.py#L131-L135

So when looking up `en`, it incorrectly returns
`Language(alpha_3='enc', name='En', scope='I', type='L')`
instead of
`Language(alpha_2='en', alpha_3='eng', name='English', scope='I', type='L')`.
This sounds like a bug to me.

In the mean time, a fix could be patching the `indices` dict and making it check `alpha_2` before `name`, so the right data gets returned first. The language input could also be checked against `^[a-z]{2}$` and calling `languages.get(alpha_2=language)` in that case, but I feel like that's just an ugly hack. The tests will pass with that hacky workaround.

bastimeyer:
I've opened a bug report on the pycountry repo (with a better problem description).

PR #3053, opened 2020-07-02
plugins.beattv: remove plugin

Billy2011:
be-at.tv is redirected to: [https://www.youtube.com/user/BeAtTvChannel](https://www.youtube.com/user/BeAtTvChannel)

PR #3055, opened 2020-07-03
utils.l10n: use DEFAULT_LANGUAGE_CODE if locale lookup fails

Billy2011:
set `language_code` to `DEFAULT_LANGUAGE_CODE` if `_parse_locale_code()` fails.

### DIFF
--- a/src/streamlink/utils/l10n.py
+++ b/src/streamlink/utils/l10n.py
@@ -141,7 +141,7 @@ class Localization(object):
                 language_code = None
             if language_code is None or language_code == "C":
                 # cannot be determined
-                language_code = DEFAULT_LANGUAGE
+                language_code = DEFAULT_LANGUAGE_CODE
 
         try:
             self.language, self.country = self._parse_locale_code(language_code)


bastimeyer:
This won't fix the whole issue. It's only relevant for the default system locale, as it changes the string from "en" to "en_US", and thus works around the issue of the latest pycountry release I've described here and in the linked bug report:
https://github.com/streamlink/streamlink/pull/3054#issuecomment-653523388

Billy2011:
> This won't fix the whole issue.

Thanks for your info, but I just want my PR to be an update and not a fix.

If I had known that there was the Workaround #3054 for yesterday’s test bugs in PR #3053, I would not have had to close my issue on the pycountry repo when the tests went through today without errors.

bastimeyer:
> I just want my PR to be an update and not a fix

Hm, yeah, it would improve it by not having to unnecessarily raise a LookupError and doing the default value handling in the except block down below. Legitimate change.

> I would not have had to close my issue on the pycountry repo

pycountry has removed its py2 support in the recent 20.x releases.
https://github.com/flyingcircusio/pycountry/commit/71a3b4275c8ce20433708f19bb4187bc39bac08d#diff-2eeaed663bd0d25b7e608891384b7298

This means we will have to do it here now as well (finally) once pycountry gets fixed and #3054 reverted. There have been a few PRs already for that, but nobody has actively pushed it so far and the PRs might be outdated by now. Removing py2 from the CI should be pretty straightforward though (#2920).

Billy2011:
This information is not necessarily of interest to you, but to me  and I think it will take months to a usable version:
[Status der Python2 -> Python3 Migration](https://www.opena.tv/openatv-6-5-python3-entwicklung/52654-status-der-python2-python3-migration-post443945.html#post443945)

back-to:
> This means we will have to do it here now as well

not necessarily, we actually are not using pycountry that much.

https://github.com/streamlink/streamlink/blob/da4d885f9d18e56e89031645ab2ad9fb44052064/setup.py#L36-L41

bastimeyer:
Some packages have pycountry set as a dependency (#567), which means unless you want to force every packager to change that, we'll have to fix this problem eventually. And fixing it, either here or waiting for an upstream fix, means that py2 has to be dropped. Since py2 is dead since the beginning of this year, there is nothing wrong with finally pushing the removal here, even if some applications/systems for some reason still rely on py2. If those maintainers have not managed to upgrade their build infrastructure since many years and have only just started working on it after the plug has already been pulled by the python devs, then I don't think that them not being able to upgrade to the latest Streamlink version will be a big concern.

https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/streamlink#n11
https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=streamlink-git#n10
https://packages.debian.org/sid/python3-streamlink
https://gitweb.gentoo.org/repo/gentoo.git/tree/net-misc/streamlink/streamlink-9999.ebuild

back-to (approved):

PR #3057, opened 2020-07-04
utils.l10n: fix pycountry language lookup

bastimeyer:
See #3054 and the comments in #3055

~~This will require py2 to be dropped first (tests will fail - will rebase once done), as pycountry's recent release has removed it.~~ The added lookup workaround can be removed as soon as pycountry fixes the issue itself (see flyingcircusio/pycountry#41).

The reason why pycountry had to be downgraded with a fixed version in #3054 was an upgrade of their database with the addition of the En language, causing generic "en" language lookups to return this instead of English. Since the returned language record doesn't have an `alpha_2` code, trying to access this attribute resulted in a raised AttributeError, causing the tests to fail.

Pycountry is being used by the packages on Arch (community repo and AUR), Debian/Ubuntu/etc and Gentoo and thus needs to be fixed at some point.

### DIFF
--- a/script/install-dependencies.sh
+++ b/script/install-dependencies.sh
@@ -6,5 +6,6 @@ set -ex
 
 python -m pip install --disable-pip-version-check --upgrade pip setuptools
 python -m pip install --upgrade -r dev-requirements.txt
-python -m pip install pycountry==19.8.18
+python -m pip install 'pycountry==19.8.18;python_version<"3.0"'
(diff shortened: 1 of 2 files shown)


back-to:
> This will require py2 to be dropped first

using `pycountry==19.8.18` for python2 and just `pycountry` for python3 might be easier for a quick fix,
instead of dropping python2 rapidly just for `pycountry`

Issue #3062, opened 2020-07-05
rasie AttributeError


adamac2:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description

I'm not sure what caused this error. I'm using an Arch-based distribution, and one day, I would imagine after a `pacman -Syu` whenever I run streamlink  I get some sort of `AttributeError` and I'm not sure what is causing it.

### Expected / Actual behavior

I would Image that after running `streamlink "{enter url here}"` the stream would open in mpv and play, but after it finds the matching plugin I get this error.

### Reproduction steps / Explicit stream URLs to test

1. `pacman -Syu` to use the latest version of streamlink available. (1.4.1-1)
2. Try using any link with `streamlink "{enter url here}"`

Some of the links I tried are:
[Destiny's Twitch Channel](https://www.twitch.tv/destiny)
[Formal's Twitch Channel](https://www.twitch.tv/formal)
[ABC News Live](https://www.youtube.com/watch?v=w_Ma8oQLmSM)
[Casualty's DLive Stream](https://dlive.tv/Casualty)

All give me a similar outout to the error below

### Log output

```
[milky@art ~vv]$ streamlink --loglevel debug "twitch.tv/destiny" 
[cli][debug] OS:         Linux-5.7.6-artix1-1-x86_64-with-glibc2.2.5 
[cli][debug] Python:     3.8.3 
[cli][debug] Streamlink: 1.4.1 
[cli][debug] Requests(2.24.0), Socks(1.7.1), Websocket(0.57.0) 
[cli][info] Found matching plugin twitch for URL twitch.tv/destiny 
[cli][debug] Plugin specific arguments: 
[cli][debug]  --twitch-disable-hosting=True (disable_hosting) 
[cli][debug]  --twitch-disable-ads=True (disable_ads) 
[cli][debug]  --twitch-low-latency=True (low_latency) 
[plugin.twitch][debug] Getting live HLS streams for destiny 
Traceback (most recent call last): 
  File "/usr/bin/streamlink", line 11, in <module> 
    load_entry_point('streamlink==1.4.1', 'console_scripts', 'streamlink')() 
  File "/usr/lib/python3.8/site-packages/streamlink_cli/main.py", line 1024, in main 
    handle_url() 
  File "/usr/lib/python3.8/site-packages/streamlink_cli/main.py", line 585, in handle_url 
    streams = fetch_streams(plugin) 
[...]

PR #3061, opened 2020-07-05
Plugin.nicolive:resolve API format change

tnira:
Currently, “broadcast_id” field of Nicolive api is not needed.
fix #3028

Issue #3029, opened 2020-06-15
[New Plugin] Add cl-live.com


SGRK901:
## Plugin Request

- [x] This is a plugin request and I have read the contribution guidelines and plugin request requirements.

### Description

cl-live.com is a new video streaming service under LDH entertainment company, providing artist live streams, variety shows on demand and live concerts.

### Example stream URLs

1. https://www.cl-live.com/programs/ondemand/uaTj4ds5VuZVsCMva7C7Hv
2. https://www.cl-live.com/programs/cast/3Sb3i75Y1KDDRhriMxoTsW

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

wiresp33d:
Hi all,

I found that cl-live is using AES-128 to encrypt the streaming file. Is there any reference to check?

back-to:
> I found that cl-live is using AES-128 to encrypt the streaming file. Is there any reference to check?

nothing special, you just need the correct cookies and headers

for this website you will need to open an api for some cookies

```json
            "api": {
                "url": "https:\u002F\u002Fapi.cl-live.com",
                "key": "AAAAAAAA",
                "authDomain": "cl-live.com"
            },
```

```py
        self.session.http.get('https://api.cl-live.com/v1/license/ondemand/AAAAAAA',
                              headers={
                                  'X-Api-Key': 'AAAA',
                              })
```

wiresp33d:
Thanks @back-to 

I have retrieved the key and relevant cookie, but I am not sure how to put that cookie when I use HLSStream, plus each segment of stream seems encrypted by aes.key. 

I tried to use aes as key to find the clue. There are few example in plugins and I still cannot verified how to decrypt the stream after downloading xxx.ts. 
```
grep -ir aes * ./src/streamlink/plugins/*
```

Once again, thank you for your help!

back-to:
@wiresp33d 
you don't need to do anything besides the cookies,
just use HLSStream with it.

>  how to put that cookie when I use HLSStream,

you don't need to, it is a session.

wiresp33d:
@back-to
I will take time to look at it! Thank you, friend! You are the best!!

wiresp33d:
Hi all,

Thanks to @back-to, I finally done! However, I haven'y found any entry point to live video.
So, I cannot parse any info from https://www.cl-live.com/programs/cast/3Sb3i75Y1KDDRhriMxoTsW

[...]

PR #3063, opened 2020-07-06
Release 1.5.0

bastimeyer:
Minor release which includes the workaround for the latest pycountry version.

As I've said in https://github.com/streamlink/streamlink/issues/3062#issuecomment-654008488, it's a bit awkward since we said 1.4.0 would be the last release with py2 support.

If there's stuff still missing or if there are other concerns, post them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,89 @@
 # Changelog
 
+## streamlink 1.5.0 (2020-07-07)
+
+A minor release with fixes for `pycountry==20.7.3` ([#3057](https://github.com/streamlink/streamlink/pull/3057)) and a few plugin additions and removals.
+
+And of course the usual plugin fixes and upgrades, which you can see in the git shortlog down below. Thank you to everyone involved!
+
+Support for Python2 has not been dropped yet (contrary to the comment in the last changelog), but will be in the near future.
+
+
+```text
+Alexis Murzeau <amubtdx@gmail.com> (1):
+      docs: update debian install instructions
+
+Billy2011 <kschmidt2007@googlemail.com> (8):
+      plugins.nbcsports: fix embed_url_re (#2980)
+      plugins.olympicchannel: fix/rewrite (#2981)
+      plugins.foxtr: fix playervars_re (#3013)
+      plugins.huya: fix _hls_re (#3007)
+      plugins.ceskatelevize: add new api for some links (#2991)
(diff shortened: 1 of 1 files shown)


back-to:
some of the open plugin pull requests should be included before you get a new release for `1.4.2`

https://github.com/streamlink/streamlink/pull/3049
https://github.com/streamlink/streamlink/pull/3052
https://github.com/streamlink/streamlink/pull/3061

(they still need some small changed, I will take a look later)

---

or if you want to go for `1.5.0` we could also merge some of the feature requests

https://github.com/streamlink/streamlink/pull/2925

https://github.com/streamlink/streamlink/pull/2892

bastimeyer:
> if you want to go for `1.5.0` we could also merge some of the feature requests

I'd say let's keep it at 1.4.2 and only do a patch release...

Billy2011:
> I'd say let's keep it at 1.4.2 and only do a patch release...

As I see it, 1.4.2 is not just a patch release, because with #2927 the plugin nbcnews was added.

I see no reason to merge #2702, #3044 also.

bastimeyer:
> not just a patch release

True. We've also removed some plugins, which would require a minor release instead. The line here is however a bit blurry, because minor releases have only been done so far when a feature was added/removed/upgraded that is part of Streamlink itself, and not its plugins. A regular semver scheme doesn't really fit into a plugin system like Streamlink's and our current release timing/scheduling, which is the reason why I'd like to see this changed at some point. But yes, technically, you're correct, @Billy2011, and I should change it to `1.5.0`.

Regarding #3044, this will probably not be merged due to lack of any live streams on this site (https://github.com/streamlink/streamlink/issues/3029#issuecomment-653743382) and #2702 is stale and in an unknown state and it also has merge conflicts.

gravyboat:
@bastimeyer I'm moving discussion here from #3062 to keep everything in one place. I'm not much of a fan of doing a second release after we said we'd deprecate py2 but I don't see much of a choice in this situation, this bug has really screwed us and it means more work for the package maintainers too which is a huge bummer. I don't think it's realistic to get all the py2 removal changes done and it would add stress for us to try and rush it.

Let's just do this release and maybe add in the changelog that this is only being released without the py2 removal due to a bug that impacts some users? As long as we're clear that this release doesn't deprecate py2 yet I think it's fine.

bastimeyer:
Anything else or is this fine now?
If there are no more concerns, please merge and tag it.

PR #3067, opened 2020-07-07
plugins.vaughnlive: removed

back-to:
they use google reCAPTCHA for the session id

closes https://github.com/streamlink/streamlink/issues/2178


gravyboat:
@back-to Thanks.

fadster:
That's really unfortunate. I still use it by providing the required info manually. There must be a way to do the same programmatically.

PR #3072, opened 2020-07-11
Twitch access_token requests fix

mmkhitaryan:


bastimeyer:
Not sure what you're trying to achieve with that change.

First, API client IDs are not OAuth access tokens, and second, OAuth access tokens - if they were supported in Streamlink - need to be set as HTTP headers.

Regarding the Twitch OAuth token support in Streamlink, this has been removed in #2846, since any kind of authentication with Twitch has become pointless due to the main API endpoint for retrieving the stream URL (where Twitch would see your account as a viewer) not accepting OAuth tokens generated from 3rd party client IDs like Streamlink's. See https://github.com/streamlink/streamlink/issues/2680#issuecomment-557605851

If you want to manually add an OAuth token generated from Twitch's API client ID, the linked thread has all the information you need. Automating this however is not possible and is therefore unsupported.

PR #3076, opened 2020-07-16
Trovo Support

TopoIogist:
This adds support for viewing trovo streams, example:

```
C:\Users\Topologist>streamlink https://trovo.live/Ler_GG
[cli][info] Found matching plugin trovo for URL https://trovo.live/Ler_GG
Available streams: 360p (worst), 480p, 720p, 1080p (best)

C:\Users\Topologist>streamlink https://trovo.live/Ler_GG best
[cli][info] Found matching plugin trovo for URL https://trovo.live/Ler_GG
[cli][info] Opening stream: 1080p (http)
[cli][info] Starting player: "C:\Program Files (x86)\VideoLAN\VLC\vlc.exe" --file-caching=5000
```

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -213,6 +213,7 @@ tga                     - star.plu.cn        Yes   No
 theplatform             player.thepl... [7]_ No    Yes
 tigerdile               tigerdile.com        Yes   --
 tlctr                   tlctv.com.tr         Yes   No
+trovo                   trovo.live           Yes   --
 trt                     trt.net.tr           Yes   No    Some streams may be geo-restricted to Turkey.
 trtspor                 trtspor.com          Yes   No    Some streams are geo-restricted to Turkey.
 turkuvaz                - atv.com.tr         Yes   No
--- a/src/streamlink/plugins/trovo.py
+++ b/src/streamlink/plugins/trovo.py
@@ -1,0 +1,20 @@
+import re
+
+from streamlink.stream import HTTPStream
+from streamlink.plugin import Plugin
+
+class Trovo(Plugin):
+    url_re = re.compile(r"https?://(?:www\.)?trovo\.live/")
+    streams_re = re.compile(r'playUrl:"([^"]+)",desc:"([^"]+)"', re.DOTALL)
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls.url_re.match(url) is not None
+
+    def _get_streams(self):
+        res = self.session.http.get(self.url)
+        for m in self.streams_re.finditer(res.text):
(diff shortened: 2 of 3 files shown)


bastimeyer:
Look fine, thanks for the PR.

Not sure though whether we should merge this yet, because the site is apparently still in beta and providing streams as a progressive HTTPStream seems very backwards, so I expect that they will change that at some point. Also not sure how long this streaming site will last (see Mixer and basically every other Twitch clone that has shut down in the past 5 years). And while talking about clones, it looks like they went all in with cloning Twitch's website, which is quite ridiculous.

gravyboat:
@bastimeyer Considering I just closed https://github.com/streamlink/streamlink/issues/3083 I would say this is a no on the merge. Anything in beta we shouldn't be supporting, same with sites constantly changing their URLs or other such nonsense. The progressive HTTPStream usage is also a joke and makes no sense. There is no way they will keep streaming like that, and if they do they're DOA any way.

Them cloning the Twitch layout doesn't factor in for me at all, it's just a sidebar nav with a carousel and a gallery in the middle. For now I'm going to close this. @TopoIogist Thanks for taking the time to write up a PR for this, please understand that by closing this I'm not shutting down the idea of adding them, but it seems like there will likely be a lot of maintenance with this plugin and we don't want to deal with that. If they make it out of beta, get decent viewership, and most importantly stop using progressive HTTPStreaming we'll be happy to merge in a PR.

nedsed:
@TopoIogist Thank you for this plugin. It basically works and is useful for me. I just have the problem when downloading a stream to a file that when the stream ends Streamlink is not closing the file and is not responding anymore. It appears to freeze without throwing errors. The code seems not too complex but I'm not familiar enough with the Streamlink project to know what's wrong. I would be grateful for a solution to avoid manually terminating the process each time. Thanks again.

TopoIogist:
@nedsed handling the progressive HTTPStream output is up to the program that displays the video, streamlink merely pipes everything to the process. For me, I do not observe a hang when recording via ffmpeg to a file, here are my parms:
`streamlink trovo.live/abc best -O | ffmpeg  -abort_on empty_output -err_detect explode -i pipe:0 -c copy -map 0 -segment_time 01:00:00 -f segment $FILE`

nedsed:
@TopoIogist I see, I had only tried to write the transport stream directly into a file using --output and without using ffmpeg. This had worked for me for Twitch in the past. The way you propose seems to work for Trovo though, in a first test Streamlink exited without hanging. Thank you for your quick help. :)

tr0n64:
i get this error error: No plugin can handle URL: trovo.live/name
your plugin outdate?

TopoIogist:
@tr0n64 the developers decided against adding this plugin to streamlink for the reasons mentioned above. Therefore it is not included in the official builds.

PR #3086, opened 2020-07-21
Add OBS-Streamlink into thirdparty.rst

Afanyiyu:
Add [OBS-Streamlink](https://github.com/dd-center/obs-streamlink) into thirdparty.rst

### DIFF
--- a/docs/thirdparty.rst
+++ b/docs/thirdparty.rst
@@ -72,14 +72,14 @@ Wtwitch
 OBS-Streamlink
 --------------
 
-:Description: OBS source plugin to receive stream using streamlink.
+:Description: OBS source plugin for embedding streams using Streamlink
 :Type: OBS Plugin
 :OS: |Windows|
 :Author: `DD Center <https://github.com/dd-center>`_
 :Website: https://github.com/dd-center/obs-streamlink
-:Info: OBS-Streamlink is an OBS plugin which allows streaming live streams into OBS |br|
-  as an OBS source. This is useful for broadcast workers of multi-platform live |br|
-  broadcasts and live commentators.
+:Info: OBS-Streamlink is a plugin for OBS (Open Broadcaster Software) which allows
+  embedding streams directly as a scene source. This is useful for broadcasters of
+  multi-platform live streams and for live commentators.
 
 .. content list end
 
--- a/docs/thirdparty.rst
+++ b/docs/thirdparty.rst
@@ -69,6 +69,18 @@ Wtwitch
   subscribe to. Open streams easily with settings that save your preferred quality |br|
   and player. A man page comes with the AUR package or can be compiled manually.
 
+OBS-Streamlink
+--------------
+
+:Description: OBS source plugin to receive stream using streamlink.
+:Type: OBS Plugin
+:OS: |Windows|
+:Author: `DD Center <https://github.com/dd-center>`_
+:Website: https://github.com/dd-center/obs-streamlink
+:Info: OBS-Streamlink is an OBS plugin which allows streaming live streams into OBS |br|
+  as an OBS source. This is useful for broadcast workers of multi-platform live |br|
+  broadcasts and live commentators.
+
 .. content list end
 
 


gravyboat:
I will not merge this, especially with the currently minimal comment you have provided. Other maintainers may have different thoughts, but what use is this? This is straight up allowing someone to steal a stream for restreaming, nothing more. Show me some valid, legal use cases and we'll consider it. None of the existing third party applications support this sort of action, and even liveproxy is only for local network access so you can use Kodi or whatever in your personal network.

Afanyiyu:
Okay, let me explain more about this software.

In fact, before creating this PR, I have carefully considered the problem you mentioned - yes, this software may be used for illegal purposes. But this software does meet the needs of some special scenarios:

## Multi-platform Broadcast

Usually, most live broadcasters will only push their live streams to one live broadcast platform, but there are many other situations. One of these situations is "streaming towards China". In China and some other countries, access to international video/live-broadcast websites such as YouTube and Twitch is not allowed, and the network connection speed is very slow. In this case, many broadcast teams with viewers in China need relevant person working there to stream their broadcasts to Chinese video websites. In the past, workers re-streaming the broadcast needed to open the live in a browser and use the "Window Capture" function of OBS to capture the screen for streaming. OBS-Streamlink was created in order to improve this inconvenient broadcasting solution.

## Multiplayer Gaming Broadcast

Another use of this software is for the "unified broadcast of multiplayer games". Imagine that several players are playing in the same room in the same game. The scenery they see and the experience of the game are similar. In this case, it is completely unnecessary to start OBS, turn on screen capture and start streaming on every player's computer. Only one player needs to start screen capture on one computer, and other players only need to open OBS-Streamlink. They can show the audience the same (and the best) game scene with different logo, banner, comment box and other interactive elements configured in OBS.

## Broadcast & Commentary

The third situation is "explanation of the live broadcast." In some scenarios, the live streaming official allows some well-known live broadcasters to broadcast live images with real-time commentary, such as some sports games and some electronic product launches. In the past, the solution to achieve this kind of broadcast was still "browser + window capture". Using OBS-Streamlink can greatly simplify the process.

The third situation is "commentary on live broadcast." In some scenarios, live streaming officially allows some well-known live broadcasters to broadcast live images with real-time commentary, such as some sports games and some electronic product launches. In the past, the solution to achieve this kind of broadcast was still "browser + window capture". Using OBS-Streamlink can greatly simplify the process.

The third situation is "commentary on live broadcast." In some scenarios, broadcast companies allow some well-known live broadcasters to re-stream the live with real-time commentary, such as sports games and "launches of electronic products". In the past, the solution to achieve this kind of broadcast was still "browser + window capture". OBS-Streamlink can greatly simplify the process.

---

## To sum up

This software is not very useful under normal circumstances, but it has a very important role in some situations that require "re-streaming", and these situations do require the support of Streamlink. Therefore, I created this PR and added OBS-Streamlink as a "3rd-party application" to this list.

As for "stream stealing", I don't think this software will increase that situation. If someone really wants to steal a stream, he only needs to Google one line of ffmpeg command instead of using OBS-Streamlink's huge set of solution. Even if the software was never born, those people still have many ways to do that. "Stream stealing" should be prevented from the aspect of platform supervision, not from the technical level.

Therefore, I don't think "stream stealing" is the reason to prevent OBS-Streamlink from becoming a 3rd-party application of Streamlink. But such a situation does exist. Maybe we can add additional prompts in the `Description` or `README.md` to avoid that.

beardypig:
Some good points there. It sounds ridiculous, but sometimes you forget how some shoddy workaround becomes the standard that no one tries to improve...

gravyboat:
Thanks for the clarification. These are all valid use cases to me and your reasoning around them is sound. @bastimeyer any additional input? If not I think we can merge this.

docs/thirdparty.rst
```diff
@@ -69,6 +69,18 @@ Wtwitch
   subscribe to. Open streams easily with settings that save your preferred quality |br|
   and player. A man page comes with the AUR package or can be compiled manually.
 
+OBS-Streamlink
+--------------
+
+:Description: OBS source plugin to receive stream using streamlink.
```
bastimeyer, line 75:
```suggestion
:Description: OBS source plugin for embedding streams using Streamlink
```

docs/thirdparty.rst
```diff
@@ -69,6 +69,18 @@ Wtwitch
...
+:Type: OBS Plugin
+:OS: |Windows|
+:Author: `DD Center <https://github.com/dd-center>`_
+:Website: https://github.com/dd-center/obs-streamlink
+:Info: OBS-Streamlink is an OBS plugin which allows streaming live streams into OBS |br|
+  as an OBS source. This is useful for broadcast workers of multi-platform live |br|
+  broadcasts and live commentators.
```
bastimeyer, line 82:
```suggestion
:Info: OBS-Streamlink is a plugin for OBS (Open Broadcaster Software) which allows
  embedding streams directly as a scene source. This is useful for broadcasters of
  multi-platform live streams and for live commentators.
```

gravyboat:
Merging since the comments are fixed up.

Issue #2323, opened 2019-02-19
drdk plugin is not working anymore


nedkellddv:
## Plugin Issue

- [drdk.py ] This is a plugin issue and I have read the contribution guidelines.

### Description

When I try to use streamlink for VOD, I get an error message, as if the webpage source has changed

### Reproduction steps / Explicit stream URLs to test

1. ...copy any VOD url like https://www.dr.dk/tv/se/xxxxxx 
2. ...in a terminal use streamlink with the url
3. ... error message

The output I get in the terminal is
```
error: Unable to validate JSON: Unable to validate key 'Data': Unable to validate key 'Assets': Unable to validate key 'Links': Unable to validate key 'Uri': Type of None should be 'str' but is 'NoneType'

```

back-to:
> Unable to validate key 'Data':

your version is outdated

`streamlink --version`

https://streamlink.github.io/latest/install.html

nedkellddv:
I am sorry I do not believe I have an outdated version `streamlink --version` say 0.14.2

beardypig:
Latest version is currently `1.0.0` (https://github.com/streamlink/streamlink/releases)

nedkellddv:
You're right now installed version 1.0.0 but still get this error

 `error: Unable to validate JSON: Unable to validate key 'Data': Unable to validate key 'Assets': Unable to validate key 'Links': Unable to validate key 'Uri': Type of None should be 'basestring' but is 'NoneType'`

bastimeyer:
> now installed version 1.0.0

what is the output of
```
streamlink --version
```

nedkellddv:
`streamlink 1.0.0`

back-to:
@nedkellddv 

it is just a misunderstanding,
because your first post doesn't have enough details

you are talking about a VOD not a Livestream

**Live**

```
$ streamlink -l debug https://www.dr.dk/tv/live/dr1 best
[cli][debug] OS:         Linux
[cli][debug] Python:     3.6.7
[cli][debug] Streamlink: 1.0.0+21.g34abc43
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.54.0)
[cli][info] Found matching plugin drdk for URL https://www.dr.dk/tv/live/dr1
[cli][info] Available streams: 364k (worst), 400k, 270p, 928k, 360p, 1020k, 432p, 1228k, 1350k, 576p, 1827k, 2010k, 720p, 3128k, 3440k (best)
[...]

PR #3090, opened 2020-07-27
plugins.drdk: fix/rewrite

mkbloke:
I was initially looking to fix and close issue #2323, but since that was opened, the live channels have stopped working too.  This PR fixes live channels again, but not the VoD access, so it does not close that issue.  I will hopefully return to this at some point and get the VoD access working as well.


src/streamlink/plugins/drdk.py
```diff
@@ -1,135 +1,75 @@
...
-    def _get_vod_streams(self, program):
-        res = self.session.http.get(VOD_API_URL.format(program))
-        video = self.session.http.json(res, schema=_video_schema)
+        for playlist_name, playlist_url in playlists.items():
+            if selected_playlist == playlist_name:
+                log.debug("{0}={1}".format(playlist_name, playlist_url))
+                return HLSStream.parse_variant_playlist(
+                    self.session,
+                    playlist_url,
+                )
```
bastimeyer, line 63: Please don't add a plugin parameter for that. Either just use the subtitles playlist, as I don't think there will be a difference content-wise, or combine the streams of both playlists and give the streams of one playlist different names, if necessary. There also seem to be playlists with an `Eu` suffix. Is there a difference other than maybe server location? Combining those sounds like something that could be done. I think it'd also be good to make those additional keys of the API reponse optional in the scheme and only use them when available.

Alternatively, there also seem to be DASH streams available, but I don't think it's necessary to include them.

I can't access any of the streams though due to 403 errors (geo-blocked).

mkbloke: OK on the plugin parameter.  Regarding the ```Eu``` playlists, I'm not sure what the difference is - the content looked to be the same.  I decided not to use them in the code I've submitted simply because they looked somewhat temporary, with hostnames of ```drtesthls``` and ```drtesttexthls```.

Funnily enough, I did initially combine them before in the way I think you're suggesting, but I noticed some things weren't appearing in the final list.  A second look shows these are only the ```_alt``` URLs that are being trimmed though.

I went through my backups to find what I had before.  Using:
```
    _live_data_schema = validate.Schema(
        {'item': {'customFields': {
            validate.optional('hlsURL'): validate.url(),
            validate.optional('hlsWithSubtitlesURL'): validate.url(),
            validate.optional('hlsURLEu'): validate.url(),
            validate.optional('hlsWithSubtitlesURLEu'): validate.url(),
        }}},
        validate.get('item'),
        validate.get('customFields'),
    )

    def _get_live(self, path):
        res = self.session.http.get(self.api_url.format(path))
        play_lists = self.session.http.json(res, schema=self._live_data_schema)

        streams = {}
        for name, url in play_lists.items():
            streams.update(HLSStream.parse_variant_playlist(
                self.session,
                url,
                name_prefix="{0}_".format(name),
            ))

        return streams
```

To get:
```
Available streams: hlswithsubtitlesurleu_720p, hlsurl_360p, hlswithsubtitlesurl_576p, hlswithsubtitlesurl_360p, hlsurl_576p, hlsurleu_432p, hlsurleu_360p, hlsurl_432p, hlswithsubtitlesurleu_360p, hlsurl_720p, hlsurl_270p, hlsurleu_576p, hlswithsubtitlesurl_432p, hlswithsubtitlesurl_720p, hlsurleu_720p, hlswithsubtitlesurl_270p
```

If that looks OK, I'll push it.  Thanks.

mkbloke: I think I prefer this, as it brings back the ```worst```, ```best``` and various ```_alt``` options in the available streams list:
```
    def _get_live(self, path):
        res = self.session.http.get(self.live_api_url.format(path))
        play_lists = self.session.http.json(res, schema=self._live_data_schema)

        streams = {}
        for name, url in play_lists.items():
            name_prefix = ''
            if name == 'hlsWithSubtitlesURL':
                name_prefix="{0}_".format('subtitled')

            streams.update(HLSStream.parse_variant_playlist(
                self.session,
                url,
                name_prefix=name_prefix,
            ))

        return streams
```
```
Available streams: subtitled_432p, subtitled_720p, subtitled_432p_alt, subtitled_270p_alt, subtitled_576p, subtitled_360p_alt, subtitled_360p, subtitled_720p_alt, subtitled_270p, subtitled_576p_alt, 270p_alt (worst), 270p, 360p_alt, 360p, 432p_alt, 432p, 576p_alt, 576p, 720p_alt, 720p (best)
```

src/streamlink/plugins/drdk.py
```diff
@@ -1,135 +1,64 @@
...
-        /[TtVv]+/
-        (?:
-            live/(?P<channel>[^/]+)
-        )?
-        (?:
-            se/(.+/)?(?P<program>[^/&?]+)
-        )?
+class DRDK(Plugin):
+    live_api_url = (
+        'https://www.dr-massive.com/api/page?device=web_browser'
```
back-to, line 13: instead of using such a long URL use params

https://github.com/streamlink/streamlink/blob/e67c68c6457246d6072d9e5cd2353ea1d21c3c11/src/streamlink/plugins/mediaklikk.py#L30-L35

mkbloke: Thanks, that's much nicer.  It also makes it easier to test stripping out unneeded parameters, which I have also now done, reducing the number to 2 now.  I've gone with ```dict()``` rather than ```{}``` because it's quicker to format.

mkbloke:
Well, that was a bad idea...

PR #3104, opened 2020-08-09
plugins.huomao: proposal to remove

### LINKED ISSUE #2341: Huomao plugin do not work
## Plugin Issue

### Description

I tried on two computers but the huomao plugin seems not work,it looks like same as issue#2132,maybe the website updated again.

1. https://www.huomao.com/9418

### Log output

```
[cli][debug] OS:         Linux-4.20.7-1.el7.elrepo.x86_64-x86_64-with-centos-7.6.1810-Core
[cli][debug] Python:     2.7.5
[cli][debug] Streamlink: 1.0.0
[cli][debug] Requests(2.21.0), Socks(1.6.7), Websocket(0.55.0)
[cli][info] Found matching plugin huomao for URL https://www.huomao.com/9418
[plugin.huomao][error] Failed to extract stream_info.
error: No playable streams found on this URL: https://www.huomao.com/9418

```
And here is my streamlink version, I think it's the latest version.
```
[root@ live]# pip show streamlink
Name: streamlink
Version: 1.0.0

```

mkbloke:
I was looking to fix the huomao plugin and close issue #2341, but I've come to the conclusion that it's not really going to be ~~possible~~ reasonably maintainable to generate the necessary token by reimplementing the logic in Python.

The token is used in a POST to ```huomao.com/swf/live_data```, which returns the streams available, complete with their own access tokens.

The token is generated via md5:
```
var _0x58d6x3a = md5(_0x58d6x38[__Ox7f0a3[13]] + _0x58d6x39 + _0x58d6x32 + _0x58d6x2b());
```
which can be translated to:
```
token = md5(video_id + from_string + time_secs + _0x58d6x2b())
```
It's that last function which is the problem.  All sorts of things happen involving loops and arrays of numbers embedded in ```sea.js```, things that no ordinary human being should have to witness...

So, without the ability to use something like Js2Py (and I know how you currently feel about that), it seems removing the huomao plugin is the best way forward for now.

closes #2341


bastimeyer:
> _0x58d6x2b()

That function seems to always return `6FE26D855E1AEAE090E243EB1AF73685`, and I don't see anything that would indicate that there's a side-effect input parameter somewhere other than some is-mobile check before that call, which might not even be releated (didn't check).

But since the entire script is totally obfuscated, has a large hex-encoded string dictionary for all string usages among the entire script and has large arrays of magic numbers, I can't tell what's exactly going on. It calls its outer interface `aes` though with `init` and `refresh` methods.

If it's indeed static or if one figures out which input parameters are used other than the ones you've mentioned, then a plugin fix could be simple, but why would they go through the efforts of obfuscating everything then?

> without the ability to use something like Js2Py

JS parsers would be okay for Streamlink, if one was actively maintained and properly packaged, because accessing an AST is always better than using weird regexes for extracting content. However, scripts that can't be analyzed statically for the required content and need to be run, like this one for example, will never work here, as JS runtimes are completely out of the question. It's simply not possible to fully implement all DOM APIs which a website could access and even if, it would be impossible to run scripts securely.

mkbloke:
You're right, it does seem to return the same value each time.  I was pretty sure I'd actually checked for that before, but perhaps I got confused...  It seems like it's rather pointless in that case.

I'll have another look at it.

Who knows why they have gone to all the effort to obfuscate it?!

Understood regarding the JS stuff.

mkbloke:
Thanks to @bastimeyer for checking my homework, I now have a plugin that works for huomao, as long as you don't mind that it contains a var called ```magic_val```!  I'll test a bit more, then create a PR.  Cheers.

mkbloke:
I'm going to close this for now, as I think I'll have a working plugin soon.

PR #3108, opened 2020-08-13
fix quantum tv streaming

derFogel:
quantum-tv had changed how their login worked:
- client token is hidden behind a obscure url
- they use zapi/v3
- session hello must only consist of token, uuid, app_version

### DIFF
--- a/src/streamlink/plugins/zattoo.py
+++ b/src/streamlink/plugins/zattoo.py
@@ -216,12 +216,9 @@ class Zattoo(Plugin):
             raise e
 
         data = self.session.http.json(res)
-
         self._authed = data['success']
         log.debug('New Session Data')
         self.save_cookies(default_expires=self.TIME_SESSION)
-
-
         self._session_attributes.set('power_guide_hash',
                                      data['session']['power_guide_hash'],
                                      expires=self.TIME_SESSION)
--- a/src/streamlink/plugins/zattoo.py
+++ b/src/streamlink/plugins/zattoo.py
@@ -1,6 +1,7 @@
 import logging
 import re
 import uuid
+import json
 
 from requests.cookies import cookiejar_from_dict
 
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/zattoo.py
```diff
@@ -175,21 +184,31 @@ def _hello(self):
...
         }
-        res = self.session.http.post(hello_url, headers=self.headers, data=params)
+
+        if self.base_url == 'https://www.quantum-tv.com':
+            params['app_version'] = '3.2028.3'
+        else:
+            params['lang'] = 'en'
+            params['format'] = 'json'
+
+        res = self.session.http.post(hello_url, headers=self.headers, data=params, cookies=self.session.http.cookies)
```
back-to, line 195:
```
cookies=self.session.http.cookies
```

are you sure this is needed, since this is already a session.

src/streamlink/plugins/zattoo.py
```diff
@@ -175,21 +184,31 @@ def _hello(self):
...
-        login_url = self.API_LOGIN.format(self.base_url)
         params = {
             'login': email,
             'password': password,
             'remember': 'true'
         }
 
+        if self.base_url == 'https://quantum-tv.com':
+            login_url = self.API_LOGIN_V3.format(self.base_url)
+            params['remember'] = 'True'
```
back-to, line 208: it is already there in line 203, is it with a upper case now instead of a lower case?

src/streamlink/plugins/zattoo.py
```diff
@@ -157,13 +160,19 @@ def _hello(self):
...
+            app_token_url = 'https://www.quantum-tv.com/token-4d0d61d4ce0bf8d9982171f349d19f34.json'
         else:
             app_token_url = self.base_url
         res = self.session.http.get(app_token_url)
         match = self._app_token_re.search(res.text)
 
-        app_token = match.group(1)
-        hello_url = self.API_HELLO.format(self.base_url)
+        if self.base_url == 'https://www.quantum-tv.com':
+            app_token = json.loads(res.text)["session_token"]
```
back-to, line 171: you don't need **json.loads**

use **self.session.http.json** or **parse_json**

https://github.com/streamlink/streamlink/blob/6ec0ab028baaae23cb2eb803f05ed26a6b3934c2/src/streamlink/plugins/zattoo.py#L306
https://github.com/streamlink/streamlink/blob/6ec0ab028baaae23cb2eb803f05ed26a6b3934c2/src/streamlink/plugin/api/http_session.py#L100-L106

https://github.com/streamlink/streamlink/blob/6ec0ab028baaae23cb2eb803f05ed26a6b3934c2/src/streamlink/utils/__init__.py#L48-L67

src/streamlink/plugins/zattoo.py
derFogel: they were in some cases extremely picky about the things they accept, so I used the same values as the webapp
but it works also with a lower case True

src/streamlink/plugins/zattoo.py
derFogel: no it isn't needed, was a leftover from testing

Issue #3107, opened 2020-08-11
plugin turkuvaz broken


worldmediaman:
## Error Report

<! - Replace [] with [x] to activate the check box ->
- [x] This is a bug report and I have read the contribution guidelines.

### Description

Unfortunately, live streams cannot be started via Turkuaz
ATV is affected

### Expected / actual behavior

I usually see the live stream, it doesn't happen here anymore

### Replication Steps / Explicit Stream URLs to Test

1.www.atv.com.tr/webtv/canli-yayin

### Log output

/ usr / sbin / streamlinksrv: 336: PythonDeprecatedWarning: Python 2.7 has reached the end of its life. A future version of streamlink will drop support for Python 2.7. Please upgrade your Python to at least 3.5.
  streamlink = Streamlink2 ()
Failed to load plugin ustvnow:
  File "/usr/lib/python2.7/site-packages/streamlink/plugins/ustvnow.py", line 12, in <module>
    from Crypto.Util.Padding import pad, unpad
ImportError: No module named Padding

[streamlinksrv] [info] Processing URL: http://www.atv.com.tr/webtv/canli-yayin
[streamlinksrv] [debug] url args: []
[streamlinksrv] [info] Found matching plugin turkuvaz for URL http://www.atv.com.tr/webtv/canli-yayin
[plugin.turkuvaz] [debug] Found HLS URL: https://trkvz-live.ercdn.net/atvhd/atvhd.m3u8?st=j5n-GHNVZ6dEjMLMN9cuJQ&e=1597219664
[utils.l10n] [debug] Language code: en_GB
[streamlinksrv] [error] Unable to open URL: https://trkvz-live.ercdn.net/atvhd/atvhd.m3u8?st=j5n-GHNVZ6dEjMLMN9cuJQ&e=1597219664 (401 Client Error: Unauthorized for url: https: // trkvz -live.ercdn.net/atvhd/atvhd.m3u8?st=j5n-GHNVZ6dEjMLMN9cuJQ&e=1597219664)
[streamlinksrv] [info] Processing URL: http://www.atv.com.tr/webtv/canli-yayin
[streamlinksrv] [debug] url args: []
[streamlinksrv] [info] Found matching plugin turkuvaz for URL http://www.atv.com.tr/webtv/canli-yayin
[...]

PR #3114, opened 2020-08-17
docs: update turkuvaz plugin matrix entry

mkbloke:
In relation to: #3107


bastimeyer:
Thanks

bastimeyer:
Btw, @mkbloke, you may want to fix your git config's `user.name` and `user.email` values, because both commits have an unknown author, which messes up squash merge commits like this when done via the Github interface (in case you care about that, otherwise, just ignore it).
https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address

mkbloke:
Thanks, @bastimeyer, I did not know about that.  Hopefully it is now fixed for future commits.

PR #3115, opened 2020-08-18
plugins.drdk: fix/rewrite

mkbloke:
New PR to fix previous screw up... (previously #3090)

I was initially looking to fix and close issue #2323, but since that was opened, the live channels have stopped working too. This PR fixes live channels again, but not the VoD access, so it does not close that issue. I will hopefully return to this at some point and get the VoD access working as well.

### DIFF
--- a/tests/plugins/test_drdk.py
+++ b/tests/plugins/test_drdk.py
@@ -15,6 +15,8 @@ class TestPluginDRDK(unittest.TestCase):
 
     def test_can_handle_url_negative(self):
         should_not_match = [
+            'https://www.dr.dk/tv/live/dr1',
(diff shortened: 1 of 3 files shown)


tests/plugins/test_drdk.py
```diff
@@ -6,9 +6,9 @@
 class TestPluginDRDK(unittest.TestCase):
     def test_can_handle_url(self):
         should_match = [
-            'https://www.dr.dk/tv/live/dr1',
```
back-to, line 9: put the old urls into `test_can_handle_url_negative`

also remove the **yes** for vods in `docs/plugin_matrix.rst`

PR #3129, opened 2020-08-26
Update radiko.py

kokarare1212:
Changed to be able to watch nationwide radio anywhere


back-to:
this is unnecessary, using the whole `_fullKey` in the plugin file is a bit weird.

PR #3132, opened 2020-08-28
cli.main: improvements

Billy2011:
related to #785

- improves output_stream()
- improves alt_streams selection

w/o the changes it's not possible to stream an alternate stream on an open error or e.g. if the main stream is a http-stream and the "_alt" stream is a hls-stream.


back-to:
is there a plugin where this can be easily tested?

Billy2011:
You can test it with livestream.com

PR #3140, opened 2020-09-01
plugins.europaplus: remove plugin

back-to:
Stream can be used with the VK plugin.

https://vk.com/europaplus

https://vk.com/video-19043_456239355


gravyboat:
Thanks @back-to.

PR #3155, opened 2020-09-07
plugins.svtplay: rewrite/fix

### LINKED ISSUE #3133: Svtplay's Plugin is currently broken
## Plugin Issue

### Description

The svtplay plugin seems to be broken. Every url I have tested yield the same error. Other Swedish geo-restricted sites work fine.

I'm on the latest streamlink 1.5.0 on Windows 10.

1. Go to Svtplay.se with a Swedish Ip, Vpn or Proxy.
2. Pick some show and it's URL
3. Try to stream it with streamlink

### Log output

```
streamlink https://www.svtplay.se/video/27096341/sportnytt/sportnytt-28-aug-22-00-3
[cli][info] Found matching plugin svtplay for URL https://www.svtplay.se/video/27096341/sportnytt/sportnytt-28-aug-22-00-3
error: Unable to parse JSON: Expecting value: line 1 column 1 (char 0) ('<!DOCTYPE html>\n\n<head>\n    <me ...)
```

Thanks a lot for your support!

mkbloke:
Tested under both Py2 and Py3 against supported sites within the plugin.

There's no necessity for an account, but streams are geo-IP restricted to Sweden.

closes #3133

### DIFF
--- a/src/streamlink/plugins/svtplay.py
+++ b/src/streamlink/plugins/svtplay.py
@@ -1,92 +1,146 @@
+import logging
 import re
 
-from streamlink.plugin import Plugin
-from streamlink.plugin.api import StreamMapper, validate
-from streamlink.stream import HLSStream, HDSStream
+from streamlink.compat import urljoin
+from streamlink.exceptions import PluginError
+from streamlink.plugin import Plugin, PluginArguments, PluginArgument
+from streamlink.plugin.api import validate
+from streamlink.stream import DASHStream, HTTPStream
+from streamlink.stream.ffmpegmux import MuxedStream
 
-API_URL = "http://www.svt.se/videoplayer-api/video/{0}"
-
-_url_re = re.compile(r"""
-    http(s)?://
-    (www\.)?
-    (?:
-        svtplay |
-        svtflow |
-        oppetarkiv
-    )
-    .se
-""", re.VERBOSE)
-
-# Regex to match video ID
-_id_re = re.compile(r"""data-video-id=['"](?P<id>[^'"]+)['"]""")
-_old_id_re = re.compile(r"/(?:video|klipp)/(?P<id>[0-9]+)/")
-
-# New video schema used with API call
-_video_schema = validate.Schema(
-    {
-        "videoReferences": validate.all(
-            [{
-                "url": validate.text,
-                "format": validate.text
-            }],
-        ),
-    },
-    validate.get("videoReferences")
-)
-
-# Old video schema
-_old_video_schema = validate.Schema(
-    {
-        "video": {
-            "videoReferences": validate.all(
-                [{
-                    "url": validate.text,
-                    "playerType": validate.text
-                }],
-            ),
-        }
-    },
-    validate.get("video"),
-    validate.get("videoReferences")
-)
+log = logging.getLogger(__name__)
 
 
 class SVTPlay(Plugin):
-    @classmethod
-    def can_handle_url(self, url):
-        return _url_re.match(url)
+    api_url = 'https://api.svt.se/videoplayer-api/video/{0}'
 
-    def _create_streams(self, stream_type, parser, video):
-        try:
-            streams = parser(self.session, video["url"])
-            return streams.items()
-        except IOError as err:
-            self.logger.error("Failed to extract {0} streams: {1}",
-                              stream_type, err)
+    author = None
+    category = None
+    title = None
+
+    url_re = re.compile(r'''
+        https?://(?:www\.)?(?:svtplay|oppetarkiv).se(?P<path>.*)
+    ''', re.VERBOSE)
+
+    latest_episode_url_re = re.compile(r'''
+        class="play_titlepage__latest-video"\s+href="(?P<url>[^"]+)"
+    ''', re.VERBOSE)
+
+    live_id_re = re.compile(r'.*/(?P<live_id>[^?]+)')
+
+    vod_id_re = re.compile(r'''
+        (?:DATA_LAKE\s+=\s+{"content":{"id":|data-video-id=)
+        "(?P<vod_id>[^"]+)"
+    ''', re.VERBOSE)
+
+    _video_schema = validate.Schema({
+        validate.optional('programTitle'): validate.text,
+        validate.optional('episodeTitle'): validate.text,
+        'videoReferences': [{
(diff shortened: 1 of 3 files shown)


src/streamlink/plugins/svtplay.py
```diff
@@ -1,92 +1,146 @@
...
+        }],
+    })
+
+    arguments = PluginArguments(
+        PluginArgument(
+            'mux-subtitles',
+            action='store_true',
+            help="Automatically mux available subtitles in to the output stream.",
+        ),
+    )
```
bastimeyer, line 55: Is it really necessary to add a plugin parameter for having muxed output streams with subtitles? We don't have this anywhere else, but if there is a need for having muxed subtitles, then maybe we could add a global parameter that could then be checked by individual plugins which have access to external subtitles. Having parameters for each plugin doesn't make much sense.

src/streamlink/plugins/svtplay.py
```diff
@@ -1,92 +1,146 @@
...
+        path = self.url_re.match(self.url).group('path')
+        log.debug("Path={0}".format(path))
+
+        if path.startswith('/kanaler/'):
+            match = self.live_id_re.search(path)
+            if match is not None:
+                live_id = "ch-{0}".format(match.group('live_id'))
+                log.debug("Live ID={0}".format(live_id))
+            else:
+                raise PluginError("Failed to get live ID")
```
bastimeyer, line 83: Just return `None` or an empty dict if there are no streams on the input URL. This shouldn't be a PluginError.

src/streamlink/plugins/svtplay.py
```diff
@@ -1,92 +1,146 @@
...
-            mapper = StreamMapper(cmp=lambda format, video: video["format"] == format)
-            mapper.map("hls", self._create_streams, "HLS", HLSStream.parse_variant_playlist)
-            mapper.map("hds", self._create_streams, "HDS", HDSStream.parse_manifest)
-        else:
-            res = self.session.http.get(self.url, params=dict(output="json"))
-            videos = self.session.http.json(res, schema=_old_video_schema)
+        path = self.url_re.match(self.url).group('path')
+        log.debug("Path={0}".format(path))
+
+        if path.startswith('/kanaler/'):
```
bastimeyer, line 77: I think it may be better to add the different paths to the regex and check for the capture groups. This way a reader of the plugin code can check for the supported paths much more easily.

Also, please move this into class methods instead of having a large if-else block.

src/streamlink/plugins/svtplay.py
```diff
@@ -1,92 +1,146 @@
...
+                    urljoin(self.url, match.group('url')),
+                )
+
+            match = self.vod_id_re.search(res.text)
+
+            if match is not None:
+                vod_id = match.group('vod_id')
+                log.debug("VOD ID={0}".format(vod_id))
+            else:
+                raise PluginError("Failed to get VOD ID")
```
bastimeyer, line 114: Same stuff with the PluginError.

src/streamlink/plugins/svtplay.py
```diff
@@ -1,92 +1,146 @@
...
+            res = self.session.http.get(self.api_url.format(vod_id))
+            api_data = self.session.http.json(res, schema=self._video_schema)
+
+            if 'programTitle' in api_data:
+                self.author = api_data['programTitle']
+
+            self.category = 'VOD'
+
+            if 'episodeTitle' in api_data:
+                self.title = api_data['episodeTitle']
```
bastimeyer, line 125: The author, category and title logic could be made more DRY. Move this to a separate method, eg. `_set_metadata`.

bastimeyer (commented):
Thanks for the plugin rewrite. I can't test/validate it though, since I don't have access to a Swedish IP.

src/streamlink/plugins/svtplay.py
mkbloke: Thanks for the review.

I only did it that way because I thought it was the 'right way to do it'!  It is the way it seems to have been done in the funimation, pluzz, rtve and vimeo plugins.

I added the subtitle support because I though it would be 'nice to have', but I have no great attachment to it either, so am happy to remove the support depending on how you want to proceed.

Cheers.

bastimeyer: > it seems to have been done in the funimation, pluzz, rtve and vimeo plugins

Apologies, this is indeed true.

We should change this though. It doesn't make sense for each plugin to have its own subtitle-muxing parameter. I hope you and anyone else agree with this.

I don't want to pressure you to change this now though. If you want to keep it, I guess that's fine considering we already have plugins with the same pattern. Deprecating or removing parameters however is always a bit tough. I don't want to decide this on my own.

mkbloke: > We should change this though. It doesn't make sense for each plugin to have its own subtitle-muxing parameter. I hope you and anyone else will agree with this.

Seems perfectly reasonable to me.

> I don't want to pressure you to change this now though. If you want to keep it, I guess that's fine considering we already have plugins with the same pattern.

I'm happy either way.

> Deprecating or removing parameters however is always a bit tough. I don't want to decide this on my own.

It seems like an issue for you and the other main devs to discuss.  I guess if you decide to go the route of adding a global parameter, it will mean suppressing the ```--*-mux-subtitles``` entries in the above mentioned plugins for backwards compatibility and modifying them to use the new global version.

mkbloke: I've been thinking about the implications of a global subtitle parameter.  Going that way also means asking why not do the same thing with other current plugin specific parameters which could be used by the plugin, or not, as necessary if they were global instead.  Likely candidates which are in no way absolutely plugin specific are email/username, password and purge-credentials, for example.

back-to: > whereas a global parameter specified in the config file would apply to all plugins

not an issue, use this instead

https://streamlink.github.io/cli.html#plugin-specific-configuration-file

mkbloke: Good point, I'd forgotten about the plugin specific config files.

mkbloke:
I left the subtitle stuff as is for the time being, but am happy to change it pending future decisions.

Cheers.

bastimeyer:
Thanks!

Issue #3145, opened 2020-09-03
tvplayer.com plugin fails


whittinghamj:
## Bug Report

- [ x] This is a bug report and I have read the contribution guidelines.

### Description

Im using the following streamlink command.

streamlink --hls-live-restart --player-external-http --player-external-http-port '2223' --tvplayer-email 'user@gmail.com' --tvplayer-password 'hidden_for_github' --url 'https://tvplayer.com/watch/channel4' --default-stream '1080p,720p,480p,best'

### Expected / Actual behavior

stream to start

### Reproduction steps / Explicit stream URLs to test

1. ... run the above code using a valid username and password

### Log output

```
[cli][info] streamlink is running as root! Be careful!
[cli][debug] OS:         Linux-4.15.0-112-generic-x86_64-with-Ubuntu-18.04-bionic
[cli][debug] Python:     3.6.9
[cli][debug] Streamlink: 1.5.0
[cli][debug] Requests(2.18.4), Socks(1.6.5), Websocket(0.44.0)
[cli][info] Found matching plugin tvplayer for URL https://tvplayer.com/watch/bbctwo
[cli][debug] Plugin specific arguments:
[cli][debug]  --tvplayer-email=user@gmail.com (email)
[cli][debug]  --tvplayer-password=******** (password)
[plugin.tvplayer][debug] Logging in as user@gmail.com
[plugin.tvplayer][debug] Got stream attributes: {}
[plugin.tvplayer][debug] Missing 'expiry' from stream attributes
[plugin.tvplayer][debug] Missing 'key' from stream attributes
[plugin.tvplayer][debug] Missing 'token' from stream attributes
[plugin.tvplayer][debug] Missing 'uvid' from stream attributes
error: No playable streams found on this URL: https://tvplayer.com/watch/bbctwo
```

### Additional comments, screenshots, etc.

I think they might have changed something?

Here is an example of the missing vars pulled from their website

[...]

PR #3160, opened 2020-09-09
plugins.tvplayer: update regex and tests for /uk/ URLs

mkbloke:
https://github.com/streamlink/streamlink/issues/3145#issuecomment-689650652


gravyboat:
Looks good, thanks.

PR #3161, opened 2020-09-09
plugins.dogus: add startv URL

hymer-up:

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -74,6 +74,7 @@ dogus                   - ntvspor.net        Yes   No
                         - kralmuzik.com.tr
                         - ntv.com.tr
                         - eurostartv.com.tr
+                        - startv.com.tr
 dommune                 dommune.com          Yes   --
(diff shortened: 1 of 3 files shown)


hymer-up:
https://github.com/streamlink/streamlink/commit/dcdb08a354d0c248930e366dc9e2dc18fc8cbf2a

mkbloke:
You should also update: https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugins/.removed for removed plugins.

I don't know for sure in this particular case, but it's possible that you'll be asked to split the plugin update and the plugin removal into separate PRs by the devs.

Hope that's useful.  Cheers.

gravyboat:
Yep please split the removal PR and the plugin update just so we have it correctly for the git history.

hymer-up:
#3163

bastimeyer (changes requested):
Please add regex tests for the new URL

bastimeyer (approved):

PR #3169, opened 2020-09-12
plugins.twitch: refactor worker, parser and tests

bastimeyer:
- Removes the custom parameters and properties from the parser and adds
  the "prefetch" field to the namedtuple Segment.
- Filters out prefetch segments in the worker's process_sequences method
- Replaces the playlist_reloads counter with a playlist_sequence == -1
  check, which is sufficient for checking the beginning of the stream.
- Adds had_content property to worker, which checks for segments which
  are not ads until the first segment was found.
- Refactors the info log messages and fixes the
  "This is not a low latency stream" message when preroll ads exist.
- Cleans up tests
- Adds more tests for certain low-latency and disable-ads conditions.

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -119,38 +119,23 @@ _video_schema = validate.Schema(
     }
 )
 
-Segment = namedtuple("Segment", "uri duration title key discontinuity ad byterange date map")
+Segment = namedtuple("Segment", "uri duration title key discontinuity ad byterange date map prefetch")
 
 LOW_LATENCY_MAX_LIVE_EDGE = 2
 
 
 class TwitchM3U8Parser(M3U8Parser):
-    def __init__(self, base_uri=None, disable_ads=False, low_latency=False, **kwargs):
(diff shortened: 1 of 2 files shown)


bastimeyer:
This PR is a combination of three things
1. a clean up of the TwitchM3U8Parser, which makes it independent of the TwitchHLSStream parameters
2. a clean up of the TwitchHLSStreamWorker and the info log messages
3. and a fix regarding the "This is not a low latency stream" info log message which gets incorrectly emitted when there are preroll ads.

And I've cleaned up the tests and added a few more for certain low-latency and disable-ads conditions.

----

This doesn't fix the timeout "issue" when filtering out mid-stream ads, which I've talked about in the other thread.

gravyboat (approved):
@bastimeyer This looks good, I'm almost tempted to say we should include a link to the pinned issue about commercials in the text for the commercial break segment to try and head off any additional issues/comments. If you don't agree feel free to merge this as is.

bastimeyer:
No, let's not add links to debug/info log messages. These links will become obsolete quickly.

Issue #3165, opened 2020-09-11
Streamlink Crashing after Commercial


ryankashi:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description

Streamlink will either crash after a 15 second commercial break or will operate slowly after the 15 second commercial break at startup. 

### Expected / Actual behavior

After typing in my streamlink command:
streamlink twitch.tv/xqcow best --hls-live-edge 1 --twitch-low-latency

I should jump right into the live edge of the stream. Instead, I will end up watching a commercial placeholder, and then most of the time crash my VLC media player afterwards. I am currently building an application as well that pipes my streamlink output into FFMPEG, and this pipe will break as well.

### Reproduction steps / Explicit stream URLs to test

1. ... Type in the command streamlink twitch.tv/channel_name best --hls-live-edge 1 --twitch-low-latency
2. ... Wait out the new 15 second commercial timeout
3. ... By this point it should have crashed. If it did not then exit VLC media player and try again

### Log output

streamlink twitch.tv/xqcow best --hls-live-edge 1 --twitch-low-latency --loglevel debug
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.6.6
[cli][debug] Streamlink: 1.5.0
[cli][debug] Requests(2.24.0), Socks(1.7.1), Websocket(0.57.0)
[cli][info] Found matching plugin twitch for URL twitch.tv/xqcow
[cli][debug] Plugin specific arguments:
[cli][debug]  --twitch-low-latency=True (low_latency)
[plugin.twitch][debug] Getting live HLS streams for xqcow
[utils.l10n][debug] Language code: en_US
[cli][info] Available streams: audio_only, 160p (worst), 360p, 480p, 720p, 720p60, 1080p60 (best)
[cli][info] Opening stream: 1080p60 (hls)
[plugin.twitch][info] Low latency streaming (HLS live edge: 1)
[stream.hls][debug] Reloading playlist
[plugin.twitch][info] This is not a low latency stream
[stream.hls][debug] First Sequence: 0; Last Sequence: 2
[stream.hls][debug] Start offset: 0; Duration: None; Start Sequence: 2; End Sequence: None
[stream.hls][debug] Adding segment 2 to queue
[cli][debug] Pre-buffering 8192 bytes
[cli][info] Starting player: "C:\Program Files\VideoLAN\VLC\vlc.exe"
[...]

PR #3170, opened 2020-09-12
plugins.twitch: block ads using the same method as uBlock

### LINKED ISSUE #3120: Twitch showing ads/commercials again
## Bug Report

- [X] This is a bug report and I have read the contribution guidelines.

### Description

Using: --player-passthrough hls --twitch-disable-ads
Since early today I noticed that the player started again showing ads interruptions. Not the ads themselves but a screen replacing them.

By the way, I noticed someone else opened a similar issue. This is new. I was using the command line the last few months. Only today I started seeing this screen. I've also updated Streamlink to the last version and the problem persists.

samfundev:
Implements [uBlock's method](https://github.com/gorhill/uBlock/blob/da7ff2b382c0382e044a4fc2523018a3ea388322/assets/resources/scriptlets.js#L1221-L1225) of blocking ads on twitch. Which basically involves putting the `platform=_` parameter in the `access_token` URLs. Doing that removes ad segments from the stream.

I removed the tests related to removing ad segments since that isn't done anymore. I'm not sure how I'd test the new way of disabling ads so I didn't add any tests.

Also closes #3120 as passthrough should now be compatible with disabling ads.

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -663,8 +663,10 @@ class Twitch(Plugin):
                 endpoint = "vods"
                 value = self.video_id
 
+            disable_ads = self.session.get_plugin_option("twitch", "disable-ads")
             sig, token = self.api.access_token(endpoint, value,
-                                               schema=_access_token_schema)
+                                               schema=_access_token_schema,
+                                               platform="_" if disable_ads else None)
         except PluginError as err:
             if "404 Client Error" in str(err):
                 raise NoStreamsError(self.url)


bastimeyer:
We've already had the `platform=_` parameter added once and it worked for less than two days until Twitch made further changes.
#2358
https://github.com/streamlink/streamlink/issues/2357#issuecomment-474044718
Then we removed it again:
#2692

Even if this seems to be working right now, I doubt that it will keep working.

If we decide to re-add this request parameter, then the only change should be the parameter itself (#2358), and the `--twitch-disable-ads` logic should not be removed.

samfundev:
It does seem like the parameter isn't a very consistent fix, sorry I didn't see that. I've changed it so the only changes are adding the parameter if the user asks for ads to be disabled.

laichiaheng:
It really needs to be merged as soon as possible, otherwise the low latency streaming with no ads is unusable.

bastimeyer:
> adding the parameter if the user asks for ads to be disabled

The `twitch-disable-ads` parameter shouldn't be checked here, because its purpose is to change the HLS client logic. Acquiring an access token and setting the `platform=_` request parameter however should also work when the user is querying the resolved HLS stream URL when setting the `--stream-url` or `--player-passthrough=hls` parameters for example, which means the `platform=_` request parameter should be applied in all cases (see 1bb3f86f3f27c07c07d67a3c01884f654646b5e4). Please fix this, otherwise I'll have to re-apply 1bb3f86f3f27c07c07d67a3c01884f654646b5e4, which I don't want to due because you made this suggestion of (re-)adding the request parameter. Thanks.

> otherwise the low latency streaming with no ads is unusable.

Not true. The LL streaming still works fine when there are preroll ads.
See https://github.com/streamlink/streamlink/issues/3120#issuecomment-691104049
I've submitted #3169 yesterday with a fix for that incorrect info log message, but it hasn't been reviewed yet. Just ignore the incorrect info message for now.

laichiaheng:
> > adding the parameter if the user asks for ads to be disabled
> 
> The `twitch-disable-ads` parameter shouldn't be checked here, because its purpose is to change the HLS client logic. Acquiring an access token and setting the `platform=_` request parameter however should also work when the user is querying the resolved HLS stream URL when setting the `--stream-url` or `--player-passthrough=hls` parameters for example, which means the `platform=_` request parameter should be applied in all cases (see [1bb3f86](https://github.com/streamlink/streamlink/commit/1bb3f86f3f27c07c07d67a3c01884f654646b5e4)). Please fix this, otherwise I'll have to re-apply [1bb3f86](https://github.com/streamlink/streamlink/commit/1bb3f86f3f27c07c07d67a3c01884f654646b5e4), which I don't want to do because you made this suggestion of (re-)adding the request parameter. Thanks.
> 
> > otherwise the low latency streaming with no ads is unusable.
> 
> Not true. The LL streaming still works fine when there are preroll ads.
> See [#3120 (comment)](https://github.com/streamlink/streamlink/issues/3120#issuecomment-691104049)
> I've submitted #3169 yesterday with a fix for that incorrect info log message, but it hasn't been reviewed yet. Just ignore the incorrect info message for now.

I saw another tread said that it was a player problem, but MPV didn't have this problem with streamlink, why?

bastimeyer:
This is unrelated to this PR and shouldn't be discussed here.
I've also answered this plenty of times.
https://github.com/streamlink/streamlink/issues/3165#issuecomment-691366440

samfundev:
My apologizes, I was busy yesterday but I'm glad this got fixed.

PR #3172, opened 2020-09-14
plugins.navernow: new plugin for https://now.naver.com/

SeonjaeHyeon:
Add plugin for Naver Now.

Naver Now is 24/7 Radio streaming service operated by Naver.

You can check the site here: [https://now.naver.com/](https://now.naver.com/)

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -121,6 +121,7 @@ mitele                  mitele.es            Yes   No    Streams may be geo-rest
 mjunoon                 mjunoon.tv           Yes   Yes
 mrtmk                   play.mrt.com.mk      Yes   Yes   Streams may be geo-restricted to North Macedonia.
 n13tv                   13tv.co.il           Yes   Yes   Streams may be geo-restricted to Israel.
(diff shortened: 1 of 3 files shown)


bastimeyer:
Apologies for the lack of feedback so far. Unfortunately, I'll have to close this PR, as we don't want to merge plugins for audio-only streams (radio stations, etc.) because it doesn't fit into the scope of Streamlink:

Before starting to work on something which you intend to contribute at some point, it's usually a good idea to ask first, so you don't waste your time if your pull request gets rejected. Closing this feels a bit bad, because the contribution guidelines haven't been updated in a while, and cases like this were not covered correctly.

You and anyone else who wants to use your plugin can still do so by sideloading it:
https://streamlink.github.io/cli.html#sideloading-plugins

SeonjaeHyeon:
@bastimeyer
It's my fault that I didn't give more information about the site. There are two different live channels. One is video channel(includes audio, of course) and the other is audio only channel.
So, the former has video stream as well as audio stream. This site also offers video streams.

Maybe.. I think you have checked audio only channel. Or as you said, Streamlink doesn't allow this at all.

PR #3173, opened 2020-09-14
plugins.twitch: platform=_ in access_token request

bastimeyer:
Since there hasn't been another response in #3170 and I don't want to wait any longer, I'm opening a PR myself with the platform=_ "workaround".

As I've said in the other PR, we've already had added the `platform=_` request parameter to the access token API endpoint when embedded ads were first implemented, but we removed it again when it became obsolete after a few days when Twitch made further changes.

This is basically the same situation and I don't expect this "workaround" to keep working indefinitely. We'll see.

Added first in: #2358
Stopped working in: #2368
Got removed in: #2692

Closes #3170

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -328,7 +328,7 @@ class TwitchAPI(object):
     # Private API calls
 
     def access_token(self, endpoint, asset, **params):
-        return self.call("/api/{0}/{1}/access_token".format(endpoint, asset), private=True, **params)
+        return self.call("/api/{0}/{1}/access_token".format(endpoint, asset), private=True, **dict(platform="_", **params))
 
     def hosted_channel(self, **params):
         return self.call_subdomain("tmi", "/hosts", format="", **params)


beardypig:
If it works, it works. The difference between this and the other PR seems to be that this changes it to always send `platform=_` and the other PR only sends it when the disable ads option is used?

bastimeyer:
> If it works, it works.

Yes, I was asking for a quick review so that this can be merged, because I'm getting spammed on various channels due to the embedded ads. I won't merge my own PRs without some kind of response or acknowledgement.

> difference

See https://github.com/streamlink/streamlink/pull/3170#issuecomment-691649263

It should be this way, because otherwise it would require to set `--twitch-disable-ads` when using `--stream-url` or `--player-passthrough=hls`, and that's not the point of the disable ads parameter. Since I got no response yesterday and today, I've opened a new PR myself with the previous implementation.

So if there are no objections, please merge, so that this issue can be solved. We may have to publish a new release because of this, not sure. My other PR with the rewrite of the custom Twitch HLS stuff is not that important. It fixes the incorrect log message though.

gravyboat:
Looks fine to me, merged. If it only fixes it for a small amount of time oh well. Better that than people constantly creating issues we have to review.

PR #3182, opened 2020-09-17
docs: remove easy_install and clean up install page

bastimeyer:
easy_install removal:
https://github.com/streamlink/streamlink/issues/2896#issuecomment-623272899

Commits can be squashed if you want


docs/install.rst
```diff
@@ -191,6 +189,9 @@ via `setuptools`_ after cloning the `git`_ repository.
 
 Using :command:`easy_install` is no longer recommended.
```
back-to, line 190: should also be removed

bastimeyer: Removed the left over recommendation against using easy_install, so that easy_install isn't mentioned at all anymore on the install page.

Also fixed punctuation and added the `pip --version` note for checking the pip version first.

gravyboat:

PR #3187, opened 2020-09-19
Add FilteredHLSStream

### LINKED ISSUE #3165: Streamlink Crashing after Commercial

### LINKED ISSUE #3178: Twitch plugin read timeout
## Plugin Issue

### Description

after a certain amount of time that the stream is opened the plugin is not able to download a new segment and the stream freezes

I'm on Debian Buster with Streamlink from backports (1.5.0+dfsg-1~bpo10+1)

1. streamlink -p "mpv --no-cache" --twitch-disable-hosting --twitch-disable-ads --twitch-low-latency --hls-live-edge 1 --loglevel debug twitch.tv/pokelawls best (or twitch.tv/lirik)
2. wait some time
3. the stream freezes

### Log output

```
[stream.hls][debug] Download of segment 523 complete
[stream.hls][debug] Adding segment 525 to queue
[stream.hls][debug] Download of segment 524 complete
[stream.hls][debug] Adding segment 526 to queue
[stream.hls][debug] Download of segment 525 complete
[stream.hls][debug] Adding segment 527 to queue
[stream.hls][debug] Download of segment 526 complete
[stream.hls][debug] Adding segment 528 to queue
[stream.hls][debug] Download of segment 527 complete
[stream.hls][debug] Adding segment 529 to queue
[stream.hls][debug] Download of segment 528 complete
[stream.hls][debug] Adding segment 530 to queue
[stream.hls][debug] Download of segment 529 complete
[stream.hls][debug] Download of segment 530 complete
[stream.hls][debug] Adding segment 531 to queue
[stream.hls][debug] Adding segment 532 to queue
[stream.hls][debug] Download of segment 531 complete
[stream.hls][debug] Adding segment 533 to queue
[stream.hls][debug] Download of segment 532 complete
[stream.hls][debug] Adding segment 534 to queue
[stream.hls][debug] Download of segment 533 complete
[stream.hls][debug] Adding segment 535 to queue
[stream.hls][debug] Download of segment 534 complete
[stream.hls][debug] Adding segment 536 to queue
[stream.hls][debug] Download of segment 535 complete
[stream.hls][debug] Adding segment 537 to queue
[stream.hls][debug] Download of segment 536 complete
[stream.hls][debug] Adding segment 538 to queue
[stream.hls][debug] Download of segment 537 complete
[stream.hls][debug] Adding segment 539 to queue
[stream.hls][debug] Download of segment 538 complete
[stream.hls][debug] Adding segment 540 to queue
[stream.hls][debug] Download of segment 539 complete
[stream.hls][debug] Adding segment 541 to queue
[stream.hls][debug] Download of segment 540 complete
[stream.hls][debug] Adding segment 542 to queue
[stream.hls][debug] Adding segment 543 to queue
[stream.hls][debug] Adding segment 544 to queue
[stream.hls][debug] Adding segment 545 to queue
[stream.hls][debug] Adding segment 546 to queue
[stream.hls][debug] Adding segment 547 to queue
[stream.hls][debug] Adding segment 548 to queue
[stream.hls][debug] Adding segment 549 to queue
[stream.hls][debug] Adding segment 550 to queue
[stream.hls][debug] Adding segment 551 to queue
[stream.hls][debug] Adding segment 552 to queue
[stream.hls][debug] Adding segment 553 to queue
[stream.hls][debug] Adding segment 554 to queue
[stream.hls][debug] Adding segment 555 to queue
[stream.hls][debug] Adding segment 556 to queue
[stream.hls][debug] Adding segment 557 to queue
[stream.hls][debug] Adding segment 558 to queue
[stream.hls][debug] Adding segment 559 to queue
[stream.hls][debug] Adding segment 560 to queue
[stream.hls][debug] Adding segment 561 to queue
[stream.hls][debug] Adding segment 562 to queue
[stream.hls][debug] Adding segment 563 to queue
[stream.hls][debug] Adding segment 564 to queue
[stream.hls][debug] Adding segment 565 to queue
[stream.hls][debug] Adding segment 566 to queue
[stream.hls][debug] Adding segment 567 to queue
[stream.hls][debug] Adding segment 568 to queue
[stream.hls][debug] Adding segment 569 to queue
[stream.hls][debug] Adding segment 570 to queue
[stream.hls][debug] Adding segment 571 to queue
[stream.hls][debug] Adding segment 572 to queue
error: Error when reading from stream: Read timeout, exiting
[cli][info] Stream ended
benny@debian:~$ 
```

### Additional comments

I'm not sure if this problems relates to the ads problem that have come up recently. If it is the case i'm sorry and you can delete this issue.

bastimeyer:
Closes #3178
Closes #3165

This adds the general purpose `FilteredHLSStreamWriter` and `FilteredHLSStreamReader` classes which filter out HLS segments and prevent the [`IOError("Read timeout")` from being raised by the `RingBuffer`](https://github.com/streamlink/streamlink/blob/3a57a3ac6211b22843505c53a3466f8719dc3bdf/src/streamlink/buffers.py#L104-L112) while segments are being filtered out and the buffer not being filled / written to.

I'm not sure if this is the best or even a good implementation. The problem is that I don't want to subclass the `RingBuffer` for this and don't want to modify the [`streamlink_cli` logic](https://github.com/streamlink/streamlink/blob/3a57a3ac6211b22843505c53a3466f8719dc3bdf/src/streamlink_cli/main.py#L268-L390) either. This means that the `IOError` needs to be caught and prevented from being raised by adding another blocking event in the reader thread while filtering is done in the writer thread. I am not sure if I have missed some edge cases with my implementation.

This can be "tested" in production via
```bash
streamlink --loglevel=trace --hls-timeout=5 --twitch-disable-ads twitch.tv/CHANNEL best --stdout > /dev/null
```

The `--hls-timeout` value needs of course to be shorter than the stream's ads, but still high enough to not starve the read iteration in between playlist refreshes while waiting for the next segments (`--twitch-low-latency` or `--hls-playlist-reload-time` helps here).

Also, while testing on Twitch, the current ads workaround needs to be reverted first:
```diff
diff --git a/src/streamlink/plugins/twitch.py b/src/streamlink/plugins/twitch.py
index 6bdfd452..a327d405 100644
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -317,7 +317,7 @@ class TwitchAPI(object):
     # Private API calls
 
     def access_token(self, endpoint, asset, **params):
-        return self.call("/api/{0}/{1}/access_token".format(endpoint, asset), private=True, **dict(platform="_", **params))
+        return self.call("/api/{0}/{1}/access_token".format(endpoint, asset), private=True, **params)
 
     def hosted_channel(self, **params):
         return self.call_subdomain("tmi", "/hosts", format="", **params)
```

### DIFF
--- a/tests/streams/test_hls_filtered.py
+++ b/tests/streams/test_hls_filtered.py
@@ -1,0 +1,263 @@
+import requests_mock
+from tests.mock import MagicMock, call, patch
+import unittest
+
+import itertools
+from textwrap import dedent
+from threading import Event, Thread
+
+from streamlink.session import Streamlink
+from streamlink.stream.hls import HLSStream
+from streamlink.stream.hls_filtered import FilteredHLSStreamWriter, FilteredHLSStreamReader
+
+
+class _TestSubjectFilteredHLSStreamWriter(FilteredHLSStreamWriter):
+    def __init__(self, *args, **kwargs):
+        super(_TestSubjectFilteredHLSStreamWriter, self).__init__(*args, **kwargs)
+        self.write_wait = Event()
+        self.write_done = Event()
+
+    def write(self, *args, **kwargs):
+        # only write once per step
+        self.write_wait.wait()
+        self.write_wait.clear()
+
+        # don't write again during cleanup
+        if self.closed:
+            return
+
+        super(_TestSubjectFilteredHLSStreamWriter, self).write(*args, **kwargs)
+
+        # notify main thread that writing has finished
+        self.write_done.set()
+
+
+class _TestSubjectFilteredHLSReader(FilteredHLSStreamReader):
+    __writer__ = _TestSubjectFilteredHLSStreamWriter
+
+
+class _TestSubjectReadThread(Thread):
+    """
+    Run the reader on a separate thread, so that each read can be controlled from within the main thread
+    """
+    def __init__(self, segments, playlists):
+        Thread.__init__(self)
+        self.daemon = True
+
+        self.mocks = mocks = {}
+        self.mock = requests_mock.Mocker()
+        self.mock.start()
+
+        def addmock(method, url, *args, **kwargs):
+            mocks[url] = method(url, *args, **kwargs)
+
+        addmock(self.mock.get, TestFilteredHLSStream.url_playlist, [{"text": p} for p in playlists])
+        for i, segment in enumerate(segments):
+            addmock(self.mock.get, TestFilteredHLSStream.url_segment.format(i), content=segment)
+
+        session = Streamlink()
+        session.set_option("hls-live-edge", 2)
+        session.set_option("hls-timeout", 0)
+        session.set_option("stream-timeout", 0)
+
+        self.read_wait = Event()
+        self.read_done = Event()
+        self.data = []
+        self.error = None
+
+        self.stream = HLSStream(session, TestFilteredHLSStream.url_playlist)
+        self.reader = _TestSubjectFilteredHLSReader(self.stream)
+        self.reader.open()
+
+    def run(self):
+        while True:
+            # only read once per step
+            self.read_wait.wait()
+            self.read_wait.clear()
+
+            # don't read again during cleanup
+            if self.reader.closed:
+                return
+
+            try:
+                data = self.reader.read(-1)
+                self.data.append(data)
+            except IOError as err:
+                self.error = err
+                return
+            finally:
+                # notify main thread that reading has finished
+                self.read_done.set()
+
+    def cleanup(self):
+        self.reader.close()
+        self.mock.stop()
+        # make sure that write and read threads halts on cleanup
+        self.reader.writer.write_wait.set()
+        self.read_wait.set()
+
+    def await_write(self):
+        writer = self.reader.writer
+        if not writer.closed:
+            # make one write call and wait until write call has finished
+            writer.write_wait.set()
+            writer.write_done.wait()
+            writer.write_done.clear()
+
+    def await_read(self):
+        if not self.reader.closed:
+            # make one read call and wait until read call has finished
+            self.read_wait.set()
+            self.read_done.wait()
+            self.read_done.clear()
+
+
+@patch("streamlink.stream.hls.HLSStreamWorker.wait", MagicMock(return_value=True))
+class TestFilteredHLSStream(unittest.TestCase):
+    url_playlist = "http://mocked/path/playlist.m3u8"
+    url_segment = "http://mocked/path/stream{0}.ts"
+
+    @classmethod
+    def get_segments(cls, num):
+        return ["[{0}]".format(i).encode("ascii") for i in range(num)]
+
+    @classmethod
+    def get_playlist(cls, media_sequence, items, filtered=False, end=False):
+        playlist = dedent("""
+            #EXTM3U
+            #EXT-X-VERSION:5
+            #EXT-X-TARGETDURATION:1
+            #EXT-X-MEDIA-SEQUENCE:{0}
+        """.format(media_sequence))
+
+        for item in items:
+            playlist += "#EXTINF:1.000,{1}\nstream{0}.ts\n".format(item, "filtered" if filtered else "live")
+
+        if end:
+            playlist += "#EXT-X-ENDLIST\n"
+
+        return playlist
+
+    @classmethod
+    def filter_sequence(cls, sequence):
+        return sequence.segment.title == "filtered"
+
+    def subject(self, segments, playlists):
+        thread = _TestSubjectReadThread(segments, playlists)
+        self.addCleanup(thread.cleanup)
+        thread.start()
+
+        return thread, thread.reader, thread.reader.writer
+
+    @patch("streamlink.stream.hls_filtered.FilteredHLSStreamWriter.should_filter_sequence", new=filter_sequence)
+    @patch("streamlink.stream.hls_filtered.log")
+    def test_filtered_logging(self, mock_log):
+        segments = self.get_segments(8)
+        thread, reader, writer = self.subject(segments, [
+            self.get_playlist(0, [0, 1], filtered=True),
+            self.get_playlist(2, [2, 3], filtered=False),
+            self.get_playlist(4, [4, 5], filtered=True),
+            self.get_playlist(6, [6, 7], filtered=False, end=True)
+        ])
+
+        self.assertTrue(reader.filter_event.is_set(), "Doesn't let the reader wait if not filtering")
+
+        for i in range(2):
+            thread.await_write()
+            thread.await_write()
+            self.assertEqual(len(mock_log.info.mock_calls), i * 2 + 1)
+            self.assertEqual(mock_log.info.mock_calls[i * 2 + 0], call("Filtering out segments and pausing stream output"))
+            self.assertFalse(reader.filter_event.is_set(), "Lets the reader wait if filtering")
+
+            thread.await_write()
+            thread.await_write()
+            self.assertEqual(len(mock_log.info.mock_calls), i * 2 + 2)
+            self.assertEqual(mock_log.info.mock_calls[i * 2 + 1], call("Resuming stream output"))
+            self.assertTrue(reader.filter_event.is_set(), "Doesn't let the reader wait if not filtering")
+
+            thread.await_read()
+
+        self.assertEqual(
+            b"".join(thread.data),
+            b"".join(list(itertools.chain(segments[2:4], segments[6:8]))),
+            "Correctly filters out segments"
+        )
+        for i, _ in enumerate(segments):
+            self.assertTrue(thread.mocks[TestFilteredHLSStream.url_segment.format(i)].called, "Downloads all segments")
+
+    # don't patch should_filter_sequence here (it always returns False)
+    def test_not_filtered(self):
+        segments = self.get_segments(2)
+        thread, reader, writer = self.subject(segments, [
+            self.get_playlist(0, [0, 1], filtered=True, end=True)
+        ])
+
+        thread.await_write()
+        thread.await_write()
+        thread.await_read()
+        self.assertEqual(b"".join(thread.data), b"".join(segments[0:2]), "Does not filter by default")
+
+    @patch("streamlink.stream.hls_filtered.FilteredHLSStreamWriter.should_filter_sequence", new=filter_sequence)
+    def test_filtered_timeout(self):
+        segments = self.get_segments(2)
+        thread, reader, writer = self.subject(segments, [
+            self.get_playlist(0, [0, 1], filtered=False, end=True)
+        ])
+
+        thread.await_write()
+        thread.await_read()
+        self.assertEqual(thread.data, segments[0:1], "Has read the first segment")
+
+        # simulate a timeout by having an empty buffer
+        # timeout value is set to 0
+        thread.await_read()
+        self.assertIsInstance(thread.error, IOError, "Raises a timeout error when no data is available to read")
+
+    @patch("streamlink.stream.hls_filtered.FilteredHLSStreamWriter.should_filter_sequence", new=filter_sequence)
+    def test_filtered_no_timeout(self):
+        segments = self.get_segments(4)
+        thread, reader, writer = self.subject(segments, [
+            self.get_playlist(0, [0, 1], filtered=True),
+            self.get_playlist(2, [2, 3], filtered=False, end=True)
+        ])
+
+        self.assertTrue(reader.filter_event.is_set(), "Doesn't let the reader wait if not filtering")
+
+        thread.await_write()
+        thread.await_write()
+        self.assertFalse(reader.filter_event.is_set(), "Lets the reader wait if filtering")
+
+        # make reader read (no data available yet)
+        thread.read_wait.set()
+        # once data becomes available, the reader continues reading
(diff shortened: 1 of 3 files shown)


bastimeyer:
I've added info log messages when filtering starts and ends and changed the initial event lock state, as it's unnecessary to have it locked initially.

Even though tests still need to be written/finished, I'd really appreciate a review, because the filtering is working fine in production.

gravyboat (approved):
This looks fine to me @bastimeyer. Should probably get a second review in case I missed something, but I don't see anything I'm concerned about.

bastimeyer:
Well, I was hoping to get some feedback rather than a plain approval. :disappointed: Did you run the code changes with the suggested examples? I'm struggling a bit writing tests for this where I can advance the writer's write calls and the reader's read calls step by step from the main thread, so I can test the IOError handling and thread events/locks with full coverage. I want it to be a real integration test with a full HLS stream, but unfortunately my Python expertise isn't that great regarding threading and I'm running into some timing issues here.

I really want to get this done as soon as possible, so that we can publish a new release with an embedded ads workaround and a working filtering that doesn't time out, if Twitch decides to block the workaround again. I don't want to deal with questions about the embedded ads on Twitch anymore. This is extremely annoying. If I haven't made any progress on this until tomorrow, I'll open a PR with a changelog for a new release. If you want, we could also do it today and ignore the filtering timeout issue.

gravyboat:
@bastimeyer Sorry for the lack of clarity, yes I did run it and I used your example as well as running it with `--twitch-low-latency` and results are actually better for me with this implementation than they currently are on the live build as I get stuttering using low latency because of how the segments are pulled once the ad is finished.

I'm not quite sure how to handle writing the tests for this either as I've always struggled with python threading. Timing issues are hard to solve and I haven't been writing enough python recently to provide the knowledge we need. @back-to Any insight or suggestions on how to handle this?

I agree with your annoyance. I don't want to deal with this again either and I'm sick of it. My issue with doing a release before handling the filtering timeout issue is that you and I both know people are going to start complaining about the filtering timeout issue. If we can't come up with a good way to test this I'd rather get this in without tests and deal with that not working for people than listen to complaints about filtering timeout.

bastimeyer:
OK, I've managed to write the tests. Not pretty and lots of verbose setup code, but whatever.

This is still unfinished. Tests are currently not working in py2 (didn't pay any attention to that, so I'm not sure what fails/stalls and how it can be fixed) and the requests mocker seems to have some issues as well, but only if the entire test suite is run for some reason.

Other than that, there's one race condition with the test setup in one of the tests which I could work around. Consecutively ran the tests 300 times and had no other race conditions.

Will try to fix the remaining issues tomorrow.

gravyboat:
I just looked the tests over, they're a bit messy as you noted but decent enough in structure that I think we can consider this acceptable. If someone comes up with a better idea they can propose a commit on to your branch, but if not this will do. I'll check in on this issue around 23:00 your time tomorrow when I take my lunch break and if no one else has weighed in by then let's merge this.

bastimeyer:
Switching from threading.Condition to threading.Event made it work in py2 now and the one race condition with the workaround now doesn't exist anymore.

The only issue I've noticed while running the tests is the requests_mocker, which doesn't seem to properly intercept all network requests to `http://mocked/path/...`, but for some reason this only happens when running all tests with coverage report. On the CI runners, this doesn't happen, which is weird. It's possible though that it's caused by the Twitch plugin tests and not the ones here.

Overall, I think I'm satisfied with it now. I should have tested everything and the coverage is 100%. Let's see if @back-to wants to review this, otherwise we can merge this later today / tomorrow.

Btw, if you're wondering why it's messy, hooking into the right methods of the HLS classes is a bit difficult and the other HLS tests are nowhere near as detailed / "thorough". I have annotated more lines in the latest push, and since every test assertion has a message, I don't think it's too bad.

gravyboat:
@bastimeyer My comment wasn't meant as a criticism merely trying to note that your concern over them not being pretty wasn't a big deal. I just looked over the tests again and they look good to me. The additional comments are also helpful.

bastimeyer:
Added one final refinement, so that the default `should_filter_sequence` method actually gets tested.

And the requests_mock issue I was talking about when generating a coverage report is caused by a different test (tested it on master). That's probably the Twitch plugin tests, which I will try to rewrite and simplify later at some point, not sure when. That's the error, btw:
> `[stream.hls][warning] Failed to reload playlist: Unable to open URL: http://mocked/path/playlist.m3u8 (HTTPConnectionPool(host='mocked', port=80): Max retries exceeded with url: /path/playlist.m3u8 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f4955825df0>: Failed to establish a new connection: [Errno -2] Name or service not known')))`

gravyboat:
@bastimeyer I just looked over the tests and am happy with them. Do you still want another review? If not let's merge this.

bastimeyer:
If we're going to publish a new release today or tomorrow, then you can go ahead and merge if you want.
I will write a quick changelog and open a PR then.

gravyboat:
@bastimeyer Done. I can check the changelog PR after work.

bastimeyer:
@gravyboat There's a problem with a race condition in one of the tests which I didn't spot...

And now two commits on the master branch have failed already:
https://github.com/streamlink/streamlink/runs/1146925732#step:5:239

That's a bit unfortunate, because I let my tests run for at least 500 times in both py3 and py2 because I wasn't too sure as I've said. I guess my CPU is too fast for the issue to occur, so I didn't notice it. Let me see if I can fix this quickly.

gravyboat:
@bastimeyer It's not ideal but you could put some sleeps in just to ensure the race conditions don't happen. It's dirty but if you can't get it working it's an option.

tests/streams/test_hls_filtered.py
```diff
@@ -0,0 +1,263 @@
...
+        thread, reader, writer = self.subject(segments, [
+            self.get_playlist(0, [0, 1], filtered=True),
+            self.get_playlist(2, [2, 3], filtered=False, end=True)
+        ])
+
+        self.assertTrue(reader.filter_event.is_set(), "Doesn't let the reader wait if not filtering")
+
+        thread.await_write()
+        thread.await_write()
+        self.assertFalse(reader.filter_event.is_set(), "Lets the reader wait if filtering")
```
bastimeyer, line 228: *Commenting here, so that it's clear where the issue is.*

Adding sleeps is a last resort solution, and I don't really want this there.

I don't understand the issue if I'm being honest. The writer and main thread should both be synced properly after the await_write() call by the main thread, and the writer thread should have cleared the reader event, meaning that filter_event.is_set() should be false at this point.

The only thing that could go wrong here is the patched should_filter_sequence method, which could have returned a False value and altered the intended code path. And it's confusing to me that this has only happened in py2 so far.

bastimeyer: The test failure has started to happen after my last change with the patch decorator (if we've not been unlucky with successful runs prior to that). Is it possible that it gets replaced lazily and too late in some cases?

Why has the `test_filtered_closed` test not failed yet? It's basically the same setup at the beginning of both tests.

gravyboat: I don't see how we could have been that unlucky with previous successful runs. Lazy replacement is possible. I just ran the tests in my local (Windows) environment and the new twitch tests passed perfectly fine. It seems pretty suspicious as well that this issue is limited to the CI.

bastimeyer: I've run this specific test on my test-account's fork 100 times and it didn't fail:
https://github.com/bastimeyer-test/streamlink-test/actions/runs/266135453

Running all tests 100 times however did fail eventually (with failfast enabled to cancel the other runners):
https://github.com/bastimeyer-test/streamlink-test/actions/runs/266138718

I will revert the last change now and then run it again.

bastimeyer: Looks like there's something else wrong with the await_write() method or the general thread syncing method that's being used, because it now failed at a different place:
https://github.com/bastimeyer-test/streamlink-test/runs/1147228622?check_suite_focus=true#step:5:1246
that's annoying

bastimeyer: Did some more debugging by adding detailed log messages and found the source of the issue (but not the cause yet).

In the `test_filtered_no_timeout` test for example, the two segments of the first playlist are supposed to be filtered out and the two segments of the second playlist are supposed to be written. This means that when trying to write the first two segments, it should ignore them and lock the reader thread. Sometimes this isn't done though, because the writer already starts with the third sequence, the first segment of the second playlist. Since this segment should not get filtered out, the lock in the reader thread is of course not set and the test fails.

So the question is, why does the writer start writing the third segment sometimes? The worker, which is putting / queuing up segments into the writer is for some reason starting with the second playlist.

According to the log output, this isn't the fault of the worker, so it must be `requests_mock`'s fault:
https://github.com/bastimeyer-test/streamlink-test/runs/1147915591?check_suite_focus=true#step:5:639

I've already tried moving the requests_mock setup from the reader thread where it currently is to the main thread and into the unittest.TestCase class's `setUp` and `tearDown` methods, but it didn't help. The playlist request responses are set up correctly:
https://requests-mock.readthedocs.io/en/latest/response.html#response-lists

Is it possible that it's interference with the other tests, most likely the Twitch plugin's tests, where the requests_mock doesn't get reset properly? That would explain the network error messages I was talking about earlier.

bastimeyer: I have fixed some issues with stalled threads on teardown in the hls_filtered tests by rewriting them and joining all spawned threads in the main thread's tearDown method and waiting for their termination, but the issue still occurs, now much rarer though, which is interesting.

When printing the name of all running threads after a clean teardown, it looks like some `ThreadPoolExecuter` segment-fetch-threads of one of the previously run stream workers are still running, even though all other tests are calling close() on their spawned streams every time. Those left over fetch-threads are bleeding into the current test run and are making the requests on the mocked URLs, which are the same between tests:
https://github.com/bastimeyer-test/streamlink-test/runs/1149311834?check_suite_focus=true#step:5:1418

If I change the mocked URLs in just the hls_filtered tests, then there are no failures :tada::tada::tada:, which means these bleeding threads are coming from other tests runs. Those are probably test_twitch.py or test_hls.py, because those are the only ones which are opening HLSStreams and reading them, the latter being more likely since it's run right before test_hls_filtered.py.
https://github.com/bastimeyer-test/streamlink-test/actions/runs/266894833 (no failed tests after 30 mins of execution - canceled the workflow after that, because I didn't want to wait any longer)

I will open a PR with fixes later after cleaning up all my debugging. This will be a fix and a workaround at the same time, because I won't touch the other "faulty" tests with their bleeding threads now. This can be done later.

bastimeyer: Fix/workaround (commit ID will be different than my PR):
https://github.com/bastimeyer-test/streamlink-test/commit/8be37ef6fb145ccb83ad11b51fa0e1453b5e0cdd
All tests running 50 times in a row on all CI runners (currently still running since 20 mins while I'm writing this):
https://github.com/bastimeyer-test/streamlink-test/actions/runs/266975807

PR #3188, opened 2020-09-20
plugins.mico: new plugin for http://www.micous.com/

### LINKED ISSUE #3136: MICO LIVE plugin request
## Plugin Request

- [ x] This is a plugin request and I have read the contribution guidelines and plugin request requirements.

### Description

 This is a popular live stream app's website which is known as **MICO LIVE** similar to bigo live

### Example stream URLs

1. https://www.micous.com/live/70314292
2. https://www.micous.com/live/68928340

mkbloke:
closes #3136

### DIFF
--- a/src/streamlink/plugins/mico.py
+++ b/src/streamlink/plugins/mico.py
@@ -1,0 +1,72 @@
+import logging
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import validate
+from streamlink.stream import HLSStream
+from streamlink.utils import parse_json
+from streamlink.utils.url import update_scheme
+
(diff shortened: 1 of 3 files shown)


src/streamlink/plugins/mico.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def get_title(self):
+        if self.title is not None:
+            return self.title
+
+    def _get_streams(self):
+        json_data = self.session.http.get(self.url, schema=self._json_data_schema)
+
+        if not json_data:
+            log.error("Failed to get JSON data")
```
back-to, line 58: don't mix `"` and `'` in a new plugin,
use just one of them.

Issue #3191, opened 2020-09-21
Update codecov/coverage config and add lgtm.com check


bastimeyer:
As an improvement to the code coverage reports, I'm proposing the following changes:

1. **Remove codecov-bot comment from pull requests**
   The comment with the status table nowadays gets added to the PR **checks** page ([example](https://github.com/streamlink/streamlink/pull/3187/checks?check_run_id=1144194464)) and can be read there. This will reduce the spamming (notifications as well) and also the audit log spamming when the bot usually updates its comment. The report page on codecov.io should be linked in the **checklist** next to the merge button, but for some reason it doesn't do that on the repo here and just links to their homepage. I don't see this behavior on the Twitch GUI repo and it correctly links to the generated report. That's probably a configuration issue here.
2. **Add branch coverage**
   `pytest-cov` currently only generates line coverage data, which isn't that detailed and useful. Branch/partial coverage data can be generated by adding [`--cov-branch`](https://pytest-cov.readthedocs.io/en/latest/config.html) to the [`pytest` command](https://github.com/streamlink/streamlink/blob/3a57a3ac6211b22843505c53a3466f8719dc3bdf/.github/workflows/main.yml#L39).
3. **Split the coverage scopes**
   The coverage reports can be [split up by codecov into different "projects"](https://docs.codecov.io/docs/commit-status), eg. "overall" (default - should be disabled), "streamlink", "streamlink_cli", "plugins", "tests".
   This helps checking which parts are tested well enough and which are not (plugins for example), and different targets + thresholds can be defined. The split coverage projects will be listed separately in the PR checklist and commit status, but differentiating between "streamlink", "streamlink_cli" and "tests" is what we should do at least IMO. The tests [need to be added](https://github.com/streamlink/streamlink/blob/3a57a3ac6211b22843505c53a3466f8719dc3bdf/.coveragerc#L1-L3) by the coverage generator of course.
4. **Rename `codecov.yml` to `.codecov.yml`**
   CI provider config files should be a dotfile.

----

Regarding lgtm.com, this is a CI service which semantically analyzes the code ("AI based" :roll_eyes:) in addition to the code-linter's static analysis. This helps finding more potential issues and the service is free for open source projects.

We currently have over 50 unnoticed alerts, most of them being issues with regexes in plugins:
https://lgtm.com/projects/g/streamlink/streamlink/alerts/?mode=list

[...]

PR #3192, opened 2020-09-21
plugins: fix unescaped literal dots in url_re entries

mkbloke:
As identified in: https://github.com/streamlink/streamlink/issues/3191#issue-705642714


bastimeyer:
Thank you, appreciated!

PR #3194, opened 2020-09-22
stream.hls_filtered: fix tests

bastimeyer:
- Refactor test setup and properly tear down threads
- Re-order tests
- Change mocked request URLs to temporarily work around bleeding threads
  from previously run HLSStream tests which are sometimes still making
  requests on the same URLs and causing the mocked playlist URL to
  incorrectly advance one call and breaking the current stream worker

Resolves #3187
See https://github.com/streamlink/streamlink/pull/3187#discussion_r492716895

As mentioned in the linked comment, test_hls.py and test_twitch.py need to be fixed at some point, so that they gracefully tear down their spawned HLSStream threads, so that the same mocked request URLs don't get queried in tests which are run (directly) afterwards. Using unique mocked request URLs in each test would make this extra safe.

----

All tests successfully ran 50 times in a row on all CI runners on my test-account's fork:
https://github.com/bastimeyer-test/streamlink-test/actions/runs/266975807

### DIFF
--- a/tests/streams/test_hls_filtered.py
+++ b/tests/streams/test_hls_filtered.py
@@ -5,6 +5,7 @@ import unittest
 import itertools
 from textwrap import dedent
 from threading import Event, Thread
+from time import sleep
 
 from streamlink.session import Streamlink
 from streamlink.stream.hls import HLSStream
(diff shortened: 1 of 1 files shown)


bastimeyer:
Btw, I believe the slightly incorrect coverage results in some PRs are a result of the other HLSStream related tests not tearing down properly. The code lines which are run sometimes and sometimes not are these:
- https://codecov.io/gh/streamlink/streamlink/pull/3174/changes
- https://github.com/streamlink/streamlink/blob/19eb7c9c084a77b287add0cc36062f5b09e5bdef/src/streamlink/stream/hls.py#L133-L135
- https://github.com/streamlink/streamlink/blob/19eb7c9c084a77b287add0cc36062f5b09e5bdef/src/streamlink/stream/hls.py#L322-L323

gravyboat:
Nice this looks good, thanks @bastimeyer.

PR #3196, opened 2020-09-22
plugins.ine: fix unescaped literal dots in js_re

mkbloke:
@bastimeyer, I don't know if these test failures under Py2 are something you might need to look into.  Perhaps they're already fixed in #3194 though?


bastimeyer:
@mkbloke please rebase onto the new master HEAD after #3194 got just merged. Your PR is based on the faulty tests from yesterday.

bastimeyer:
Thanks, appreciated!

Issue #2906, opened 2020-04-17
[New Plugin] Add DTV


roshavagarga:
## Plugin Request
- [x] This is a plugin request and I have read the contribution guidelines and plugin request requirements.

### Description
DTV, a.k.a. Diyana TV is a regional TV station (Yambol, Bulgaria). It can be considered a polythematic channel - featuring news, tv shows, documentaries, etc.

### Example stream URLs
1. http://dk-tv.com/dtv-%D0%BD%D0%B0-%D0%B6%D0%B8%D0%B2%D0%BE/
2. http://live.dk-tv.com

### Additional comments, screenshots, etc.
It seems to be using an HLS address, but opening that directly with streamlink doesn't work.
https://hastebin.com/ejejexirur.xml

mkbloke:
You closed?  I thought this was probably a reasonable request and likely to be approved by the devs if a plugin was created...

roshavagarga:
@mkbloke Honestly it's a small channel compared to TV999, isn't even available on satellite unlike the latter. I could open this up, but not sure it's worth the work :)

mkbloke:
Fair enough.  There certainly isn't a shortage of new plugin requests.  :-D

Issue #2331, opened 2019-02-27
doki.live


mamushi00:
## Plugin Request

- [x] This is a plugin request and I have read the contribution guidelines and plugin request requirements.

### Description

streaming service https://www.doki.live/. Is it possible to add a plugin for it?

### Example stream URLs

1. https://www.doki.live/
2. https://www.doki.live/2727773
3. https://livecdn.dokidokilive.com/live/dokidoki/mlinkm/6915512168_/seg0/index.m3u8

### Additional comments, screenshots, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

PR #3199, opened 2020-09-23
plugins.tv999: new plugin for http://tv999.bg/live.html

### LINKED ISSUE #2905: [New Plugin] Add TV999
## Plugin Request

### Description
Bulgarian TV channel that mainly features Russian content with Bulgarian subtitles.

### Example stream URLs
http://tv999.bg/live.html

mkbloke:
closes #2905

### DIFF
--- a/src/streamlink/plugins/tv999.py
+++ b/src/streamlink/plugins/tv999.py
@@ -1,0 +1,51 @@
+import logging
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import validate
+from streamlink.stream import HLSStream
+from streamlink.utils.url import update_scheme
+
+log = logging.getLogger(__name__)
+
+
+class TV999(Plugin):
+    url_re = re.compile(r'https?://(?:www\.)?tv999\.bg/live\.html')
+    iframe_re = re.compile(r'<iframe.*src="([^"]+)"')
+    hls_re = re.compile(r'src="([^"]+)"\s+type="application/x-mpegURL"')
+
+    iframe_schema = validate.Schema(
+        validate.transform(iframe_re.search),
+        validate.any(None, validate.all(
+            validate.get(1),
+            validate.url(),
+        )),
+    )
+
+    hls_schema = validate.Schema(
+        validate.transform(hls_re.search),
+        validate.any(None, validate.all(
+            validate.get(1),
+            validate.transform(lambda x: update_scheme('http:', x)),
+            validate.url(),
+        )),
+    )
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls.url_re.match(url) is not None
+
+    def _get_streams(self):
+        iframe_url = self.session.http.get(self.url, schema=self.iframe_schema)
+
+        if not iframe_url:
+            log.error('Failed to find IFRAME URL')
+            return
+
+        hls_url = self.session.http.get(iframe_url, schema=self.hls_schema)
+
+        if hls_url:
+            return {'live': HLSStream(self.session, hls_url)}
+
+
+__plugin__ = TV999
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -233,6 +233,7 @@ tv5monde                - tv5monde.com       Yes   Yes   Streams may be geo-rest
                         - tv5mondepl... [8]_
 tv8                     tv8.com.tr           Yes   No
 tv360                   tv360.com.tr         Yes   No
+tv999                   tv999.bg             Yes   --    Streams are geo-restricted to Bulgaria
 tvibo                   player.tvibo.com     Yes   --
 tvp                     tvpstream.vod.tvp.pl Yes   No    Streams may be geo-restricted to Poland.
 tvplayer                tvplayer.com         Yes   No    Streams may be geo-restricted to Great Britain. Premium streams are not supported.
--- a/tests/plugins/test_tv999.py
+++ b/tests/plugins/test_tv999.py
@@ -1,0 +1,24 @@
+import unittest
+
+from streamlink.plugins.tv999 import TV999
+
+
+class TestPluginTV999(unittest.TestCase):
+    def test_can_handle_url(self):
+        should_match = [
+            'http://tv999.bg/live.html',
+            'http://www.tv999.bg/live.html',
+            'https://tv999.bg/live.html',
+            'https://www.tv999.bg/live.html',
+        ]
+        for url in should_match:
+            self.assertTrue(TV999.can_handle_url(url))
+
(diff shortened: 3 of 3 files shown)


gravyboat:
@mkbloke Let's be careful about adding too many plugins. When people say content is "popular" I want to see actual stats showing the site is popular. As well as ensuring it's not just stolen content or VODs crushed together in to a live stream, since that doesn't really count.

mkbloke:
@gravyboat, fair points, for sure.  I have done some digging following your comment and have found that it *seems* to be a live channel available via satellite and cable in Bulgaria, according to these pages (via translation in Chrome):

- http://tv999.bg/%D0%B7%D0%B0-%D0%BD%D0%B0%D1%81.html
- https://bg.wikipedia.org/wiki/999_%D0%A2%D0%92

As for popularity, that seems harder to ascertain.  :-/

gravyboat:
@mkbloke Yeah when sites don't show viewers like many live streaming websites do it makes it hard to determine. The real problem I've been finding with people submitting these sites lately is what we consider to be "live" and people are trying to skirt the rules to get their crappy VOD site in because "but this is live TV in country/state XYZ!" which isn't live at all and I don't want to add more of these styles of plugins.

roshavagarga:
@mkbloke @gravyboat The plugin works and I can confirm that it's a small TV station that is available nationally - I believe it was from the city of Yambol, might be a former regional TV that went from being public to being private. I'd say the amount of people using it wouldn't be high, but the amount of work it'll require will be minimal, as the website hasn't been changed in ages.

It's an entirely entertainment channel - a few documentaries here and there, but mostly movies and TV series, often foreign ones with Bulgarian subtitling.

src/streamlink/plugins/tv999.py
```diff
@@ -0,0 +1,40 @@
...
+
+class TV999(Plugin):
+    url_re = re.compile(r'https?://(?:www\.)?tv999\.bg/live\.html')
+    iframe_re = re.compile(r'<iframe.*src="([^"]+)"')
+    hls_re = re.compile(r'src="([^"]+)"\s+type="application/x-mpegURL"')
+
+    iframe_schema = validate.Schema(
+        validate.transform(iframe_re.search),
+        validate.get(1),
+        validate.any(None, validate.url()),
```
bastimeyer, line 16: If the regex search fails, it'll return a `None` value, and trying to access `[1]` on a `None` value will raise a `TypeError`. That's what the `validate.any(None, validate.all(...))` call is for.

src/streamlink/plugins/tv999.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+
+    hls_schema = validate.Schema(
+        validate.transform(hls_re.search),
+        validate.get(1),
+        validate.any(None, validate.all(
```
bastimeyer, line 22: Same here

mkbloke:
@roshavagarga, I've just pushed some small cosmetic changes to the plugin, so if you could test again, it would be much appreciated.

By the way, I deleted the comment I made in your plugin request issue because I wanted to make another comment with the link to the newly pushed commit, rather than as a result of the conversation here.

Unfortunately, I can no longer seem to test directly myself.  I was using Tor configured with an exit node in Bulgaria, but perhaps I just got lucky yesterday as this no longer seems to be working for me now.  If the current commit works for you, I'll leave it as is and hopefully it can be merged, given the info you have provided above.

Cheers.

src/streamlink/plugins/tv999.py
```diff
@@ -0,0 +1,40 @@
...
+            validate.transform(lambda x: "http:{0}".format(x)),
```
bastimeyer, line 23: use `update_scheme`:

```python
from streamlink.utils.url import update_scheme
update_scheme("http:", "//foo.bar/baz") == "http://foo.bar/baz"
update_scheme("http:", "http://foo.bar/baz") == "http://foo.bar/baz"
update_scheme("http:", "https://foo.bar/baz") == "https://foo.bar/baz"
```

src/streamlink/plugins/tv999.py
```diff
@@ -0,0 +1,40 @@
...
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import validate
+from streamlink.stream import HLSStream
+
+
+class TV999(Plugin):
+    url_re = re.compile(r'https?://(?:www\.)?tv999\.bg/live\.html')
+    iframe_re = re.compile(r'<iframe.*src="([^"]+)"')
+    hls_re = re.compile(r'src="([^"]+)"\s+type="application/x-mpegURL"')
```
bastimeyer, line 11: Just saying this in general, and it isn't strictly meant in regards to your changes, but maybe we should stop using those wonky regexes for extracting data from HTML nodes and instead use a proper XML/HTML parser to get the nodes via xpath queries or CSS selectors and then read the attribute values. But ignore this now, just wanted to mention it.

bastimeyer (changes requested):

mkbloke:
@gravyboat, I see that roshavagarga has another plugin request open (#2906) which looks fairly reasonable to me, but the above conversation makes me unsure about creating a plugin for it.

I also wonder about sites like Mico, which I created a plugin for in #3188.  There's another plugin request (much older) for what appears to be a similar chat site in issue #2331.  Obviously these are live streaming sites, but it could be a case that these will pop up more and more often, which I guess might result in reluctance to accept streamlink plugins for them.

Perhaps some of these old plugin request issues should be reviewed again and perhaps closed.  It might also help if additional labels were used on these plugin request issues, say ```pending review``` and ```approved``` to indicate which ones are actually going to be accepted and merged into master if they are created?

bastimeyer:
> I see that roshavagarga has another plugin request open

Nothing unusual... :thinking:
https://github.com/streamlink/streamlink/issues?q=is%3Aissue+author%3Aroshavagarga+sort%3Aupdated-desc+label%3A%22plugin+request%22

There have been a ton of similar requests made by him over the past years and most of those small Bulgarian TV streams are covered by the `cdnbg` plugin. It's always the same or a similar structure:
https://github.com/streamlink/streamlink/blob/4fa902fc93470af30616e29e0fd49692a05ae84c/src/streamlink/plugins/cdnbg.py

Looks like this new plugin might be a duplicate and it could be added to the `cdnbg` plugin.

mkbloke:
Maybe I should just close this PR and consider trying to add this site to the cdnbg plugin, which as it is now seems to require a fix and update as per #3201...

mkbloke:
@bastimeyer, I've looked at the cdnbg plugin and that seems designed specifically for finding IFRAMEs that are sourced from cdn.bg.  While I could modify the IFRAME regex to include sources from elsewhere, would that really be the right thing to do?

PR #3200, opened 2020-09-24
Improve coverage reports on codecov

### LINKED ISSUE #3191: Update codecov/coverage config and add lgtm.com check

bastimeyer:
Resolves #3191

1. Enables branch coverage (proper line hit numbers are still unsupported)
2. Gathers coverage from tests
3. Turns codecov config file into a dotfile
4. Splits up coverage reports into "projects"
   - `default` (overall data)
   - `streamlink`
   - `streamlink_cli`
   - `plugins`
     Plugins are excluded from the streamlink project, because they are rarely fully tested and unnecessarily distort the coverage results
   - `tests`
5. Sets coverage change thresholds
   - `default`
     Doesn't have a threshold and sets an overall **target of 50%**
   - `streamlink` **1%**
   - `streamlink_cli` **1%**
   - `plugins`
     Since plugins are rarely fully tested or tested at all, no threshold here as well and instead a **target of 30%**
   - `tests` **0%**
     New tests should always be fully covered. if it's not possible, lines can be annotated with `# pragma: no cover`
6. Disables the PR comment made by the codecov bot, to reduce noise.
   When disabling the PR comments, the `default` project needs to be kept, because it's the only project which has a summary text similar to the comment, and the other projects don't have this.
7. Increases the `codecov.notify.after_n_builds` config to the exact number of test runners, so that the reports will only be posted once everything is ready. Otherwise, failures may be reported for a short while until every runner has finished, and we don't want that.
   To make sure that the coverage PR checks get created, even during service outages on codecov or when an upload has stalled and is in limbo, we've made the `default` status project a requirement for merging, which means it'll always await the coverage reports. This `default` check has a coverage target of 50% and it's unlikely to be undershot by a contributor's changes and therefore won't fail. We could also set an optional `informational` flag here, so that it always passes.

**One thing to note is that by splitting up the coverage reports and adding threshold values, this PR's status will fail.**
A workaround would be submitting two PRs, where the second one sets the thresholds and the first one doesn't have any success conditions, but that's unnecessary.


gravyboat:
@bastimeyer Looks good, pretty dumb that `after_n_builds` doesn't support a negative value to achieve what you want, but oh well.

bastimeyer:
As you can see, it's failing in some cases, because it doesn't have a reference point on the PR base yet (the master branch) and the delta is too high for the set threshold of 1%:
https://github.com/streamlink/streamlink/pull/3200/checks?check_run_id=1158219285

And regarding the `codecov.notify.after_n_builds`, I think we should set this to 10, the overall number of runners, so that it doesn't show failures while running.

bastimeyer:
What a great timing, there are no statuses after my push because apparently there's a "minor service outage"...
https://codecov.io/gh/streamlink/streamlink/commit/9f9fbab16df9a03413cb3f54e82a7d4bff61592b

And btw, if it's sometimes only showing "Codecov report" in the check list instead of the coverage delta, that's a bug which was reported on codecov's support forum yesterday or today.

bastimeyer:
@gravyboat Should we set the overall default report (the 50% target) as a master branch protection requirement? I think that a missing status like right now counts as a failure and blocks from (accidental / early) merging.

gravyboat:
@bastimeyer I'm not really a fan of that because it increases the overhead and easily confuses contributors, but if you're passionate about it I'm fine with that.

bastimeyer:
With the coverage of the tests being included now, the overall coverage is at 62.73%. That means that 50% won't be undershot unless lots of stuff gets removed, which is very unlikely. That won't confuse contributors, because it "can't fail". Remember that this check only exists for the summary similar to the removed comment. Requiring this check is purely for cases like this right now where the report is still pending.

gravyboat:
@bastimeyer Okay that's fine by me then, let's do it. 👍

bastimeyer:
> **codecov/project** Expected — Waiting for status to be reported

OK, done.

Let me restart the runners so they can re-upload the reports. I don't think this was properly queued on codecov's end and the missing reports are lost. I could also force a notification push on the PR page on codecov.io, and it would probably be enough since the only missing reports were ubuntu 3.5 and 3.6, but let's not do this.

bastimeyer:
@gravyboat If you feel more comfortable with the `informational: true` flag on the default project, so that it always succeeds/passes, and only *truly* serves as a report-upload-guarantee, then I could quickly change that.

Otherwise, we should be good to go here if we also ignore the status failure due to the split up and thresholds added, as mentioned above/earlier.

gravyboat:
@bastimeyer Nah I don't care that much. It's fine since the chance of it falling below is extremely small. I'll give the others a couple hours to respond and if they don't will merge later tonight.

bastimeyer:
That squash merge wasn't ideal. Now the history of the codecov config is gone due to the rename+change and the other individual changes are now part of one commit :/
A simple merge commit or rebase would've been better to preserve the semantic changes of the PR.

gravyboat:
Sorry @bastimeyer :( I didn't think it was something we'd consider reverting.

PR #3204, opened 2020-09-25
docs: use recommonmark as an extension

amurzeau:
Sphinx v1.8 has [deprecated source_parsers](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-source_parsers) which is used for recommonmark.

This commit replace the use of source_parsers for recommonmark with the
use of recommonmark as an extension (supported by recommonmark since
0.5.0).

Note: I've made a version requirement on recommonmark of >= 0.5.0 as recommonmark can't be used as a sphinx extension before that version.

If you want to keep compatibility with recommonmark v0.4.0 and older, just close this PR :-)

### DIFF
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,2 +1,2 @@
 sphinx>=1.6,<1.7
-recommonmark
+recommonmark>=0.5.0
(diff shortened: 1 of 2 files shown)


bastimeyer (approved):

gravyboat:
I thought I already approved and merged this. Any way thanks for the contribution.

PR #3209, opened 2020-09-28
tests: clean up existing HLS stream tests

bastimeyer:
Sorry for another wall of text, but this is a fix / cleanup for all existing HLS integration tests which I've mentioned in #3194. Please bear with me, because that are quite a lot of changes, although not too difficult and also split up into multiple commits. Except for a few cases, the logic of the individual tests was not changed and just cleaned up / simplified.

Description of the individual commits:

1. Adds the `TestMixinStreamHLS` class and related ones for simplification of all HLS integration tests. The module was added to the `tests.mixins.stream_hls` import path, but I'm not sure if it's a good choice or if it breaks any conventions. If so, this needs to be moved elsewhere.

   The mixin provides methods for creating a fake HLS stream, mocking the requests, creating a dedicated read thread which can be controlled from the main thread and it adds proper tear-down methods, so that all threads properly terminate first before continuing with the next test. Its logic is partly taken from the previous implementation of the `hls_filtered` tests.

   Instead of manually building the fake HLS streams, they can now be created via extensible `Playlist`, `Tag` and `Segment` classes. This is rather basic right now and doesn't cover every existing HLS tag, but it provides a much simpler interface which makes the test more clean and easier to write.

2. Refactors the `streams.test_hls_filtered` test and makes use of the new mixin. The logic for stepping through each write call individually (due to the test's hls-timeout value of 0) is not part of the mixin and has been kept here, as it's not needed anywhere else.

   The changes also resolve the sloppy `time.sleep(0.1)` workaround from the previous PR by mocking the `filter_event.wait()` call of the reader thread with another `threading.Event.wait()` call on the main thread. The solution is a bit weird unfortunately, because Python doesn't support wait-until-executed Mocks (there's a stale PR on cpython for that exact thing).

3. Refactors the `plugins.test_twitch` tests. Not much to see here apart from the removed unnecessary master playlist setup, the added test assertion messages and overall cleanup.

4. Refactors the `TestHlsPlaylistReloadTime` tests in `streams.test_hls` and makes use of the new mixin. This should fix the test coverage wonkiness, because the HLS streams previously kept running for half a second or so, still trying to make network requests.

5. Separates the `HLSStream.parse_variant_playlist` tests from the existing HLS tests in `streams.test_hls`. Since the master playlist has nothing to do with those stream tests, it needs to be tested in its own testcase class.

   I've added a few more assertions here which required me to change the return value of the `HLSStream.parse_variant_playlist` method to be an `OrderedDict` instead of a `dict`, because there was a difference between py2 and py3 when not sorting the streams (like the plugin API does). Also the usual unicode nonsense in py2.

6. Separates the regular and encrypted HLS stream tests in `streams.test_hls` and makes use of the new mixin.

   I've added a couple more test assertions and changed the parameters a bit.

   The encryption key override test has exposed a minor issue with the `{query}` variable and the quotation mark had to be added to the URL override template.

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -2,7 +2,7 @@ import logging
 import re
 import struct
 
-from collections import defaultdict, namedtuple
+from collections import defaultdict, namedtuple, OrderedDict
 from Crypto.Cipher import AES
 from requests.exceptions import ChunkedEncodingError
 
@@ -429,7 +429,7 @@ class HLSStream(HTTPStream):
         except ValueError as err:
             raise IOError("Failed to parse playlist: {0}".format(err))
 
-        streams = {}
+        streams = OrderedDict()
         for playlist in filter(lambda p: not p.is_iframe, parser.playlists):
             names = dict(name=None, pixels=None, bitrate=None)
             audio_streams = []
--- a/tests/mixins/stream_hls.py
+++ b/tests/mixins/stream_hls.py
@@ -1,0 +1,215 @@
+from binascii import hexlify
+from collections import OrderedDict
+from functools import partial
+from threading import Event, Thread
+
+import requests_mock
+import unittest
+
+from streamlink import Streamlink
+from streamlink.stream.hls import HLSStream
+
+
+class HLSItemBase(object):
+    path = ""
+
+    def url(self, namespace):
+        return "http://mocked/{namespace}/{path}".format(namespace=namespace, path=self.path)
+
+
+class Playlist(HLSItemBase):
+    path = "playlist.m3u8"
+
+    def __init__(self, mediasequence=None, segments=None, end=False, targetduration=0, version=7):
+        self.items = [
+            Tag("EXTM3U"),
+            Tag("EXT-X-VERSION", int(version)),
+            Tag("EXT-X-TARGETDURATION", int(targetduration))
+        ]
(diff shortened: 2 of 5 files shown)


gravyboat:
Phew this took a while to go through but it looks good to me @bastimeyer. Let's see if we can get a second pair of eyes on this from @beardypig. If not we'll merge. Seems like a lot of solid improvements.

bastimeyer:
Thanks for the review. Just pushed another small change because I noticed an inconsistency in one of the tests of the Twitch plugin.

I think this PR should get merged first before the ad filtering gets updated to fix the latest issues, because otherwise it will conflict with each other. Writing new tests or making changes to existing ones will also be much easier with this PR being merged.

gravyboat:
@bastimeyer Latest commit changes look good. Want me to merge this tonight if we don't get a second review?

bastimeyer:
@gravyboat Sure.

As I've described above, the only real change here is the return value of `HLSStream.parse_variant_playlist`, which is now an `OrderedDict` instead of a `dict`. Everything else is test related and I've preserved the same test logic everywhere except in the main HLS tests, which I've extended with more assertions.

Whether anyone else agrees with my changes, I can't say, as it's a lot of stuff. But since all HLS tests are integration tests and not unit tests, mocking calls to the various interfaces with custom return values and only testing partial code paths would be even more setup work than my method with building HLS playlists and only mocking the network requests.

Issue #3210, opened 2020-09-29
Twitch.tv embedded ads (2020-09-29)


bastimeyer:
Since we're seeing reports of new changes on Twitch regarding the embedded ads situation (once again), I feel the need of opening a new thread before it gets too messy again with the issues on the issue tracker here. I will update this post once we've got new informations, so that the current state can be found and read easily without digging through issue threads.

----

### Edit 10 (2020-12-23)

Streamlink `2.0.0` has been released yesterday with the workaround from edit 9, but it doesn't seem to be working reliably anymore. Ads can still be filtered out via `--twitch-disable-ads`.

### Edit 9 (2020-11-28)

Added a new access token request workaround in #3373 which again prevents ads from being embedded into the stream. (additional note: the access token requests have been moved to the GQL API already in #3344).

Like usual when there's a new change, you can install Streamlink from the master branch, or wait for the next nightly installer to be built if you're on Windows, or wait for the next release.

### Edit 8 (2020-11-05)

The ad prevention request parameter workaround has stopped working again, as predicted, and the ad "placeholder" screen seems to have been replaced with real ads now :boom: 

Filtering out ads via `--twitch-disable-ads` is however still working.

### Edit 7 (2020-10-28)

A new ad prevention has been discovered (#3301) and has been merged into `master` on 2020-10-28.
If you don't want to wait until the next release (`2.0.0` - release date unknown), you can [install Streamlink from the master branch](https://streamlink.github.io/install.html#pypi-package-and-source-code), or apply the small change yourself (cf978c5), or as a Windows user, [wait for the next nightly build of the installer (>= 2020-10-29)](https://github.com/streamlink/streamlink/actions?query=event%3Aschedule+is%3Asuccess+branch%3Amaster).

If you're using the Twitch GUI, you will have to use a debug / pre-release due to other breaking Streamlink changes in regards to the upcoming `2.0.0` release:
streamlink/streamlink-twitch-gui#757

### Edit 6 (2020-10-18)

[`1.7.0`](https://github.com/streamlink/streamlink/releases) has been released on 2020-10-18.

### Edit 5 (2020-10-03)

#3220 has been merged as well, which should prevent the embedded ads again (for now (edit: broken again)).

### Edit 4 (2020-10-03)

[...]

PR #3213, opened 2020-09-30
plugins.twitch: filter ads by EXT-X-DATERANGE tag

bastimeyer:
As mentioned in https://github.com/streamlink/streamlink/issues/3210#issuecomment-700571743, simply filtering ads by comparing each segment's title isn't good enough anymore and checking for the title not being equal to "live" can break the implementation if Twitch decides to change the title of live stream segments or if they decide to add titles to VOD segments.

So here's a better filtering approach, based on the [`EXT-X-DATERANGE`](https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.2.7) and [`EXT-X-PROGRAM-DATE-TIME`](https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.2.6) tags.

Example playlists can be found here:
- https://gist.github.com/drmthrowaway/99ddc2308ba431e76945e456da837b75
- https://gist.github.com/tp0/f4e6798301fb8361c192e17734a0ca16

----

Basic description of each commit:

1. Adds the `EXT-X-DATERANGE` tag to the HLS playlist parser and makes the `EXT-X-PROGRAM-DATE-TIME` tag return a datetime object instead of a string. Since I couldn't find any usage of the `EXT-X-PROGRAM-DATE-TIME` tag in the existing code, changing its return value should be fine. Also adds support for returning a custom M3U8 class by the parser and adds a `is_date_in_daterange` method to the existing `M3U8` class. This isn't an ideal solution, but I couldn't find a better place for this method so that it can be called when parsing is done apart from turning the `Segment` namedtuple into a regular class. Also worth to note is that I haven't implemented all possible attribute values of the `EXT-X-DATERANGE tag`, because I didn't need them (scte35 attributes).

2. Makes the Twitch plugin filter the parsed dateranges by attributes. It tries to find "ad dateranges" by the values of the ID and CLASS attributes and key names of custom attributes by Twitch. This is the crucial part of the ad detection. If a segment is then part of this daterange, it is marked as an ad and filtered out by the existing methods.

----

I've added tests for all "ad daterange" values I know of, **but since I still don't get any ads by Twitch, it's hard to know if my implementation is working correctly**. So please go ahead and test this on live data. Thank you!


gravyboat:
I'm still seeing the following error from a random "big" stream pulled off twitch:
```
[cli][debug] Python:     3.8.4
[cli][info] Found matching plugin twitch for URL twitch.tv/ludwig
[cli][debug] Plugin specific arguments:
[cli][debug]  --twitch-disable-ads=True (disable_ads)
[cli][debug]  --twitch-low-latency=True (low_latency)
[plugin.twitch][debug] Getting live HLS streams for ludwig
[plugin.twitch][info] Will skip ad segments
[plugin.twitch][info] Low latency streaming (HLS live edge: 2)
[cli][error] Try 1/1: Could not open stream <HLSStream('https://video-weaver.sea01.hls.ttvnw.net/v1/playlist/CvYD5-nuyHKhhuV5zmXiVfVH9PjoXmeyFFvmStbY5iRPPTWO6ArtWycw3wwb9tPgFbDLQP3D10SnixFmyfoS-uMDvD7CLxTxGv8daUybYYNo1pKnPuxOcwlyIe5ns0t0Ysw5jp-Td30eXpa_uVO2cZhEjlDgu-n6osjGZH9dzrzdJoefH3yWqTHMwlAWlSjmJElqRkG_RkbGoSw0-6Jb9ccLuU6-1wpt3dVI5n9l-af_T36wq__5UZ3G4V_nNbAIE6V_Qvi7rXdywN4S1N-XU1cpKlGrlTyU3q2rB-H_6zyI-Idhq8o5B_Myo7BeFOJBCzNUz9DgKtoXgIRot7NGdW-LvorDiFOknnmCO41NwDTdTBPXD8nwo8r7YiCOiAu0-7dAvbRqLVwJdAFXq_1d7q1L0uaV6EpNthepMWYDFjY68lYMi_Jp5HqqivDKEgJsG_7VM4LP0BAm1r25818W7SSch6xL9Yx8BeWQQ5qDpNbnK5Ojp0rJQYVf_nos_DErya9e5gbGk-xXCVZTGAq0CsQjDxFnLYgZKdpJrr-2Hn3UZlDG0Wfc_HaKsjRCxxJZLUSGbFAGIUZfdrfTdIn7yv_RBNsTwV_DKDRcetFtzZ6EHXLw_7BHw8OD-IXpx9Ish3jOUL9Sub9gHdC-hYxn0aj4YVieZd1AQxIQLmf5mJRpJTbwzOTajgzEERoMmTIkBI1gQnXYCk4F.m3u8')> (Could not open stream: invalid literal for int() with base 10: '15.183')
error: Could not open stream <HLSStream('https://video-weaver.sea01.hls.ttvnw.net/v1/playlist/CvYD5-nuyHKhhuV5zmXiVfVH9PjoXmeyFFvmStbY5iRPPTWO6ArtWycw3wwb9tPgFbDLQP3D10SnixFmyfoS-uMDvD7CLxTxGv8daUybYYNo1pKnPuxOcwlyIe5ns0t0Ysw5jp-Td30eXpa_uVO2cZhEjlDgu-n6osjGZH9dzrzdJoefH3yWqTHMwlAWlSjmJElqRkG_RkbGoSw0-6Jb9ccLuU6-1wpt3dVI5n9l-af_T36wq__5UZ3G4V_nNbAIE6V_Qvi7rXdywN4S1N-XU1cpKlGrlTyU3q2rB-H_6zyI-Idhq8o5B_Myo7BeFOJBCzNUz9DgKtoXgIRot7NGdW-LvorDiFOknnmCO41NwDTdTBPXD8nwo8r7YiCOiAu0-7dAvbRqLVwJdAFXq_1d7q1L0uaV6EpNthepMWYDFjY68lYMi_Jp5HqqivDKEgJsG_7VM4LP0BAm1r25818W7SSch6xL9Yx8BeWQQ5qDpNbnK5Ojp0rJQYVf_nos_DErya9e5gbGk-xXCVZTGAq0CsQjDxFnLYgZKdpJrr-2Hn3UZlDG0Wfc_HaKsjRCxxJZLUSGbFAGIUZfdrfTdIn7yv_RBNsTwV_DKDRcetFtzZ6EHXLw_7BHw8OD-IXpx9Ish3jOUL9Sub9gHdC-hYxn0aj4YVieZd1AQxIQLmf5mJRpJTbwzOTajgzEERoMmTIkBI1gQnXYCk4F.m3u8')>, tried 1 times, exiting
```

Billy2011:
> Every answer to any question about this I've found is "use python-dateutil", but I can't just add a dependency here, so a custom parser had to be written

Why don’t you use the isodate module for parsing, which is also used for the dash manifest parser?

azizLIGHT:
I think it's working sometimes (line 58)  but I might be doing this wrong but here's my output. Let me know if I should adjust something: https://gist.github.com/azizLIGHT/e60250ecb8091daa5619a4902a449efe

bastimeyer:
@Billy2011 Thanks... Don't ask how I've missed that. :disappointed: 
@azizLIGHT Regarding the error, you could try it again with the new changes. I will check this later, since I have to go now.

azizLIGHT:
> @azizLIGHT Regarding the error, you could try it again with the new changes. I will check this later, since I have to go now.

I got the latest version and tried again but got same errors about [`Could not open stream <HLSStream`]( https://gist.github.com/azizLIGHT/5863d9bf4382df59f69e98d8634bdc8a).
I will try again when there's more changes added

gravyboat:
@bastimeyer Most recent changes look good on a quick test. I get the pre-roll notification on the CLI then the stream opens correctly once they're in.

gravyboat:
@bastimeyer Yes there anything else you want to add here? Otherwise I want to squash and merge.

bastimeyer:
> anything else you want to add here?

No, that should be pretty much it. As said in the OP, some tag attributes haven't been implemented, but that doesn't matter.

One thing to note though is that I've also ignored the `END-ON-NEXT` attribute in the `is_date_in_daterange` method, which technically isn't 100% correct. That attribute basically says that if there's a second daterange with the same class name as the first one and that second one starts later than the first one, the first one ends at the start of the second one, regardless its own end or duration attributes. For filtering out ads on Twitch, this isn't a problem, because it doesn't matter which ad-daterange (of a similar kind) a segment is in if they overlap.

But before merging, I would prefer if we could get some feedback from other users as well, because I still don't get ads on Twitch from my IP address and have basically implemented something purely on the HLS spec and the reported playlist contents, and therefore don't know if I have missed anything in regards to the ad-daterange detection.
https://github.com/streamlink/streamlink/pull/3213/files#diff-195bebeb65dac1f6a8e94f1957a95ec1R143-R147

> I want to squash and merge

Don't squash. Both commits change two different components and therefore shouldn't be squashed (and they also have proper commit messages). Either rebase onto master or create a merge commit.

src/streamlink/plugins/twitch.py
```diff
@@ -125,27 +125,37 @@
...
+
     def get_segment(self, uri):
         byterange = self.state.pop("byterange", None)
         extinf = self.state.pop("extinf", (0, None))
         date = self.state.pop("date", None)
         map_ = self.state.get("map")
         key = self.state.get("key")
         discontinuity = self.state.pop("discontinuity", False)
-        ad = self.state.pop("ad", False)
+        ad = any([self.m3u8.is_date_in_daterange(date, daterange) for daterange in self.m3u8.dateranges_ads])
```
jtamagnan, line 158: I stumbled across this pull request while looking at the related issue. I figured I'd look at the code to see what was going and then figured I'd leave a small piece of feedback. If this wasn't the type of feedback you were hoping for then please ignore me :)

 Here and throughout this diff we are using `any([... for ... in ...])` and `all([... for ... in ...])`, these instances can be replaced by `any(... for ... in ...)` and `all(... for ... in ...)`. By removing the inner `[ ]` we are no longer creating intermediate lists, instead we are feeding a generator directly into `any` and `all`. This means that in the best case scenario we are iterating over fewer values since `any` and `all` will exit once they encounter the first `True` or `False` values respectively.

Thank you for all of your work on this, really love the software and appreciate the work you put into it!

bastimeyer: > If this wasn't the type of feedback you were hoping for then please ignore me :)

No, I appreciate it, thanks. It's just my inexperience with Python. I thought this was part of the iterator syntax. And yes, you're totally right, we don't need to feed an iterator into a list constructor to gather all of its results and then get the list's iterator to check for any/all matching results. That is wasteful.

bastimeyer:
@gravyboat I think we can merge this. Should be fine.

PR #3220, opened 2020-10-03
plugins.twitch: remove platform access token param

bastimeyer:
This once again removes the `platform=_` parameter from the stream access token request.

See #3210
As mentioned there, we need a better solution for this. Am happy to discuss this.

----

History of the parameter:
1. Added in #2358
2. Removed in #2692
3. Added in #3173

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -317,7 +317,7 @@ class TwitchAPI(object):
     # Private API calls
 
     def access_token(self, endpoint, asset, **params):
-        return self.call("/api/{0}/{1}/access_token".format(endpoint, asset), private=True, **dict(platform="_", **params))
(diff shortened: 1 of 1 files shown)


gravyboat:
@bastimeyer I'm not in favor of doing this. again. As you noted:
> Any volunteers for this work? Because I am already annoyed.

I don't feel like doing more, I already have a day job for that.

Maybe a param with CLI output for when it catches an ad and we link to the documentation talking about how to use the param? We need a way to keep people out of our hair on this as much as possible, that's the real problem. The less configuration required the better.

bastimeyer:
Let's get this merged, so that we can have a nightly build tomorrow which includes this change and the new filtering.

How the ads workaround should get handled in the future can be discussed later. A note when an ad is playing for the first time with a link to a page with instructions on setting custom request parameters doesn't sound too bad though.

PR #3224, opened 2020-10-04
plugins.mbcmini: New Plugin for http://mini.imbc.com

hyperbora:
mbcmini is South Korea radio station.

- http://mini.imbc.com

### DIFF
--- a/src/streamlink/plugins/mbcmini.py
+++ b/src/streamlink/plugins/mbcmini.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+import logging
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import useragents, validate
+from streamlink.stream import HTTPStream
+from streamlink.utils import parse_json
+
+log = logging.getLogger(__name__)
+
+
+class Mbcmini(Plugin):
+
+    _url_re = re.compile(r'https?://miniplay\.imbc\.com/WebLiveURL\.ashx\?channel=(?P<channel_id>[a-z]+)')
+
+    _api_url = '''http://miniplay.imbc.com/WebLiveURL.ashx?channel={0}&protocol=M3U8\
+&agent=android&androidversion=18&callback=jarvis.miniInfo.loadOnAirComplete'''
+
+    _json_re = re.compile(r'jarvis\.miniInfo\.loadOnAirComplete\(({.*?})', re.DOTALL)
+
+    _stream_schema = validate.Schema(
+        validate.transform(_json_re.search),
+        validate.all(
+            validate.get(1),
+            validate.transform(lambda v: v.replace("\n", "")),
+            validate.transform(lambda v: v.replace("\t", "")),
+            validate.transform(parse_json),
+            {
+                "AACLiveURL":
+                validate.url()
+            }, validate.get("AACLiveURL"),
+        )
+    )
+
+    @ classmethod
+    def can_handle_url(cls, url):
+        return cls._url_re.match(url) is not None
+
+    def _get_streams(self):
+        match = self._url_re.match(self.url)
+        channel_id = match.group("channel_id")
+
+        headers = {
+            'User-Agent': useragents.CHROME,
+            'Referer': '''http://mini.imbc.com/webapp_v3/minifloat.html?\
+src=http://www.imbc.com/broad/radio/minimbc/index.html&ref='''
+        }
+        url = self._api_url.format(channel_id)
+
+        stream_url = self.session.http.get(url, headers=headers, schema=self._stream_schema)
+
+        if stream_url:
+            yield "live", HTTPStream(self.session, stream_url)
+
+
+__plugin__ = Mbcmini
--- a/src/streamlink/plugins/mbcmini.py
+++ b/src/streamlink/plugins/mbcmini.py
@@ -1,57 +1,49 @@
-# -*- coding: utf-8 -*-
-import logging
 import re
 
 from streamlink.plugin import Plugin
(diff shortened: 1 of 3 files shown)


mkbloke:
I presume the reason you have added the ```--mbcmini-id``` parameter is because there's no direct URL exposed to users for each station?  That's the logical solution for such a case, but I'll be interested to see if this is acceptable to the streamlink devs as it's counter to the way other plugins work, with the exception of just one - the sbscokr plugin which implements ```--sbscokr-id``` presumably for the same reason.

hyperbora:
Hi, thanks for the review. I modified it to work without parameters as you said.

mkbloke:
I'm sure they'll be happier with that. :+1:

src/streamlink/plugins/mbcmini.py
```diff
@@ -0,0 +1,44 @@
...
+        }
+        url = self._api_url.format(channel_id)
+
+        res = self.session.http.get(url, headers=headers)
+
+        json_text = '{' + ''.join(res.text.split()[1:-1])
+        json_data = json.loads(json_text)
+
+        if json_url_key in json_data:
+            return self.session.streams(json_data[json_url_key])
```
bastimeyer, line 41: You are recursively calling the plugin's `_get_streams` method by calling `self.session.streams` ([1](https://github.com/streamlink/streamlink/blob/6de76fca8f66dac8a9c686a5fabc6b65de1fe43e/src/streamlink/session.py#L449-L458), [2](https://github.com/streamlink/streamlink/blob/6de76fca8f66dac8a9c686a5fabc6b65de1fe43e/src/streamlink/plugin/plugin.py#L280-L317)).

In `Plugin._get_streams`, you are supposed to return an iterator or a list of objects which implement the [`Stream`](https://github.com/streamlink/streamlink/blob/6de76fca8f66dac8a9c686a5fabc6b65de1fe43e/src/streamlink/stream/stream.py) class, eg. `HLSStream`, `DashStream`, `HTTPStream`, `RTMPStream`, etc., depending on what kind of streams the website you're writing the plugin for provides.

src/streamlink/plugins/mbcmini.py
```diff
@@ -0,0 +1,44 @@
...
+            'User-Agent': useragents.CHROME,
+            'Referer': '''http://mini.imbc.com/webapp_v3/minifloat.html?\
+src=http://www.imbc.com/broad/radio/minimbc/index.html&ref='''
+        }
+        url = self._api_url.format(channel_id)
+
+        res = self.session.http.get(url, headers=headers)
+
+        json_text = '{' + ''.join(res.text.split()[1:-1])
+        json_data = json.loads(json_text)
```
bastimeyer, line 38: Plugins should use Streamlink's validator API to return results from network requests, if possible. The API response appears to be a JSONP call, which means you should extract the JSON payload via a regex instead of splitting the response text by newline characters.

bastimeyer (changes requested):
Thank you for the pull request. Is this a radio station only without any live video content?

Before writing a plugin and submitting a PR, it's usually best to open a plugin request or ask first whether it'll be accepted, because otherwise, you may waste the time you've spent on it if it gets rejected. I'm just mentioning it, because I don't know what the current policy is for audio-only streams (even though it's currently not stated in the [contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#plugin-requests)). I personally don't have a problem with this, but let's hear from the other maintainers as well.

mkbloke:
- [ ] CB
- [ ] DC

src/streamlink/plugins/mbcmini.py
hyperbora: Hi, thanks for the review. I modified it to use HTTPStream.

src/streamlink/plugins/mbcmini.py
hyperbora: Modified to use Streamlink's validation API.

hyperbora:
This radio station only transmits audio content.

bastimeyer (commented):
I've added a commit to the PR branch to fix up the plugin.
Whether it's good for merging depends on whether radio stations (audio only streams) are acceptable.

```
$ streamlink -l debug 'https://miniplay.imbc.com/WebLiveURL.ashx?channel=sfm'
[cli][debug] OS:         Linux-5.9.0-rc8-1-git-x86_64-with-glibc2.2.5
[cli][debug] Python:     3.8.5
[cli][debug] Streamlink: 1.6.0+23.g49f009f
[cli][debug] Requests(2.24.0), Socks(1.7.1), Websocket(0.56.0)
[cli][info] Found matching plugin mbcmini for URL https://miniplay.imbc.com/WebLiveURL.ashx?channel=sfm
Available streams: 81k (worst, best)
```

gravyboat:
@bastimeyer My concern is that allowing a radio plugin opens up a pretty big can of worms. There are a million online radio stations, are we going to accept all those because "well you accepted mbcmini". One of the big reasons for putting a site through Streamlink is to make it less resource intensive (Twitch is always my go to example because the site is so awful) or terrible to use in the event of a mediocre video player. Radio stations don't suffer from that issue, it's just streaming audio and presumably isn't that resource intensive though I can't confirm because the site is geo-blocked.

I'm leaning towards no on this, Streamlink is designed specifically for video content, and live content at that. I don't think a radio station really counts any more than a playlist of audio files in my local audio player does. I'm open to discussing it but these are my thoughts. It's not a video, it's not especially resource intensive, so what need does Streamlink serve in this case?

hyperbora:
I understand. Then I'll close this PR.

Issue #3167, opened 2020-09-11
No audio track selection in arte plugin


sysmso:
## Plugin Issue

- [x ] This is a plugin issue and I have read the contribution guidelines.

### Description

No audio track selection (or subtitles) in streams from arte.tv. For example, on this stream (https://www.arte.tv/fr/videos/087934-006-A/le-tueur-de-l-ombre-6-8/) there are 5 versions available: VOST, VF, Audiodescription, hearing impaired and VO. 

### Reproduction steps / Explicit stream URLs to test

1. ... streamlink https://www.arte.tv/fr/videos/087934-006-A/le-tueur-de-l-ombre-6-8/ best

### Log output

```
[cli][debug] OS:         Linux-5.8.8-arch1-1-x86_64-with-glibc2.2.5
[cli][debug] Python:     3.8.5
[cli][debug] Streamlink: 1.5.0
[cli][debug] Requests(2.24.0), Socks(1.7.1), Websocket(0.57.0)
[cli][info] Found matching plugin artetv for URL https://www.arte.tv/fr/videos/087934-006-A/le-tueur-de-l-ombre-6-8/
[utils.l10n][debug] Language code: fr_FR
[utils.l10n][debug] Language code: fr_FR
[cli][info] Available streams: 720p_http, 720p_http_alt, 360p_http, 216p_http, 406p_http, 216p_http_alt, 360p_http_alt, 406p_http_alt, 61k_alt (worst), 61k, 216p_alt, 216p, 360p_alt, 360p, 406p_alt, 406p, 720p_alt, 720p (best)
[cli][info] Opening stream: 720p (hls)
[stream.hls][debug] Reloading playlist
[stream.hls][debug] First Sequence: 1; Last Sequence: 249
[stream.hls][debug] Start offset: 0; Duration: None; Start Sequence: 1; End Sequence: 249
[stream.hls][debug] Adding segment 1 to queue
[cli][debug] Pre-buffering 8192 bytes
[stream.hls][debug] Adding segment 2 to queue
[stream.hls][debug] Adding segment 3 to queue
[stream.hls][debug] Adding segment 4 to queue
[stream.hls][debug] Adding segment 5 to queue
[stream.hls][debug] Adding segment 6 to queue
[stream.hls][debug] Adding segment 7 to queue
[stream.hls][debug] Adding segment 8 to queue
[stream.hls][debug] Adding segment 9 to queue
[stream.hls][debug] Adding segment 10 to queue
[...]

PR #3228, opened 2020-10-06
plugins.artetv: only pick the first variant of the stream

beardypig:
Removes the f4m and http streams from the returned list, and only returns the default variant for the stream. I believe this will match the expected behaviour. 

see #3167

### DIFF
--- a/src/streamlink/plugins/artetv.py
+++ b/src/streamlink/plugins/artetv.py
@@ -1,16 +1,14 @@
 """Plugin for Arte.tv, bi-lingual art and culture channel."""
 
+import logging
 import re
+from operator import itemgetter
 
-from itertools import chain
-
-from streamlink.compat import urlparse
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
-from streamlink.stream import HDSStream
 from streamlink.stream import HLSStream
-from streamlink.stream import HTTPStream
 
+log = logging.getLogger(__name__)
 JSON_VOD_URL = "https://api.arte.tv/api/player/v1/config/{0}/{1}?platform=ARTE_NEXT"
 JSON_LIVE_URL = "https://api.arte.tv/api/player/v1/livestream/{0}"
 
@@ -32,7 +30,8 @@ _video_schema = validate.Schema({
                     "height": int,
                     "mediaType": validate.text,
                     "url": validate.text,
-                    "versionShortLibelle": validate.text
+                    "versionProg": int,
+                    "versionLibelle": validate.text
                 },
             },
         )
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/artetv.py
```diff
@@ -46,43 +46,29 @@ def can_handle_url(cls, url):
...
-
-                        for stream in streams.items():
-                            yield stream
-                    except IOError as err:
-                        self.logger.error("Failed to extract HLS streams: {0}", err)
-                else:
-                    yield stream_name, HTTPStream(self.session, stream_url)
-
-            elif stream_type == "f4m":
+            if stream_type in "hls" and urlparse(stream_url).path.endswith("m3u8"):
```
bastimeyer, line 65: `stream_type in "hls"` should probably be changed to a strict comparison.

src/streamlink/plugins/artetv.py
```diff
@@ -46,43 +46,29 @@ def can_handle_url(cls, url):
...
-                    streams = HDSStream.parse_manifest(self.session, stream_url)
-
-                    for stream in streams.items():
-                        yield stream
+                    streams = HLSStream.parse_variant_playlist(self.session, stream_url)
+                    for name, stream in streams.items():
+                        yield name, stream
                 except IOError as err:
-                    self.logger.error("Failed to extract HDS streams: {0}", err)
+                    self.logger.error("Failed to extract HLS streams: {0}", err)
```
bastimeyer, line 71: Could you also please quickly fix the `self.logger` calls?

bastimeyer:
There are still potentially up to 4 additional alternative HLS streams for each stream input which would override each other's stream selections. Just like I've described in https://github.com/streamlink/streamlink/issues/3167#issuecomment-691299267.

If there really is a need for those alt-streams other than the main one, then we should add a plugin parameter for that, which selects those alt-streams.

bastimeyer:
Please check the commit message prefix. There's a typo in the plugin name.

beardypig:
> There are still potentially up to 4 additional alternative HLS streams for each stream input which would override each other's stream selections. Just like I've described in [#3167 (comment)](https://github.com/streamlink/streamlink/issues/3167#issuecomment-691299267).

I'm not sure which ones would come up as `alt`, I believe the list of codes that are filtered out are only the regular audio tracks. If you could provide a counter case that would be helpful. 

> If there really is a need for those alt-streams other than the main one, then we should add a plugin parameter for that, which selects those alt-streams.

A plugin parameter could be added to select which stream you wanted, you'd also need an option to list them. Regardless, that should be a separate PR as this is intended to fix the immediate issue highlighted in #3167.

bastimeyer:
> I'm not sure which ones would come up as `alt`

I think renaming the yielded streams could just work. Add the code as prefix with lowercase text and underscores instead of dashes. There will be a lot more streams to select, but at least they don't override each other. A plugin parameter would then not be needed.

beardypig:
After closer inspection I see there is a much better way to do it by selecting the default variant, I will also prepare a PR with a variant selection option that can be review separately, I think it makes more sense than listing 20 streams.

src/streamlink/plugins/artetv.py
```diff
@@ -45,44 +44,18 @@ class ArteTV(Plugin):
...
-
-                    for stream in streams.items():
-                        yield stream
-                except IOError as err:
-                    self.logger.error("Failed to extract HDS streams: {0}", err)
+                        streams = HLSStream.parse_variant_playlist(self.session, stream["url"])
+                        for s in streams.items():
+                            yield s
+                    except IOError:
+                        log.exception("Failed to extract HLS streams")
```
back-to, line 58:
```py
log.exception
```

this will add a **new line**, which looks weird.

beardypig: I changed it to a warning and included some more useful information.

src/streamlink/plugins/artetv.py
```diff
@@ -45,44 +44,21 @@ class ArteTV(Plugin):
...
-        elif language == "pl":
-            language = ["POL", "VOST-POL"]
-
-        if stream_language in language:
-            if stream_type in ("hls", "mp4"):
-                if urlparse(stream_url).path.endswith("m3u8"):
+    def _create_stream(self, streams):
+        variant, variantname = min([(stream["versionProg"], stream["versionLibelle"]) for stream in streams.values()],
+                                   key=itemgetter(0))
+        log.debug("Using the '{0}' stream variant".format(variantname))
```
back-to, line 50: error with py2 because of `Français`

```
$ streamlink https://www.arte.tv/fr/direct/ -l info
src/streamlink_cli/main.py:760: PythonDeprecatedWarning: Python 2.7 ...
[cli][info] Found matching plugin artetv for URL https://www.arte.tv/fr/direct/
error: 'ascii' codec can't encode character u'\xe7' in position 11: ordinal not in range(128)
```

beardypig: Good spot. This should fix it. 
```suggestion
        log.debug(u"Using the '{0}' stream variant".format(variantname))
```

beardypig: Fixed it on the error log too.

quantenschaum:
Nice, now selecting the `best` stream consistently returns the stream with the "right" language. But what, if I want to manually select another stream (different language/subtitles)? The `720_alt` or `720p_alt2` are no longer available now.

beardypig:
@quantenschaum as a compromise there is only the primary language stream available right now, to change the language we need an extra option.

quantenschaum:
Yes, please do so. The option `--hls-audio-select` does not have an effect on the arte plugin. How can I download the stream in a different language? Maybe by changing the URL to point the page page in a different language? Is the order of streams different between the different language versions of the page?

PR #3230, opened 2020-10-07
ci: add python 3.9 test runners

bastimeyer:
Python 3.9 has been released yesterday / two days ago and Github has added Python 3.9 builds a few hours ago. This requires a bump of the `actions/setup-python` action to `v2`, since it's not available in the default virtual environments of the CI runners.
- https://docs.python.org/release/3.9.0/whatsnew/3.9.html
- https://github.com/actions/setup-python
- https://github.com/actions/python-versions/commit/1430296346bab896c21a97ff3396425c26ad87b1

----

Python 3.5 is also no longer supported now and should be dropped now once we finally drop 2.7:
- https://blog.python.org/2020/10/python-35-is-no-longer-supported.html

That's currently an overall number of 12 test runners now. A bit much. Dropping 2.7 and 3.5 would reduce it to 8.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -3,7 +3,7 @@ codecov:
     require_ci_to_pass: true
     # wait until at all test runners have uploaded a report (see the test job's build matrix)
     # otherwise, coverage failures may be shown while some reports are still missing
-    after_n_builds: 10
+    after_n_builds: 12
 comment:
   # this also configures the layout of PR check summaries / comments
   layout: "reach, diff, flags, files"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ jobs:
       # when changing the build matrix and changing the number of test runners
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python: [2.7, 3.5, 3.6, 3.7, 3.8]
+        python: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
 #       include:
 #         - python: X.Y-dev
 #           continue: true
@@ -29,7 +29,7 @@ jobs:
     timeout-minutes: 60
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-python@v1
+      - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python }}
       - name: Install dependencies


mkbloke:
Just out of interest... are there any particular reasons as to why Py2 support has yet to be dropped?

Could:
https://github.com/streamlink/streamlink/blob/bf1460c8c0a75026a4824dfb5c32134e6daac293/src/streamlink/session.py#L55-L59

just be altered to state it's unsupported, tell folks to upgrade and then exit, or are there more things that need to be done?

bastimeyer:
> are there any particular reasons as to why Py2 support has yet to be dropped?

I'm not trying to blame anyone here, but nobody except me has been trying to push the py2 removal for months now and I have been ignored for at least 4 times now :disappointed:. I can't do this alone though and don't want to be responsible for potential issues afterwards. We have a PR for removing py2 from the CI and setup.py, and one or two for removing compatibility related and deprecated stuff, but those are outdated now and either need an update or someone needs to open a new PR. I will probably do this myself soon though, because I really want this to be finally over and be done with.

> state it's unsupported, tell folks to upgrade and then exit

Simply exiting the program on py2 isn't enough and also shouldn't be done. Dropping py2 will require a major version bump and the removal from setup.py. Everything else is optional. However, the idea was to use that opportunity and remove APIs and features deprecated by Streamlink and make some other changes which would otherwise require a major version bump, but that is a bit of additional work. The overall motivation for making changes is unfortunately a bit low at the moment as it seems. Maybe with Hacktoberfest, we will see some more activity.

And since the preparation for a major version bump will take some time, this will conflict with the regular plugin bugfixing that may require a patch release, like the Twitch plugin for example. Developing on two branches simultaneously is stupid for a project this size with the current number of active maintainers, which means it's not that simple to organize.

Then there are also some projects which are using Streamlink and are still stuck on py2, I believe. That shouldn't be our problem though.

gravyboat:
> Just out of interest... are there any particular reasons as to why Py2 support has yet to be dropped?

@mkbloke  It's a lot of work and I am hesitant to spend even more time on Streamlink than I already do. As @bastimeyer noted they're the only one that has really been pushing it with PRs. If you just want to update those PRs let's do it Bastimeyer, before we were waiting but screw it. If it breaks it breaks, let's just get it done for those existing PRs.

I'm not concerned about any project that is on Python 2. We've been supporting Python 3 for multiple years now, people should upgrade. If their app breaks and they're locked to an old release that is out of our hands, it's open source, they can go back to the prior commit and fork if they need to.

gravyboat:
This looks good so I'm merging it @bastimeyer.

bastimeyer:
Btw, this adds about 10 seconds of setup time to the runners, as the python envs need to be downloaded and extracted now.

> Maybe with Hacktoberfest, we will see some more activity.

Also just wanted to mention that I've added the `hacktoberfest` tag to the repo yesterday, because that's apparently required now unless a `hacktoberfest-accepted` label gets set on an individual PR. PRs still need to be merged or approved by one of the project's maintainers in order to count towards the 4 PR goal.

gravyboat:
@bastimeyer I'm not sure if we should add the label, there were a lot of reports last week of extremely low quality PRs this year and we may get spammed with those. I guess we'll see what happens.

mkbloke:
Thanks for the responses.  I'm not a Python expert, as I am sure is evident, but I'll be looking into this a bit more shortly.  I'm going to digest #2896, #2919 and #2920 and then look at how I *might* be able to help.  At the very least, even if I can't do the really clever stuff, I can perhaps do some of the tedious donkey-work.

PR #3234, opened 2020-10-08
plugins.dlive: remove plugin

Billy2011:
- google recaptcha protected
- dlive.tv is on youtube: [DLive.tv](https://www.youtube.com/channel/UCmPXtUceJsdYe5LcL2M5PDA)


bastimeyer:
Codecov still seems to be totally broken after their outage yesterday. I've re-triggered the latest commit on master (and with that accidentally made it fail in the upload docs step) and the PR commit, but it still shows garbage data and makes the checks fail here.

@gravyboat I think we should change the codecov config and set the threshold of all checks to 100%, so that it can never fail. This is too annoying to deal with.

Apologies for this nonsense...

----

Regarding the plugin removal, I don't see any captcha protection.

The HLS playlist URLs are actually very simple to obtain and a plugin fix looks rather straightforward:
`https://live.prd.dlive.tv/hls/live/{username}.m3u8?web=true`

Billy2011:
The URL has changed: "https://playback.prd.dlivecdn.com/live/pdp/1556554403/vod.m3u8?web=true"

It's not possible for me to obtain the username **pdp** and the following tag **1556554403**

bastimeyer:
There is a GQL API that can be used to retrieve the URLs of past broadcasts, highlights and clips:
`https://graphigo.prd.dlive.tv/`

Billy2011:
I see no way to use it.

Billy2011:
> Regarding the plugin removal, I don't see any captcha protection.

It's an invisible vue-recaptcha.

back-to:
livestream

```python
import json
import requests
username = ""
data=json.dumps({'query': '''query {{
  userByDisplayName(displayname:"{0}") {{
    username
  }}
}}'''.format(username)})
res = requests.post('https://graphigo.prd.dlive.tv/', data=data).text
print(res)
```

or

https://github.com/ytdl-org/youtube-dl/blob/master/youtube_dl/extractor/dlive.py#L59-L97

gravyboat:
@bastimeyer I'm fine with changing the codecov threshold. There's nothing to apologize for. I don't anyone anticipated it would work so poorly.

Billy2011:
@back-to, thanks, I only saw the vue-recaptcha and then did not expect that you can handle it so easily:)

PR #3227, opened 2020-10-06
plugins.twitch: refactor plugin

bastimeyer:
Opening this as a draft PR for now, because it's still WIP.
As already said in #3206, I wanted to refactor the entire Twitch plugin, as it hasn't been done in years and there's a ton of cruft.

I've split this up into multiple commits for reviewing reasons, but this should be squashed if it gets merged. Commits are therefore probably not 100% optimal and also don't have a proper commit message.

I can rebase this into fewer proper commits though, if you feel like it's too many changes for a single commit if squashed.

----

**My goals with the code-refactor are the following:**

1. Move most from the module's global context to class attributes, where it belongs
2. Remove old and now unsupported stuff (also affects plugin's URL regex)
   - Remove different video types
     [There's only one type now](https://dev.twitch.tv/docs/v5/reference/videos)
   - Remove VOD playlists
     Removed in favor of ["collections"](https://dev.twitch.tv/docs/v5/reference/collections) in 2017, but doesn't matter, since Streamlink can output only a single stream anyway and videos in a collection have their own canonical URL.
   - Remove FLV stuff
     Which year is it again?
3. Improve Twitch plugin class
   - Split up `_get_hls_streams` and `_access_token` by stream type ("live" and "video")
     Simplifies logic and removes dead code paths
   - Improve hosted channel logic
     This is a private API call which can break at any time. It's currently being used on each live stream and doesn't have a try/except block. Also currently calls `_get_hls_streams` multiple times instead of just resolving the hosted channel chain in a loop.
     Now it also automatically updates the metadata on channel switch.
   - Refactor rerun logic and its tests
   - Refactor metadata code (already done that in #3223)
     This PR moves it to its correct position within the plugin class.
   - WIP: Improve main API calls
     Validation schemas should be added and redundancies removed.
   - Various minor changes
4. Improve TwitchAPI class
   - Remove unused API calls and API versions/subdomains
     Keep the new Twitch API in mind
   - Add GQL API calls
   - WIP: move validation schemas here and add methods for each request type, not for each API endpoint
4. Add tests
   At least for the stuff that doesn't require mocking every network request

PR #3237, opened 2020-10-09
tests: fix early writer close in stream_hls mixin 

bastimeyer:
This should fix the following issue in #3227:
https://github.com/streamlink/streamlink/runs/1227523417?check_suite_focus=true#step:5:257

----

Two issues fixed in one PR (don't squash):

1. Sometimes the writer thread was too fast for the reader thread of the stream_hls test mixin to read, and it closed before the first read occured. The result was an empty read, since the buffer got closed by the writer thread too early. This PR makes the writer thread wait before closing until the reader thread has at least attempted to read once.

2. Currently, Streamlink doesn't wait for the spawned threads of the `ThreadPoolExecutor` to shut down and immediately clears all queued thread pool threads globally. This is not a particularly graceful way to close the writer thread:
   https://github.com/streamlink/streamlink/blob/cb5cee6109b98e61f98b42967185a379a4748e0a/src/streamlink/stream/segmented.py#L107-L109

   In Python 3.9, the `cancel_futures` parameter was just added to the `ThreadPoolExecutor.shutdown` method, which only clears the queue of the current ThreadPoolExecutor instance:
https://github.com/python/cpython/blame/3.9/Lib/concurrent/futures/thread.py#L210-L229

   This can be backported easily, so that the current ungraceful method can be removed. Also setting `wait=True` here will ensure that the writer thread has fully closed. For the HLS tests, this is a needed fix, and for streamlink_cli, I believe this won't make a difference, since it doesn't join the daemon thread anyway, but I'm not sure.

### DIFF
--- a/src/streamlink/stream/segmented.py
+++ b/src/streamlink/stream/segmented.py
@@ -1,13 +1,38 @@
-import concurrent.futures.thread
-import logging
 from concurrent import futures
+from concurrent.futures.thread import ThreadPoolExecutor
+import logging
 from threading import Thread, Event
+from sys import version_info
 
 from .stream import StreamIO
 from ..buffers import RingBuffer
 from ..compat import queue
 
 log = logging.getLogger(__name__)
+
+
+class CompatThreadPoolExecutor(ThreadPoolExecutor):
+    if version_info < (3, 9):
+        def shutdown(self, wait=True, cancel_futures=False):
+            with self._shutdown_lock:
+                self._shutdown = True
+                if cancel_futures:
+                    # Drain all work items from the queue, and then cancel their
+                    # associated futures.
+                    while True:
+                        try:
+                            work_item = self._work_queue.get_nowait()
+                        except queue.Empty:
+                            break
+                        if work_item is not None:
+                            work_item.future.cancel()
+
+                # Send a wake-up to prevent threads calling
+                # _work_queue.get(block=True) from permanently blocking.
+                self._work_queue.put(None)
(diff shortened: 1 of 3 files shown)


bastimeyer:
@beardypig Could you please take a look at the second commit of this PR? I've just noticed that you've added d285e3a09e13d92c7acb285fca38b3e2a4ec0f4f two and a half years ago which set `wait=False` in order to improve the process exit time during a `SIGINT` and also removed the thread.join calls of the worker and writer threads.

Just in case you've missed the reason for why I'm adding this back: this messes up or has been messing up the HLS integration tests (there are a lot more now) and the mocked network calls. I've written a mixin for these integration tests a few weeks ago which helped setting up the tests and also made sure that all spawned threads shut down on each test's tear-down. The removal of the thread.join calls isn't a problem because of this, but the ThreadPoolExecutor is the last remaining bit of this inconsistency, and this gets fixed by this PR.

If I had gone through the git blame I could've seen your old commit way sooner and it would have caused way less trouble debugging this. Damn... :disappointed: 

In theory, the `wait=True` parameter should at most cause a delay of the max segment download time, if there's a network issue. But since the worker thread doesn't get joined into the main thread when closing the stream reader, the KeyboardInterrupt should still work fine on the main thread, right?

beardypig (approved):
Looks good to me. I had also noticed this instability in the tests and I was going to take a look - but you beat me to it.

gravyboat:
Merging.

PR #3238, opened 2020-10-09
ci.codecov: remove additional coverage checks

bastimeyer:
The codecov coverage data is currently totally broken, because of the bug I've reported here:
https://community.codecov.io/t/bug-project-path-prefix-of-codecov-action-running-on-windows-is-incorrectly-kept-in-report-data/2141

This unfortunately also affects the latest commit of the master branch and PRs are failing because of the incorrect coverage base.

For now, I've removed the codecov "default project" status check from the master branch protection list, so that the PR checks at least don't stall.

If this bug gets fixed by codecov, we could theoretically keep the current config with the split up projects / PR statuses, but since it hasn't worked that great so far, I think reducing it to the default and having an overall coverage status report there (instead of the bot comment) is enough for now. I would have preferred having reports on the individual PR statuses, but that doesn't work currently.

This change will also make the coverage status always succeed. I've set both the threshold to 100% and informational to true.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -16,28 +16,9 @@ coverage:
   status:
     changes: false
     patch: false
-    # split up coverage reports by path
-    # don't set coverage targets and instead set coverage thresholds
     project:
-      # we can't disable the overall default project, because only this will have PR check summaries / comments
-      # this replaces the PR comment
+      # the "default project" PR status check replaces the PR comment with the coverage change message
+      # set threshold to 100% and informational to true, so that the check always passes
       default:
-        target: 50
-      streamlink:
-        threshold: 1
-        paths:
-          - "src/streamlink/"
-          - "!src/streamlink/plugins/"
-      streamlink_cli:
-        threshold: 1
-        paths:
-          - "src/streamlink_cli/"
-      plugins:
-        # don't set a threshold on plugins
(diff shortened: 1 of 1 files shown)


bastimeyer:
The bug on codecov is allegedly fixed now. Restarting the jobs of the latest commit on master unfortunately didn't remove the false additional data, so this has to be fixed in the next commit on the master branch. The failed status is still there because of the docs deploy issue that I have a fix for in the other PR.

Once we have another commit on the master branch, the PRs with wrong data need to be rebased if there are still coverage failures. Otherwise a simple restart will do the job.

This PR however should be good to go if nobody disagrees with the codecov config simplification.

gravyboat:
@bastimeyer Do you still want to merge this? They claim to have fixed the bug in your issue report.

bastimeyer:
This PR was not about the bug, it was about reducing the config and removing the split up PR checks by path and making the code coverage check always succeed. But let's keep it for now... Should there be any future issues, I'll re-open this one, so that we have a config that doesn't make PRs fail because of this nonsense.

There is another bug though and codecov currently doesn't create PR checks, which means the summary similar to the old bot comment is currently missing:
https://community.codecov.io/t/github-checks-api-and-annotation-not-working/2142

PR #3245, opened 2020-10-12
docs: add basic development setup guide

bastimeyer:
As mentioned in #3244, this adds a **basic** guide for setting up a dev environment and running tests, etc. locally. In the future, this can be extended with guides on how to write plugins, but for now, this is enough.

And of course the pull request template with a link to the new docs section.

Happy to discuss any changes/improvements/additions...

PR #3249, opened 2020-10-13
docs: add plugin addition/removal infos

mkbloke:
~~This adds what I was talking about in: https://github.com/streamlink/streamlink/pull/3245#issuecomment-707386309~~

Edit: I made this a draft PR, as there may well be things that should be changed...

### DIFF
--- a/docs/developing.rst
+++ b/docs/developing.rst
@@ -114,3 +114,22 @@ performing these checks locally avoids unnecessary build failures.
     # build the documentation
     make --directory=docs clean html
     $BROWSER ./docs/_build/html/index.html
+
+
+Plugins
+-------
+
+Adding plugins
+^^^^^^^^^^^^^^
+
+1. Implement the plugin in ``src/streamlink/plugins/``, similar to already existing plugins. Check the git log for recently
+   added or modified plugins to help you get an overview of what's needed to properly implement a plugin. A complete guide
+   is currently not available.
+2. Add at least tests for the URL regex matching in ``tests/plugins/``. Once again, check other plugin tests from the git log.
+
+Removing plugins
+^^^^^^^^^^^^^^^^
+
+1. Remove the plugin file in ``src/streamlink/plugins/`` and the test file in ``tests/plugins/``
+2. Remove the plugin entry from the documentation in ``docs/plugin_matrix.rst``
+3. Add an entry to ``src/streamlink/plugins/.removed``


bastimeyer:
I appreciate your efforts, but you are over-complicating things here.

Two simple checklists are more than enough for "adding" and "removing" plugins. A "modifying" plugins section is not needed as it's already implied by the "adding" section.

The headlines are also too long and won't get rendered properly in the menu and the bottom table is too wide as well (and if it was a good idea to have such a table, its rows and columns would need to be flipped).

----

See how much simpler this example is:

```rst
Plugins
----------------

Removing plugins
^^^^^^^^^^^^^^^^

1. Remove the plugin file in ``src/streamlink/plugins/`` and the test file in ``tests/plugins/``
2. Remove the plugin entry from the documentation in ``docs/plugin_matrix.rst``
3. Add an entry to ``src/streamlink/plugins/.removed``

Adding plugins
^^^^^^^^^^^^^^

1. Implement the plugin in ``src/streamlink/plugins/``, similar to already existing plugins. Check the git log for recently added or modified plugins to help you get an overview of what's needed to properly implement a plugin. A complete guide is currently not available.
2. Add at least tests for the URL regex matching in ``tests/plugins/``. Once again, check other plugin tests from the git log.
```

mkbloke:
A fair point, it does look overcomplicated next to what you've suggested.

Shall I change this PR to use what you've suggested or just close it?  I'm happy either way.  Cheers.

bastimeyer:
Just update it, no need to close it. Thanks.

docs/developing.rst
```diff
@@ -114,3 +114,20 @@ performing these checks locally avoids unnecessary build failures.
...
+
+1. Remove the plugin file in ``src/streamlink/plugins/`` and the test file in ``tests/plugins/``
+2. Remove the plugin entry from the documentation in ``docs/plugin_matrix.rst``
+3. Add an entry to ``src/streamlink/plugins/.removed``
+
+Adding plugins
+^^^^^^^^^^^^^^
+
+1. Implement the plugin in ``src/streamlink/plugins/``, similar to already existing plugins. Check the git log for recently added or modified plugins to help you get an overview of what's needed to properly implement a plugin. A complete guide is currently not available.
+2. Add at least tests for the URL regex matching in ``tests/plugins/``. Once again, check other plugin tests from the git log.
```
bastimeyer, line 133: Max line length defined by the editorconfig is 128:
https://github.com/streamlink/streamlink/blob/8db5786a4f8422688d8618685cc30057a2fa423c/.editorconfig#L9

bastimeyer (changes requested):

docs/developing.rst
mkbloke: Oops, sorry, I had not realised.  I didn't check it before with the docs build either, but this time I have.

docs/developing.rst
```diff
@@ -114,3 +114,22 @@ performing these checks locally avoids unnecessary build failures.
     # build the documentation
     make --directory=docs clean html
     $BROWSER ./docs/_build/html/index.html
+
+
+Plugins
+-------
+
+Removing plugins
```
back-to, line 122: might be more useful if you use **Adding** first instead of **Removing**

PR #3250, opened 2020-10-14
chore: update issue templates

mkbloke:
I don't know if it'll help or perhaps you don't think it's that much of a problem...  It is, however, nice to get log output in text, if only for copying the URL, rather than having to manually type it in if the URL has not been specified in text form elsewhere in the issue.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -52,6 +52,8 @@ https://streamlink.github.io/cli.html#cmdoption-l
 
 Make sure to **remove usernames and passwords**
 You can copy the output to https://gist.github.com/ or paste it below.
+
+Don't post screenshots of the log output and instead copy the text from your terminal application.
 -->
 
 ```
@@ -59,7 +61,7 @@ REPLACE THIS TEXT WITH THE LOG OUTPUT
 ```
 
 
-### Additional comments, screenshots, etc.
+### Additional comments, etc.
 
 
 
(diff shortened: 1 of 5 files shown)


.github/issue_template.md
```diff
@@ -47,6 +47,8 @@ TEXT LOG OUTPUT IS OPTIONAL for generic issues!
 However, depending on the issue, log output can be useful for the developers to understand the problem. It also includes information about your platform. If you don't intend to include log output, please at least provide these platform details.
 Thanks!
 
+Although not essential, if you choose to include log output, it is more helpful to the developers if you can copy and paste the text from your terminal/command window into the issue, rather than providing a screenshot.
+
```
bastimeyer, line 51: This is in theory already covered by
https://github.com/streamlink/streamlink/blame/50c8cc2f6416083aab61b237e00fe50e54c5b8e3/CONTRIBUTING.md#L22
but including it directly in the issue template may be better. I'd prefer a short single sentence though, since the template is already quite long and "daunting".

Adding

> Don't post screenshots of the log output and instead copy the text from your terminal application.

to the end of the comment block is better. Maybe also remove "screenshots" from the tier-3 headline down below.

bastimeyer (changes requested):
There's more than one issue template. If we add this to the generic template, then this needs to be added to the other ones as well.
https://github.com/streamlink/streamlink/tree/50c8cc2f6416083aab61b237e00fe50e54c5b8e3/.github/ISSUE_TEMPLATE

.github/issue_template.md
mkbloke: OK, yes, shorter is better.

Yes, removing the screenshot section seems like a great idea.  I was never sure about the utility of screenshots for a text based application.

I'll make the changes to all the relevant templates and push again soon.  Cheers.

bastimeyer:
Thanks!

PR #3251, opened 2020-10-15
plugins.cbsnews: support for live streams from CBS News

beardypig:
Support for live streams from CBS News.

Including:
- https://www.cbsnews.com/live/
- https://www.cbsnews.com/live/cbsn-local-bay-area/
- https://www.cbsnews.com/live/cbsn-local-ny/

### DIFF
--- a/src/streamlink/plugins/cbsnews.py
+++ b/src/streamlink/plugins/cbsnews.py
@@ -1,0 +1,38 @@
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import validate
+from streamlink.stream import HLSStream
+from streamlink.utils import parse_json
+
+
+class CBSNews(Plugin):
+    url_re = re.compile(r"https://www\.cbsnews\.com/live/")
+    _re_default_payload = re.compile(r"CBSNEWS.defaultPayload = (\{.*)")
+
+    _schema_items = validate.Schema(
+        validate.transform(_re_default_payload.search),
+        validate.any(None, validate.all(
+            validate.get(1),
+            validate.transform(parse_json),
+            {"items": [validate.all({
+                "video": validate.url(),
+                "format": "application/x-mpegURL"
+            }, validate.get("video"))]},
+            validate.get("items")
+        ))
+    )
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls.url_re.match(url) is not None
+
+    def _get_streams(self):
+        items = self.session.http.get(self.url, schema=self._schema_items)
+        if items:
+            for hls_url in items:
+                for s in HLSStream.parse_variant_playlist(self.session, hls_url).items():
+                    yield s
+
+
+__plugin__ = CBSNews
--- a/tests/plugins/test_cbsnews.py
+++ b/tests/plugins/test_cbsnews.py
@@ -1,0 +1,24 @@
+import unittest  # noqa: F401
+import pytest
+from streamlink.plugins.cbsnews import CBSNews
+
+
+class TestCBSNews:
+    valid_urls = [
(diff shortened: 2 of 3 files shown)


src/streamlink/plugins/cbsnews.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import print_function
```
bastimeyer, line 1: There is no `print` call in the module, so the `__future__` import is unnecessary.

src/streamlink/plugins/cbsnews.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def _get_streams(self):
+        res = self.session.http.get(self.url)
+        _re_json = re.compile(r"CBSNEWS.defaultPayload = (\{.*)")
+        m = _re_json.search(res.text)
+        if m:
+            data = parse_json(m.group(1))
+            for i in data['items']:
+                for s in HLSStream.parse_variant_playlist(self.session, i['video']).items():
+                    yield s
```
bastimeyer, line 25: The HLS URL should ideally be extracted via a validation schema.

```suggestion
_re_default_payload = re.compile(r"CBSNEWS.defaultPayload = (\{.*)")

_schema_items = validate.Schema(
    _re_default_payload.search,
    validate.any(None, validate.all(
        validate.get(1),
        validate.transform(parse_json),
        {"items": [validate.all({
            "video": validate.url(),
            "format": "application/x-mpegURL"
        }, validate.get("video"))]},
        validate.get("items")
    ))
)

def _get_streams(self):
    items = self.session.get(self.url, schema=self._schema_items)
    if items:
        for hls_url in items:
            for s in HLSStream.parse_variant_playlist(self.session, hls_url).items():
                yield s
```

Also please don't mix quotation styles.

tests/plugins/test_cbsnews.py
```diff
@@ -0,0 +1,24 @@
+import unittest  # noqa: F401
+import pytest
+from streamlink.plugins.cbsnews import CBSNews
+
+
+class TestCBSNews:
```
bastimeyer, line 6: What is the reason for not using `unittest.TestCase`?
We shouldn't be mixing test setups. Almost all tests are written with unittest and pytest is only used as the test runner. Mixing this makes it much more difficult to read and maintain.

bastimeyer (changes requested):

tests/plugins/test_cbsnews.py
beardypig: This has been used previously; we have `pytest` available and it has superior test builder functions.

src/streamlink/plugins/cbsnews.py
beardypig: I did not feel it useful or necessary to use the validation methods.  
Bit nit-picky on the quotes, but OK.

bastimeyer (changes requested):
Sorry about the previous diff. I had only written it as a rather simple comment without actually testing it and it was not meant to be copied as a whole, just a suggestion.

src/streamlink/plugins/cbsnews.py
```diff
@@ -0,0 +1,38 @@
...
+from streamlink.stream import HLSStream
+from streamlink.utils import parse_json
+
+
+class CBSNews(Plugin):
+    url_re = re.compile(r"https://www\.cbsnews\.com/live/")
+    _re_default_payload = re.compile(r"CBSNEWS.defaultPayload = (\{.*)")
+
+    _schema_items = validate.Schema(
+        _re_default_payload.search,
```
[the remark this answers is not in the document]
beardypig: Ah yeah, already fix that but didn't push it :sigh:

bastimeyer (approved):

PR #3259, opened 2020-10-16
plugins.urplay: add UR Play plugin

Tigge:
This adds support for the free and independent educational service UR's (Utbildningsradion) play site -  <http://urplay.se>. Mostly this is programs in Swedish, but some shows in different languages exist as well. The content might be geo-locked.

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -237,6 +237,7 @@ tvrplus                 tvrplus.ro           Yes   No    Streams may be geo-rest
 tvtoya                  tvtoya.pl            Yes   --
 twitcasting             twitcasting.tv       Yes   No
 twitch                  twitch.tv            Yes   Yes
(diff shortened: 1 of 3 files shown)


gravyboat:
These don't appear to be live streams but pre-recorded content?

Tigge:
Yeah, the content is live streamed on svtplay.se or sr.se usually. But on this site there is usually no live streamed content that I know of. But there seems to be one or two content only plugins in streamlink if I look at the documentation matrix. 

But feel free to close if this type of site us not allowed 😊.

bastimeyer:
Closing, as the plugin doesn't meet the requirements, sorry:
https://github.com/streamlink/streamlink/blob/72eee9f41594c1b063e4380162190228bee6df2e/CONTRIBUTING.md#plugin-requests

> 6. Sites which don't provide any real live streaming content, eg. only VODs or VODs being rebroadcasted

PR #3264, opened 2020-10-17
release: 1.7.0

### LINKED ISSUE #3258: Release 1.7.0
I'd like to have a new release this weekend, as there are important changes on the master branch for a few weeks now which should be published as a stable release.

Do we have any PRs which need to be merged first? Any other blockers or important changes still missing?

After this next release, the py2 removal and preparation for `2.0.0` can begin (#3235).

bastimeyer:
Release date is already set for tomorrow, since I don't expect a merge within the next 3 hours.

I had to manually clean up the git shortlog, since commit 50c8cc2f6416083aab61b237e00fe50e54c5b8e3 has a malformed commit message. It was probably just a mistake and everybody is aware of the basic format (and naming scheme), but please take a better look at the commit messages before merging and double-check. It's all fine though and not too much of a big deal.

Closes #3258

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,84 @@
 # Changelog
 
+## streamlink 1.7.0 (2020-10-18)
+
+Release highlights:
+
+- Added: new plugins for micous.com, tv999.bg and cbsnews.com
+- Added: new embedded ad detection for Twitch streams ([#3213](https://github.com/streamlink/streamlink/pull/3213))
+- Fixed: a few broken plugins and minor plugin issues (see changelog down below)
+- Fixed: arguments in config files were read too late before taking effect ([#3255](https://github.com/streamlink/streamlink/pull/3255))
+- Fixed: Arte plugin returning too many streams and overriding primary ones ([#3228](https://github.com/streamlink/streamlink/pull/3228))
+- Fixed: Twitch plugin error when stream metadata API response is empty ([#3223](https://github.com/streamlink/streamlink/pull/3223))
+- Fixed: Zattoo login issues ([#3202](https://github.com/streamlink/streamlink/pull/3202))
+- Changed: plugin request and submission guidelines ([#3244](https://github.com/streamlink/streamlink/pull/3244))
+- Changed: refactored and cleaned up Twitch plugin ([#3227](https://github.com/streamlink/streamlink/pull/3227))
+- Removed: `platform=_` stream token request parameter from Twitch plugin (again) ([#3220](https://github.com/streamlink/streamlink/pull/3220))
+- Removed: plugins for itvplayer, aljazeeraen, srgssr and dingittv
+
+
+```text
+Alexis Murzeau <amubtdx@gmail.com> (1):
+      docs: use recommonmark as an extension
+
+Billy2011 <kschmidt2007@googlemail.com> (3):
+      plugins.zattoo: use hello api v2 for zattoo.com (#3202)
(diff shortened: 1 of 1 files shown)


gravyboat:
What was the issue with the format, it looked like this:
```
plugins.tv3cat: update URL match, test and plugin matrix

- allow http and https schemes in URL match
- escape literal dots in URL match
- remove unnecessary pass statement in except
- update test for both http and https
- update plugin matrix with correct domain
```

First line should have been taken and interpreted properly no?

bastimeyer:
It doesn't have two NL characters after the header/summary, so the entire body is part of the header/summary.

```
$ git show --oneline -s 50c8cc2f6416083aab61b237e00fe50e54c5b8e3
50c8cc2f plugins.tv3cat: update URL match, test and plugin matrix - allow http and https schemes in URL match - escape literal dots in URL match - remove unnecessary pass statement in except - update test for both http and https - update plugin matrix with correct domain
```

gravyboat:
Ahh okay I totally missed that, my bad.

gravyboat:
It's been 3 hours so I'm merging. Will also tag.

mkbloke:
Should I be making sure there's a blank line between the header and summary in the commit message then?  I'm not sure I've ever been doing that for any of my commits, but I can't seem to recall now...

gravyboat:
@mkbloke Yep that would take care of it!

mkbloke:
@gravyboat, cool, thanks and noted for the future!

bastimeyer:
> Should I be making sure there's a blank line between the header and summary in the commit message then? I'm not sure I've ever been doing that for any of my commits, but I can't seem to recall now...

These are the very basic rules of basically every organized git project:

- https://www.git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines
- https://chris.beams.io/posts/git-commit/

Optionally, the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) scheme can be used when writing commit messages. We've already been doing this partially with the `component.subcomponent: summary` format, similar to the Linux kernel's commit message format but with python-style module names. IMO, the CC scheme is too strict and if you adopt it, custom scopes need to be added anyway depending on the project, so just using the name of the changed component as commit message scope is a way simpler method.

We could add a commit message linter to the CI runners, but I feel like this is a bit too much and would fail in lots of cases where people just mindlessly push changes to PRs without proper commit messages, and this makes submitting PRs more complicated for new contributors or beginners. It would catch malformed message though which Github still displays as correctly formatted messages for some reason and thus can confuse project maintainers, like it's been the case here.

bastimeyer:
Btw, vim has a built-in formatter (at least on Arch Linux via `vim-runtime` - and there are also lots of plugins) for commit messages, rebase todo lists, etc., and it prevents you from misformatting commit messages.

Lots of other frontends for git have this as well. When I need a git GUI, I'm using SmartGit (proprietary, but free/gratis for FOSS projects), because it's feature-rich, simple and has a nice layout for the git index and the commit history graph.

mkbloke:
@bastimeyer, thanks for the links and info.  I have read them all, and some pages linked from those like the Angular conventions.  I need to improve my git foo a bit and will try to work my way through Pro Git at some point, although for now I'm generally comfortable with the git workflow for the stuff I'm doing here.  I've been trying to generally follow what seems to be accepted conventions for commit messages in Streamlink, but I'll try to improve on them too.  Cheers.

PR #3265, opened 2020-10-18
plugins: update logging from self.logger.* to log.*

mkbloke:
- add: import logging where required
- add: log = logging.getLogger(__name__) where required
- change: instances of self.logger.* to log.*

Also (only related to some plugins with logging changes):
- reordered imports in a few plugins to alphabetical order/removed unnecessary blank lines
- fixed linting errors

### DIFF
--- a/src/streamlink/plugins/akamaihd.py
+++ b/src/streamlink/plugins/akamaihd.py
@@ -1,9 +1,12 @@
+import logging
 import re
 
 from streamlink.plugin import Plugin
 from streamlink.plugin.plugin import parse_url_params
 from streamlink.stream import AkamaiHDStream
 from streamlink.utils import update_scheme
+
+log = logging.getLogger(__name__)
 
 
 class AkamaiHDPlugin(Plugin):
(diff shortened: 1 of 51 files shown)


src/streamlink/plugins/akamaihd.py
```diff
@@ -18,7 +21,7 @@ def _get_streams(self):
         urlnoproto = self._url_re.match(url).group(1)
         urlnoproto = update_scheme("http://", urlnoproto)
 
-        self.logger.debug("URL={0}; params={1}", urlnoproto, params)
+        log.debug("URL={0}; params={1}", urlnoproto, params)
```
bastimeyer, line 24: I'm not going to mark every log call here, but the log message strings should ideally be formatted directly, so that they can be automatically translated into f-strings later on.

See the commit 001a8fab57f623116cc1dfc5e0786a1d713d7b9d which I had added to my logger removal branch after removing the deprecated logger class:
https://github.com/bastimeyer/streamlink/compare/py2-compat-removal...bastimeyer:deprecations/logger-removal
It is basically the same as this PR, but since you've already opened it, I will remove it from my branch when I open a PR for this.

mkbloke: Thanks for the info.  Let's keep your branch though, as it's complete and this is not.  It's not worthwhile for me to make the extra changes if you've already done it.  Cheers.

PR #3266, opened 2020-10-18
docs: remove entries for unmaintained Windows portable version

mkbloke:
It seems like the `streamlink/streamlink-portable` repo is dead for now.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -397,18 +397,10 @@ Windows portable version
 ==================================== ===========================================
 Maintainer                           Links
 ==================================== ===========================================
-RosadinTV                            `Latest precompiled stable release`__ |br|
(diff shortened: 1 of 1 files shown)


back-to:
might be unmaintained, but it still works for streamlink 1.7.0

there is no need to remove it

mkbloke:
Am I missing something here?  The link to https://github.com/streamlink/streamlink-portable/releases does not seem to contain any stable release since 1.0.0?

back-to:
> Am I missing something here?

Yes, you did not install it.

>  does not seem to contain any stable release since 1.0.0?

unimportant, the software can automatically download the Github version of Streamlink such as 1.7.0

**I did test it with 1.7.0 and it works.**

PR #3269, opened 2020-10-19
chore: drop support for Python 3.5

bastimeyer:
- Bump Python version requirement to 3.6
- Update pypi package classifiers
- Update docs
- Update CI test runners and codecov config


bastimeyer:
Unless there are any concerns, please go ahead and merge. Otherwise I'm doing it myself in a couple of hours. No reason to wait. This is currently blocking #3270, which in turn is blocking another PR I'd like to open once merged.

PR #3270, opened 2020-10-19
chore: remove py2 compat stuff

bastimeyer:
I've split up everything into logical chunks, so when reviewing the changes, please check the individual commits. This looks like much, but it's all rather simple.

In addition to removing the compat imports, I've also corrected the import order in some modules, but not all of them. Only where it made the most sense.


bastimeyer:
I will rebase this once #3269 got merged, so that the checks will all pass.

back-to (approved):

bastimeyer:
Forgot to remove one remaining thing from the compat module. Should be good now...

Issue #3271, opened 2020-10-19
[Facebook] No playable streams found on this URL.


DawgNewb:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.

### Description

Feel like they changing something for the last couple day. I couldn't do anything with facebook.

### Reproduction steps / Explicit stream URLs to test

1. Copy plain live's facebook link or changing from facebook.com/watch/live/?v=xxxxxxxxxxx&ref=watch_permalink to normal video link.
2. Usually even without any command it will show me the resolution.
3. Now it's only "No playable streams found on this" debug log below

### Log output

```
[Streamlink for Windows v1.7.0 - Git 8bc84d0]
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.6.3
[cli][debug] Streamlink: 1.7.0+8bc84d0
[cli][debug] Requests(2.18.4), Socks(1.6.7), Websocket(0.44.0)
[cli][info] Found matching plugin facebook for URL facebook.com/xxxxxxx/videos/xxxxxxxxx/
[plugin.facebook][debug] No meta og:video:url
[plugin.facebook][debug] Falling back to playlist regex
[plugin.facebook][debug] Falling back to tahoe player
[plugin.facebook][debug] No meta og:video:url
error: No playable streams found on this URL: facebook.com/xxxxxxx/videos/xxxxxxx/
```

### Additional comments, etc.

Love yall it's been 2 years since I last report fb issues.

I'm not testing with COOKIES yet if you could tell me about which one I need would be great! [sb datr c_user spin wd xs fr]
But so far as long as I can run the vid without my account I guess it wouldn't need cookie.
Thank you.

back-to:
it should work when you remove `self.session.http.headers.update({'User-Agent': useragents.CHROME})` from the file

https://github.com/streamlink/streamlink/pull/3272

```
streamlink https://www.facebook.com/Firerecordingstudio/videos/405703290823328/
[cli][debug] OS:         Linux-5.4.0-51-generic-x86_64-with-glibc2.29
[cli][debug] Python:     3.8.5
[cli][debug] Streamlink: 1.7.0+2.gdf6ab12.dirty
[cli][debug] Requests(2.24.0), Socks(1.7.1), Websocket(0.57.0)
[cli][info] Found matching plugin facebook for URL https://www.facebook.com/Firerecordingstudio/videos/405703290823328/
[...]

PR #3272, opened 2020-10-19
plugins.facebook: remove User-Agent

back-to:
Fixes https://github.com/streamlink/streamlink/issues/3271

### DIFF
--- a/src/streamlink/plugins/facebook.py
+++ b/src/streamlink/plugins/facebook.py
@@ -4,7 +4,6 @@ import re
 from urllib.parse import unquote_plus, urlencode
 
 from streamlink.plugin import Plugin
-from streamlink.plugin.api import useragents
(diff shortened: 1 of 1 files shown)


evanfrawley:
Is there an ETA when this would be put in a release? Would love to get this change into production!

bastimeyer:
If you take a look at the master branch, you'll see that work is being done on a major version bump and the `2.0.0` release, but that's nowhere near ready yet, so all plugin fixes that come along the way will need to wait for this to finish first, as we decided to not deal with multiple branches.

As usual, install Streamlink from the master branch if you can't wait until the next release.
https://streamlink.github.io/install.html#pypi-package-and-source-code

evanfrawley:
😄 awesome, thank you @bastimeyer for the help

PR #3274, opened 2020-10-19
logger: replace all invalid log calls

bastimeyer:
Lots of changes today... (sorry :innocent:)

----

First commit replaces the `self.logger` calls in all plugin modules with proper log calls. The `self.logger` calls in the `Plugin` base class can't be replaced, as the `logger` attribute is dynamic and based on the module name.

I've committed these changes last week and unfortunately forgot to use f-strings instead of string.format here. Doesn't matter though as there are countless of other string.format calls which can be automatically translated.

----

The second commit (hopefully) fixes all remaining log calls throughout the entire project which add custom `LogRecord` args via `log.info("foo {0}", bar)` by pre-formatting the log messages.

By doing this, the custom `LogRecord` class and `makeRecord` method in the `StreamlinkLogger` class can be removed (see #3273) and Python's default behavior of percent-formatted log messages can be restored.

After this change, the log message format becomes irrelevant for Streamlink, as all log messages should be pre-formatted now. This means that external loggers using Streamlink's root logger can keep using the percent-format syntax and Streamlink doesn't have to deal with the origin of the log message anymore.

This also means that all log calls need to be pre-formatted once #3273 has been (updated and) merged, but this isn't a big deal, especially due to the simplified f-string syntax. Or the percent-style format needs to be used, like this `log.info("foo %s", bar)`. The curly-brace syntax `log.info("foo {0}", bar)` is not going to work anymore.

### DIFF
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -370,8 +370,7 @@ class Plugin(object):
             if match:
                 name = match.group(1)
             else:
-                self.logger.debug("The stream '{0}' has been ignored "
-                                  "since it is badly named.", name)
+                self.logger.debug(f"The stream '{name}' has been ignored since it is badly named.")
                 continue
 
             # Force lowercase name and replace space with underscore.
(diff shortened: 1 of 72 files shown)


gravyboat (approved):
Looks good.

PR #3275, opened 2020-10-20
chore: remove old LIVESTREAMER_VERSION constant

bastimeyer:


back-to:
do we really need this in constant?
it won't change at all and we use most of the time just `__version__`

https://github.com/streamlink/streamlink/blob/603f4b5dc1099099f1dd2090f6ef31e331b1c682/src/streamlink_cli/main.py#L21
https://github.com/streamlink/streamlink/blob/603f4b5dc1099099f1dd2090f6ef31e331b1c682/src/streamlink/plugins/steam.py#L30
https://github.com/streamlink/streamlink/blob/603f4b5dc1099099f1dd2090f6ef31e331b1c682/docs/conf.py#L5
https://github.com/streamlink/streamlink/blob/8034b1606388e325916baf9f8ce738915fa6b528/src/streamlink/session.py#L502-L504

bastimeyer:
> do we really need this in constant?

Probably not. I don't know why it exists.
I'd say we remove it. Importing `__version__` from `streamlink` is the direct and correct way of reading the version string.

PR #3276, opened 2020-10-21
Remove OrderedDict, not needed since Python 3.6

back-to:
`dict` is sorted by default since Python 3.6

Revert `OrderedDict` from https://github.com/streamlink/streamlink/commit/66409146e32df8d68d9339f2a5c293b4fcc882b3#diff-6aad4096b88711d014b62443f4856dfed3bc2dab52e74822c62dae35e442da4b


bastimeyer:
According to the Python docs, the dict's insertion order is just an implementation detail of CPython in 3.6 and it has only become a mandatory feature of the Python language spec in 3.7.

- https://docs.python.org/3.6/whatsnew/3.6.html#new-dict-implementation
- https://docs.python.org/3.7/whatsnew/3.7.html#summary-release-highlights

How relevant is this for Streamlink?

beardypig:
Looks like we cannot do this unless we only support 3.7+.

PR #3277, opened 2020-10-22
chore: remove deprecated CLI arguments

bastimeyer:
This removes the following deprecated/suppressed CLI arguments:

- `--http-cookies`
  Multiple `--http-cookie` arguments need to be used
- `--http-headers`
  Multiple `--http-header` arguments need to be used
- `--http-query-params`
  Multiple `--http-query-param` arguments need to be used
- `--no-version-check`
  Version checking is opt-in via `--version-check` or `--auto-version-check`
- `--twitch-oauth-authenticate`
  Auth on Twitch is not possible with OAuth tokens generated by 3rd-party client-IDs like Streamlink's

And the following deprecated/suppressed plugin arguments:

- `--btv-username` / `--btv-password`
- `--crunchyroll-locale`
  Generic `--locale` argument needs to be used
- `--pixiv-username` / `--pixiv-password`
- `--twitch-oauth-token` / `--twitch-cookie`
  As a workaround, `--http-header="Authorization=OAuth oauth-token-from-twitch-website"` can be used
- `--ustvnow-station-code`
- `--youtube-api-key`

### DIFF
--- a/src/streamlink/plugins/btv.py
+++ b/src/streamlink/plugins/btv.py
@@ -1,8 +1,7 @@
-import argparse
 import logging
 import re
 
-from streamlink.plugin import Plugin, PluginArguments, PluginArgument
+from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
 from streamlink.stream import HLSStream
 from streamlink.utils import parse_json
@@ -11,18 +10,6 @@ log = logging.getLogger(__name__)
 
 
 class BTV(Plugin):
-    arguments = PluginArguments(
-        PluginArgument(
-            "username",
-            help=argparse.SUPPRESS
-        ),
-        PluginArgument(
-            "password",
-            sensitive=True,
-            help=argparse.SUPPRESS
-        )
-    )
-
     url_re = re.compile(r"https?://(?:www\.)?btvplus\.bg/live/?")
     api_url = "https://btvplus.bg/lbin/v3/btvplus/player_config.php"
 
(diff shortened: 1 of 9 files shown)


bastimeyer:
There's also the issue with the various `mux-subtitles` plugin arguments and merging them all into one global argument. Ideally, this should be done in the 2.0.0 release due to the major version bump, but it's optional.

mkbloke:
> There's also the issue with the various `mux-subtitles` plugin arguments and merging them all into one global argument. Ideally, this should be done in the 2.0.0 release due to the major version bump, but it's optional.

It really seems like the preferable way to go while breaking changes are already being made, otherwise I guess Streamlink will be stuck with the current implementation until release 3.0.0.

gravyboat:
Looks good to me!

beardypig:
@bastimeyer how would you merge those mux subtitles options?

bastimeyer:
By adding one global `--mux-subtitles` boolean parameter and removing the individual plugin parameters from
- funimationnow
- pluzz
- rtve
- svtplay
- vimeo
and making the plugins read the global parameter value instead. Looks like a simple change to be honest.

beardypig:
Yeah, my only concern is that users might assume a plugin supports `mux-subtitles` where it does not.

bastimeyer:
An info log message could be added to each of the supported plugins. Maybe even as a decorator, to unify the log message.

beardypig:
@bastimeyer I created #3282 to continue the discussion.

PR #3279, opened 2020-10-22
tests: fix log tests when run on a system with a non-UTC timezone

beardypig:
Set `tz_offset=0` so that `time.asctime` returns the expected value. This requires a more recent version of `freezegun`, so I set a minimum version in dev_requirements.txt.

### DIFF
--- a/tests/test_log.py
+++ b/tests/test_log.py
@@ -61,14 +61,14 @@ class TestLogging(unittest.TestCase):
         self.assertEqual(str(cm.exception), "Only {} and % formatting styles are supported")
 
     def test_datefmt_default(self):
-        with freezegun.freeze_time(datetime(2000, 1, 2, 3, 4, 5, 123456)):
+        with freezegun.freeze_time(datetime(2000, 1, 2, 3, 4, 5, 123456), tz_offset=0):
             log, output = self._new_logger(format="[{asctime}][{name}][{levelname}] {message}")
             logger.root.setLevel("info")
             log.info("test")
             self.assertEqual(output.getvalue(), "[03:04:05][test][info] test\n")
 
     def test_datefmt_custom(self):
-        with freezegun.freeze_time(datetime(2000, 1, 2, 3, 4, 5, 123456)):
+        with freezegun.freeze_time(datetime(2000, 1, 2, 3, 4, 5, 123456), tz_offset=0):
             log, output = self._new_logger(format="[{asctime}][{name}][{levelname}] {message}", datefmt="%H:%M:%S.%f")
             logger.root.setLevel("info")
             log.info("test")
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,5 +4,5 @@ pytest-cov
 coverage
 mock
 requests-mock
-freezegun
+freezegun>=1.0.0
 flake8


bastimeyer:
Looks like you only need to set the `tz_offset` argument in the `freezegun.freeze_time` call:
- https://github.com/spulec/freezegun#timezones
- https://github.com/streamlink/streamlink/blob/d8c8a0252a4ef39c69885367b5b0c47f7bfe4564/tests/test_log.py#L63-L75

The other tests are irrelevant.

beardypig:
That does work, I had an older version of `freezegun` where it did not work.

tests/test_log.py
```diff
@@ -1,7 +1,7 @@
-from datetime import datetime
-from io import StringIO
 import logging
 import unittest
+from datetime import datetime
+from io import StringIO
```
bastimeyer, line 4: Before we move the imports back and forth in every commit, we need to define rules for the import order, because flake8 doesn't have one and there's no PEP convention for that, and everyone wants to use their personal style which leads to unnecessary diffs, git history chaos/bloat and potential merge conflicts.
https://stackoverflow.com/questions/20762662/whats-the-correct-way-to-sort-python-import-x-and-from-x-import-y-statement

When I removed the compat imports a few days ago, I also fixed the import orders in most of the modules. Since the most common style used throughout the project appeared to be mixing `import` and `from` statements, I continued that pattern. So the order was built-ins, 3rd-party imports, and then local imports, each block ordered alphabetically regardless of `import` and `from` statements.

Please restore the import order in this PR, so that there are no unrelated changes. Then we can think about standardizing the import order in Streamlink if you want.

beardypig: The import lists are a mess. 

My preference is that the imports should be grouped in to built-in, third-party, and local. Within each group the plain imports first and from imports second, with multiple imports from the same package being on the same line, in alphabetical order. 

```python
import logging
from datetime import datetime

import requests
from Crypto.Cipher import Blowfish

from streamlink.plugin.api import useragents, validate
```

I will leave this one unchanged for now.

bastimeyer: Since we currently don't have many PRs open and are in an overall "clean-up" state for the major bump, do you think it would make sense re-ordering all the imports via an automation tool like this for example?
https://github.com/asottile/reorder_python_imports

And optionally add an import linting rule?

beardypig: `isort` looks like it sorts them nicely, we could do that. Not sure if how to do lint the imports does flake8 do it?

beardypig: @bastimeyer check out [153bb2ef](https://github.com/streamlink/streamlink/commit/153bb2ef4494a5ea03ed34f7ef18d458772e45be) to see what an epic diff it will be to re-order the imports, I'm not sure what kind of conflicts that would cause.

bastimeyer: Did you do this manually or did you use a tool for this? Without strict import linting rules, making all those changes is a bit pointless, because it'll become messy again at some point.

Also what about the blank lines after the imports? Due to the two blank lines rule between global methods as defined by PEP8, only having one blank line after the imports is a bit weird when the following code is not a method definition, and it is inconsistent.
https://stackoverflow.com/questions/2953250/python-pep8-blank-lines-convention

PR #3280, opened 2020-10-22
chore: use new py3 yield from syntax

beardypig:
Use the `yield from` syntax introduced in Python 3.3.

It's mainly used when returning streams from plugins, but there are a few other cases where the pattern is used.


bastimeyer (approved):

mkbloke:
Hmm, `yield from` looks rather nice.  Good to know about that.  Cheers.

PR #3282, opened 2020-10-22
chore: make mux-subtitles a global CLI argument

beardypig:
removes the plugin specific `mux-subtitles` arguments in favour a a global `--mux-subtitles` option.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -553,6 +553,15 @@ def build_parser():
         You will be prompted if the file already exists.
         """
     )
+    output.add_argument(
+        "--mux-subtitles",
+        action="store_true",
+        help="""
+        Automatically mux available subtitles in to the output stream.
+
+        If supported by the plugin.
+        """,
+    )
 
     stream = parser.add_argument_group("Stream options")
     stream.add_argument(
--- a/src/streamlink/plugins/funimationnow.py
+++ b/src/streamlink/plugins/funimationnow.py
@@ -177,15 +177,6 @@ class FunimationNow(Plugin):
 
             Default is "english".
             """
-        ),
-        PluginArgument(
-            "mux-subtitles",
-            argument_name="funimation-mux-subtitles",
-            action="store_true",
-            help="""
-            Enable automatically including available subtitles in to the output
-            stream.
-            """
         )
     )
 
@@ -264,7 +255,7 @@ class FunimationNow(Plugin):
                     url = item["src"]
                     if ".m3u8" in url:
                         for q, s in HLSStream.parse_variant_playlist(self.session, url).items():
-                            if self.get_option("mux_subtitles") and subtitles:
+                            if self.session.get_option("mux_subtitles") and subtitles:
                                 yield q, MuxedStream(self.session, s, subtitles, metadata=stream_metadata,
                                                      disposition=disposition)
                             else:
@@ -272,7 +263,7 @@ class FunimationNow(Plugin):
                     elif ".mp4" in url:
                         # TODO: fix quality
                         s = HTTPStream(self.session, url)
-                        if self.get_option("mux_subtitles") and subtitles:
+                        if self.session.get_option("mux_subtitles") and subtitles:
                             yield self.mp4_quality, MuxedStream(self.session, s, subtitles, metadata=stream_metadata,
                                                                 disposition=disposition)
                         else:
--- a/src/streamlink/plugins/pluzz.py
+++ b/src/streamlink/plugins/pluzz.py
@@ -3,7 +3,7 @@ import re
 import sys
 import time
 
-from streamlink.plugin import Plugin, PluginArguments, PluginArgument
+from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
 from streamlink.stream import DASHStream, HDSStream, HLSStream, HTTPStream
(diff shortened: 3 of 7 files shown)


src/streamlink_cli/argparser.py
```diff
@@ -553,6 +553,15 @@ def build_parser():
...
     )
+    output.add_argument(
+        "--mux-subtitles",
+        action="store_true",
+        help="""
+        Automatically mux available subtitles in to the output stream.
+
+        If supported by the plugin.
+        """,
+    )
```
bastimeyer, line 564: The parameter belongs into the `transport` group, I think.

You'll also have to add it to `streamlink_cli.main.setup_options` and `streamlink.plugin.plugin`. In `streamlink.plugin.plugin`, where the `self.session.get_option` call reads from, the parameter name needs to include a dash and not an underscore character. This needs to be changed in each plugin.

bastimeyer (changes requested):
I had already made roughly the same changes on my local repo and was testing the subtitle muxing before committing and pushing it, but I ran into an issue with the muxing itself, so I took a break. The issue is unrelated to the parameter changes, but the whole muxing seems to be broken when the subtitle stream ends early and the stream gets shut down. See this input for the pluzz plugin for example (requires a French IP address):
https://www.france.tv/documentaires/voyages/1081133-les-voltigeurs-du-lagon.html

----

Anyway, I had also added this decorator to the base Plugin class and applied it to the `_get_streams` methods (or the according sub-calls where subtitles are supported, eg. VODs):
```py
    @classmethod
    def supports_subtitles(cls, func):
        def decorator(self, *args, **kwargs):
            if not self.session.get_option("mux-subtitles"):
                self.logger.info("This plugin supports subtitle muxing via --mux-subtitles")
            return func(self, *args, **kwargs)

        return decorator
```

If you don't like the info log, then we need to annotate the supported plugin classes and add them to the docs (automatically).

src/streamlink_cli/argparser.py
beardypig: I figured as it affects the output it could go in there, but maybe transport makes sense.

beardypig: re:decorator, that looks like a nice idea - but maybe it could be applied to the plugin and that information could be picked up in help somewhere. There is perhaps the possibility for something generic here too.

beardypig: The more I look at it, I think that there is something we could do similar to how `PluginArguments` works for plugins, but for Streamlink that way other interfaces could interrogate streamlink to find out which options are supported.

bastimeyer:
I believe once the subtitle muxing parameters have been merged into one global parameter, 2.0.0 should be pretty much ready, at least all the mandatory changes. Everything else is very minor and optional, and this here is probably the last remaining thing which makes sense including in a major version bump.

I'm not sure, what exactly you have in mind with a plugin-supported parameter list and how it'll work, but if this is too much effort now, this can be done later on as well, I believe. The only thing that makes these changes critical right now is the removal of the individual CLI parameters.

mkbloke:
While subtitles are being addressed here, is it also worth considering better options for subtitle selection?

The problem at the moment is that not all sites will list subtitles in the standard locale format, i.e. formatted as `[language_code]_[country_code]`.  Some sites may list subtitles in the locale format, but also have multiple versions of those, such as user contributed subs.  This leads to the problem that, as far as I know, they cannot currently be selected.

There are perhaps 3 possible options to address this:
- Relax the formatting on the `--locale` parameter so it will accept any string
- Introduce another global parameter such as `--subtitle-lang` or `--subtitle-select` (this could maybe default to the system locale when it's not explicitly set)
- Do neither of the above and implement the subtitle selection via a plugin specific parameter as required for those plugins

Thoughts?

bastimeyer:
@mkbloke Different issue, let's not discuss this here.

@beardypig I've been working on adding global argument references to the plugins, like you've suggested, and I'm running into some issues due to how argparse works. My idea was this:

1. Add a `global_arguments` attribute to the `Plugin` class, similar to `arguments`, with the same usage of `Arguments` and `Argument` classes, but without any argument options, just the names. Then add a simple `mux-subtitles` "global-argument" to each of those plugins.
2. Upgrade `streamlink_cli.main.setup_plugin_args` and interate through the available `global_arguments` instances of each plugin and compare their `dest` attributes with the real global arguments. If it matches, then set the provided argument value on the plugin's `options` dict to make it available there. This makes it compatible with the existing code. Also create a `plugins` list on the matching argument and add the current plugin name to it, so that it can be used by the docs generator where it lists the available plugins of that argument. For a regular instance of Streamlink, this `plugins` list doesn't do anything as it's just meant for the docs.
3. Make `docs/ext_argparse.py` read that `plugins` attribute and add the list of supported plugins below the help text.

However, the issue I'm facing with this approach is that the docs extension is using a custom argparse "hook", because accessing the available/registered arguments on an argparse parser requires accessing private interfaces:
https://github.com/streamlink/streamlink/blob/f0b35b64d4b070c29d6ac3e30fc565a71bb9508a/docs/ext_argparse.py#L37-L59

This means that the docs extension needs to be refactored for the private argparse interfaces. This is fairly straightforward though with one exception, and this is causing some issues which I haven't fully figured out yet and it has something to do with how argparse registers/processes the arguments.

bastimeyer:
Nvm, found the issue... I will submit a PR later today or tomorrow.

PR #3285, opened 2020-10-24
flake8: add import-order linting config

bastimeyer:
As discussed in #3279, we should add rules for the module import order, because it's total chaos right now and everyone uses their own style, which leads to a lot of problems.

This PR therefore adds [`flake8-import-order`](https://github.com/PyCQA/flake8-import-order) as a plugin for flake8. The provided `pycharm` import style rules should match the import style of 153bb2ef4494a5ea03ed34f7ef18d458772e45be with 18 remaining errors.

Since we don't care about `src/streamlink/__init__.py` and `src/streamlink/packages/*`, I've put those paths onto the ignore list.

And just in case that it's not obvious, this PR will fail and can only be merged once the import orders have been fixed (in a separate PR). I will rebase it then.

Btw, should we do the linting steps before the test steps in the CI runners? That would make it fail faster if there are linting issues, which is usually the case when PRs are submitted.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -6,3 +6,4 @@ mock
 requests-mock
 freezegun>=1.0.0
 flake8
+flake8-import-order
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,7 +28,9 @@ exclude =
   versioneer.py,
   win32/,
 per-file-ignores =
-  src/streamlink/__init__.py:E402,F401,
+  src/streamlink/__init__.py:E402,F401,I100,I101,I201,I202,I666,
+  src/streamlink/packages/*:I100,I101,I201,I202,I666,
+  src/streamlink/packages/**/*:I100,I101,I201,I202,I666,
   src/streamlink/plugin/api/useragents.py:E501,
   src/streamlink/plugins/__init__.py:F401,
   src/streamlink/stream/__init__.py:F401,
@@ -43,3 +45,9 @@ builtins =
   raw_input,
   unicode,
   xrange,
+import-order-style = pycharm
+application-import-names =
+  streamlink,
+  streamlink_cli,
+  tests,
+  versioneer,
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ import codecs
 from os import environ, path
 from sys import argv, path as sys_path
 
-from setuptools import setup, find_packages
+from setuptools import find_packages, setup
 
 import versioneer
 
--- a/src/streamlink/__init__.py
+++ b/src/streamlink/__init__.py
@@ -8,7 +8,8 @@ An API is also provided that allows direct access to stream data.
 Full documentation is available at https://streamlink.github.io.
 
 """
-from ._version import get_versions
+from streamlink._version import get_versions
+
 __version__ = get_versions()['version']
 del get_versions
 __title__ = "streamlink"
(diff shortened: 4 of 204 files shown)


bastimeyer:
@beardypig, I've changed the import styles to `pycharm`. This matches the style of 153bb2ef4494a5ea03ed34f7ef18d458772e45be and only 18 minor errors are left.

beardypig:
Looks good to me. How do we feel about the relative imports? I personally don't like them because I think it make it harder to read.

beardypig (approved):
How should we proceed with making the changes needed to comply with the new rules? Should I PR the changes I made so we can continue discussion there? It seems pretty simple to fix those one or two small issues - some a bit weird, cos there are imports that are used to check if a feature is available, or to do some monkey patching.

bastimeyer:
I have a local branch here with all the fixes which I've pushed a few hours ago:
https://github.com/bastimeyer/streamlink/tree/flake8/import-order-fixed

There was one issue with your commit in `streamlink.stream.__init__` which required a specific import order to prevent a circular import for some reason. Haven't looked much into it, but the fix was adding a `# noqa: I100,I202` comment to the first absolute import and moving the relative imports to the top.

> How do we feel about the relative imports?

I'd say we remove them, because they are part of the local group and can't be split up when using the `pycharm` style. This means that mixed local imports have to look like this, which is stupid:
```py
from streamlink.stream.hls import HLSStream
from .session import Session
from ..foo import bar
from ...qux import quux
```

Only having absolute local imports fixes this.

> How should we proceed with making the changes needed to comply with the new rules?

Take my branch, apply your commit, fix all the things (or take my `import-order-fixed` branch) and push it onto the one of this PR, so we can do it all at once. One thing though: you should not make changes to the `streamlink.packages` modules (they are ignored anyway) or external stuff like docs, examples and scripts. This can be done separately and unnecessarily bloats up the commit with the intended changes. If the toplevel versioneer module needs to be ignored, then I'll add this to the config afterwards with a rebase.

> there are imports that are used to check if a feature is available

Conditional imports get ignored by flake8-import-order.

beardypig:
I took at look at the circular dependencies. It goes like this

`streamlink.stream.__init__` imports `streamlink.stream.stream.Stream` and `streamlink.stream.hds.HDSStream` which in turn imports `streamlink.stream.Stream`. This means that `streamlink.stream.hds` depends on `streamlink.stream`, which also depends on `streamlink.stream.hds` - creating the possible cycle. If `Stream` is imported first then it is available when `HDSStream` is imported and there is no problem, but if they are imported in alphabetical order it will fail. 
The fix is to import `streamlink.stream.stream.Stream` in `streamlink.stream.hds`, etc.

beardypig:
If you wanna take a look at this commit [4980a9f](https://github.com/streamlink/streamlink/commit/4980a9fb53632952193ed358154923229475ac51), or I can push it on to this PR.

bastimeyer:
4980a9f has this flake8 summary:
```
1     E501 line too long (129 > 128 characters)
36    I100 Import statements are in the wrong order. 'from streamlink.exceptions import PluginError' should be before 'from streamlink.plugin.plugin import Plugin'
31    I101 Imported names are in the wrong order. Should be find_packages, setup
12    I201 Missing newline between import groups. 'import unittest' is identified as Stdlib and 'import requests_mock' is identified as Third Party.
12    I202 Additional newline in a group of imports. 'from streamlink import Streamlink' is identified as Application and 'from tests.mixins.stream_hls import Playlist, Tag, Segment, TestMixinStreamHLS' is identified as Application.
92
```

If you want, I can fix and push it myself here.

beardypig:
I had to update `flake8` to get everything to match up :/

PR #3286, opened 2020-10-24
docs: update wtwitch in thirdparty list

krathalan:
A small PR that updates the screenshot and supported OS list for my third-party application, wtwitch.

### DIFF
--- a/docs/thirdparty.rst
+++ b/docs/thirdparty.rst
@@ -55,14 +55,14 @@ LiveProxy
 Wtwitch
 -------
 
-.. image:: https://i.imgur.com/oaxH3ss.jpg
+.. image:: https://git.sr.ht/~krathalan/wtwitch/blob/refs/heads/master/screenshot.jpg
   :align: center
   :alt: Wtwitch screenshot of checking subscriptions
 
 :Description: Browse Twitch streams; subscribe to streamers locally without an account
 :Type: TUI (terminal user interface)
-:OS: |Linux|
-:Author: `krathalan <https://git.sr.ht/~krathalan/>`_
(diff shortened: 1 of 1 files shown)


docs/thirdparty.rst
```diff
@@ -55,14 +55,14 @@ LiveProxy
 Wtwitch
 -------
 
-.. image:: https://i.imgur.com/oaxH3ss.jpg
+.. image:: https://git.sr.ht/~krathalan/wtwitch/blob/refs/heads/master/screenshot.jpg
```
bastimeyer, line 58: That's not a static image URL. If you have an image file in the git tree (which btw is always a bad idea), at least link to a static commit ID.

bastimeyer (changes requested):

docs/thirdparty.rst
back-to: 1. use a github url, not a third party domain

example https://github.com/streamlink/streamlink/blame/master/docs/applications.rst#L11

2. compress the image, your image is 730kB vs 135kB from the streamlink twitch gui

bastimeyer (commented):
IMO, that's still a bit large, considering that you're only trying to show text output.

```bash
$ curl -s https://user-images.githubusercontent.com/25400030/97094748-31449780-1647-11eb-858b-2665272a9ae9.jpg \
  | wc -c \
  | numfmt --to=iec-i --suffix=B

201KiB
```

Just as an example, you could reduce the JPEG quality to 30 and have a 70KiB image, or remove the background completely and have a PNG that's way better to read while being even smaller in size.

bastimeyer (approved):

PR #3290, opened 2020-10-25
plugins.bfmtv: fix ID & embed re, use Dailymotion

### LINKED ISSUE #3288: bfmtv - No playable streams found on this URL
## Plugin Issue

### Description

There is no playable streams for bfmtv

`streamlink https://www.bfmtv.com/en-direct/ best`

or `streamlink https://www.bfmtv.com/paris/en-direct/ best`
or any other channels supported by this plugin

### Log output

```
streamlink --loglevel debug https://www.bfmtv.com/en-direct/ best
[cli][debug] OS:         Linux-5.8.15-201.fc32.x86_64-x86_64-with-glibc2.2.5
[cli][debug] Python:     3.8.6
[cli][debug] Streamlink: 1.7.0
[cli][info] Found matching plugin bfmtv for URL https://www.bfmtv.com/en-direct/
error: No playable streams found on this URL: https://www.bfmtv.com/en-direct/
```

### Additional comments, etc.

As a workaround you can use their dailymotion stream: `streamlink https://www.dailymotion.com/embed/video/xgz4t1 best`

mkbloke:
- update matching for account/video IDs
- update matching for embed to find Dailymotion ID
- use Dailymotion plugin func for stream data retrieval

closes #3288

### DIFF
--- a/src/streamlink/plugins/bfmtv.py
+++ b/src/streamlink/plugins/bfmtv.py
@@ -1,45 +1,46 @@
+import logging
 import re
 
 from streamlink.plugin import Plugin
 from streamlink.plugins.brightcove import BrightcovePlayer
-from streamlink.stream import HLSStream
+
+log = logging.getLogger(__name__)
 
 
 class BFMTV(Plugin):
     _url_re = re.compile(r'https://.+\.(?:bfmtv|01net)\.com')
+    _dailymotion_url = 'https://www.dailymotion.com/embed/video/{}'
     _brightcove_video_re = re.compile(
-        r'data-holder="video(?P<video_id>[0-9]+)" data-account="(?P<account_id>[0-9]+)"'
-    )
-    _brightcove_video_alt_re = re.compile(
-        r'data-account="(?P<account_id>[0-9]+).*?data-video-id="(?P<video_id>[0-9]+)"',
+        r'accountid="(?P<account_id>[0-9]+).*?videoid="(?P<video_id>[0-9]+)"',
         re.DOTALL
     )
-    _embed_video_url_re = re.compile(
-        r"\$YOPLAYER\('liveStitching', {.+?file: '(?P<video_url>[^\"]+?)'.+?}\);",
+    _embed_video_id_re = re.compile(
+        r'<iframe.*?src=".*?/(?P<video_id>\w+)"',
         re.DOTALL
     )
 
     @classmethod
     def can_handle_url(cls, url):
-        return BFMTV._url_re.match(url)
+        return cls._url_re.match(url) is not None
 
     def _get_streams(self):
         # Retrieve URL page and search for Brightcove video data
         res = self.session.http.get(self.url)
-        match = self._brightcove_video_re.search(res.text) or self._brightcove_video_alt_re.search(res.text)
+        match = self._brightcove_video_re.search(res.text)
         if match is not None:
             account_id = match.group('account_id')
+            log.debug(f'Account ID: {account_id}')
             video_id = match.group('video_id')
+            log.debug(f'Video ID: {video_id}')
             player = BrightcovePlayer(self.session, account_id)
-            for stream in player.get_streams(video_id):
-                yield stream
+            yield from player.get_streams(video_id)
         else:
-            # Try to get the stream URL in the page
-            match = self._embed_video_url_re.search(res.text)
+            # Try to find the Dailymotion video ID
+            match = self._embed_video_id_re.search(res.text)
             if match is not None:
-                video_url = match.group('video_url')
-                if '.m3u8' in video_url:
-                    yield from HLSStream.parse_variant_playlist(self.session, video_url).items()
+                video_id = match.group('video_id')
+                log.debug(f'Video ID: {video_id}')
+                yield from self.session.streams(self._dailymotion_url.format(video_id)).items()
 
 
 __plugin__ = BFMTV


mkbloke:
Hmm, just looking at the cnews plugin and see that also gets a DM URL, but instead returns:
```
return self.session.streams(match.group('dm_url'))
```

Maybe that's the better (right?) way to do it?

back-to:
> Maybe that's the better (right?) way to do it?

yes might be better, you can use `dm_url` or use the ID and create a valid URL

mkbloke:
Interestingly:
```
            # Try to find the Dailymotion video embed URL
            match = self._embed_video_url_re.search(res.text)
            if match is not None:
                print(match.group('dm_url'))
                return self.session.streams(match.group('dm_url'))
```
gives:
```
[cli][info] Found matching plugin bfmtv for URL https://www.01net.com/01tv/
https://www.dailymotion.com/embed/video/k4HSurCuB7v2sQvwFiw
[plugin.dailymotion][debug] Found media ID: k4HSurCuB7v2sQvwFiw
[utils.l10n][debug] Language code: en_GB
error: No playable streams found on this URL: https://www.01net.com/01tv/
```
but I can do:
```
streamlink https://www.dailymotion.com/embed/video/k4HSurCuB7v2sQvwFiw
[cli][info] Found matching plugin dailymotion for URL https://www.dailymotion.com/embed/video/k4HSurCuB7v2sQvwFiw
Available streams: 180p_alt (worst), 180p, 288p_alt, 288p, 477p_alt, 477p, 720p_alt, 720p, 1080p_alt, 1080p (best)
```
Is there something else I should be doing here?  Thanks.

bastimeyer (commented):
This fully turns the plugin into a proxy plugin. We've only kept those proxy plugins so far which were still working, but now that this one is broken and only acts as a proxy for other sites, my opinion is that it should be removed.

Are the streams available/accessible on those other sites, or are they hidden?

mkbloke:
In the case 01net.com, it would seem that the Dailymotion URLs in the `iframe` `src` attributes can be directly accessed.  These exist in the basic HTML source returned by 01net.com.

With bfmtv.com it seems there are no direct URLs for the Brightcove player  based videos in either the basic HTML or the DOM.  It seems you need to find the account and video IDs, then GET (for example):
```
https://edge.api.brightcove.com/playback/v1/accounts/876450610001/videos/5615950982001
```
to get the master HLS playlist URL.

back-to:
@mkbloke 

> Is there something else I should be doing here? Thanks.

dailymotion doesn't like Referer, just remove it before you use `self.session.streams`
```py
del self.session.http.headers['Referer']
```

mkbloke:
I don't know what you want to do with this.  There are no obvious links to underlying content on bfmtv.com.  There is no obvious link to the underlying content for the live channel on 01net.com, but the clips do feature a share button that gives you a dai.ly short link.

If you want it to go, I'll turn this into a plugin removal.  Cheers.

mkbloke:
What are the current thoughts on this with regard to the info I have provided above?  I can turn this into a plugin removal if that's still the desired course of action?

As it is, I'm still have issues trying to `return self.session.streams(match.group('dm_url'))` that don't seem related to the `Referer:` header (although I appreciate the pointer, back-to), as there is no `Referer:` header to delete.

The interesting thing is that checking with mitmproxy, I can see in both cases that the playlist is returned (ref: https://github.com/streamlink/streamlink/pull/3290#issuecomment-716899638).  In the case of using the DM URL directly with the DM plugin the playlist is parsed, but even though the playlist is returned over the network when I `return self.session.streams(match.group('dm_url'))` it does not seem to get parsed and I cannot understand why.  If you want to keep this plugin, it may be easier for the time being to merge it as is.

src/streamlink/plugins/bfmtv.py
```diff
@@ -2,44 +2,41 @@
...
-            # Try to get the stream URL in the page
-            match = self._embed_video_url_re.search(res.text)
+            # Try to find the Dailymotion video ID
+            match = self._embed_video_id_re.search(res.text)
             if match is not None:
-                video_url = match.group('video_url')
-                if '.m3u8' in video_url:
-                    yield from HLSStream.parse_variant_playlist(self.session, video_url).items()
+                video_id = match.group('video_id')
+                for stream in DailyMotion._get_streams_from_media(self, video_id):
```
back-to, line 38: `yield from`

src/streamlink/plugins/bfmtv.py
```diff
@@ -2,44 +2,41 @@
...
     def _get_streams(self):
         # Retrieve URL page and search for Brightcove video data
         res = self.session.http.get(self.url)
-        match = self._brightcove_video_re.search(res.text) or self._brightcove_video_alt_re.search(res.text)
+        match = self._brightcove_video_re.search(res.text)
         if match is not None:
             account_id = match.group('account_id')
             video_id = match.group('video_id')
             player = BrightcovePlayer(self.session, account_id)
             for stream in player.get_streams(video_id):
```
back-to, line 31: `yield from`

src/streamlink/plugins/bfmtv.py
```diff
@@ -2,44 +2,41 @@
...
         else:
-            # Try to get the stream URL in the page
-            match = self._embed_video_url_re.search(res.text)
+            # Try to find the Dailymotion video ID
+            match = self._embed_video_id_re.search(res.text)
             if match is not None:
-                video_url = match.group('video_url')
-                if '.m3u8' in video_url:
-                    yield from HLSStream.parse_variant_playlist(self.session, video_url).items()
+                video_id = match.group('video_id')
```
back-to, line 37: could you add a debug message with the _video_id_

back-to (commented):
`from streamlink.plugins.dailymotion import DailyMotion`

is not important, just let it be like this.

mkbloke:
Great, thanks for coming back on that.  Cheers.

PR #3293, opened 2020-10-26
plugins.cnews: fix video ID search, add schema

mkbloke:


src/streamlink/plugins/cnews.py
```diff
@@ -1,25 +1,43 @@
...
 
 class CNEWS(Plugin):
     _url_re = re.compile(r'https?://www.cnews.fr/[^ ]+')
-    _embed_video_url_re = re.compile(r'class="dm-video-embed_video" src="(?P<dm_url>.*)"')
-    _embed_live_url_re = re.compile(r'class="wrapper-live-player main-live-player"><iframe src="(?P<dm_url>.*)"')
+    _json_data_re = re.compile(r'({"basePath":.*})\);')
+    _dailymotion_url = 'https://www.dailymotion.com/embed/video/{}'
+
+    _data_schema = validate.Schema(
+        validate.any(None, validate.all(
```
bastimeyer, line 14: The outer `validate.any(None, validate.all(` calls are unnecessary. `validate.Schema` is already a `validate.all` wrapper.

If the response is indeed `null`, then your code will raise a `TypeError`, so I don't think you'll need the `validate.any` call.

src/streamlink/plugins/cnews.py
```diff
@@ -1,25 +1,43 @@
...
+            validate.transform(_json_data_re.search),
+            validate.any(None, validate.all(
+                validate.get(1),
+                validate.transform(parse_json),
+                validate.filter(lambda k, v: 'dm_player_' in k),
+                {
+                    validate.text: {
+                        'video_id': validate.text,
+                    },
+                },
```
bastimeyer, line 24: Don't filter by "substring in key". Just add `validate.optional()` keys for the two properties to the dict validation. Btw, `validate.text` can be replaced with `str` now.

src/streamlink/plugins/cnews.py
```diff
@@ -1,25 +1,43 @@
...
-from streamlink.plugin.api import useragents
+from streamlink.plugin.api import validate
+from streamlink.utils import parse_json
 
 
 class CNEWS(Plugin):
     _url_re = re.compile(r'https?://www.cnews.fr/[^ ]+')
-    _embed_video_url_re = re.compile(r'class="dm-video-embed_video" src="(?P<dm_url>.*)"')
-    _embed_live_url_re = re.compile(r'class="wrapper-live-player main-live-player"><iframe src="(?P<dm_url>.*)"')
+    _json_data_re = re.compile(r'({"basePath":.*})\);')
```
bastimeyer, line 10: I think a better regex would be anchoring it with the entire function call:
```
jQuery\.extend\(Drupal\.settings, ({.*})\);
```

mkbloke:
I have a question regarding the definition of a "proxy plugin".  In relation to this plugin for example, there's no direct link to the Dailymotion URL in the basic HTML returned by the website.  It can, however, be found as an `iframe` `src` attribute URL in the browser DOM.  Does that count?  What is your definition?  Thanks.

bastimeyer:
In my previous review, I was just commenting on the code, not the proxy-plugin problem itself.

> definition of a "proxy plugin"

A plugin that provides streams from a website which are run by another website and are accessible over there via a canonical URL.

Prime example:
https://github.com/streamlink/streamlink/blob/a585e8829750d2e5bd68a3887efedf448283749e/src/streamlink/plugins/teamliquid.py

In case of
https://www.cnews.fr/le-direct
this is a bit of a problem, because the video player doesn't properly link to the source. And there are lots of other streams and VODs on this news site as well. However, if you click the "share" button, it'll link to
https://dai.ly/x3b68jn aka https://www.dailymotion.com/video/x3b68jn
and this works for every other stream/VOD as well.

mkbloke:
@bastimeyer, I think I am not explaining/have not explained my thinking very well here, so...

> A plugin that provides streams from a website which are run by another website and are accessible over there via a canonical URL.

OK, so I'm contending that this makes the cnews plugin a proxy plugin...

> because the video player doesn't properly link to the source

Could you expand on what you mean by "doesn't properly link to the source", please?

> However, if you click the "share" button, it'll link to...

Right.  So, each share button link provides a link to a canonical URL.  That link may not be the underlying source in terms of the playlist, but it's static as far as users are concerned right?   - and users do not need to be worried about the underlying true playlist source as long as the plugin that can handle it (DM in this case) can resolve it to the underlying playlist?

So, if I have explained my logic well enough, wouldn't the cnews plugin be considered a proxy plugin?  - because users can, for example, stream a particular live stream or VOD on cnews via the static dai.ly link provided via the share link.

Does that make any sense?!  Cheers.

bastimeyer:
> this makes the cnews plugin a proxy plugin...

Yes, it technically is indeed a proxy plugin

> doesn't properly link to the source

Other than clicking the "share" button, there are no indications that the stream source is from another website. Eg. by including links, overlays, etc.

This makes it a problem for the user to find the actual URL which they could/should use instead, so a proxy plugin might be justified here, even though it's unnecessary code in the repo. But I don't want to decide this on my own. News sites could also be handled as an exception here.

mkbloke:
OK, cool, thanks.  I just wanted to check my understanding here.

I'll push some fixes as per your review.  Cheers.

PR #3294, opened 2020-10-26
ci.github: run lint step before test step

bastimeyer:
As briefly mentioned in #3285, this will make the PR checks fail faster in case there are linting errors but all tests are passing.

Also added the `--statistics --count` parameters, so that a summary gets included in the output when there are linting errors.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,12 +35,12 @@ jobs:
       - name: Install dependencies
         continue-on-error: ${{ matrix.continue || false }}
         run: bash ./script/install-dependencies.sh
+      - name: Lint
+        continue-on-error: ${{ matrix.continue || false }}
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -35,12 +35,12 @@ jobs:
       - name: Install dependencies
         continue-on-error: ${{ matrix.continue || false }}
         run: bash ./script/install-dependencies.sh
+      - name: Lint
+        continue-on-error: ${{ matrix.continue || false }}
+        run: flake8 --statistics --count
```
back-to, line 40: > --statistics

it is already active, you don't need it here

https://github.com/streamlink/streamlink/blob/8034b1606388e325916baf9f8ce738915fa6b528/setup.cfg#L39

PR #3295, opened 2020-10-26
ci.github: install latest version of pynsist

bastimeyer:
This installs the latest version of pynsist instead of pinning it to `2.4`. It was pinned to `2.4` because of #2655.

According to the [pynsist release notes](https://pynsist.readthedocs.io/en/latest/releasenotes.html), the newer versions (current one is `2.5.1`) assemble the executables during build-time rather than run-time/install-time. This cleans up the install directory by a bit and is less confusing for the enduser.

The NSIS GUI has also been improved with proper DPI scaling, but that requires NSIS 3.0 which is currently unavailable on the `ubuntu-18.04` CI image and `ubuntu-latest` still points towards `18.04`. That will change soon though, and once it does, the installer will see those improvements automatically. We could also manually set it to `ubuntu-20.04`.

----

While building the installer locally, I also tried bumping the included Python version, but there's currently an issue in pynsist and it doesn't recognize the new wheel naming scheme since py38 and pycryptodome can't be found. I will see if I can submit a PR there and fix it when I get the time. There's already a PR for it, but it's stale since April and just a hack rather than a proper fix.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -83,7 +83,7 @@ jobs:
       - name: Install dependencies
         run: |
           ./script/install-dependencies.sh
-          python -m pip install pynsist==2.4
+          python -m pip install pynsist
           sudo apt update
           sudo apt install -y nsis imagemagick inkscape
       - name: Installer file name


beardypig:
I'd be fine with using `ubuntu-20.04`...

bastimeyer:
> I'd be fine with using `ubuntu-20.04`...

According to `actions/virtual-environments`, the switch will happen automatically during 2020-11-30 and the following weeks excluding holidays, but if you don't want to wait for this, we can change it now.

This should be changed in a separate PR though. The used Python version for the docs, installer and release jobs could be changed as well in one go.

bastimeyer:
Are we good here? If this gets merged, all that'll change for now are the executables, which are built during compile-time, which is a big improvement.

I will open a PR with the `ubuntu-20.04` changes later, which will then use NSIS 3.0, so that the installer will work better on high DPI displays.

I haven't had the time and motivation yet to check the issue on pynsist regarding the new wheel naming scheme, so bumping the python version above 3.7 for the Windows installers is not possible right now.

PR #3296, opened 2020-10-26
plugins.zdf_mediathek: also support 3sat mediathek

gromit1811:
3sat and ZDF mediathek essentially are the same, only the URLs and the API
URLs differ. So support 3sat URLs and check the URL to decide whether to
use the ZDF or 3sat API. Everything else is identical.

### DIFF
--- a/src/streamlink/plugins/zdf_mediathek.py
+++ b/src/streamlink/plugins/zdf_mediathek.py
@@ -1,5 +1,6 @@
 import logging
 import re
+from urllib.parse import urlparse, urlunparse
 
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
@@ -8,8 +9,6 @@ from streamlink.utils import parse_json
 from streamlink.utils.url import url_concat
 
 log = logging.getLogger(__name__)
-
-API_URL = "https://api.zdf.de"
 
 QUALITY_WEIGHTS = {
     "hd": 720,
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/zdf_mediathek.py
```diff
@@ -29,9 +30,12 @@
     )
 }
 
-_url_re = re.compile(r"""
+_url_zdf_re = re.compile(r"""
     http(s)?://(\w+\.)?zdf.de/
 """, re.VERBOSE | re.IGNORECASE)
+_url_3sat_re = re.compile(r"""
+    http(s)?://(\w+\.)?3sat.de/
+""", re.VERBOSE | re.IGNORECASE)
```
bastimeyer, line 38: Please only use one regex with a capture group for the domain names.

Then select the right API URL from the capture group value, or build the API URL dynamically.

bastimeyer (changes requested):

src/streamlink/plugins/zdf_mediathek.py
gromit1811: Done. I'm now extracting the API base URL from the 1st response and dropped the hardcoded API URL(s). Note that I'm neither very familiar with streamlink internals nor with web APIs in general, so I hope this makes sense. At least it worked fine for 2 test URLs I tried.

PR #3300, opened 2020-10-28
cli: add support for stream manifest URL output

### LINKED ISSUE #3098: Retreiving master HLS playlist/manifest of streams
Is there a way to get master HLS playlist for twitch streams. When i use the options --stream-url and --json, then I'm able to retrieve links for the media playlists. This is different than the master playlist. Where can I find the master playlist?

Or is there a way to convert to the media playlist to the master playlist? I think I would need codec information for that.

bastimeyer:
Resolves #3098

This enables the output of the input-stream's manifest/master URL via
`streamlink --stream-url URL` / `streamlink --json URL`
in addition to the already supported output when selecting a stream via
`streamlink --stream-url URL STREAM` / `streamlink --json URL STREAM`

This is primarily helpful for HLS streams with a variant playlist.

----

Things to note:
- This ignores cases where a plugin yields streams from different sources. Only the last stream in the returned `dict/OrderedDict` gets used, which is usually `best` or `best-unfiltered`. `--stream-types` can still be used though for filtering out unwanted stream types.
- I didn't add the manifest/master URL to the HLSStream's string representation and own JSON output, as it'd spam the output by a lot. Maybe it should be added though, at least to the JSON output, as it would (partly) solve the issue mentioned in the previous point.
- Added tests for both `handle_url` and `handle_stream` (output without and with stream/quality selection)
- IMO, `streamlink_cli.main.handle_url` needs to be refactored, because patching `streamlink_cli.main.args` is awkward and lots of unrelated args need to be set. The entire main CLI module needs to be simplified and modularized.
- I'm not sure why the other test methods don't mock the globals in a context manager. That introduces side-effects, no?

### DIFF
--- a/src/streamlink/stream/dash.py
+++ b/src/streamlink/stream/dash.py
@@ -262,3 +262,6 @@ class DASHStream(Stream):
 
     def to_url(self):
         return self.mpd.url
+
+    def to_manifest_url(self):
+        return self.mpd.url
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -339,7 +339,7 @@ class HLSStreamReader(SegmentedStreamReader):
 class MuxedHLSStream(MuxedStream):
     __shortname__ = "hls-multi"
 
-    def __init__(self, session, video, audio, force_restart=False, ffmpeg_options=None, **args):
+    def __init__(self, session, video, audio, url_master=None, force_restart=False, ffmpeg_options=None, **args):
         tracks = [video]
         maps = ["0:v?", "0:a?"]
         if audio:
@@ -353,6 +353,10 @@ class MuxedHLSStream(MuxedStream):
         ffmpeg_options = ffmpeg_options or {}
 
         super().__init__(session, *substreams, format="mpegts", maps=maps, **ffmpeg_options)
+        self.url_master = url_master
+
+    def to_manifest_url(self):
+        return self.url_master
 
 
 class HLSStream(HTTPStream):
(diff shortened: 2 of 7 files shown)


back-to:
> --stream-types can still be used though for filtering out unwanted stream types.

basically required

`streamlink https://www.zdf.de/nachrichten/heute-sendungen/so-wird-das-wetter-102.html --stream-priority hls --json --stream-url`

vs

`streamlink https://www.zdf.de/nachrichten/heute-sendungen/so-wird-das-wetter-102.html  --json --stream-url`

`"url_manifest": null,`

---

might return the **type** of the manifest or use a different name, not `url_manifest`

---

the output seems weird sometimes

bastimeyer:
> `--json --stream-url`

These params are mutually exclusive...

> might return the **type** of the manifest or use a different name, not `url_manifest`

And as said, it's only a problem if the plugin yields streams from different sources, but I don't see any way of fixing this in the regular `--stream-url` output. If the user hasn't selected a stream/quality, and the plugin yields different stream types, what should be done here other than selecting the last (usually the "best"/"best-unfiltered") stream? Reading the value of all streams and returning everything is not possible, because `--stream-url` is supposed to return only a single URL. Returning just one, sorted by the number of occurrences seems confusing if the number of stream types varies.

Regarding the name, the problem is that `streamlink_cli` should not have individual implementations for the various stream types, so a generic name needs to be chosen. I went with "manifest", because that works with DASH streams as well as HLS streams, and on HLS streams, "master" or "variant" are both specific terms only related to it. Not sure about the other stream types.

In the JSON output, as opposed to being `null`, the `url_manifest` property could be removed if a value doesn't exist, if you prefer it that way. That could prevent some kind of confusion, but a parser would have to deal with conditional properties. Or as a better solution, the "manifest-url" could be added to the object (and string representation) of each stream item. That would make it less legible with lots of repetitions, but would solve the stream type issue.

> the output seems weird sometimes

The output hasn't been changed at all.

bastimeyer:
Removed the `url_manifest` property from `--json` again since each stream item has its own JSON representation and HLS streams with a master playlist now have an additional `master` property.

Also changed the string representation of the HLSStream class to include the master playlist URL if it exists, and renamed the internal property and constructor parameter to `url_master`.

The `to_url_manifest` interface still exists on the abstract `Stream` class, as I couldn't find a better name that fits into every stream type. This method is needed for the `--stream-url` parameter, which reads the value from the last stream item in the resolved streams dict, which is usually "best".

gravyboat (approved):
Looks good, if @back-to has no further objections let's get it merged.

beardypig:
I'd think `to_manifest_url` is a better name, as it's the url to the manifest, not the manifest of the url?

bastimeyer:
@beardypig My idea was to have "manifest" as a suffix, so that it fits into the scheme of the existing interface. I can change it though to your suggestion which makes more sense readability-wise.

bastimeyer:
What's the decision here before I rebase? Rename it or not?

gravyboat:
@bastimeyer I'm not particularly passionate about either option, let's just rename it to `to_manifest_url` so we can get this merged.

PR #3306, opened 2020-10-29
plugins.nbcnews: fix video ID search, add schemas

mkbloke:
- add JSON data schema
- add token schema
- add JSON post data and token request

### DIFF
--- a/src/streamlink/plugins/nbcnews.py
+++ b/src/streamlink/plugins/nbcnews.py
@@ -11,15 +11,56 @@ log = logging.getLogger(__name__)
 
 class NBCNews(Plugin):
     url_re = re.compile(r'https?://(?:www\.)?nbcnews\.com/now')
-    js_re = re.compile(r'https://ndassets\.s-nbcnews\.com/main-[0-9a-f]{20}\.js')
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/nbcnews.py
```diff
@@ -28,18 +52,14 @@ def can_handle_url(cls, url):
...
 
-    @Plugin.broken(3123)
     def _get_streams(self):
-        html = self.session.http.get(self.url).text
-        match = self.js_re.search(html)
-        js = self.session.http.get(match.group(0)).text
-        match = self.api_re.search(js)
-        log.debug('API ID: {0}'.format(match.group(1)))
-        api_url = self.api_url.format(match.group(1))
+        id = self.session.http.get(self.url, schema=self.json_data_schema)
```
back-to, line 56: don't use a Built-in Function name https://docs.python.org/3.6/library/functions.html#id

mkbloke: Oh, I hadn't realised that, thanks.

mkbloke:
This needs a bit more fixing right now.

PR #3307, opened 2020-10-30
chore: implicit py3 super() calls

bastimeyer:
This changes all explicit `super(type, object)` calls to implicit py3 `super()` calls, as they are easier to read and maintain.

Most of the super calls here are in a single inheritance scheme, and only some of the tests have multiple inheritance, but the MRO is correct, so nothing will change logic-wise.

All other methods in the Streamlink codebase which are calling methods of their parent classes do this directly without calling `super()`, which may be a tiny bit faster, but it also makes it less extensible. I obviously didn't change it in this PR, but believe it would be better having it changed at some point in the future. Not important for now though.


beardypig:
Nice nice.

PR #3308, opened 2020-10-30
plugins.skai: plugin removal

mkbloke:
Links to static playlist files are embedded in the basic HTML source.


mkbloke:
I have to note here (and confess) that I've realised I failed to check the other site supported by this plugin, i.e. skaitv.gr as well.  Grrr, sorry about that.

Anyway, although the plugin doesn't work for that site either, there could be justification for supporting it, as the playlist URLs are not as apparent as those for skai.gr.

I don't know what you might feel the better decision is - to reverse the merge or for me to fix the plugin for that site and add it back again later...

Sorry again.

back-to:
> to reverse the merge or for me to fix the plugin for that site and add it back again later...

None of it, it is static.

```
    "live_url":       "https://skai-live.siliconweb.com/media/cambria4/index.m3u8",
    "live_url_gr":    "https://skai-live-gr.siliconweb.com/media/cambria4/index.m3u8",
    "live_url_gr_cy": "https://skai-live-gr-cy.siliconweb.com/media/cambria4/index.m3u8",
```

mkbloke:
Yeah, they're static, they just require the backslash escaping removed before use if manually scraped from the HTML source...

PR #3309, opened 2020-10-31
plugins.kingkong: plugin removal

mkbloke:
kingkong.com.tw now redirects to play.lang.live after a brief message:
```
重新導向

因應《金剛直播》於2020/4/29 (三) 結束營運，原平台網址將轉移至《浪PLAY 》繼續為用戶服務。此網址變更並不影響原平台用戶使用之所有權利，詳請參考網頁最新資訊公告。
				金剛運營團隊敬上
```
```
Redirect

In response to the end of operation of "King Kong Live" on 2020/4/29 (Wed), the original platform website will be transferred to "Lang PLAY" to continue serving users. This URL change does not affect all the rights of users of the original platform, please refer to the latest information announcement on the web page for details.
King Kong Operations Team
```

I figure an initial removal is probably the better option right now as it would seem like this plugin has been broken since late April.  A plugin to support play.lang.live might be better off having an appropriate name for it, but I guess that's only worth doing upon receiving a plugin request?

The other option is to look at supporting lang.live in this plugin, but given the above, is it worth doing right now?


back-to:
> I figure an initial removal is probably the better option right now as it would seem like this plugin has been broken since late April.

most likely nobody uses this plugin or doesn't want to open an issue

> A plugin to support play.lang.live might be better off having an appropriate name for it, but I guess that's only worth doing upon receiving a plugin request?

if we remove it right now, there is no point in readding it again.

> The other option is to look at supporting lang.live in this plugin, but given the above, is it worth doing right now?

might not be worth it, but a fix would be easy.

```diff
diff --git a/src/streamlink/plugins/kingkong.py b/src/streamlink/plugins/kingkong.py
index 86161661..0db24892 100644
--- a/src/streamlink/plugins/kingkong.py
+++ b/src/streamlink/plugins/kingkong.py
@@ -7,9 +7,9 @@ from streamlink.stream import HLSStream, HTTPStream
 
 log = logging.getLogger(__name__)
 
-API_URL = "https://g-api.langlive.com/webapi/v1/room/info?room_id={0}"
+API_URL = "https://game-api.lang.live/webapi/v1/room/info?room_id={0}"
 VOD_API_URL = (
-    "https://g-api.langlive.com/webapi/v1/replayer/detail?live_id={0}")
+    "https://game-api.lang.live/webapi/v1/replayer/detail?live_id={0}")
 STATUS_ONLINE = 1
 STATUS_OFFLINE = 0
 STREAM_WEIGHTS = {
@@ -20,7 +20,7 @@ STREAM_WEIGHTS = {
 }
 
 _url_re = re.compile(r"""
-    https://www\.kingkong\.com\.tw/
+    https?://play\.lang\.live/
     (?:
         video/(?P<vid>[0-9]+G[0-9A-Za-z]+)|
         (?P<channel>[0-9]+)
@@ -68,7 +68,6 @@ class Kingkong(Plugin):
             return STREAM_WEIGHTS[stream], "kingkong"
         return Plugin.stream_weight(stream)
 
-    @Plugin.broken()
     def _get_streams(self):
         match = _url_re.match(self.url)
         vid = match.group("vid")

```

mkbloke:
> most likely nobody uses this plugin or doesn't want to open an issue

I suspect the former.

> might not be worth it, but a fix would be easy.

Yeah, I'd looked before and noticed the same thing.  It just depends on your and the other main Streamlink devs' views on whether it should be done, given that it means you'll end up with a plugin that nobody has requested and might not ever be used.  That was my thinking when it came to the removal...

PR #3312, opened 2020-10-31
ci.github: set ubuntu to 20.04 and python to 3.9

bastimeyer:
As mentioned in #3295, explicitly setting the CI environment to `ubuntu-20.04` will make NSIS 3 available for `pynsist`, so that the Windows installers can benefit from proper DPI scaling.

`ubuntu-latest` will be changed from `ubuntu-18.04` to `ubuntu-20.04` next month anyway, but the transitioning period is a couple of weeks and it doesn't hurt us setting an explicit version.

This also bumps the used python version in all CI jobs from `3.8` to `3.9`, as `3.9` is available natively now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ jobs:
       # please remember to change the `codecov.notify.after_n_builds` value in .codecov.yml
       # when changing the build matrix and changing the number of test runners
       matrix:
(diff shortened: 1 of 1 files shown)


bastimeyer:
Once this PR has been merged, I will update the branch protection rules and make the python 3.9 steps a merge requirement.

Issue #2947, opened 2020-05-03
--player-args "--fullscreen" with VLC breaks playback


emerysteele:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description

VLC has a built in --fullscreen argument for auto opening in fullscreen mode. But when using `--player-args "--fullscreen"`, vlc will open, but it does not play the stream. Just goes to default playlist window.

### Expected / Actual behavior

VLC should open streams in fullscreen mode when '--fullscreen' is set as an argument.

Setting '--fullscreen' as an argument causes the stream to not open at all. Just VLC to the default window.

### Reproduction steps / Explicit stream URLs to test

1. run `streamlink --player-args "--fullscreen" <streamurl> best`

### Log output

```
>streamlink --player-args "--fullscreen" <streamurl> best
[cli][info] Found matching plugin okru for URL <streamurl>
[cli][info] Available streams: 240p (worst), 360p, 488p, 720p (best)
[cli][info] Opening stream: 720p (hls)
[cli][info] Starting player: "C:\Program Files\VideoLAN\VLC\vlc.exe"
```
```

gravyboat:
Can you look through the [fullscreen](https://trac.videolan.org/vlc/query?status=assigned&status=new&status=reopened&summary=~fullscreen&col=id&col=summary&col=status&col=owner&col=priority&col=milestone&col=component&col=version&order=priority) bug reports over on their tracker and see if anyone encountered your problem there? I don't know what we can do to solve this as we're just passing the option. If you call VLC to launch without dumping data in to it via Streamlink does the problem still occur? So maybe try either apps, or feeding it a video directly and passing those options. I'm not sure exactly what the command will look like for that as I don't use VLC.

emerysteele:
Launching vlc directly with the --fullscreen argument is working normally. `C:\Program Files\VideoLAN\VLC\vlc.exe --fullscreen <streamurl>` Updated to latest VLC version, but still same issue when launched through streamlink.

Example Streamlink Command (Not Working/No Playback):
`streamlink --player-args "--fullscreen" hls://https://b028.wpc.azureedge.net/80B028/Samples/a38e6323-95e9-4f1f-9b38-75eba91704e4/5f2ce531-d508-49fb-8152-647eba422aec.ism/Manifest(format=m3u8-aapl) best`
Note: removing '--player-args "--fullscreen"' from command will allow the stream to play, but obviously not in full screen.

[...]

PR #3313, opened 2020-10-31
cli: optional player-args input variable

bastimeyer:
- Make the player input variable in the `--player-args` parameter
  optional. If it's missing, append it to the parameter value.
- Rename it from `{filename}` to `{playerinput}`, to reduce confusion
  and add `{filename}` back as a secondary fallback variable.
- Update parameter docs accordingly and add+fix tests

See: #2947


gravyboat:
Looks good, thanks!

PR #3318, opened 2020-11-04
plugins.nrk: fix/rewrite plugin

sn4kebite:
The TV and radio sites of nrk.no has been changed some time ago which
has left the NRK plugin broken. This change makes use of [PSAPI] to
bring both TV and radio back to working state, for both live and
on-demand shows.

[PSAPI]: https://psapi-we.nrk.no/documentation/

### DIFF
--- a/src/streamlink/plugins/nrk.py
+++ b/src/streamlink/plugins/nrk.py
@@ -33,6 +33,7 @@ class NRK(Plugin):
                 validate.optional("luna"): validate.any(None, {
                     "data": {
                         "title": str,
+                        "category": validate.any(None, str),
                     },
                 }),
             },
@@ -44,9 +45,18 @@ class NRK(Plugin):
         },
     ))
 
+    category = None
+    title = None
+
     @classmethod
     def can_handle_url(cls, url):
         return cls._url_re.match(url) is not None
+
+    def get_category(self):
+        return self.category
+
+    def get_title(self):
+        return self.title
 
     def _get_streams(self):
         # Construct manifest URL for this program.
@@ -64,19 +74,22 @@ class NRK(Plugin):
             reason = manifest["nonPlayable"]["reason"]
             log.error(f"Not playable ({reason})")
             return None
+        self._set_metadata(manifest)
         asset = manifest['playable']['assets'][0]
 
         # Some streams such as podcasts are not HLS but plain files.
         if asset['format'] == 'HLS':
             return HLSStream.parse_variant_playlist(self.session, asset['url'])
         else:
-            return [(self._get_title(manifest), HTTPStream(self.session, asset['url']))]
+            return [("live", HTTPStream(self.session, asset['url']))]
 
-    def _get_title(self, manifest):
+    def _set_metadata(self, manifest):
         luna = manifest.get("statistics").get("luna")
         if not luna:
-            return None
-        return luna.get("data").get("title")
+            return
+        data = luna["data"]
+        self.category = data.get("category")
+        self.title = data.get("title")
 
 
 __plugin__ = NRK
--- a/src/streamlink/plugins/nrk.py
+++ b/src/streamlink/plugins/nrk.py
@@ -41,6 +41,7 @@ _playable_schema = validate.Schema({
     },
 })
 
+
 class NRK(Plugin):
     _psapi_url = 'https://psapi.nrk.no'
 
@@ -49,9 +50,6 @@ class NRK(Plugin):
         return _url_re.match(url)
 
     def _get_streams(self):
-        # Get the stream type from the url (tv/radio).
-        stream_type = _url_re.match(self.url).group(1).upper()
-
         # Construct manifest URL for this program.
         program_type, program_id = _id_re.search(self.url).groups()
         if program_type == 'direkte':
@@ -84,5 +82,6 @@ class NRK(Plugin):
         if not data:
             return None
         return data.get("title")
+
 
 __plugin__ = NRK
--- a/src/streamlink/plugins/nrk.py
+++ b/src/streamlink/plugins/nrk.py
@@ -1,9 +1,12 @@
+import logging
 import re
 from urllib.parse import urljoin
 
-from streamlink.plugin import Plugin, PluginError
+from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
 from streamlink.stream import HLSStream, HTTPStream
+
+log = logging.getLogger(__name__)
 
 
 class NRK(Plugin):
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/nrk.py
```diff
@@ -3,30 +3,10 @@
...
+from streamlink.stream import HLSStream, HTTPStream
 
-COOKIE_PARAMS = (
-    "devicetype=desktop&"
-    "preferred-player-odm=hlslink&"
-    "preferred-player-live=hlslink"
-)
-
-_id_re = re.compile(r"/(?:program|direkte|serie/[^/]+)/([^/]+)")
+_id_re = re.compile(r"/(program|direkte|serie|podkast)(?:/.+)?/([^/]+)")
```
bastimeyer, line 8: Could you please move the regexes and the schemas into the class as class attributes? The current plugin implementation is still using the old layout which we don't want anymore.

src/streamlink/plugins/nrk.py
```diff
@@ -35,30 +15,73 @@
...
-        baseurl = self.session.http.get(self.url, cookies=cookie, schema=_schema)
-        program_id = _id_re.search(self.url).group(1)
+        # Construct manifest URL for this program.
+        program_type, program_id = _id_re.search(self.url).groups()
+        if program_type == 'direkte':
+            manifest_type = 'channel'
+        elif program_type == 'serie':
+            manifest_type = 'program'
+        elif program_type == 'podkast':
+            manifest_type = 'podcast'
```
bastimeyer, line 60: It's probably better to use a key-value map here for readability reasons. Maybe also return `None` early if there's no match, in case the URL regex gets changed and the programm type checks do not get updated.

src/streamlink/plugins/nrk.py
```diff
@@ -35,30 +15,73 @@
...
+        manifest_url = urljoin(self._psapi_url, "playback/manifest/{0}/{1}".format(manifest_type, program_id))
```
bastimeyer, line 61: Since the min. Python version of Streamlink is 3.6, you can use an f-string here for better readability. You could also remove the `urljoin` call and direcly add `playback/manifest/{0}/{1}` to the `_psapi_url` attribute value, if you prefer it this way.

src/streamlink/plugins/nrk.py
```diff
@@ -35,30 +15,73 @@
...
+                        path=validate.any(
+                            validate.endswith(".m3u8"),
+                            validate.endswith(".mp3")
+                        ),
+                    ),
+                    "format": validate.all(validate.text),
+                }]
+            ),
+        }
+    ),
```
bastimeyer, line 34: Calling `validate.all` with just one argument is unnecessary.
Is it really necessary to limit the URL's `scheme` to `http`?
`validate.text` can nowadays be replaced by `str`, as it's an old py2 compat thing.

src/streamlink/plugins/nrk.py
```diff
@@ -35,30 +15,73 @@
...
+        statistics = manifest.get("statistics")
+        if not statistics:
+            return None
+        luna = statistics.get("luna")
+        if not luna:
+            return None
+        data = luna.get("data")
+        if not data:
+            return None
+        return data.get("title")
```
bastimeyer, line 84:
```suggestion
        return manifest.get("statistics", {}).get("luna", {}).get("data", {}).get("title")
```

src/streamlink/plugins/nrk.py
bastimeyer, line 8: It's better to have the URL regex include the ID regex, so that there's only a single URL to match the input. Not sure why it's been done this way previously, but if the line is too long, use the verbose regex option.

src/streamlink/plugins/nrk.py
bastimeyer: This is probably unnecessary though, as you already have a fully nested schema for the `statistics` object, and if it's missing or not matching, an error will be returned. You'll have to set it to optional in the schema if it's indeed optional. To do this, set the object to to `validate.optional("statistics")`.

bastimeyer (changes requested):
Thanks for the PR!
Before this can get merged, I'd like to see some parts improved/fixed first. Nothing big though.

Unfortunately, I can't test/verify this, as I don't have access to a Norwegian IP address.

src/streamlink/plugins/nrk.py
sn4kebite: I looked at this in more detail and it appears that the `luna` object can either be `null` or missing entirely. The `statistics` object is always present however, so I went with a middle ground here.

src/streamlink/plugins/nrk.py
sn4kebite: I cleaned this up a bit, also wondering if it's ok to just remove the path validation. If the manifest returns an HLS playlist the `format` field explicitly states this, otherwise it should be any playable file stream. MIME type of the URL is also present in the manifest if needed, but not included here as it's not used.

src/streamlink/plugins/nrk.py
sn4kebite: I put this in a class attribute dict, and also changed it to raise PluginError with a descriptive message if `program_type` is unknown, which I assume is better than silently failing.

bastimeyer (changes requested):

src/streamlink/plugins/nrk.py
```diff
@@ -1,64 +1,88 @@
...
+            validate.optional("luna"): validate.any(
+                None,
+                {
+                    "data": {
+                        "title": str,
+                    },
+                },
+            ),
+        },
+    })
```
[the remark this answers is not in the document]
sn4kebite: I went with removing path validation since it makes more sense to just check `format` than to try to hard code supported file extensions.

src/streamlink/plugins/nrk.py
```diff
@@ -1,64 +1,88 @@
...
-        json_url = urljoin(baseurl, "mediaelement/{0}".format(program_id))
-        res = self.session.http.get(json_url, cookies=cookie)
-        media_element = self.session.http.json(res, schema=_mediaelement_schema)
-        media_url = media_element["mediaUrl"]
+        res = self.session.http.get(manifest_url)
+        manifest = self.session.http.json(res, schema=self._playable_schema)
+        playable = manifest['playable']
+        if playable is None:
+            reason = manifest["nonPlayable"]["reason"]
+            raise PluginError(f"Not playable ({reason})")
```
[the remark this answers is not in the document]
sn4kebite: I changed both `PluginError` instances, but on second thought maybe the unknown `program_type` is considered a plugin error?

src/streamlink/plugins/nrk.py
```diff
@@ -1,64 +1,88 @@
...
-        stream_type = _url_re.match(self.url).group(1).upper()
-        cookie = {
-            "NRK_PLAYER_SETTINGS_{0}".format(stream_type): COOKIE_PARAMS
-        }
-
-        # Construct API URL for this program.
-        baseurl = self.session.http.get(self.url, cookies=cookie, schema=_schema)
-        program_id = _id_re.search(self.url).group(1)
+        # Construct manifest URL for this program.
+        program_type, program_id = self._url_re.search(self.url).groups()
```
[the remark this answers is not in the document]
sn4kebite: `re.search` has been changed to `re.match` which I assumed is what you meant.

src/streamlink/plugins/nrk.py
```diff
@@ -1,64 +1,82 @@
...
+            reason = manifest["nonPlayable"]["reason"]
+            log.error(f"Not playable ({reason})")
+            return None
+        asset = manifest['playable']['assets'][0]
+
+        # Some streams such as podcasts are not HLS but plain files.
+        if asset['format'] == 'HLS':
+            return HLSStream.parse_variant_playlist(self.session, asset['url'])
+        else:
+            return [(self._get_title(manifest), HTTPStream(self.session, asset['url']))]
```
bastimeyer, line 73: What's the title data here, by the way? `_get_streams` is supposed to return a list of tuples or yield tuples in the `name, stream-instance` format. This name however is not supposed to be an actual title, just the name of the stream quality/variant, like `"1080p60"` for example. If it's an actual title, aka stream metadata, then this title has to be returned from the `get_title` method (`get_category` and `get_author` are also available). Since you're returning an HTTPStream here, the common method is returning `"live"` as stream name.

src/streamlink/plugins/nrk.py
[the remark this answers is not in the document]
bastimeyer: Technically it should be a PluginError here, but since the regex already covers the possible values of the map and the error should thus never be raised, I don't think it's important. If you want to change it, you can do so, but if you don't then it's fine too.

src/streamlink/plugins/nrk.py
sn4kebite: Thanks for pointing this out, I wasn't aware of this. Title and category should now be properly set.

src/streamlink/plugins/nrk.py
```diff
@@ -1,64 +1,95 @@
...
+    _psapi_url = 'https://psapi.nrk.no'
+    # Program type to manifest type mapping
+    _program_type_map = {
+        'direkte': 'channel',
+        'serie': 'program',
+        'program': 'program',
+        'podkast': 'podcast',
+    }
+
+    _url_re = re.compile(r"https?://(?:tv|radio)\.nrk\.no/(program|direkte|serie|podkast)(?:/.+)?/([^/]+)")
```
back-to, line 22: this needs an update, you will catch URLs without a program_id

---

url without program_id
https://tv.nrk.no/serie/kongelige-fotografer/sesong/1/episode/2/avspiller

url with program_id
https://tv.nrk.no/serie/nytt-paa-nytt/2020/MUHH43003020

---

add new url tests in https://github.com/streamlink/streamlink/blob/master/tests/plugins/test_nrk.py

sn4kebite: Thanks, I somehow missed this. There doesn't appear to be a consistent `program_id` format according to the docs (they list 6 different formats), so I think the best way to handle this is to try to extract the `nrk:program-id` meta property, then fall back to the last part of the URL if not found. The fallback is needed for live streams under `/direkte/` to work.

PR #3324, opened 2020-11-08
plugins: turn mux-subtitles into a global argument

bastimeyer:
Closes #3282 

Split into two commits. See the commit messages for the descriptions.

----

Since the argparse docs extension had to be refactored, here's a demonstration of the overall CLI docs changes, to make sure that nothing else changed:

```bash
diff <(curl -s https://streamlink.github.io/latest/cli.html) <(cat ./docs/_build/html/cli.html)
```

```diff
10c10
<   <title>Command-Line Interface &mdash; Streamlink 1.7.0+52.gf0b35b6 documentation</title>
---
>   <title>Command-Line Interface &mdash; Streamlink 1.7.0+54.g49ce436 documentation</title>
31c31
<     <link rel="top" title="Streamlink 1.7.0+52.gf0b35b6 documentation" href="index.html"/>
---
>     <link rel="top" title="Streamlink 1.7.0+54.g49ce436 documentation" href="index.html"/>
56c56
<   <span class="wy-side-nav-versions-current"><i class="fa fa-book"></i>1.7.0+52.gf0b35b6</span>
---
>   <span class="wy-side-nav-versions-current"><i class="fa fa-book"></i>1.7.0+54.g49ce436</span>
1382a1383,1390
> <dl class="option">
> <dt id="cmdoption-mux-subtitles">
> <code class="descname">--mux-subtitles</code><code class="descclassname"></code><a class="headerlink" href="#cmdoption-mux-subtitles" title="Permalink to this definition">¶</a></dt>
> <dd><p>Automatically mux available subtitles into the output stream.</p>
> <p>Needs to be supported by the used plugin.</p>
> <p><strong>Supported plugins:</strong> funimationnow, pluzz, rtve, svtplay, vimeo</p>
> </dd></dl>
> 
1610,1616d1617
< <dt id="cmdoption-funimation-mux-subtitles">
< <code class="descname">--funimation-mux-subtitles</code><code class="descclassname"></code><a class="headerlink" href="#cmdoption-funimation-mux-subtitles" title="Permalink to this definition">¶</a></dt>
< <dd><p>Enable automatically including available subtitles in to the output
< stream.</p>
< </dd></dl>
< 
< <dl class="option">
1688,1699d1688
< <dt id="cmdoption-pluzz-mux-subtitles">
< <code class="descname">--pluzz-mux-subtitles</code><code class="descclassname"></code><a class="headerlink" href="#cmdoption-pluzz-mux-subtitles" title="Permalink to this definition">¶</a></dt>
< <dd><p>Automatically mux available subtitles in to the output stream.</p>
< </dd></dl>
< 
< <dl class="option">
< <dt id="cmdoption-rtve-mux-subtitles">
< <code class="descname">--rtve-mux-subtitles</code><code class="descclassname"></code><a class="headerlink" href="#cmdoption-rtve-mux-subtitles" title="Permalink to this definition">¶</a></dt>
< <dd><p>Automatically mux available subtitles in to the output stream.</p>
< </dd></dl>
< 
< <dl class="option">
1742,1747d1730
< <dt id="cmdoption-svtplay-mux-subtitles">
< <code class="descname">--svtplay-mux-subtitles</code><code class="descclassname"></code><a class="headerlink" href="#cmdoption-svtplay-mux-subtitles" title="Permalink to this definition">¶</a></dt>
< <dd><p>Automatically mux available subtitles in to the output stream.</p>
< </dd></dl>
< 
< <dl class="option">
1813,1818d1795
< <dt id="cmdoption-vimeo-mux-subtitles">
< <code class="descname">--vimeo-mux-subtitles</code><code class="descclassname"></code><a class="headerlink" href="#cmdoption-vimeo-mux-subtitles" title="Permalink to this definition">¶</a></dt>
< <dd><p>Automatically mux available subtitles in to the output stream.</p>
< </dd></dl>
< 
< <dl class="option">
1925c1902
<             VERSION:'1.7.0+52.gf0b35b6',
---
>             VERSION:'1.7.0+54.g49ce436',
```

### DIFF
--- a/docs/ext_argparse.py
+++ b/docs/ext_argparse.py
@@ -2,9 +2,6 @@
 
 Inspired by sphinxcontrib.autoprogram but with a few differences:
 
-- Instead of relying on private argparse structures uses hooking
-  to extract information from a argparse parser.
-
 - Contains some simple pre-processing on the help messages to make
   the Sphinx version a bit prettier.
 
@@ -13,7 +10,6 @@ Inspired by sphinxcontrib.autoprogram but with a few differences:
 import argparse
 import re
 
-from collections import namedtuple
 from textwrap import dedent
 
 from docutils import nodes
@@ -23,9 +19,6 @@ from docutils.statemachine import ViewList
 from sphinx.util.nodes import nested_parse_with_titles
 
 
-_ArgumentParser = argparse.ArgumentParser
-_Argument = namedtuple("Argument", ["args", "options"])
-
 _block_re = re.compile(r":\n{2}\s{2}")
 _default_re = re.compile(r"Default is (.+)\.\n")
 _note_re = re.compile(r"Note: (.*)(?:\n\n|\n*$)", re.DOTALL)
@@ -34,27 +27,8 @@ _option_re = re.compile(r"(?:^|(?<=\s))(--\w[\w-]*\w)\b")
 _prog_re = re.compile(r"%\(prog\)s")
 
 
-class ArgumentParser(object):
-    def __init__(self, *args, **kwargs):
-        self.args = args
-        self.kwargs = kwargs
-        self.groups = []
-        self.arguments = []
-
-    def add_argument(self, *args, **options):
-        if not options.get("help") == argparse.SUPPRESS:
-            self.arguments.append(_Argument(args, options))
-
-    def add_argument_group(self, *args, **options):
-        group = ArgumentParser(*args, **options)
-        self.groups.append(group)
-        return group
-
-
 def get_parser(module_name, attr):
-    argparse.ArgumentParser = ArgumentParser
     module = __import__(module_name, globals(), locals(), [attr])
-    argparse.ArgumentParser = _ArgumentParser
     parser = getattr(module, attr)
     return parser if not(callable(parser)) else parser.__call__()
 
@@ -110,34 +84,45 @@ class ArgparseDirective(Directive):
         return indent(help)
 
     def generate_group_rst(self, group):
-        for arg in group.arguments:
-            help = arg.options.get("help")
-            metavar = arg.options.get("metavar")
+        for action in group._group_actions:
+            # don't document suppressed parameters
+            if action.help == argparse.SUPPRESS:
+                continue
 
+            metavar = action.metavar
             if isinstance(metavar, tuple):
                 metavar = " ".join(metavar)
 
-            if metavar:
-                options = []
-                for a in arg.args:
-                    if a.startswith("-"):
-                        if arg.options.get("nargs") == "?":
-                            metavar = "[{0}]".format(metavar)
-                        options.append("{0} {1}".format(a, metavar))
-                    else:
-                        options.append(metavar)
+            options = []
+            # parameter(s) with metavar
+            if action.option_strings and metavar:
+                for arg in action.option_strings:
+                    # optional parameter value
+                    if action.nargs == "?":
+                        metavar = f"[{metavar}]"
+                    options.append(f"{arg} {metavar}")
+            # positional parameter
+            elif metavar:
+                options.append(metavar)
+            # parameter(s) without metavar
             else:
-                options = arg.args
+                options += action.option_strings
 
-            yield ".. option:: {0}".format(", ".join(options))
+            yield f".. option:: {', '.join(options)}"
             yield ""
-            for line in self.process_help(help).split("\n"):
+            for line in self.process_help(action.help).split("\n"):
                 yield line
             yield ""
+            if hasattr(action, "plugins") and len(action.plugins) > 0:
+                yield f"    **Supported plugins:** {', '.join(action.plugins)}"
+                yield ""
 
     def generate_parser_rst(self, parser):
-        for group in parser.groups:
-            title = group.args[0]
+        for group in parser._action_groups:
+            # Exclude empty groups
+            if len(group._group_actions) == 0:
+                continue
+            title = group.title
             yield ""
             yield title
             yield "^" * len(title)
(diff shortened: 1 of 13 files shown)


back-to:
might be useful if you also add this plugin option to the log output

`plugin.global_arguments`

https://github.com/streamlink/streamlink/blob/49ce43614e7c364775fe640500048cba8dab399f/src/streamlink_cli/main.py#L557-L564

maybe not important for `mux-subtitles`,
but if we decide to add `purge-credentials` `username` `password` in the future as global args it would be better with a log output.

bastimeyer:
That would require refactoring the `Argument` class or adding a separate class for the global arguments, because the arguments of the global arguments list are just references and will be logged differently (incorrectly). And arguments with sensitive data make this even more complicated, because the global argument which is being referenced needs to be looked up.

Regarding the other arguments which you've mentioned, I'm not sure if making them global is a good idea, because they depend on the input URL what's being done, and this can be confusing or lead to user issues when used incorrectly, especially purge-credentials. Global username/email/password/token/etc arguments could work, but since there's no unique scheme/pattern, that would require having a lot of global arguments and the list of supported plugins would be confusing. The mux-subtitles argument works in a global context because it's an independent thing.

src/streamlink/plugins/vimeo.py
```diff
@@ -50,12 +50,8 @@ class Vimeo(Plugin):
         validate.any(None, validate.Schema(validate.get(1), _config_schema)),
     )
 
-    arguments = PluginArguments(
-        PluginArgument(
-            "mux-subtitles",
-            action="store_true",
-            help="Automatically mux available subtitles in to the output stream.",
-        )
+    global_arguments = PluginArguments(
```
beardypig, line 53: Instead of adding `global_arguments` what about adding a flag to `PluginArgument` to mark it as global. That way there is no extra `Plugin` parameter/attribute. The processing of the arguments would remain almost identical but I think the Plugin interface would be nicer.

bastimeyer: My initial idea was to not touch the Argument class, hence the `global_arguments` attribute, but I agree that this would be a better solution. Even better by subclassing the Argument class, so no parameter has to be set and the distinction becomes more clear. I will take a look at this later.

bastimeyer:
Added the `is_global` parameter to the `Argument` class and refactored the `streamlink_cli.main` methods. The global plugin arguments will now correctly be debug-logged (what @back-to mentioned), just like the regular plugin arguments. Documentation-wise, nothing has changed from the previous implementation.

Just noticed some changes I've missed which should have been reverted. Will force-push again in a minute...

beardypig:
Looks good, but I'd like to add somethings to the tests. Lemme see if it's feasible....

beardypig:
As an additional note, for this to really make sense I think we need to move the argument definition from the CLI module in to the main `streamlink` module otherwise the global arguments are coupled to the CLI.

beardypig:
The test I added shows the strong coupling between the CLI and the main module.

docs/ext_argparse.py
```diff
@@ -110,34 +84,45 @@ def process_help(self, help):
...
-            yield ".. option:: {0}".format(", ".join(options))
+            yield f".. option:: {', '.join(options)}"
             yield ""
-            for line in self.process_help(help).split("\n"):
+            for line in self.process_help(action.help).split("\n"):
                 yield line
             yield ""
+            if hasattr(action, "plugins") and len(action.plugins) > 0:
+                yield f"    **Supported plugins:** {', '.join(action.plugins)}"
+                yield ""
```
beardypig, line 118: nice touch!

beardypig (approved):
Looks good to me. @bastimeyer should take a look at the meta-test I added though.

bastimeyer:
Test looks good... Theoretically this could even be extended a bit by checking that no other argument parameters are set, as the global arguments are not supposed to receive any parameters other than the name, but I don't think it's necessary.

> for this to really make sense I think we need to move the argument definition from the CLI module in to the main streamlink module otherwise the global arguments are coupled to the CLI.

That was already the case previously with the regular plugin arguments, because they are reading their values from `streamlink_cli.main.args` as well. Strictly speaking, there is no dependency on `streamlink_cli`, as `streamlink_cli` itself sets the values on the plugin's `options` dict, and without it, the dict is just empty, but that doesn't affect the plugin logic.

Why do we even have a separate `args` variable, when the `session.options` dict exists? Isn't that supposed to contain all argument values? Right now, it's just an additional dict with hand-picked arguments and a little bit of value mapping/customization (which btw isn't particularly efficient).

beardypig:
I mean that the definition of the "global arguments" should be moved to the `streamlink` module, probably using a class similar to `PluginArgument` to define them. The important aspect here is the definition of the arguments. There is no definition of what `mux-subtitles` is outside of the `streamlink_cli` module, which is why the new test I wrote has to import `streamlink_cli.argparser.build_parser`. 

`streamlink_cli.main.args` is the direct output from the `argparse` parser, there still needs to be a separation between the CLI and `Streamlink` arguments, all the player options, file writing options, etc. are interface specific but all the HLS options, HTTP options, etc. could have their definitions moved to `Streamlink`. I am imagining the `Streamlink`/`Pluigin` arguments just have to be defined in one place, this would make a new front-end implementation a bit easier.

bastimeyer:
> there still needs to be a separation between the CLI and Streamlink arguments, all the player options, file writing options

True, I didn't think about that while writing my previous comment.

If you want to move all "global arguments" to `streamlink`, then this would include every argument that's not strictly CLI-related, so basically everything that's in the `Session` options, and you'd have to feed those arguments to `streamlink_cli` in a similar fashion as the plugin arguments.

That's not an issue of this PR though.

Btw, decoupling plugin arguments from their plugin modules is another issue and would probably an annoyance later on if we change/improve the plugin import system.

beardypig:
That is what I was thinking. Perhaps I will solidify my ideas in to an issue/PR to be worked on/discussed.

beardypig:
That is what I was thinking. Perhaps I will solidify my ideas in to an issue/PR to be worked on/discussed.

bastimeyer:
Any further concerns? I think this should be good now. All other things mentioned here are unrelated to the specific changes here.

Issue #3317, opened 2020-11-04
Streaming www.huya.com/kpl results in only one frozen frame


RasmusJohns:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description

Last week, using streamlink on huya.com/kpl was not an issue. However, since last night is has not been working for me on any machine. The stream launches, but does not seem to generate any new frames -- it is simply stuck showing a rather old frame.

Logs suggest packages are not sent (if I understand TS discontinuity correctly), but running the stream in the browser works fine for me.

### Expected / Actual behavior

Should generate a videostream with valid frames

### Reproduction steps / Explicit stream URLs to test

run `streamlink https://www.huya.com/kpl best`

### Log output

Here are two logs: one using `--verbose` and one using `--loglevel debug`

```
 % streamlink --verbose https://www.huya.com/kpl best
[cli][info] Found matching plugin huya for URL https://www.huya.com/kpl
[cli][info] Available streams: live (worst, best)
[cli][info] Opening stream: live (hls)
[cli][info] Starting player: /Applications/VLC.app/Contents/MacOS/VLC
VLC media player 3.0.10 Vetinari (revision 3.0.10-0-g7f145afa84)
[00007ffd0d104950] main libvlc: Running vlc with the default interface. Use 'cvlc' to use vlc without interface.
[00007ffd122495e0] videotoolbox decoder: Using Video Toolbox to decode 'h264'
[00007ffd1211eea0] mpeg4audio packetizer: AAC channels: 1 samplerate: 22050
[00007ffd122495e0] videotoolbox decoder: Using Video Toolbox to decode 'h264'
[00007ffd122495e0] videotoolbox decoder: vt cvpx chroma: 420v
[00007ffd0bdee710] ts demux error: libdvbpsi error (PSI decoder): TS discontinuity (received 0, expected 4) for PID 17
[00007ffd0bdee710] ts demux error: libdvbpsi error (PSI decoder): TS discontinuity (received 0, expected 4) for PID 17
[00007ffd0bdee710] ts demux error: libdvbpsi error (PSI decoder): TS discontinuity (received 0, expected 4) for PID 17
[00007ffd0bdee710] ts demux error: libdvbpsi error (PSI decoder): TS discontinuity (received 0, expected 4) for PID 17
...
```

```
% streamlink --loglevel debug https://www.huya.com/kpl best
[cli][debug] OS:         macOS 10.15.5
[cli][debug] Python:     3.8.6
[...]

PR #3325, opened 2020-11-08
plugins.huya: use FLV stream with multiple mirrors

back-to:
Not every mirror works good for every user location,
the users must select the mirror which works for them.

```
$ streamlink https://www.huya.com/kpl -l info
[cli][info] Found matching plugin huya for URL https://www.huya.com/kpl
[cli][info] Available streams: source_al13, source_al (worst), source_tx, source_bd, source_tx15 (best)
[cli][info] Opening stream: source_tx15 (http)
```
closes https://github.com/streamlink/streamlink/issues/3317


RasmusJohns:
I can confirm that this works as expected. Thanks @back-to

PR #3326, opened 2020-11-08
plugins.cdnbg: simplify and fix iframes without schema

enilfodne:
Currently `plugins.cdnbg` has a bug where `urlparse.scheme` is being passed as arg to `urlparse`  (used in  `update_schema` ), but `urlparse` treats it as path (as its simply the protocol name, not a URL). Thus for a URL without a specified protocol, the script fails.

Changes:
- Remove a redundant `if` statement as `update_scheme` handles both cases (with/without scheme)
- Remove a redundant `if` statement as `iframe_url` is always defined
- Use `html.unescape` to unescape HTML-encoded strings (should me more of a robust)
- More robust check (via parsing), whether the supplied URL is already a target URL

### DIFF
--- a/src/streamlink/plugins/cdnbg.py
+++ b/src/streamlink/plugins/cdnbg.py
@@ -1,5 +1,6 @@
 import logging
 import re
+from html import unescape as html_unescape
 from urllib.parse import urlparse
 
 from streamlink.plugin import Plugin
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/cdnbg.py
```diff
@@ -43,29 +44,25 @@ def can_handle_url(cls, url):
...
-                iframe_url = iframe_url.replace("&#58;", ":")
-                if iframe_url.startswith("//"):
-                    return update_scheme(p.scheme, iframe_url)
-                else:
-                    return iframe_url
+                return update_scheme(self.url, iframe_url)
 
     def _get_streams(self):
-        if "i.cdn.bg/live/" in self.url:
+
```
back-to, line 54: remove this line

```suggestion
```

src/streamlink/plugins/cdnbg.py
```diff
@@ -43,29 +44,25 @@ def can_handle_url(cls, url):
     def find_iframe(self, url):
         self.session.http.headers.update({"User-Agent": useragents.CHROME})
         res = self.session.http.get(self.url)
-        p = urlparse(url)
         for iframe_url in self.iframe_re.findall(res.text):
             if "googletagmanager" not in iframe_url:
+                iframe_url = html_unescape(iframe_url)
                 log.debug(f"Found iframe: {iframe_url}")
```
back-to, line 50:
```suggestion
```

src/streamlink/plugins/cdnbg.py
```diff
@@ -43,29 +44,25 @@ def can_handle_url(cls, url):
...
             iframe_url = self.find_iframe(self.url)
 
-        if iframe_url:
-            res = self.session.http.get(iframe_url, headers={"Referer": self.url})
-            stream_url = update_scheme(self.url, self.stream_schema.validate(res.text))
-            log.warning("SSL Verification disabled.")
-            return HLSStream.parse_variant_playlist(self.session,
-                                                    stream_url,
-                                                    verify=False)
+        res = self.session.http.get(iframe_url, headers={"Referer": self.url})
```
back-to, line 60:
```suggestion
        log.debug(f"Found iframe: {iframe_url}")
        res = self.session.http.get(iframe_url, headers={"Referer": self.url})
```

PR #3327, opened 2020-11-09
Add a note about dropping old Windows support

### LINKED ISSUE #3323: Installer incompatible with Windows 7
Tried latest development build (streamlink-1.7.0_52.gf0b35b6-win32), issue is installer doesn't add the path environment variable. latest stable (1.7.0) adds it without a problem.

second issue is python 3.9+ does not work on Windows 7 anymore. 3.8 branch still works on Windows 7, maybe using 3.8.6 instead of 3.9 would be better? or at least specify in docs that streamlink windows installer won't work on Win7 anymore.
But is 3.9 really needed?

gravyboat:
Closes #3323

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -40,6 +40,12 @@ Portable                             See the `Windows portable version`_ section
 
 .. _Installing Winget packages: https://docs.microsoft.com/en-us/windows/package-manager/
 
+.. note::
+
+    As of Streamlink version 2.0.0 this software will no longer operate on Windows 7 or Windows Vista.
+    These operationg systems do not support Python 3.9 which is now required. We recommend upgrading your
+    operating system as both versions of Windows noted above are now EoL and receive no support by Microsoft.
+
 macOS
 -----
 


docs/install.rst
```diff
@@ -40,6 +40,12 @@ Portable                             See the `Windows portable version`_ section
 
 .. _Installing Winget packages: https://docs.microsoft.com/en-us/windows/package-manager/
 
+.. note::
+
+    As of Streamlink version 2.0.0 this software will no longer operate on Windows 7 or Windows Vista.
```
back-to, line 45: technically it is just the installer, not Streamlink

bastimeyer (changes requested):
This is what I had written yesterday before committing my changes:

```diff
diff --git a/docs/install.rst b/docs/install.rst
index 59382d5b..78453020 100644
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -177,6 +177,7 @@ Windows port. version                RosadinTV <RosadinTV at outlook.com> |br|
                                      beardypig <beardypig at protonmail.com>
 ==================================== ===========================================
 
+.. _PyPI package and source code:
 
 PyPI package and source code
 ----------------------------
@@ -285,6 +286,8 @@ using `virtualenv`_, which creates a user owned Python environment instead.
 
 .. _virtualenv: https://virtualenv.readthedocs.io/en/latest/
 
+.. _Dependencies:
+
 Dependencies
 ^^^^^^^^^^^^
 
@@ -357,6 +360,17 @@ Release                              Notes
                                      See the `commit log`_ for a list of changes since the last stable release.
 ==================================== ====================================
 
+.. note::
+
+    **Both Streamlink installers for Windows require at least Windows 8.** |br|
+    Windows 7, Vista and XP are not supported due to limitations of Python 3.9+.
+
+    Be aware that the packages for `Chocolatey`_ and the `Windows Package Manager`_ are just wrappers
+    around the stable installer and thus depend on Windows 8+ as well.
+
+    Alternatively, :ref:`Streamlink can be installed via python-pip<PyPI package and source code>`
+    in a :ref:`compatible<Dependencies>` Python environment.
+
 .. _Stable release:
 .. _GitHub releases page: https://github.com/streamlink/streamlink/releases/latest
 .. _Development build:
```

Instead of adding a note at the top, this adds one at the bottom where the installers are actually linked. We don't have any notes in the macOS and Linux+BSD package sections, so adding it to the bottom is much cleaner.

This note also includes links to the mentioned sections or packages. Unfortunately, the section headers had to be updated with custom reference points, because this isn't done automatically, and when adding an extension to sphinx for this auto-header-linking, other issues need to be resolved first, because there are duplicate header names.

If we decide to downgrade the installer's Python version to 3.8, the text obviously needs to be changed/corrected.

mkbloke:
Seeing this and with reference to my comment on Gitter earlier today, would there be any value in also adding a note for Linux users, just to try and stave off any unnecessary issues when 2.0.0 is released?  Perhaps something like:

---

note:

For Streamlink versions >= 2.0.0:

Users of older but currently supported Linux distributions with Python 3.5 will need to install a Python package with a minimum version of 3.6 that can coexist with the system version.

For Python up to version 3.9 on Ubuntu 16.04, install the PPA by following the instructions [here](https://github.com/deadsnakes/issues/blob/master/README.md).  Then `sudo apt-get update` and `sudo apt-get install python3.9 python3.9-venv`.  Create the virtual environment with `virtualenv -p3.9 venv_dir` and then follow the pip install instructions for Streamlink in a virtual environment for either a release version or from master on GitHub.

For Python up to version 3.7 on Debian Stretch, add the repository by following the instructions [here](https://people.debian.org/~paravoid/python-all/).  Then `sudo apt-get update` and `sudo apt-get install python3.7 python3.7-venv`.  Create the virtual environment with `virtualenv -p3.7 venv_dir` and then follow the pip install instructions for Streamlink in a virtual environment for either a release version or from master on GitHub.

gravyboat:
@bastimeyer Your reasoning on the section makes sense to me and the text is fine, I'll close this.

bastimeyer:
@mkbloke No, this doesn't make much sense, as the available Streamlink packages on various Linux distros are built against the distro's default Python env. If this Python env is too old, then there won't be any new Streamlink packages available. Ubuntu 16.04 for example doesn't have any Streamlink packages, and Debian Stretch's last Streamlink package version is 1.0.0, which shows you that it's irrelevant that Python 2.7 and 3.5 have been dropped. It also doesn't matter if old distro releases which "supported" are still on Python 3.5, because this version is not maintained by the Python devs anymore. We don't need to bloat up the install docs for old and irrelevant Linux distro releases. There already is an entire section on how to install Streamlink via python-pip and in a virtualenv and it already tells the user to do this if a newer package version is not available on their distro.

@gravyboat You could have just updated your PR with my diff suggestion. But since you've already closed yours and deleted your branch, I'm going to open one myself now with an updated note.

gravyboat:
@bastimeyer Ahh sorry I thought you already had a branch waiting and just hadn't opened the PR.

docs/install.rst
beardypig: the version installed by the installer ;)

karlo2105:
Can't you allow Windows 7 support at least till 2023 as it's still supported by ESU updates ?

gravyboat:
@karlo2105 We're not going to do that. Extended security updates are typically only for businesses and they cost money. Streamlink was never intended or designed for businesses to use on their old outdated machines. If anyone is using Streamlink in a business context they can upgrade.

mkbloke:
@karlo2105, pip installs require Python >= 3.6.

mkbloke:
@karlo2105, another thought:

In *theory* this won't really affect you anyway.  The Windows installer was downgraded to Python 3,8 to allow for installation on Windows 7, it's just **not officially supported** by Streamlink and requires a **fully patched** Windows 7 system to work.

bastimeyer:
#3330, #3331

https://github.com/streamlink/streamlink/blame/master/docs/install.rst#L358-L367

PR #3330, opened 2020-11-09
installer: downgrade to python 3.8

### LINKED ISSUE #3323: Installer incompatible with Windows 7

bastimeyer:
Resolves #3323 

This re-enables support for Windows 7, even though it has already seen its EOL in January this year.

Downgrading does only make sense considering that Windows 7 still has a high share of users among the overall number of Windows users. Forcing Python 3.9 and making the installers incompatible would result in us seeing more threads like #3323, with possibly even more complaints and/or rude/emotional comments, and I don't have any interest in dealing with this and wasting my time even more.

As said, if someone even has issues with the upgrade to Python 3.8, they can install Streamlink via pip in a separate Python 3.6 environment, as it's the lowest Python version that's currently supported. Or they can just be a responsible person and upgrade/switch their unsupported operating system.


gravyboat:
I disagree with this and think that the logic of `Windows 7 still has a high share of users among the overall number of Windows users.` is bad. It only has a high share because software continues supporting that insecure operating system, and I say this as someone running Windows 10 who doesn't like it and held on to running Windows 7 for as long as I reasonably could.

Having said that the verdict right now is 2-1 so let's merge this and be done with it.

beardypig:
I didn't actually suggest using 3.8 to support Windows 7 :) 

@gravyboat I don't know why so many people to Windows 7, maybe because they think Windows 10 is bad or whatever. But, I do know that some people use it because software _isn't_ supported on Windows 7+. I know several people that do not want to upgrade from Windows 7 because that would mean their older versions of software they use will no longer work, and they'd have to upgrade to the newest versions which are all subscription based. Also, that old cliche about change being unpopular or whatever it is ;)

gravyboat:
@beardypig There are a fair number of security related concerns with Microsoft and monitoring on Windows 10 (pretty easy to work around though which I have done myself) so for a long time that was a valid reason to stay on 7 since it was still supported. When the EoL date was announced I started the process of upgrading and researching what needed to be done to protect my privacy with Windows 10. I imagine this is why many people are still using Windows 7. The point about subscription based software is kind of valid, but if you don't like a company's business model just stop using their product or run the software in a VM.

Bastimeyer's note about additional issues being created is probably the most valid after this recent Twitch debacle. No one pays us for Streamlink so I don't care if we don't support someone's operating system (even if it was my own OS if the usage is low), it's open source software, deal with it, fork us, or find other software. I'm really against a small number of users (~25% of people are still on Windows 7 apparently) deciding how we run this project simply because they are loud in their complaints.

beardypig:
@gravyboat I think we are on the same page ;D

PR #3331, opened 2020-11-10
docs: add note about supported Windows versions

bastimeyer:
See #3323, #3330

Linking to certain sections of the document unfortunately requires adding custom references.

A better solution would be adding the `sphinx.ext.autosectionlabel` Sphinx extension to the docs config, but this requires fixing some headers in other sections first, because it doesn't allow duplicate header names (which is stupid).


gravyboat:
Looks good to me.

PR #3333, opened 2020-11-11
docs: fix links

bastimeyer:
**First commit:**
- adds `autosectionlabel` Sphinx extension, so that headlines don't need to be labelled manually
  see https://www.sphinx-doc.org/en/master/usage/extensions/autosectionlabel.html
- sets `autosectionlabel_prefix_document` to True, so that no duplicates exist
- removes custom headline labels
- fixes all `:ref:` links with properly prefixed references

**Second commit:**
- replaces all `http://` links with `https://` where possible
  Debian repos and Ubuntu PPAs are all non-TLS, so those are the last remaining http links
- sets new/proper links where the site was either outdated, unofficial or offline
  - "What are cookies" replaced with Wikipedia HTTP-Cookies page
  - MS doesn't document its old Windows command prompt, link should probably be removed, as it's documenting the available commands/utilities
  - Added Windows Terminal docs link
  - Switched from Wikipedia to Apple's Terminal app guide page
- re-orders players table, moves its links to the section footer and adds OMXPlayer link

### DIFF
--- a/docs/_applications.rst
+++ b/docs/_applications.rst
@@ -1,5 +1,3 @@
-.. _applications:
-
 :orphan:
 
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,5 +1,3 @@
-.. _api:
-
 API Reference
 =============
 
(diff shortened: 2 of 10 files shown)


gravyboat (approved):
This looks good to me! If no one else leaves a comment or request for changes let's merge.

beardypig (approved):

PR #3335, opened 2020-11-13
docs: switch theme to furo

### LINKED ISSUE #2851: Suggestion to replace custom sphinx theme with official sphinx rtd theme

bastimeyer:
Resolves #2851 

This switches the docs theme from the custom `sphinx_rtd_theme_violet` theme to `furo`:
https://github.com/pradyunsg/furo

- Sets `furo` docs requirement to `==2020.12.09.beta21`, as it's changing quickly with breaking changes due to not having a major release yet
- Sets `sphinx` docs requirement to `>=3.0`, as it's a requirement of furo

As usual, everything is grouped into logical commits with a proper description in the commit messages.

----

#### Current theme modifications

- Added logo, project title, oneliner and version selection, which now remembers the current page when switching. Since this all pushes the main menu down by a bit, I've tried to minimize the height as much as possible.
- Added proper brand colors, so that it matches the Streamlink icon colors. This makes the menu links a bit darker, but this is preferred, so that it's "out of focus" while reading the page content.
- Fixed the font sizes of various components and removed the 110% font-size media query on "large screens", as it's stupid and makes the font a bit blurry (`16px*1.1 = 17.6px`).
- Improved the font-stack a bit by adding more names for Apple's San Francisco font.
- Fixed scrollbar issues in both sidebars with custom styles overrides. This should now work on all display resolutions.

----

#### Additional docs changes

- Changed the style of the inline `:command:`s from bold text to inline code.
- Split the plugin arguments list into nested groups, so that it's easier to find them via the second TOC menu.

----

#### Remaining issues

None

----

#### Thoughts

- The CLI argument default values should be placed right under the argument names, so that they don't appear in the middle of the description. This can be done by improving the regex in the argparse extension and by adding a special syntax to the argument descriptions for this. Not sure how to connect these changes with the HTML output, as the man pages need to keep working.

----

image

image


docs/_static/styles/custom.css
```diff
@@ -0,0 +1,155 @@
...
+  --admonition-title-font-size: var(--font-size--normal);
+}
+
+/*
+  Generic style overrides
+*/
+
+html {
+  /* unset @media (min-width: $full-width + $sidebar-width) query which sets font-size to 110% */
+  font-size: 100% !important;
```
pradyunsg, line 25: Suggestion: don't override this, since that's a pattern followed by many other themes -- it makes the reading experience on larger format displays and ultra-wide monitors better.

pradyunsg: See https://github.com/squidfunk/mkdocs-material/blob/ef5f2930003d1db53e5f7131ddcfff4046663a23/src/assets/stylesheets/main/layout/_base.scss#L37, for another example (although it's burried in a bunch of hacks).

gravyboat: @bastimeyer Do you want to update this?

bastimeyer: No, because that's an opinionated design choice which I don't share. People using bigger and/or wider screen resolutions can choose to increase their desktop's DPI scaling or font scaling if reading text in the browser's default text size is a problem for them, which they probably already have done, making this media query highly questionable, especially for those others who dislike large texts. Apart from that, all text sizes defined on the default scaling are carefully chosen to match whole pixel values, and setting the base scaling to 110% breaks this logic.

Regarding the overall state of the PR, which you're probably not aware of, there are still some unresolved issues with the theme left, like for example scrollbar issues with both sidebars, which I don't want to fix in a custom workaround here and have therefore opened issue on its issue tracker a few days ago. Once this has all been fixed on the theme side, I will update the PR branch and check the remaining issues like the message box colors and the CLI parameter layout.

gravyboat: @bastimeyer I'm aware of some of the outstanding issues per your `to-do` section but I wasn't aware you had opened PRs in the other issue trackers. Thanks for the update.

src/streamlink_cli/main.py
```diff
@@ -866,10 +866,11 @@ def setup_plugin_args(session, parser):
     plugin_args = parser.add_argument_group("Plugin options")
     for pname, plugin in session.plugins.items():
         defaults = {}
+        group = plugin_args.add_argument_group(pname.capitalize())
 
         for parg in plugin.arguments:
             if not parg.is_global:
-                plugin_args.add_argument(parg.argument_name(pname), **parg.options)
+                group.add_argument(parg.argument_name(pname), **parg.options)
```
bastimeyer, line 873: I've just noticed that Python's argparse doesn't properly format help texts when adding nested argument groups, as it only includes the first level of groups. This currently ignores all plugin arguments in the `--help` output.

That means that `argparse.ArgumentParser.format_help` needs to be overridden, which is a bit annoying:
https://github.com/python/cpython/blame/761c5a1ce477b82cca35a79917c88d72d57ed776/Lib/argparse.py#L2521-L2526

The reason why I've nested the plugin arguments this is that it's easier to navigate/browse on the rendered HTML docs. The rendered manpage of course also includes the nested argument groups.

bastimeyer:
The recent push
- fixed remaining scrollbar and content width issues in the sidebars due to the sidebar customization
- fixed the brand colors on the dark theme
- fixed the `--help` output with nested argument groups
- reduced the size of the logo a bit, so that the menu moves further up
- added favicons and a PWA manifest (progressive web apps)

Once furo has had its next release (see remaining issues), I'm going to rebase and resolve the merge conflict (#3379). The PR should then be pretty much ready from my side at least. Please build the docs locally and review.

```sh
pip install -U -r docs-requirements.txt
xdg-open ./docs/_build/html/index.html
```

bastimeyer:
Please build the docs on your systems and give this a final review.

gravyboat:
@bastimeyer I'm getting the following build error on my Windows box, might be okay elsewhere:
```
\docs\index.rst:62:toctree contains reference to document 'changelog' that doesn't have a title: no link will be generated
make: *** [Makefile:45: html] Error 2
```

bastimeyer:
```
$ stat -c %N docs/changelog.md 
'docs/changelog.md' -> '../CHANGELOG.md'
```

Your git-for-Windows is misconfigured and doesn't create softlinks.
https://stackoverflow.com/a/42137273
```
core.symlinks=true
```

gravyboat:
@bastimeyer No it's not that. I already have developer mode enabled and have symlinks enabled for git both globally and for the repo itself. I'll continue investigating.

bastimeyer:
You'll have to re-clone the repo, because otherwise it'll just create file copies, and you probably have an older version of that file there and/or maybe it's empty, because it complains about a missing document title in the changelog file. I don't know how this is handled on Windows. Just use a POSIX environment to build it if you're having trouble getting it to work.

gravyboat:
I did. It was some problem with make itself. Uninstalling and installing the latest via chocolatey fixed it. The docs look good even on my 1440p monitor. I'm marking this as approved.

gravyboat (approved):
Looks good, if no one else has feedback let's merge this.

bastimeyer:
We've talked about this in another PR (can't remember which one), but a docs preview deploy config for PRs would be very useful sometimes.

Just had a look at netlify for the past hour or so:
https://www.netlify.com/blog/2016/07/20/introducing-deploy-previews-in-netlify/

Here's a PR on my test repo with docs deploy statuses:
https://github.com/bastimeyer-test/streamlink-test/pull/8
https://deploy-preview-8--streamlink-test.netlify.app/

Unless I'm mistaken, there doesn't seem to be a way though to make it only build on PR, because we don't want to build on push to master. Auto-building can be disabled so that the content has to be deployed manually from GH actions, but according to some comments on the netlify GH issues, their CLI tool for manually deploying content doesn't support setting PR statuses. This means that the preview-URL would only be printed to stdout in the GH actions build log, which isn't ideal.

Another issue is that it also always triggers a build, which doesn't make sense for non-docs related changes. We could add another GH actions workflow that [only triggers on change in certain paths](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths), and make it build and deploy the docs via the netlify CLI tool. The preview URL could be added via a comment from the streamlink-bot account via a custom script.

gravyboat:
@bastimeyer I want to merge this tomorrow. Do you want to squash down some of these commits or do you prefer they stay?

Regarding Netlify it seems like a pretty good option, unfortunately we can't used the shared account as it would violate their [terms of service](https://www.netlify.com/legal/terms-of-use):

`Each user must have unique login credentials that may not to be shared by multiple users.`

That would make it pretty annoying to manage unless we want to pay and current monthly opencollective contributions would be entirely eaten by just two users at $20 USD each.

bastimeyer:
Fixed an issue with the TOC tree and the current item having bold font-weight, which caused some texts to wrap due to the larger width. The current item in the TOC tree now has a little icon next to it ("TRIANGULAR BULLET" - U+2023).

The commits are grouped into logical chunks and should not be squashed any further. All I can do is squash the :command: commit into the fonts commit, which I just did.

Regarding netlify, FOSS projects can apply for a different plan. They only show this option if you're signed in.
https://opensource-form.netlify.app/

gravyboat:
@bastimeyer That's what I figured regarding the commits but wanted to double check. I'll rebase and merge.

Regarding netlify, I'm not a big fan of some of their requirements for this:

```
Features a Code of Conduct at the top level directory of the project repository or prominently in the documentation (with a link in the navigation, footer, or homepage).
```

We've never had a code of conduct and I don't see a reason to create one now to appease some company.

```
Must feature a link to our service on your main page, or all internal pages. You have two options:

    We have premade badges for your convenience, or
    You may create your own link, which should read “This site is powered by Netlify”, and include a link back to our home page.

```

This requirement is ridiculous. They want prominent placement on the documentation? Absolutely not assuming that this has to apply to more than just the preview build. If they want us to advertise they can sponsor us.

The plan is also still limited to the free usage tier per the bottom of that document so we don't even get anything out of it other than multiple accounts, and let's be real how many people are going to be logging in regularly? It's just a way for them to get free advertising out of us without actually offering anything of significance.

bastimeyer:
@gravyboat Yes, the requirements are a bit ridiculous, but I guess this is mostly meant for projects which want their website to be hosted by them. We only want to preview simple documentation builds, so the requirements don't necessarily have to fully apply to us. If you're okay with this, I'll send them a support email first and ask if it's possible for Streamlink to apply to their open source plan without meeting all strict requirements. If they still insist on us having a CoC, I've already got one for you "be nice and respectful". And if this doesn't work out, I can manage the account on my own just fine, I guess. Nobody else needs to login there anyway and if something breaks or gets annoying, the netlify integration can be removed from the repo here.

> The plan is also still limited to the free usage tier per the bottom of that document 

Their open source plan is the same as the pro plan with unlimited members (see the top), which would be nice in regards to the monthly build time limitation. The link at the bottom just says that they can disable the account at any time if they want, similar to the free/gratis tier.

gravyboat:
@bastimeyer I'm fine with sending them a support email with your proposed details. Let's see what they say. I'm also fine with that CoC if we have to add one. I don't want to put more management work on you is my main problem with the whole process, it just sucks when one person has to handle everything related to a specific aspect of the project.

> The link at the bottom just says that they can disable the account at any time if they want, similar to the free/gratis tier.

Ahh my misunderstanding then.

bastimeyer:
I'm a little bit annoyed. You can't send them a regular support email, not even via the support form on their site unless you're a paying customer and you'll have to use the community forums instead. I don't want to post this there. Sent it on the enterprise contact form instead, lol...

gravyboat:
Yikes. I can understand not wanting to support customers who don't pay but that seems a bit extreme.

PR #3338, opened 2020-11-16
plugins.mitele: update plugin to support new website APIs

### LINKED ISSUE #3336: Mitele changed
## Plugin Issue

Mitele has changed its code.

### Description

Plugin not able to handle the new configuration.

### Log output

```
[cli][info] Found matching plugin mitele for URL https://mitele.es/directo/telecinco
error: Unable to open URL: https://gatekeeper.mediaset.es (HTTPSConnectionPool(host='gatekeeper.mediaset.es', port=443): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x7ff07c967220>: Failed to establish a new connection: [Errno -2] Name or service not known')))
```
### Additional comments, etc.
My english and knowledge are poor, sorry.

beardypig:
The API changed a bit and required a few updates. This should work for all the channels except the one premium channel which requires a subscription and uses DRM so cannot be supported. 

fixes #3336

### DIFF
--- a/src/streamlink/plugins/mitele.py
+++ b/src/streamlink/plugins/mitele.py
@@ -73,7 +73,7 @@ class Mitele(Plugin):
                                         headers={"origin": "https://www.mitele.es"},
                                         schema=self.cerbero_schema)
 
-        if "code" in pdata:
+        if "code" in tokens:
             log.error("Could not get stream tokens: {} ({})".format(tokens["code"],
                                                                     self.token_errors.get(tokens["code"], "unknown error")))
             return
--- a/src/streamlink/plugins/mitele.py
+++ b/src/streamlink/plugins/mitele.py
@@ -4,7 +4,8 @@ import re
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
 from streamlink.stream import HLSStream
-from streamlink.utils import parse_json
+from streamlink.utils import parse_json, parse_qsd
+from streamlink.utils.url import update_qsd
 
 log = logging.getLogger(__name__)
 
@@ -12,40 +13,42 @@ log = logging.getLogger(__name__)
 class Mitele(Plugin):
     _url_re = re.compile(r"https?://(?:www\.)?mitele\.es/directo/(?P<channel>[\w-]+)")
 
-    pdata_url = "https://indalo.mediaset.es/mmc-player/api/mmc/v1/{channel}/live/html5.json"
-    gate_url = "https://gatekeeper.mediaset.es"
+    caronte_url = "https://caronte.mediaset.es/delivery/channel/mmc/{channel}/mtweb"
+    gbx_url = "https://mab.mediaset.es/1.0.0/get?oid=mtmw&eid=%2Fapi%2Fmtmw%2Fv2%2Fgbx%2Fmtweb%2Flive%2Fmmc%2F{channel}"
 
-    error_schema = validate.Schema({
-        "code": validate.any(validate.text, int),
-        "message": validate.text,
-    })
-    pdata_schema = validate.Schema(validate.transform(parse_json), validate.any(
-        validate.all(
-            {
-                "locations": [{
-                    "gcp": validate.text,
-                    "ogn": validate.any(None, validate.text),
-                }],
-            },
-            validate.get("locations"),
-            validate.get(0),
-        ),
+    error_schema = validate.Schema({"code": int})
+    caronte_schema = validate.Schema(validate.transform(parse_json), validate.any(
+        {
+            "cerbero": validate.url(),
+            "bbx": str,
+            "dls": [{
+                "lid": validate.all(int, validate.transform(str)),
+                "format": validate.any("hls", "dash", "smooth"),
+                "stream": validate.url(),
+                validate.optional("assetKey"): str,
+                "drm": bool,
+            }],
+        },
         error_schema,
     ))
-    gate_schema = validate.Schema(
+    gbx_schema = validate.Schema(
+        validate.transform(parse_json),
+        {"gbx": str},
+        validate.get("gbx")
+    )
+    cerbero_schema = validate.Schema(
         validate.transform(parse_json),
         validate.any(
-            {
-                "mimeType": validate.text,
-                "stream": validate.url(),
-            },
+            validate.all(
+                {"tokens": {str: {"cdn": str}}},
+                validate.get("tokens")
+            ),
             error_schema,
         )
     )
-
-    def __init__(self, url):
-        super().__init__(url)
-        self.session.http.headers.update({"Referer": self.url})
+    token_errors = {
+        4038: "User has no privileges"
+    }
 
     @classmethod
     def can_handle_url(cls, url):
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/mitele.py
```diff
@@ -54,28 +59,41 @@ def can_handle_url(cls, url):
...
-        gdata = self.session.http.post(self.gate_url,
-                                       acceptable_status=(200, 403, 404),
-                                       data=pdata,
-                                       schema=self.gate_schema)
-        log.trace("{0!r}".format(gdata))
-        if gdata.get("code"):
-            log.error("{0} - {1}".format(gdata["code"], gdata["message"]))
+        tokens = self.session.http.post(pdata["cerbero"],
+                                        acceptable_status=(200, 403, 404),
+                                        data=json.dumps({"bbx": pdata["bbx"], "gbx": gbx}),
```
beardypig, line 76: `json.dumps` might not actually be required here, lemme check that.

beardypig (commented):
And some linting errors to fix.

beardypig:
Just wanted to create this PR to show it was being worked on.

src/streamlink/plugins/mitele.py
```diff
@@ -54,28 +59,40 @@ def can_handle_url(cls, url):
...
 
-        log.debug("Stream: {0} ({1})".format(gdata["stream"], gdata.get("mimeType", "n/a")))
-        for s in HLSStream.parse_variant_playlist(self.session,
-                                                  gdata["stream"],
-                                                  name_fmt="{pixels}_{bitrate}").items():
-            yield s
+        for stream in pdata["dls"]:
+            if stream['drm']:
+                log.warning("Stream may be protected by DRM")
+            else:
```
bastimeyer, line 91: I don't know what the `dls` response looks like, but you're filtering out all stream objects which have the `drm` attribute set, and I don't know if it makes sense logging each of them with the same message.

src/streamlink/plugins/mitele.py
```diff
@@ -54,28 +59,40 @@ def can_handle_url(cls, url):
...
+                log.debug("Stream: {0} ({1})".format(stream["stream"], stream.get("format", "n/a")))
+                cdn_token = tokens[str(stream['lid'])]["cdn"]
```
bastimeyer, line 93: This might raise a KeyError since there's no check for the available tokens.

src/streamlink/plugins/mitele.py
```diff
@@ -54,28 +59,40 @@ def can_handle_url(cls, url):
...
+                                        acceptable_status=(200, 403, 404),
+                                        data=json.dumps({"bbx": pdata["bbx"], "gbx": gbx}),
+                                        headers={
+                                            "content-type": "application/json;charset=UTF-8",
+                                            "origin": "https://www.mitele.es",
+                                            "referer": "https://www.mitele.es"},
+                                        schema=self.cerbero_schema)
+
+        if tokens.get("code"):
+            code = tokens['code']
```
bastimeyer, line 84: Nitpicky, I know :stuck_out_tongue_winking_eye:, but you're mixing quotation styles (down below as well). Also a double attribute lookup.

src/streamlink/plugins/mitele.py
```diff
@@ -54,28 +59,40 @@ def can_handle_url(cls, url):
...
+            log.error(f"Could not get stream tokens: {code} ({self.token_errors.get(code, 'unknown error')})")
```
bastimeyer, line 85: You could use f-strings on the other str.format calls if you're already using an f-string here

bastimeyer (commented):
> Just wanted to create this PR to show it was being worked on.

Hope it's okay if I just quickly comment on the current code changes...

Btw, we should add a convenience method to the http session class which automatically requests and parses json, so that `validate.transform(parse_json)` calls at the beginning of a schema or 	 
`res = session.http.request(); data = session.http.json(res)` don't have to be used.

src/streamlink/plugins/mitele.py
beardypig: It's a `bool`, so only those set to `True` will display the warning.

src/streamlink/plugins/mitele.py
beardypig: Sometimes the f-strings get a bit ugly.

beardypig: I actually dislike this use of f-string, and think I will change it.

src/streamlink/plugins/mitele.py
```diff
@@ -54,28 +59,40 @@ def can_handle_url(cls, url):
...
-        gdata = self.session.http.post(self.gate_url,
-                                       acceptable_status=(200, 403, 404),
-                                       data=pdata,
-                                       schema=self.gate_schema)
-        log.trace("{0!r}".format(gdata))
-        if gdata.get("code"):
-            log.error("{0} - {1}".format(gdata["code"], gdata["message"]))
+        tokens = self.session.http.post(pdata["cerbero"],
+                                        acceptable_status=(200, 403, 404),
+                                        data=json.dumps({"bbx": pdata["bbx"], "gbx": gbx}),
```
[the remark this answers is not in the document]
beardypig: Yep, works great.

src/streamlink/plugins/mitele.py
```diff
@@ -1,51 +1,56 @@
...
+    gbx_url = "https://mab.mediaset.es/1.0.0/get?oid=mtmw&eid=%2Fapi%2Fmtmw%2Fv2%2Fgbx%2Fmtweb%2Flive%2Fmmc%2F{channel}"
 
-    error_schema = validate.Schema({
-        "code": validate.any(validate.text, int),
-        "message": validate.text,
-    })
-    pdata_schema = validate.Schema(validate.transform(parse_json), validate.any(
+    error_schema = validate.Schema({"code": int})
+    caronte_schema = validate.Schema(validate.transform(parse_json), validate.any(
         validate.all(
```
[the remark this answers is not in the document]
beardypig: True. Copy-pasta.

src/streamlink/plugins/mitele.py
beardypig: Very. It's cos I refactored an f-string which would have used `'` quotes.

src/streamlink/plugins/mitele.py
beardypig: I removed all the f-strings in the end, it just makes the code less readable.

src/streamlink/plugins/mitele.py
```diff
@@ -54,28 +57,37 @@ def can_handle_url(cls, url):
...
-        log.trace("{0!r}".format(gdata))
-        if gdata.get("code"):
-            log.error("{0} - {1}".format(gdata["code"], gdata["message"]))
+        tokens = self.session.http.post(pdata["cerbero"],
+                                        acceptable_status=(200, 403, 404),
+                                        json={"bbx": pdata["bbx"], "gbx": gbx},
+                                        headers={"origin": "https://www.mitele.es"},
+                                        schema=self.cerbero_schema)
+
+        if "code" in pdata:
```
back-to, line 76:
```suggestion
        if "code" in tokens:
```

src/streamlink/plugins/mitele.py
```diff
@@ -4,48 +4,51 @@
...
+            ),
             error_schema,
         )
     )
-
-    def __init__(self, url):
-        super().__init__(url)
-        self.session.http.headers.update({"Referer": self.url})
+    token_errors = {
+        4038: "User has no privileges"
```
back-to, line 50: might be also useful

```
4033 ERROR_USER_NOT_VALID
4035 ERROR_CONTENT_GEOBLOCKED
4036 ERROR_CONTENT_GEOBLOCKED
4037 ERROR_CONTENT_GEOBLOCKED
4038 ERROR_USER_NO_PRIVILEGES
4039 ERROR_USER_MAX_SESSIONS_REACHED
```

beardypig: I was going to add these too :P

back-to: it was not that important

PR #3342, opened 2020-11-18
docs: fix mpv property-list link in --title description

azizLIGHT:
Fixed URL pointing to list of mpv format codes

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -468,7 +468,7 @@ def build_parser():
             inserted inside your --title string.
 
             A full list of the format codes mpv uses is available here:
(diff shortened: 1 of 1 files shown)


src/streamlink_cli/argparser.py
```diff
@@ -468,7 +468,7 @@ def build_parser():
             inserted inside your --title string.
 
             A full list of the format codes mpv uses is available here:
-            https://mpv.io/manual/stable/#property-expansion
+            https://mpv.io/manual/master/#property-list
```
beardypig, line 471: I'd leave it pointing to stable.

azizLIGHT: My mistake

PR #3348, opened 2020-11-19
plugin.api.useragents: update User-Agent's

back-to:
- update all User-Agent's
- use newer User-Agent's for some duplicates,
  for backwards compatibility they are not removed


bastimeyer:
Would probably better to split this up into two commits

back-to:
>  and fix streann plugin

currently not worth a new commit, **removed it from this PR**

---

the plugin itself needs an update, it would only work for `www.ellobo106.com` but they don't use it anymore https://github.com/streamlink/streamlink/pull/3305

https://github.com/streamlink/streamlink/blob/054b760ce7e9f43451eed08e9f39de440c3e5add/src/streamlink/plugins/streann.py#L59

but that is something for a different PR

Issue #3235, opened 2020-10-08
The road towards 2.0.0


bastimeyer:
Let's prepare a checklist for the py2 removal and the 2.0.0 release. As said, we should also clean up all the deprecated stuff and remove the remaining cruft after bumping the required Python version, and include that in the 2.0.0 release.

Since nobody wants to do this on a separate branch and deal with conflicts later on, we'll have to publish another release first with the current changes and then get the changes merged on the master branch one by one.

I really want to push the progress on this now, so that it's finally over.

----

- [x] **Publish the last minor or patch release on 1.x**
  Just so that the current plugin fixes and other unrelated changes are not stuck on the master branch until work on 2.0.0 has been completed.
  https://github.com/streamlink/streamlink/releases/tag/1.7.0

- [x] **Python version requirements**
  - [x] Remove Python 2 support
    #3232 
  - [x] Set min Python version to 3.6
    #3269
  - [x] Remove py2 stuff from compatibility modules
    #3270
  - [ ] Remove remaining compatibility code from various modules, eg. `python-requests` related stuff
    Optional, `streamlink.packages` can be ignored.
  - [x] Upgrade to modern py3 syntax
    Optional, can mostly be automated via [`pyupgrade`](https://github.com/asottile/pyupgrade).
    `streamlink.packages` can be ignored.
    - [x] u-string removal
      #3311
    - [ ] ~~f-strings / implicit str.format syntax~~
    - [x] `yield from`
      #3280, #3368
    - [x] super calls
      #3307
    - [x] implicit class inheritance
      #3368
    - [x] dict comprehensions and set literals
      #3368
    - [x] OSError
      #3368

- [x] **Changes which otherwise require a major version bump**
  - [x] Remove all suppressed argparse parameters, including plugin parameters
    #3277
  - [x] Replace various `mux-subtitles` plugin parameters with a global one
    #3324
  - [x] Fix old log calls
    #3274
  - [x] Remove deprecated logger classes and compat classes
    #3273
  - [x] Remove `LIVESTREAMER_VERSION` constant from streamlink_cli
    #3275
  - [x] Remove compat parameters from `HLSStream.parse_variant_playlist`
    #3347
  - [x] Remove compat key names from `Session.{g,s}et_option`
    #3349
  - [x] Remove `plugin.api.support_plugin`, as it's unused and unnecessary
    #3398

- [x] **Changes which were blocked by py2, but are optional for 2.0.0**
[...]

PR #3349, opened 2020-11-19
chore: remove old rtmpdump/subprocess CLI args

bastimeyer:
and remove compat key name translations in the Session options.

Removes the following CLI arguments:
- `--rtmpdump-proxy` (use `--rtmp-proxy` instead)
- `-c` / `--cmdline` (use `--subprocess-cmdline` instead)
- `-e` / `--errorlog` (use `--subprocess-errorlog` instead)
- `--erorlog-path` (use `--subprocess-errorlog-path` instead)

Keeps `--rtmpdump` as a fallback for `--rtmp-rtmpdump`, as the default
streamlinkrc config file of the Windows installer is still using the
old parameter name.

----

https://github.com/streamlink/streamlink/issues/3235#issuecomment-730251337

### DIFF
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -246,16 +246,6 @@ class Streamlink(object):
 
         """
 
-        # Backwards compatibility
-        if key == "rtmpdump":
-            key = "rtmp-rtmpdump"
-        elif key == "rtmpdump-proxy":
-            key = "rtmp-proxy"
-        elif key == "errorlog":
-            key = "subprocess-errorlog"
-        elif key == "errorlog-path":
-            key = "subprocess-errorlog-path"
-
         if key == "http-proxy":
             self.http.proxies["http"] = update_scheme("http://", value)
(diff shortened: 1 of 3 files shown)


back-to:
`--rtmpdump` is used in streamlinkrc

https://github.com/streamlink/streamlink/blob/1.7.0/win32/streamlinkrc#L48-L51

would be bad removing this

bastimeyer:
Would it make sense adding a suppressed argparse argument for this and dealing with this in `streamlink_cli.main.setup_options`, so that the `--rtmpdump` argument doesn't appear in the docs?

Btw, the entire streamlinkrc file desperately needs an overhaul.

back-to:
> Would it make sense adding a suppressed argparse argument

might be the best solution, the command itself doesn't even need to work as `--rtmp-rtmpdump`

as long as we don't need to deal with `streamlink: error: unrecognized arguments: --rtmpdump`

Issue #3359, opened 2020-11-24
Pixiv.net not log in properly


hatabbich:
Hello

I can't log in in Pixiv.net site, trying many of accounts still not effect someone have the same issue???

t374:
Having the same issue. 

Using the methods from #2515 no longer work for me.
I first used --pixiv-purge-credentials and then used --pixiv-sessionid & --pixiv-devicetoken to set my sessionid & devicetoken and it still doesn't recognize r18 streams, so it seems I'm still not logged in, even though it says its set my sessionid & devicetoken .

![Screenshot (37)](https://user-images.githubusercontent.com/68221225/100074880-9d9efa80-2e82-11eb-92fe-fade77e75fda.png)

bastimeyer:
Post the log output **IN TEXT FORMAT** and use the **LATEST DEVELOPMENT VERSION** from the master branch in a clean install.

hat3l0v3:
Hello. Seems I have same issue.
Today **R18** streams stop working with pixiv plugin. It says:

`error: No playable streams found on this URL: `  ([Screenshot](https://i.imgur.com/KiG3Vcl.png))

When I try open regular (not **R18**) steam - everything works fine.
So I guess a problem is a login again. 

It's happened on 1.4.1 version for me. I've install 1.7.0 version - same issue. 
Tried to use _--pixiv-username_, _--pixiv-password_ commands    - nothing helped.

hat3l0v3:
Tested changes to pixiv.py - everything works **fine**. 
**_Thank you_** for quick response.

PR #3364, opened 2020-11-25
plugins.pixiv: set headers for stream data, fixed login issue

back-to:
@hatabbich 
@t374
@hat3l0v3

could you test it

closes https://github.com/streamlink/streamlink/issues/3359


hat3l0v3:
I tried, seems works fine to me.

t374:
Looks good to me, thank you for the fix.

hatabbich:
Thank you so much, its working ^^/

PR #3366, opened 2020-11-26
session: replace usage of deprecated imp module

bastimeyer:
Refactor `utils.load_module`:
- move it to the top and add it to `__all__`
- move imports and loader_details to the global scope

Refactor `session`:
- replace plugin imports via `imp` with `utils.load_module`
- fix incorrect plugin module names
  "streamlink.plugin.{name}" -> "streamlink.plugins.{name}"
- use `log.exception()` to print plugin load failures
- remove unused `print_small_exception`

Add / fix tests:
- remove session dependency from test_plugins
- add proper assertions for plugin modules, methods and their signatures
- add tests for invalid and for overridden plugins

----

Based on the changes of #2570

**Things to note**

- I've decided to keep the implementation of `importlib.machinery.FileFinder` in the `utils` module and use that for loading the plugins, as I had some issues with the module names while trying out the `pkgutils`-based imports of #2570, which led to plugins log calls not being recognized by the StreamlinkLogger, as their module names did not begin with `streamlink.`.
- `test_session` is really messy as it imports everything in `tests.plugins`, including all plugin test modules, which are irrelevant there. The actual `testplugin`s should be moved somewhere else. Mocking certain things is also impossible when everything is run in the test's `setUp`, and I couldn't mock the log call of the plugin override without patching the whole test class which would require adding an unused argument to every test method. That's why plugin load failures are also not tested.
- Why does `plugin.api.support_plugin` even exist? It's not being used and it looks totally unnecessary. This module is the reason why `utils.load_module` existed, but now it's being used by `session` as well.

### DIFF
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -1,8 +1,5 @@
-import imp
 import logging
 import pkgutil
-import sys
-import traceback
 from collections import OrderedDict
 
 import requests
@@ -12,33 +9,13 @@ from streamlink.compat import is_win32
 from streamlink.exceptions import NoPluginError, PluginError
 from streamlink.logger import StreamlinkLogger
 from streamlink.options import Options
-from streamlink.plugin import api
-from streamlink.utils import memoize, update_scheme
+from streamlink.plugin import Plugin, api
+from streamlink.utils import load_module, memoize, update_scheme
 from streamlink.utils.l10n import Localization
 
 # Ensure that the Logger class returned is Streamslink's for using the API (for backwards compatibility)
 logging.setLoggerClass(StreamlinkLogger)
 log = logging.getLogger(__name__)
-
-
-def print_small_exception(start_after):
-    type, value, traceback_ = sys.exc_info()
-
-    tb = traceback.extract_tb(traceback_)
-    index = 0
-
-    for i, trace in enumerate(tb):
-        if trace[2] == start_after:
-            index = i + 1
-            break
-
-    lines = traceback.format_list(tb[index:])
-    lines += traceback.format_exception_only(type, value)
-
-    for line in lines:
-        sys.stderr.write(line)
-
-    sys.stderr.write("\n")
 
 
 class PythonDeprecatedWarning(UserWarning):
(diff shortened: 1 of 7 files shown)


back-to:
https://github.com/streamlink/streamlink/blob/5e7a04b006d0e4d7a884b13bf2aa8047791b837b/src/streamlink/session.py#L413-L417

it is not always `ImportError` which will change the behavior of Streamlink a bit,
broken plugins will error Streamlink instead of skipping the plugin.

Not that important but just to be noted.

---

> Why does plugin.api.support_plugin even exist?

we don't need it, could probably be removed,
same for the `common_` plugins

bastimeyer:
> it is not always `ImportError`

According to the docs, both the finder and loader will raise an `ImportError` (or one of their subclasses) on failure. Same does the `load_module` method if there's no spec or spec.loader. What else is there to be aware of?

- https://docs.python.org/3/reference/import.html
- https://docs.python.org/3/library/importlib.html

back-to:
Example: if you sideload a broken plugin

```

  File "venv/bin/streamlink", line 33, in <module>
    sys.exit(load_entry_point('streamlink', 'console_scripts', 'streamlink')())
  File "src/streamlink_cli/main.py", line 1012, in main
    setup_plugins(args.plugin_dirs)
  File "src/streamlink_cli/main.py", line 739, in setup_plugins
    load_plugins([PLUGINS_DIR])
  File "src/streamlink_cli/main.py", line 640, in load_plugins
    streamlink.load_plugins(directory)
  File "src/streamlink/session.py", line 414, in load_plugins
    mod = load_module(module_name, path)
  File "src/streamlink/utils/__init__.py", line 25, in load_module
    spec.loader.exec_module(mod)
  File "<frozen importlib._bootstrap_external>", line 779, in exec_module
  File "<frozen importlib._bootstrap_external>", line 916, in get_code
  File "<frozen importlib._bootstrap_external>", line 846, in source_to_code
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/.../.config/streamlink/plugins/testplugin.py", line 36
    __plu gin__ = Teamliquid
          ^
SyntaxError: invalid syntax
```

```
  File "venv/bin/streamlink", line 33, in <module>
    sys.exit(load_entry_point('streamlink', 'console_scripts', 'streamlink')())
  File "src/streamlink_cli/main.py", line 1012, in main
    setup_plugins(args.plugin_dirs)
  File "src/streamlink_cli/main.py", line 739, in setup_plugins
    load_plugins([PLUGINS_DIR])
  File "src/streamlink_cli/main.py", line 640, in load_plugins
    streamlink.load_plugins(directory)
  File "src/streamlink/session.py", line 414, in load_plugins
    mod = load_module(module_name, path)
  File "src/streamlink/utils/__init__.py", line 25, in load_module
    spec.loader.exec_module(mod)
  File "<frozen importlib._bootstrap_external>", line 779, in exec_module
  File "<frozen importlib._bootstrap_external>", line 916, in get_code
  File "<frozen importlib._bootstrap_external>", line 846, in source_to_code
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/.../.config/streamlink/plugins/testplugin.py", line 19
    def _get_streams(self):
                          ^
IndentationError: unindent does not match any outer indentation level
```

bastimeyer:
If you want, I can "revert this" change of logic by adding a secondary `except SyntaxError:` block and adding back `print_small_exception`, so that custom plugin errors can be written to stderr while continuing Streamlink's execution.

I personally am OK though with Streamlink failing if code of a custom plugin causes a parsing error.

back-to:
> If you want, I can "revert" this change

no it's fine, can stay like this.

back-to:
well, lets test this with a large user base.

bastimeyer:
Maybe we should add a note to the sideloading plugins section in the docs that plugins which fail parsing will make Streamlink terminate its process beginning with 2.0.0.
I'll open a PR later or tomorrow. Maybe I'll also refactor the tests (see my comment above).

Will also take a look at the support_plugin thing again. I don't think it makes sense keeping it.

PR #3368, opened 2020-11-26
chore: upgrade codebase to py36

bastimeyer:
Upgraded the entire codebase to py36 via `pyupgrade --py36-plus` in separate commits. This should be pretty much it for now with code translations and cleanups (ignoring two or three dependency compat imports, I believe).

There are three things left which I did not include, but that's only one irrelevant blank line, one implicit "read" argument of `open()` and two unnecessary parenthesis in one of the tests which get cleaned up by my other open PR anyway.

f-strings and the implicit str.format syntax are excluded from these changes as well, as it's not always applicable in a sensible way.

Not sure if we should add linting configs for py36+ after this.

### DIFF
--- a/src/streamlink/buffers.py
+++ b/src/streamlink/buffers.py
@@ -15,7 +15,7 @@ class Chunk(BytesIO):
         return self.tell() == self.length
 
 
-class Buffer(object):
+class Buffer:
     """Simple buffer for use in single-threaded consumer/filler.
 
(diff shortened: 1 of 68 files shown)


gravyboat (approved):
Looks good and is much cleaner.

> Not sure if we should add linting configs for py36+ after this.

Might be worthwhile just to ensure people aren't submitting old style code.

PR #3373, opened 2020-11-28
plugins.twitch: fix ads

back-to:
See https://github.com/streamlink/streamlink/issues/3210


bastimeyer:
Here we go again... Round number four.
Looks like it's working fine, at least for now. Let's not close 3210 though.

tarkov2213:
I have noticed this yesterday, yet for some reason ublock had no difficulty with the ads.
Do they use a different approach or just respond quickly?

arrowgent:
confirmed embed player.twitch.tv workaround works (linux streamlink v1.7.0)
https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugins/twitch.py
-- atleast for now

i was using this in a browser recently also

Issue #3257, opened 2020-10-16
Twitch: Detection of reruns no longer working


skulblakka:
## Plugin Issue

- [X] This is a plugin issue and I have read the contribution guidelines.

### Description

The detection of reruns (used for the `--twitch-disable-reruns` command) is no longer working. This is because the [API](https://dev.twitch.tv/docs/v5/reference/streams#get-live-streams) changed and delivers the same response for reruns as for offline streams. 

### Reproduction steps / Explicit stream URLs to test

1. Run streamlink for twitch stream currently broadcasting a rerun with `--twitch-disable-reruns`.
2. Streams starts normally.

### Log output

I edited streamlink to output the API request results to the debug log.

**Log for a stream currently broadcasting a rerun:**
```
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.8.5
[cli][debug] Streamlink: 1.6.0+42.g8db5786.dirty
[cli][debug] Requests(2.22.0), Socks(1.7.1), Websocket(0.57.0)
[cli][info] Found matching plugin twitch for URL https://www.twitch.tv/a_couple_streams
[cli][debug] Plugin specific arguments:
[cli][debug]  --twitch-disable-reruns=True (disable_reruns)
[plugin.twitch][debug] {'_total': 1, 'users': [{'display_name': 'a_couple_streams', '_id': '83828553', 'name': 'a_couple_streams', 'type': 'user', 'bio': 'aeseaes on Spotify/YouTube/etc. | aeseaes is husband and wife duo Allie and Travis | aeseaes = A-C-S = a_couple_streams', 'created_at': '2015-02-26T17:53:44.81295Z', 'updated_at': '2020-10-16T05:12:57.458573Z', 'logo': 'https://static-cdn.jtvnw.net/jtv_user_pictures/50a52b22-df5c-4b95-8064-169baeae792c-profile_image-300x300.png'}]}
[plugin.twitch][debug] {'hosts': [{'host_id': 83828553, 'host_login': 'a_couple_streams', 'host_display_name': 'a_couple_streams'}]}
[plugin.twitch][debug] {'stream': None}
[plugin.twitch][debug] Getting live HLS streams for a_couple_streams
[...]

PR #3375, opened 2020-11-28
plugins.twitch: fix --twitch-disable-reruns

back-to:
closes https://github.com/streamlink/streamlink/issues/3257

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -232,24 +232,6 @@ class TwitchAPI:
             validate.transform(int)
         ))
 
-    def stream_rerun(self, channel_id):
-        return self.call("/kraken/streams/{0}".format(channel_id), schema=validate.Schema(
-            {
-                "stream": validate.any(None, validate.all({
-                    "stream_type": validate.text,
-                    "broadcast_platform": validate.text,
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/twitch.py
```diff
@@ -595,8 +607,8 @@ def _check_for_rerun(self):
             return False
 
         try:
-            stream_type, broadcast_platform, broadcaster_software = self.api.stream_rerun(self.channel_id)
-            if stream_type != "live" or broadcast_platform == "rerun" or broadcaster_software == "watch_party_rerun":
+            stream_type = self.api.stream_rerun(self.channel)
+            if stream_type == "rerun":
```
bastimeyer, line 611:
```suggestion
            if stream_type != "live":
```

This is better, as Twitch has changed the stream type names several times, and using the GQL API doesn't ensure that it'll stay "rerun". "live" however will certainly not change.

bastimeyer (commented):
Let me add some of the tests back before merging this. Just mocking the `self.api.stream_rerun` call is enough, I guess. Will do this later today when I have the time.

bastimeyer:
Are you okay with these changes, @back-to ?

PR #3379, opened 2020-11-29
docs: split CLI args in HTML output into rows

bastimeyer:
instead of separating them with a comma, so that it adds anchors to
secondary arguments. The rendered manpage already splits arguments into
multiple rows if there are multiple arguments for a specific command.

Fix CSS of the rtd_theme_violet theme and hide empty argument values.

### DIFF
--- a/docs/_themes/sphinx_rtd_theme_violet/static/css/theme.css
+++ b/docs/_themes/sphinx_rtd_theme_violet/static/css/theme.css
@@ -4827,6 +4827,21 @@ table.table-custom-layout tbody tr td {
   white-space: unset;
 }
 
+/* fix CLI arguments */
+.rst-content dl.option:not(.docutils) > dt {
+  float: left;
+}
+.rst-content dl.option:not(.docutils) > dt,
+.rst-content dl.option:not(.docutils) > dd {
+  clear: left;
+}
+.rst-content dl.option:not(.docutils) > dt + dt {
+  margin-top: 0;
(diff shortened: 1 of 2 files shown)


bastimeyer:
This is cherry-picked from the furo-theme PR, with fixes for the current theme.

The CSS fixes are rather ugly, but that's because of the layout of the docutils output and theme styling, and it's supposed to work everywhere.

----

Btw, I'd love to split the CLI docs page into two pages:
1. "Command-line tutorial" on a new page and 
2. "Command-line usage" on the current page, so that most CLI links will keep working.

The reason for this is that finding CLI arguments is always a bit annoying and you have to scroll down. On the current theme, this is especially a problem, because the individual CLI argument seconds are not listed in a TOC menu, like on the furo theme.

bastimeyer:
Just in case it's not clear what this change actually does:

----

Previously:
Both args on one line with only one anchor and an empty block if there's no metavar
image
- `/cli.html#cmdoption-q`


----

After:
Each argument on one line with their own anchor
image
- `/cli.html#cmdoption-q`
- `/cli.html#cmdoption-quiet`

----

I'm aware that this may look a bit weird on the current theme due to the background+border style, but the arguments should not be separated with a comma here and putting two boxes on the same line is weird, because the anchors only appear on hover and hovering the first box would move the secondary or third boxes, which is bad. On the furo theme, this won't be a problem.

PR #3382, opened 2020-11-29
docs: fix CLI page

bastimeyer:
- move plugin-sideloading section up (under the plugin-config section)
- two blank lines for 2nd layer headlines, one for 3rd layer headlines
- fix (most) punctuation and grammar errors
- fix outdated description in proxy-support section

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -107,7 +107,7 @@ You can also specify the location yourself using the :option:`--config` option.
 
 .. note::
 
-  On Windows there is a default config created by the installer but on any
+  On Windows, there is a default config created by the installer, but on any
   other platform you must create the file yourself.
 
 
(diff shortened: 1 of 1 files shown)


docs/cli.rst
```diff
@@ -107,7 +107,7 @@ You can also specify the location yourself using the :option:`--config` option.
 
 .. note::
 
-  On Windows there is a default config created by the installer but on any
+  On Windows, there is a default config created by the installer, but on any
```
gravyboat, line 110: I don't think the commas are needed here.

gravyboat (approved):
Looks good, one very minor piece of feedback, everything else is a major improvement.

docs/cli.rst
bastimeyer: No, both commas are correct here. The first one comes after the introductory and the second one separates an independent clause.

beardypig: I believe English is taught better in non-english speaking countries... 😞

gravyboat: @beardypig I'm a native English speaker and while grammatically it may be correct it sounds odd when read aloud. English is a garbage language.

beardypig: @gravyboat as am I and i agree with everything you said.

Issue #2998, opened 2020-05-30
1spotmedia.com


speakerboar:
## Plugin Request

- [x] This is a plugin request and I have read the contribution guidelines and plugin request requirements.

### Description
1spotmedia.com, contain Jamaican and Caribbean channels.

### Example stream URLs

1. https://www.1spotmedia.com/#!/live-stream/79336998495
2. https://www.1spotmedia.com/#!/live-stream/77999142467
3. https://www.1spotmedia.com/#!/live-stream/79336998577

### Additional comments, screenshots, etc.
Requires registering for free account but viewing only free from Jamaican IP
[When accessing in Jamaica](https://i.imgur.com/po5szL3.png)
[Outside of  Jamaica](https://i.imgur.com/ybAiuaf.png)

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

gravyboat:
@speakerboar Unfortunately this site violates a lot of the rules outlined in Our [plugin request guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#plugin-requests). @back-to opened a PR with the plugin here: https://github.com/streamlink/streamlink/pull/3389/files if you want to [sideload it](https://streamlink.github.io/cli.html?highlight=side%20load#sideloading-plugins).

mkbloke:
Off-topic for this issue, but this is the reason I'm reluctant to take on any of the existing plugin requests.  Most of the rules are straightforward; it's number 10 that's most problematic.  Not all sites will have viewer figures (if they can be trusted anyway) and what constitutes an acceptable number to justify the inclusion of a new plugin into master...

gravyboat:
@mkbloke This is one of the big problems since we don't track user install or usage stats at all. There's no way to even see if a plugin will be widely used, or if it's just one person who wants a website. Viewer numbers are the easiest way to track that, but as you said a lot of sites don't include these details. Maybe we should consider a new repo of side load plugins that don't meet these requirements?

bastimeyer:
> Maybe we should consider a new repo of side load plugins that don't meet these requirements?

I'd suggest creating a new orga `streamlink-contrib` (or something similar) for this purpose. Then you can decide whether there should be repos for individual plugins or one plugin for a collection of community plugins.

By moving such plugins to a different orga, this protects Streamlink from invalid/incorrect DMCA takedown notices and the likes. And it makes it immediately clear to the user that it's nothing official.

[...]

PR #3389, opened 2020-12-04
plugins.onespotmedia: new plugin for 1spotmedia.com

back-to:
```
streamlink https://www.1spotmedia.com/#!/live-stream/79336998495 best
[cli][info] Found matching plugin onespotmedia for URL https://www.1spotmedia.com/live-stream/79336998495
[cli][info] Available streams: 240p (worst), 360p, 480p, 720p (best)
```

closes https://github.com/streamlink/streamlink/issues/2998

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -126,6 +126,7 @@ ntv                     ntv.ru               Yes   No
 okru                    ok.ru                Yes   Yes
 olympicchannel          olympicchannel.com   Yes   Yes   Only non-premium content is available.
 oneplusone              1plus1.video         Yes   No
+onespotmedia            1spotmedia.com       Yes   No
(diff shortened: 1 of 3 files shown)


gravyboat:
Looking at the site it seems pretty questionable regarding how much of this content is live. At least 75% of the channels seem to be prerecorded content that is being played in a specific order. This violates #3 and #6 from https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#plugin-requests. It also requires a subscription, which violates #4, and one of the only real live pieces is a radio show which violates #7. How many viewers are these channels getting during prime time? This seems like a plugin that should be side loaded if anything.

back-to:
> This seems like a plugin that should be side loaded if anything.

true, it is a small website.

gravyboat:
Sorry for making you waste your time @back-to, I didn't look at the original issue.

PR #3394, opened 2020-12-05
plugins: fix radiko.py url

smallbomb:
1. modify url regexp to support 'https'.
2. timefree m3u8 default url had been modified.

test url success:
https://radiko.jp/#!/ts/QRR/20201205210000 
https://radiko.jp/#!/ts/YFM/20201206010000

### DIFF
--- a/src/streamlink/plugins/radiko.py
+++ b/src/streamlink/plugins/radiko.py
@@ -49,7 +49,6 @@ class Radiko(Plugin):
         return url, token
 
     def _timefree(self, station_id, start_at):
-        # m3u8_url = 'https://radiko.jp/v2/api/ts/playlist.m3u8' ## old url?
         m3u8_url = 'https://tf-rpaa.smartstream.ne.jp/tf/playlist.m3u8'
         token, area_id = self._authorize()
         lsid = hashlib.md5(str(random.random()).encode('utf-8')).hexdigest()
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/radiko.py
```diff
@@ -49,7 +49,8 @@ def _live(self, station_id):
         return url, token
 
     def _timefree(self, station_id, start_at):
-        m3u8_url = 'https://radiko.jp/v2/api/ts/playlist.m3u8'
+        # m3u8_url = 'https://radiko.jp/v2/api/ts/playlist.m3u8' ## old url?
+        m3u8_url = 'https://tf-rpaa.smartstream.ne.jp/tf/playlist.m3u8'
```
bastimeyer, line 53: Don't comment out old stuff. All changes will be preserved in the git history.

bastimeyer (changes requested):
Other than the commented out line, I can't review these changes, because the site doesn't load on my system / from my connection.

Is there a reason why the "timefree" URL is on another host?

src/streamlink/plugins/radiko.py
smallbomb: I understand. Thanks for your suggestions very much. :)

smallbomb:
I'm sorry. I don't know why m3u8 url was modified. :(
Old m3u8 url has been invaild when I used steamlink for radiko site first. (include 'http')
So, I guess this plugin script was not updated.

PR #3390, opened 2020-12-04
refactor: test_session, move testplugin files

bastimeyer:
- move testplugin and its related files into their own `tests/plugin`
  directory, so that tests don't have to unnecessarily import regular
  plugin test files
- refactor test_session
  - don't initialize session in `setUp`
  - add test cases for ImportError and SyntaxError
  - add assertion for plugin override debug log call
  - rename "load_plugins" test cases
  - rename some confusing variable names
- fix remaining tests in regards to the testplugin path changes

PR #3399, opened 2020-12-07
tests: fix test_cmdline{,_title}

bastimeyer:
This fixes the issues with `tests/test_cmdline{,_title}.py` when running those test files independently by properly mocking the `subprocess` method calls and the global `streamlink_cli.main.streamlink` variable, so that there are no side effects from other test cases.

See https://github.com/streamlink/streamlink/pull/3390#issuecomment-739636791

```
$ pytest -q --tb=no tests/test_cmdline.py 
.....ssssss                                                                                                                              [100%]
5 passed, 6 skipped in 0.51s

$ pytest -q --tb=no tests/test_cmdline_title.py 
......ssssssss                                                                                                                           [100%]
6 passed, 8 skipped in 0.41s
```

----

The testplugin path for the `load_plugins(plugin_path)` call could ideally be defined in `tests/__init__.py`, so that it doesn't have to be defined in every test file separately, but that's not important here.

### DIFF
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -1,19 +1,10 @@
 import os.path
 import unittest
-from unittest.mock import ANY, patch
+from unittest.mock import ANY, Mock, patch
 
 import streamlink_cli.main
 from streamlink import Streamlink
 from streamlink_cli.compat import is_win32
-
-
-PluginPath = os.path.join(os.path.dirname(__file__), "plugin")
-
-
-def setup_streamlink():
-    streamlink_cli.main.streamlink = Streamlink()
-    streamlink_cli.main.streamlink.load_plugins(PluginPath)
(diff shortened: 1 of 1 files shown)


bastimeyer:
hm? why are there so many coverage misses all of a sudden now (the HLS/segmented ones can be ignored, as usual)? even in some tests...
https://codecov.io/gh/streamlink/streamlink/pull/3399/changes

bastimeyer:
Looks like it's a codecov bug, because the Windows tests failed in the previous commit and the current coverage results have not been updated with the new data. At least that's what I'm thinking because all lines which are missing coverage are Windows related... The pytest runner logs show that those tests were run and passed.

beardypig (approved):
Looks good. I didn't have a problem running the test cases individually, but I can see where there would be a problem.

PR #3403, opened 2020-12-09
add basic HLS fMP4 support

### LINKED ISSUE #3328: Modern fmp4 HLS Incompatibility
## Bug Report

### Description
Streamlink is unable to playback modern fmp4 HLS playlists referenced from a master playlist because it doesn't load the initialization mp4 specified on the `#EXT-X-MAP:URI="init_0.mp4"` line. This support was introduced several years ago and seems to be the only type of HLS container Apple is moving forward with. The tested HLS stream works in VLC, videojs, hls.js, and mpv, but fails in streamlink.

I expected HLS support. The documentation could also be improved with specification of what version of HLS is supported by streamlink, although fmp4 HLS is some years old now.

1. Generate an fmp4 HLS playlist with a master playlist with ffmpeg. The key option is `-hls_segment_type fmp4` to generate the fmp4 playlist, and `-master_playlist_name` to make a master playlist for it.
2. Try to load the stream with streamlink
3. Observe the logged failure indicating the init.mp4 wasn't loaded first. Manual inspection of the playlist file shows the `#EXT-X-MAP:URI="init_0.mp4"` is correctly set, the file exists, and the playlist works in VLC, videojs, hls.js, and mpv.

### Log output

```
[cli][debug] OS:         Linux-5.8.18-gentoo-x86_64-Intel-R-_Core-TM-_i7-8750H_CPU_@_2.20GHz-with-gentoo-2.7
[cli][debug] Python:     3.7.9
[cli][debug] Streamlink: 1.7.0
[cli][info] Found matching plugin hls for URL https://streamurl.net/hls/stream.m3u8
[plugin.hls][debug] URL=https://streamurl.net/hls/stream.m3u8; params={}
[cli][info] Available streams: 720p (worst), 1080p (best)
[stream.hls][debug] First Sequence: 35; Last Sequence: 40
[stream.hls][debug] Start offset: 0; Duration: None; Start Sequence: 38; End Sequence: None
[stream.hls][debug] Adding segment 38 to queue
[stream.hls][debug] Adding segment 39 to queue
[stream.hls][debug] Adding segment 40 to queue
[cli][info] Starting player: mpv --cache=yes --demuxer-max-bytes=2048k
[stream.hls][debug] Download of segment 38 complete
[cli.output][debug] Opening subprocess: mpv --cache=yes --demuxer-max-bytes=2048k --title=https://streamurl.net/hls/stream.m3u8 -
[stream.hls][debug] Download of segment 39 complete
[file] Reading from stdin...
[ffmpeg/demuxer] mov,mp4,m4a,3gp,3g2,mj2: could not find corresponding track id 1
[ffmpeg/demuxer] mov,mp4,m4a,3gp,3g2,mj2: could not find corresponding track id 2
[ffmpeg/demuxer] mov,mp4,m4a,3gp,3g2,mj2: could not find corresponding trex (id 1)
[ffmpeg/demuxer] mov,mp4,m4a,3gp,3g2,mj2: could not find corresponding track id 0
[ffmpeg/demuxer] mov,mp4,m4a,3gp,3g2,mj2: trun track id unknown, no tfhd was found
[ffmpeg/demuxer] mov,mp4,m4a,3gp,3g2,mj2: error reading header
[lavf] avformat_open_input() failed
[ffmpeg/demuxer] mov,mp4,m4a,3gp,3g2,mj2: could not find corresponding track id 1
[ffmpeg/demuxer] mov,mp4,m4a,3gp,3g2,mj2: could not find corresponding track id 2
[ffmpeg/demuxer] mov,mp4,m4a,3gp,3g2,mj2: could not find corresponding trex (id 1)
[ffmpeg/demuxer] mov,mp4,m4a,3gp,3g2,mj2: could not find corresponding track id 0
[ffmpeg/demuxer] mov,mp4,m4a,3gp,3g2,mj2: trun track id unknown, no tfhd was found
[ffmpeg/demuxer] mov,mp4,m4a,3gp,3g2,mj2: error reading header
[lavf] avformat_open_input() failed
Failed to recognize file format.

Exiting... (Errors when loading file)
[cli][info] Player closed
[stream.hls][debug] Download of segment 40 complete
[cli][info] Stream ended

```

### Additional comments, etc.
I have no problem getting the exact same playlist generation command in ffmpeg working in streamlink if I remove `-hls_segment_type fmp4` to generate the older mpegts playlists, but I'd like to move to fmp4 for the lower overhead and new codec support Apple is specifying for it.

Billy2011:
you can test it with a swiss ip and [this](https://streams.wilmaa.com/m3u8/get?channelId=ard_hd) link.
closes #3328

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -170,6 +170,7 @@ class HLSStreamWorker(SegmentedStreamWorker):
         self.stream = self.reader.stream
 
         self.init_map = None
+        self.maps = None
         self.playlist_changed = False
         self.playlist_end = None
         self.playlist_sequence = -1
@@ -232,6 +233,7 @@ class HLSStreamWorker(SegmentedStreamWorker):
         if playlist.iframes_only:
             raise StreamError("Streams containing I-frames only is not playable")
 
+        self.maps = playlist.maps
         media_sequence = playlist.media_sequence or 0
         sequences = [Sequence(media_sequence + i, s)
                      for i, s in enumerate(playlist.segments)]
@@ -301,12 +303,15 @@ class HLSStreamWorker(SegmentedStreamWorker):
         while not self.closed:
             for sequence in filter(self.valid_sequence, self.playlist_sequences):
                 if self.init_map != sequence.segment.map:
-                    self.init_map = sequence.segment.map
-                    _segment = hls_playlist.Segment(
-                        self.init_map.uri, 0, None, None, False, self.init_map.byterange, None, None)
-                    _sequence = Sequence(sequence.num, _segment)
-                    log.info(f"Adding map segment {sequence.num} to queue")
-                    yield _sequence
+                    if sequence.segment.map is not None:
+                        self.init_map = sequence.segment.map
+                        _segment = hls_playlist.Segment(
+                            self.maps[self.init_map].uri, 0, None, None, False, self.maps[self.init_map].byterange, None, None)
+                        _sequence = Sequence(sequence.num, _segment)
+                        log.debug(f"Adding map segment {sequence.num} to queue")
+                        yield _sequence
+                    else:
+                        log.warning(f"Missing segment init map, maybe corrupted playlist, sequence num: {sequence.num}")
                 log.debug(f"Adding segment {sequence.num} to queue")
                 yield sequence
                 total_duration += sequence.segment.duration
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -169,6 +169,7 @@ class HLSStreamWorker(SegmentedStreamWorker):
         SegmentedStreamWorker.__init__(self, *args, **kwargs)
         self.stream = self.reader.stream
 
+        self.init_map = None
         self.playlist_changed = False
         self.playlist_end = None
         self.playlist_sequence = -1
@@ -299,6 +300,13 @@ class HLSStreamWorker(SegmentedStreamWorker):
         total_duration = 0
         while not self.closed:
             for sequence in filter(self.valid_sequence, self.playlist_sequences):
+                if self.init_map != sequence.segment.map:
+                    self.init_map = sequence.segment.map
+                    _segment = hls_playlist.Segment(
+                        self.init_map.uri, 0, None, None, False, self.init_map.byterange, None, None)
+                    _sequence = Sequence(sequence.num, _segment)
+                    log.info(f"Adding map segment {sequence.num} to queue")
+                    yield _sequence
                 log.debug(f"Adding segment {sequence.num} to queue")
                 yield sequence
                 total_duration += sequence.segment.duration
--- a/src/streamlink/stream/hls_playlist.py
+++ b/src/streamlink/stream/hls_playlist.py
@@ -256,7 +256,7 @@ class M3U8Parser:
     def parse_tag_ext_x_map(self, value):
         attr = self.parse_attributes(value)
         byterange = self.parse_byterange(attr.get("BYTERANGE", ""))
-        self.state["map"] = Map(attr.get("URI"), byterange)
+        self.state["map"] = Map(self.uri(attr.get("URI")), byterange)
 
     def parse_tag_ext_x_i_frame_stream_inf(self, value):
         attr = self.parse_attributes(value)
--- a/src/streamlink/stream/hls_playlist.py
+++ b/src/streamlink/stream/hls_playlist.py
@@ -60,6 +60,7 @@ class M3U8:
         self.playlists = []
         self.dateranges = []
         self.segments = []
+        self.maps = []
 
     @classmethod
     def is_date_in_daterange(cls, date, daterange):
@@ -256,7 +257,8 @@ class M3U8Parser:
     def parse_tag_ext_x_map(self, value):
         attr = self.parse_attributes(value)
         byterange = self.parse_byterange(attr.get("BYTERANGE", ""))
-        self.state["map"] = Map(self.uri(attr.get("URI")), byterange)
+        self.state["map"] = len(self.m3u8.maps)
+        self.m3u8.maps.append(Map(self.uri(attr.get("URI")), byterange))
 
     def parse_tag_ext_x_i_frame_stream_inf(self, value):
         attr = self.parse_attributes(value)


src/streamlink/stream/hls.py
```diff
@@ -299,6 +300,13 @@ def iter_segments(self):
         total_duration = 0
         while not self.closed:
             for sequence in filter(self.valid_sequence, self.playlist_sequences):
+                if self.init_map != sequence.segment.map:
```
beardypig, line 303: I think a bit of a comment describing the mechanism might be useful here.

Billy2011: The map segment has to be put in the queue only once, and is then valid for all following segments until an `ext_x_discontinuity` tag arrives (in this case, a possibly new map segment follows) or the stream ends.

Billy2011: I want to save the init maps in a separate list and not in each `Segment` as before, because I want only the index of the init. map store, so less memory is used, especially with very long playlist of vod’s - do you have any objections, @beardypig, @bastimeyer?

bastimeyer (changes requested):
`HLSStreamWorker.iter_segments()` should not create new segments and yield them with the same num. This is bad, because first, it's not the job of the worker to create segments, it's the job of the parser, and second, you're re-downloading the same media initialization over and over for each segment that comes after the `EXT-X-MAP` tag, which is wasteful.

Also not sure what the purpose of the `init_map` and `maps` list is. Each `Segment`'s `map` attribute gets set on all segments following the `EXT-X-MAP`, as defined by the HLS spec:
- https://tools.ietf.org/html/rfc8216#section-4.3.2.5
- https://github.com/streamlink/streamlink/blob/83fb57cc4266f9b5450369131935a3750e51aa14/src/streamlink/stream/hls_playlist.py#L333
- https://github.com/streamlink/streamlink/blob/83fb57cc4266f9b5450369131935a3750e51aa14/src/streamlink/stream/hls_playlist.py#L256-L259
- https://github.com/streamlink/streamlink/blob/83fb57cc4266f9b5450369131935a3750e51aa14/src/streamlink/stream/hls_playlist.py#L246-L248

A much better implementation would be fetching the media initialization once and writing it to the buffer in the `HLSStreamWriter.write` method before it writes the fetched segment data. The map-fetching would require some modifications of the writer's fetch method and making it so that the media initialization data of the map attribute gets stored somewhere and shared between the segments with the same map.

Proper tests are also a requirement before any feature extensions to the streaming classes can get merged, ideally with full coverage.

Billy2011:
> you're re-downloading the same media initialization over and over for each segment that comes after the EXT-X-MAP tag, which is wasteful.

> Also not sure what the purpose of the init_map and maps list is. Each Segment's map attribute gets set on all segments following the EXT-X-MAP, as defined by the HLS spec:

Maybe you missed my comments:

> The map segment has to be put in the queue only once, and is then valid for all following segments until an ext_x_discontinuity tag arrives (in this case, a possibly new map segment follows) or the stream ends.

> I want to save the init maps in a separate list and not in each Segment as before, because I want only the index of the init. map store, so less memory is used, especially with very long playlist of vod’s - do you have any objections, @beardypig, @bastimeyer?

Billy2011:
With [this apple demo](https://devstreaming-cdn.apple.com/videos/streaming/examples/img_bipbop_adv_example_fmp4/master.m3u8) it comes to many errors in the output stream:

> [mpegts @ 000002026bb82900] Non-monotonous DTS in output stream 0:1; previous: 1028520, current: 896040; changing to 1028521. This may result in incorrect timestamps in the output file.

Billy2011:
@bastimeyer,

> A much better implementation would be fetching the media initialization once and writing it to the buffer in the HLSStreamWriter.write method before it writes the fetched segment data.

That was my thought at first, but I’ve been dealing with it for days and my current solution is already quite optimal and you will probably find that too, if you take it seriously. The handling of the `byterange` requires in the fetcher to create a `sequence` if you want to use the existing function for it, apart from the additional code this does not bring any advantage over my solution.

bastimeyer:
> The map segment has to be put in the queue only once

Wrong. The media initialization data of the `EXT-X-MAP` tag needs to be written to the output buffer before **every** segment which comes after the tag. This has to be done until the playlist ends or until a discontinuity tag is set. By only putting the "map segment" with the media initialization to the writer once (yielding it in the worker's `iter_segments()`), you are corrupting all following segments in the output except the first one.

See here:
https://tools.ietf.org/html/rfc8216#section-4.3.2.5

> The EXT-X-MAP tag specifies how to obtain the Media Initialization Section (Section 3) required to parse the applicable Media Segments. It applies to every Media Segment that appears after it in the Playlist until the next EXT-X-MAP tag or until the end of the Playlist.

The `init_map` attribute is not needed, as each segment already carries the map information set by the parser (the "map" state does not get popped in `get_segment()`, only read). The memory optimization by storing an index of all maps is also unneeded, as the map attribute on each segment only stores a reference to the instance of the `Map` namedtuple which carries the URL and byterange information.

----

Sorry, but I don't think that your proposed solution is optimal. You are abusing the current implementation of the Worker/Writer and how the `Sequence`s/`Segment`s are used to fetch data and how it's written to the output buffer. Map data needs to be fetched once, cached and then properly applied to the output buffer depending on each Segment's map attribute value. A proper implementation requires a bit, maybe even lots of refactoring of the methods of the `{HLS,Segmented}{Writer,Worker}` classes and maybe also the HLS parser, because there's currently no caching logic built in. Not to mention timing issues between the threads.

Adding hack after hack without any tests will make this an unmaintainable and inextensible mess.

Billy2011:
> The media initialization data of the EXT-X-MAP tag needs to be written to the output buffer before every segment which comes after the tag. This has to be done until the playlist ends or until a discontinuity tag is set. 

If that were the case, all my swiss wilmaa links would not run.

> The memory optimization by storing an index of all maps is also unneeded, as the map attribute on each segment only stores a reference to the instance of the Map namedtuple which carries the URL and byterange information.

You are partly right, but an` int` uses significantly less memory than a `namedtuple`.

> Adding hack after hack without any tests will make this an unmaintainable and inextensible mess.

I think the best tests are real life tests. The python tests only check the rudimentary functions, but that does not mean that those tests are useless.

bastimeyer:
Well, in case of MPEG-TS segments, [section 3.2](https://tools.ietf.org/html/rfc8216#section-3.2) of the RFC 8216 says this:

> The Media Initialization Section of an MPEG-2 Transport Stream Segment is a Program Association Table (PAT) followed by a Program Map Table (PMT).
> **Each** Transport Stream Segment MUST contain a PAT and a PMT, or have an EXT-X-MAP tag (Section 4.3.2.5) applied to it.

Which means that when an `EXT-X-MAP` tag is present with MIS data, it has to be applied to all segments, not just the first one.

For fMP4 segments in [section 3.3](https://tools.ietf.org/html/rfc8216#section-3.3), it's more complicated. The MIS via `EXT-X-MAP` is a requirement (hence your PR for resolving #3328), but since fMP4 is not a transport stream and since the paragraphs don't make it perfectly clear, I am not sure if applying the MIS to all segments is a necessity here. @beardypig, do you know more? Also, what happens to start offsets or the live-edge value here?

> an int uses significantly less memory than a namedtuple

The same instance of the `Map` namedtuple (stored on the `state` dict) is shared between all `Segment`s in their `map` attribute. As I've already said and linked, it does not get removed from the `state` when a segment gets created, unlike other attributes on the state.

> I think the best tests are real life tests.

No. Unit and integration tests ensure that the implementation is correct and that future changes to any part of the project won't break anything.

gravyboat:
No response in ~3 months. Closing.

PR #3404, opened 2020-12-09
add ffmpeg-copyts option

Billy2011:
`hls-multi` streams occasionally have an input offset between the audio and video streams and the sound is then out of sync. With `copyts`, the offset is not changed when outputting to the player so that the sound is in sync.

### DIFF
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -61,6 +61,7 @@ class Streamlink:
             "ffmpeg-video-transcode": None,
             "ffmpeg-audio-transcode": None,
             "ffmpeg-start-at-zero": False,
+            "ffmpeg-copyts": False,
             "mux-subtitles": False,
             "locale": None,
             "user-input-requester": None
@@ -202,6 +203,8 @@ class Streamlink:
         ffmpeg-start-at-zero     (bool) When used with ffmpeg and copyts,
                                  shift input timestamps so they start at zero
                                  default: ``False``
+
+        ffmpeg-copyts            (bool) When used with ffmpeg, do not shift input timestamps.
 
         mux-subtitles            (bool) Mux available subtitles into the
                                  output stream.
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -60,8 +60,8 @@ class Streamlink:
             "ffmpeg-fout": None,
             "ffmpeg-video-transcode": None,
             "ffmpeg-audio-transcode": None,
-            "ffmpeg-start-at-zero": False,
             "ffmpeg-copyts": False,
+            "ffmpeg-start-at-zero": False,
             "mux-subtitles": False,
             "locale": None,
             "user-input-requester": None
@@ -200,11 +200,11 @@ class Streamlink:
                                  audio when muxing with ffmpeg
                                  e.g. ``aac``
 
+        ffmpeg-copyts            (bool) When used with ffmpeg, do not shift input timestamps.
+
         ffmpeg-start-at-zero     (bool) When used with ffmpeg and copyts,
                                  shift input timestamps so they start at zero
                                  default: ``False``
-
-        ffmpeg-copyts            (bool) When used with ffmpeg, do not shift input timestamps.
 
         mux-subtitles            (bool) Mux available subtitles into the
                                  output stream.
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -1134,18 +1134,18 @@ def build_parser():
         """
     )
     transport.add_argument(
-        "--ffmpeg-start-at-zero",
-        action="store_true",
-        help="""
-        Enable the -start_at_zero ffmpeg option when using copyts.
-        """
-    )
-    transport.add_argument(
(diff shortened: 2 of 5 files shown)


src/streamlink_cli/argparser.py
```diff
@@ -1116,6 +1116,14 @@ def build_parser():
         when using copyts.
         """
     )
+    transport.add_argument(
+        "--ffmpeg-copyts",
+        action="store_true",
+        help="""
+        Do not remove the initial start time offset value.
+        ffmpeg-start-at-zero option is set to "false".
+        """
```
bastimeyer, line 1125:
```suggestion
        help="""
        Forces the -copyts ffmpeg option and does not remove
        the initial start time offset value.

        --ffmpeg-no-start-at-zero will be set to True.
        """
```

The CLI docs should only reference other CLI arguments and not session option names, as it's not clear to the user what exactly will be done. The argument should also ideally be defined before `--ffmpeg-no-start-at-zero` and adding a double-dash will make the doc's argparser turn the referred arg into a link.

src/streamlink/session.py
```diff
@@ -197,6 +198,9 @@ def set_option(self, key, value):
         ffmpeg-start-at-zero     (bool) When used with ffmpeg and copyts,
                                  shift input timestamps so they start at zero
 
+        ffmpeg-copyts            (bool) When used with ffmpeg, do not shift input
+                                 timestamps - start-at-zero is therefore set to false.
```
bastimeyer, line 202:
```suggestion
        ffmpeg-copyts            (bool) When used with ffmpeg, do not shift input timestamps.
```

Only `streamlink_cli` overrides values, not the `Session` or `FFMPEGMuxer`.

src/streamlink_cli/main.py
```diff
@@ -851,6 +851,9 @@ def setup_options():
         streamlink.set_option("ffmpeg-audio-transcode", args.ffmpeg_audio_transcode)
     if args.ffmpeg_no_start_at_zero:
         streamlink.set_option("ffmpeg-start-at-zero", False)
+    if args.ffmpeg_copyts:
+        streamlink.set_option("ffmpeg-copyts", True)
+        streamlink.set_option("ffmpeg-start-at-zero", False)
```
bastimeyer, line 856: I'm not entirely sure, but I think it makes more sense overriding the CLI argument value of `args.ffmpeg_no_start_at_zero` (setting it to `True`) and moving the `if args.ffmpeg_copyts` block up. Then the session option `ffmpeg-start-at-zero` does only get set in one single place and it aligns with the argument help text which should be referring to the other CLI argument.

Billy2011: @beardypig, do you agree with the proposal of bastimeyer, I quarrel with the proposal?

src/streamlink_cli/main.py
```diff
@@ -851,6 +851,8 @@ def setup_options():
         streamlink.set_option("ffmpeg-audio-transcode", args.ffmpeg_audio_transcode)
     if args.ffmpeg_no_start_at_zero:
         streamlink.set_option("ffmpeg-start-at-zero", False)
+    if args.ffmpeg_copyts:
+        streamlink.set_option("ffmpeg-copyts", True)
```
back-to, line 855:
```suggestion
        streamlink.set_option("ffmpeg-copyts", args.ffmpeg_copyts)
```

back-to (commented):
could you add some tests for

```py
session.options.set("ffmpeg-copyts", False)
session.options.set("ffmpeg-copyts", True)
```

similar to https://github.com/streamlink/streamlink/blob/master/tests/streams/test_ffmpegmux.py#L56-L117

bastimeyer (changes requested):
Please move all the definitions of `ffmpeg-copyts` up, so that it comes before `ffmpeg-start-at-zero`, as `start-at-zero` depends on `copyts`. In the Session, FFMPEGMuxer, argparser and cli.main.

Billy2011:
>  as start-at-zero depends on copyts

Please why, I see no dependence?

bastimeyer:
https://github.com/streamlink/streamlink/blob/ca1b9665ef3ddd34f7a1afab5f55cc5784ddc0ad/src/streamlink/stream/ffmpegmux.py#L114-L117

Billy2011:
> start_at_zero = session.options.get("ffmpeg-start-at-zero") or options.pop("start_at_zero", False)
> copyts = session.options.get("ffmpeg-copyts") or options.pop("copyts", False)
> 

It doesn’t matter which option comes first.

back-to:
it is basically just style

**git diff, git history** would look better then now

https://github.com/streamlink/streamlink/pull/3404/files#diff-04516d5f6b73bc1d25a0bc7385dc2b3e1993578ff01d3a34ec040a271d71b73d

bastimeyer:
> it is basically just style

What's important here is the order of the CLI args in the docs / man page / help text. And since that definitely should be corrected, it doesn't make sense having the other code lines in a different order.

Billy2011:
Have a good excuse ;)

bastimeyer (approved):

Issue #3231, opened 2020-10-07
Set mpv's title to "streamer - streamtitle" by default for twitch


txtsd:
## Feature Request

- [x] This is a feature request and I have read the contribution guidelines.

### Description

When watching a twitch stream, set mpv's title to the streamer's name and the stream title.

### Expected / Actual behavior

Currently, mpv just shows the link in the title.

### Additional comments, screenshots, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

bastimeyer:
https://streamlink.github.io/cli.html#cmdoption-t

guihkx:
Correct me if I'm wrong, but streamlink's `-t` / `--title` option is only useful to set the title of the player's window, and not the title of the current media playing. So, at least with mpv, the media title will always be displayed as `-`:

![image](https://user-images.githubusercontent.com/626206/101595668-84459480-39d2-11eb-8ffe-299847cb0b52.png)

![image](https://user-images.githubusercontent.com/626206/101595703-99222800-39d2-11eb-82f7-fcfb1a1e5b8d.png)


The proper way would be setting mpv's `--force-media-title` option in streamlink's `--player-args`, but as far I as know, there's  currently no way of using `{title}` or `{author}` in `--player-args`. :(

My "workaround" for now, is to set the stream title to `streamlink`, which is much better than `-` already. My config. file:

`~/.config/streamlink/config`:
```ini
player=mpv
title="{author} - {title}"
player-args="--force-media-title=streamlink" {filename}
```

streamlink 1.7.0+103.g401440c

bastimeyer:
https://github.com/streamlink/streamlink/issues/3036#issuecomment-647053067

guihkx:
Hah, thanks for linking that comment. I guess I should've searched for `force-media-title` before commenting here. Oh well. If I knew Python, I'd give it try implementing it myself... :c

bastimeyer:
Just open a pull request which changes the MPV parameter from `--title={self.title}` to `--force-media-title={self.title}`. `--force-media-title` does both and also sets the window title.
- https://github.com/streamlink/streamlink/blob/401440c34d6fb63b3f09c5fc8517ce08d59461fa/src/streamlink_cli/output.py#L202
- https://github.com/streamlink/streamlink/blob/401440c34d6fb63b3f09c5fc8517ce08d59461fa/tests/test_cmdline_title.py#L30-L31
[...]

PR #3405, opened 2020-12-10
cli.output: replace MPV player title parameter

bastimeyer:
Replace MPV's `--title` parameter with `--force-media-title`, as it also
changes the name of the playlist item and not just the window title.

----

See https://github.com/streamlink/streamlink/issues/3231#issuecomment-741636677

Also changed the content of some test assertions in one go, as they were not testing the entire player argument logic (`subprocess.list2cmdline` on Windows).


gravyboat:
Hmm, do we have to consider this a breaking change? I don't think so since it's only when the title is not set, but it could be problematic if anyone is hooking in.

bastimeyer:
> do we have to consider this a breaking change?

No, this doesn't change anything Streamlink related. User-provided player args get appended to the player args list/string, so there's nothing to worry about here.

Issue #3407, opened 2020-12-11
afreecatv.com


pulbitz:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.
- [x] I am using the latest development version from the master branch.

### Description

Only certain afreecatv streamer get this error. 

### Reproduction steps / Explicit stream URLs to test

http://play.afreecatv.com/ryuri79

### Log output

```
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.6.6
[cli][debug] Streamlink: 1.7.0
[cli][debug] Requests(2.24.0), Socks(1.7.1), Websocket(0.57.0)
[cli][info] Found matching plugin afreeca for URL http://play.afreecatv.com/ryuri79
[cli][debug] Plugin specific arguments:
[cli][debug]  --afreeca-username=******** (username)
[cli][debug]  --afreeca-password=******** (password)
[plugin.afreeca][debug] Attempting login as ********
[plugin.afreeca][info] Successfully logged in as ********
error: Unable to open URL: http://resourcemanager.afreecatv.com:9090/broad_stream_assign.html (400 Client Error: Bad Request for url: http://resourcemanager.afreecatv.com:9090/broad_stream_assign.html?return_type=gs_cdn&broad_key=229108890-flash-original-hls)

```

### Additional comments, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

pulbitz:
I tried a test with the modified afeeca.py file. The problem has been fixed.

https://github.com/streamlink/streamlink/pull/3408/files/886be309a6f7653274bca46b8baba72e9dc57a70#diff-3b4336061632b5600984e648cc9bf2a4eec1f02bcdf259c7648f8275994f8c42

PR #3408, opened 2020-12-11
plugins.afreeca: update '_get_channel_info' with 'bno', plugin cleanup

back-to:
- use broadcast number for _get_channel_info
- use FilteredHLSStream instead of --hls-segment-ignore-names
- save cookies for a user session
- add new option --afreeca-purge-credentials
- general plugin cleanup

closes https://github.com/streamlink/streamlink/issues/3407

### DIFF
--- a/src/streamlink/plugins/afreeca.py
+++ b/src/streamlink/plugins/afreeca.py
@@ -4,55 +4,66 @@ import re
 from streamlink.plugin import Plugin, PluginArgument, PluginArguments
 from streamlink.plugin.api import validate
 from streamlink.stream import HLSStream
+from streamlink.stream.hls_filtered import FilteredHLSStreamReader, FilteredHLSStreamWriter
 
 log = logging.getLogger(__name__)
 
-CHANNEL_API_URL = "http://live.afreecatv.com:8057/afreeca/player_live_api.php"
-STREAM_INFO_URLS = "{rmd}/broad_stream_assign.html"
 
-CHANNEL_RESULT_ERROR = 0
-CHANNEL_RESULT_OK = 1
+class AfreecaHLSStreamWriter(FilteredHLSStreamWriter):
+    def should_filter_sequence(self, sequence):
+        return "preloading" in sequence.segment.uri
 
-QUALITYS = ["original", "hd", "sd"]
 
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/afreeca.py
```diff
@@ -4,55 +4,66 @@
...
+        reader.open()
+        return reader
 
 
 class AfreecaTV(Plugin):
-    login_url = "https://member.afreecatv.com:8111/login/LoginAction.php"
+    _re_bno = re.compile(r"var nBroadNo = (?P<bno>\d+);")
+    _re_url = re.compile(r"https?://play\.afreecatv\.com/(?P<username>\w+)(?:/(?P<bno>:\d+))?")
+
+    CHANNEL_API_URL = "http://live.afreecatv.com/afreeca/player_live_api.php"
```
bastimeyer, line 32: Not sure if it's worth it, but the API URL could be changed to `https` and `https` added to `_schema_stream` down below. What's required for this to work is disabling the TLS validation via `self.session.set_option("http-ssl-verify", False)`.

src/streamlink/plugins/afreeca.py
```diff
@@ -4,55 +4,66 @@
...
+
+    _schema_channel = validate.Schema(
+        {
+            "CHANNEL": {
+                "RESULT": validate.transform(int),
+                validate.optional("BPWD"): validate.text,
+                validate.optional("BNO"): validate.text,
+                validate.optional("RMD"): validate.text,
+                validate.optional("AID"): validate.text,
+                validate.optional("CDN"): validate.text
```
bastimeyer, line 49: I know you've only moved it from the module's global scope to the plugin class, but you could change `validate.text` to `str` in one go here (down below as well).

bastimeyer (commented):
@pulbitz 
> Only certain afreecatv streamer get this error.

Could you please check whether these changes are working for you, so that this can be merged and #3407 be closed?

src/streamlink/plugins/afreeca.py
back-to: there website uses `http` in the web browser,
adding `https` here with `http-ssl-verify", False` is currently unnecessary

Issue #3365, opened 2020-11-25
Shared resolution streams undetected by dash plugin


GBeauregard:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.
- [x] I am using the latest development version from the master branch.
confirmed in master

### Description

When multiple DASH entries share a resolution (w/o extra metadata to distinguish?), streamlink won't be able to play the highest quality/highest bitrate stream because it doesn't list it.

### Expected / Actual behavior
Streamlink to play the highest bitrate/highest quality stream

### Reproduction steps / Explicit stream URLs to test
Generate a dash playlist with ffmpeg's dash muxer (may need to use master, there's been some changes with the dash muxer) with two different bitrates encoded at the same resolution. The ffmpeg dash muxer was contributed by those invested in the dash forum so it's all up to spec. Try to play the "best" quality with streamlink, you'll note it will play the lower quality of the two bitrates in the dash muxer playlist and the dash plugin will only list one stream at the given resolution, preventing playing high quality stream. If there's a way to workaround this bug in streamlink by adding metadata in ffmpeg I haven't discovered it yet, but I'll update if I do.

### Log output

Here's an example of a stream being loaded where manual inspection of the mpd shows there's two streams at 800p. Note one is listed, and the one that gets loaded is the lower quality of the two. This is relevant not just for multiple bitrates at the same resolution, but situations where you transcode from e.g. HEVC into H264 for varying platform support.

```
❯ ~/.scripts/loadstream.sh
[cli][debug] OS:         Linux-5.8.18-gentoo-x86_64-Intel-R-_Core-TM-_i7-8750H_CPU_@_2.20GHz-with-gentoo-2.7
[cli][debug] Python:     3.7.9
[cli][debug] Streamlink: 1.7.0
[cli][debug] Requests(2.24.0), Socks(1.7.1), Websocket(0.57.0)
[cli][info] Found matching plugin dash for URL https://example.com/stream/stream.mpd
[plugin.dash][debug] Parsing MPD URL: https://example.com/stream/stream.mpd
[utils.l10n][debug] Language code: en_US
[stream.dash][debug] Available languages for DASH audio streams: NONE (using: n/a)
[cli][info] Available streams: 212p (worst), 320p, 426p, 532p, 746p, 800p (best)
[cli][info] Opening stream: 800p (dash)
```

beardypig:
Any chance you could post an minimal example MPD?

GBeauregard:
[...]

PR #3410, opened 2020-12-12
stream.dash: allow '_alt' streams with the same resolution

back-to:
- create a list of stream_names and rename streams with the same resolution
- `defaultdict` seems the easy way without changing a lot of tests
- added `test_10.mpd`, so we have an actual test file for future updates

closes https://github.com/streamlink/streamlink/issues/3365

### DIFF
--- a/src/streamlink/stream/dash.py
+++ b/src/streamlink/stream/dash.py
@@ -258,7 +258,7 @@ class DASHStream(Stream):
                 elif n == 1:
                     ret_new[f'{q}_alt'] = items[n]
                 else:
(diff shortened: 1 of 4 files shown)


bastimeyer (commented):
Do you want to keep this naming scheme? `Plugin.streams()` renames the duplicates like this, without the second underscore character:
`foo`, `foo_alt`, `foo_alt2`
https://github.com/streamlink/streamlink/blob/275febc93cdab52309afa34dded51906a5703471/src/streamlink/plugin/plugin.py#L365-L366

back-to:
changed it to `_alt2`

PR #3413, opened 2020-12-13
stream.ffmpegmux: disable -start_at_zero for -copyts as default

back-to:
- don't use `-start_at_zero` as default for `-copyts`
- revert https://github.com/streamlink/streamlink/pull/2559

- change command `--ffmpeg-no-start-at-zero` to `--ffmpeg-start-at-zero`
  (there was no release in between this command)

- use `--ffmpeg-start-at-zero` for the facebook plugin,
  because the original fix was for it.
  https://github.com/streamlink/streamlink/issues/2488

---

as this might have some downsides

Might be helpful for https://github.com/streamlink/streamlink/pull/3404

### DIFF
--- a/src/streamlink/plugins/facebook.py
+++ b/src/streamlink/plugins/facebook.py
@@ -79,6 +79,8 @@ class Facebook(Plugin):
                 yield from DASHStream.parse_manifest(self.session, manifest).items()
 
     def _get_streams(self):
+        self.session.set_option("ffmpeg-start-at-zero", True)
+
         done = False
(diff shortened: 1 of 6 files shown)


bastimeyer (approved):
Changes looking good. One thing to note though: `ffmpeg-start-at-zero` on the session should default to `False`, not to `None`. It's already documented as bool.

bastimeyer:
Just for reference, `--ffmpeg-no-start-at-zero` was added in #2892

Issue #2497, opened 2019-06-11
Ignore segments of different resolution than rest of stream


ischmidt20:
## Feature Request

- [x] This is a feature request and I have read the contribution guidelines.

### Description

I would like streamlink to ignore segments of a different resolution than a main stream, particularly when recording/downloading with `-o`.

For some players, notably [FOX](https://www.fox.com/live/channel/FS1/) and [WatchESPN](http://www.espn.com/watch/) (login required for both), the segments during commercial breaks are often of a different resolution and codec than the segments for the actual content. This is not a problem if the stream is playing in a player, but when the goal is to save to a file with `-o`, the resulting .ts file with different resolutions is tricky to work with.

Note that I am using the "manual" method, with a command like so: `streamlink 'hlsvariant://https://content.uplynk.com/channel/d0695f6fa2f143eb95c7a41d96d581c4.m3u8?oid=d09b16c953aa40c98dd8c513526aca5a&ad=espnlive&ad.csid=watchespn:web:espn:live:en&ad.caid=8a474211-a0d0-45c3-a6f3-82d38a9dc9b4&ad.kv=_fw_ae,5525ff59adcaac313923ab89d0a618c5,_fw_euid,41D65629-4BD2-4779-A7E2-85FF9405478E&ct=c&tc=1&exp=1560293022&rn=646822596&ptid=ESPN_tvOS_VDMS&linearv=4&euid=SEC_VDMS&eid=espn_espn1&sig=1c6d9b0859f81575a92d9932b9cc18d9803a38bc84d4ae7ae1f28960775a1ae0' best -o espn.ts`

It appears such segments come up when there is an `#EXT-X-DISCONTINUITY` in the playlist file, in case that helps. There is also an `#UPLYNK-SEGMENT` tag, which will end with `,ad` in the case of an advertisement (see [here](https://github.com/ytdl-org/youtube-dl/issues/21363), you can tell I'm really after this problem), but this is not all-encompassing and ignoring segments with such a tag beforehand unfortunately does not completely solve the problem.

### Expected / Actual behavior

The end goal here is that when recording with streamlink, the .ts file should consistently have the same resolution throughout. Any segments of a different resolution should be ignored. `ffmpeg -i file.ts -c copy file.mp4` should result in an .mp4 file playable in dummy players such as QuickTime, which does not happen in the case of a multi-resolution file.

### Additional comments, screenshots, etc.

[...]

PR #3414, opened 2020-12-13
stream.hls: remove --hls-segment-ignore-names command

back-to:
It has some downsides such as `--hls-timeout`,
Plugin specified FilteredHLSStream should be used instead.

https://github.com/streamlink/streamlink/pull/3187

---

Afreeca Plugin as an Example

https://github.com/streamlink/streamlink/pull/3408

---

Revert https://github.com/streamlink/streamlink/pull/1432

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -1,5 +1,4 @@
 import logging
-import re
 import struct
 from collections import OrderedDict, defaultdict, namedtuple
 from urllib.parse import urlparse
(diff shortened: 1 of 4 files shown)


bastimeyer (commented):
I agree with the removal of this feature.

However, would it maybe make more sense merging the FilteredHLS logic into the regular HLS classes instead? By doing this, we could keep the CLI argument and custom filtering feature. The drawback would be the need of refactoring the regular HLS classes a bit and making it a bit more slow due to the additional filtering checks on each write call, which was the reason why I implemented this separately. The classes which depend on FilteredHLS also need to get refactored, which is currently only the HLSStream stuff of the Twitch plugin (and AfreecaTV once that PR gets merged).

back-to:
> However, would it maybe make more sense merging the FilteredHLS logic into the regular HLS classes instead?

maybe, but probably to much work for streamlink 2.0.0

Ref https://github.com/streamlink/streamlink/issues/2497

bastimeyer:
I don't think it's too much work. If we keep the filtered tests separated from the other HLS tests, it should be pretty simple. Let me give this a try...

PR #3417, opened 2020-12-14
plugins.plugin: use the same cls.logger 'plugins'

back-to:
since https://github.com/streamlink/streamlink/pull/3366
https://github.com/streamlink/streamlink/blob/5e7a04b006d0e4d7a884b13bf2aa8047791b837b/src/streamlink/session.py#L412
the logger is `[plugins.afreeca]` which was `[plugin.afreeca]` before

update the cls.logger so it will be `[plugins.afreeca]` also

after this PR

```
[plugins.afreeca][debug] Restored cookies: ...
...
[plugins.afreeca][debug] Attempting to authenticate using cached cookies
```


gravyboat:
Good catch.

PR #3419, opened 2020-12-17
cli: move --stream-url to different args group

bastimeyer:
`--stream-url` is currently listed in the "Stream transport options" args group, which doesn't make sense, as it's CLI output related, and not stream transport related. This PR moves it to the "stream" args group, right under `--url` and `--default-stream`.

See the current latest CLI docs:
https://streamlink.github.io/latest/cli.html#cmdoption-stream-url


gravyboat:
Good catch.

PR #3420, opened 2020-12-18
ci: add netlify docs preview deploy config

bastimeyer:
As discussed in https://github.com/streamlink/streamlink/pull/3335#issuecomment-742131264, here's a config for docs previews in PRs via netlify.com. This will make reviewing docs changes way more easy.

This config will be required before enabling the repo integration, otherwise it'd create production builds on pushes to the master branch, which is not what we want. I've added comments to the config, so that it's clear what everything's for.

----

So far, I haven't received a response from them yet regarding the upgrade to their "open source" plan. If they are so kind and update the account from the free "starter" plan, that would enable us having multiple account maintainers and an increased build time (which doesn't matter so much).

Since I feel like they probably won't answer my message, as I had to send it on a different contact form, I will try this again via their community forums, which is the only contact method that's supported for starter accounts, as I've already mentioned. There's a post on their forums though where you can discuss stuff in private after asking first. I will try that later today.

The upgrade would probably require us adding a CoC and a link to their site in the docs. I've already explained in my contact message that I don't know whether these are hard requirements for being able to apply to the open source plan. If they are, then we'll have to figure something out or keep the free account, which is fine, too. Personally, I'm fine with adding a **very** simple CoC and adding a "thank you" note/link to the index page or even adding a new "thank you" page and listing them there.

----

Please don't merge until the account stuff is resolved.

This is how the PR status checks look like:
https://github.com/bastimeyer-test/streamlink-test/pull/8

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,7 +9,7 @@ insert_final_newline = true
 max_line_length = 128
 trim_trailing_whitespace = true
 
-[*.{cfg,svg,yml}]
+[*.{cfg,svg,toml,yml}]
 indent_size = 2
 
 [*.{md,markdown}]
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,27 @@
+# Streamlink documentation preview config for pull requests via netlify.com
+# Some settings like PR statuses for example can only be configured via the netlify website
+
+# https://docs.netlify.com/configure-builds/file-based-configuration/#build-settings
+[build]
+  command = "pip install -U pip setuptools && pip install -e . && pip install -U -r docs-requirements.txt && make --directory=docs clean html"
+  publish = "docs/_build/html"
+  # Since production builds can't be disabled without disabling PRs as well, pushes to the master branch need to be ignored.
+  # Also ignore pull requests which don't include any changes that would alter the docs:
+  # - docs/
+  # - docs-requirements.txt
+  # - src/streamlink_cli/argparser.py
+  # https://docs.netlify.com/configure-builds/file-based-configuration/#ignore-builds
+  ignore = "[ $(git rev-parse --abbrev-ref HEAD) = master ] || git diff --quiet master HEAD docs/ docs-requirements.txt src/streamlink_cli/argparser.py"
+
+[build.environment]
+  # Set the latest natively available Python version in the Ubuntu env (currently 16.04 / Xenial)
+  # https://docs.netlify.com/configure-builds/manage-dependencies/#python
+  # https://github.com/netlify/build-image/blob/xenial/included_software.md
+  PYTHON_VERSION = "3.7"
+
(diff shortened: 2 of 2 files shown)


bastimeyer:
> a bit over the top for a project like Streamlink
> building it locally isn't a big problem.

And it's always annoying and has not been done by the reviewer(s) most of the time anyway so far. Having the docs built and hosted for everyone to see and review, even the PR author, is an improvement and it encourages making improvements to the docs. Only one always passing PR status check gets added to PRs, and only to those with actual changes to the docs, so there's no need to worry about spam.

I would very much like to give this a try at least. If there are major problems, the integration can be removed again.

If you want, we can wait until the 2.0.0 release, so that it doesn't cause unnecessary noise.

----

Now to their answer to my questions. They said that having a CoC is mandatory, but having a simple/minimal one is acceptable. Adding a small link somewhere on the index page for the second requirement is enough.

Not sure if you guys agree with this though. As said, I don't have a problem with these two requirements, but if one does, we could go with the free account as well.

My CoC suggestions are this:
1. https://github.com/domgetter/NCoC/blob/master/CODE_OF_CONDUCT.md
2. https://github.com/dorssel/usbipd-win/blob/d68b39d4af46288eb4601dd4ecf77643f0adf91a/CODE_OF_CONDUCT.md (the only version of this I found on Github)
3. Or a simplified/shortened version of 2

The link diff could look like this:
```diff
diff --git a/docs/index.rst b/docs/index.rst
index 530b643a..bf1c254a 100644
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -75,3 +75,9 @@ See their respective sections for more information on how to use them.
     donate
     applications
     thirdparty
+
+Thank you
+---------
+
+- `Github <https://github.com/>`_
+- `Netlify <https://www.netlify.com/>`_
```

gravyboat:
@bastimeyer Thanks for getting the details back from them. I'm in favor of CoC #2 or a slightly shortened version, and the doc link is fine since it's only on the index and only text.

bastimeyer:
Can we give this a go then? We've already merged the CoC without even adding the netlify integration here. As said, this config is required before I can add the integration, as it would otherwise run production builds on pushes onto the master branch.

What I've also noticed is that the ignore option currently doesn't catch any changes to PluginArguments. That's a bit difficult to catch in a git diff though. All that could be done is increasing the diff size and checking the output for `PluginArgument`, eg. like this `git diff -U8 | grep PluginArgument`, but it would catch unrelated changes too.

gravyboat:
> Can we give this a go then?

Let's do it!

bastimeyer:
I've added the netlify integration to the repo and opened #3463 where you can see the docs preview in the PR status check list.

I'm going to open another PR with the thank you note and will reference the PR link in the application form.

PR #2861, opened 2020-04-04
docs: fix and rewrite various sections and pages

bastimeyer:
I don't think we need to (temporarily) include the old bintray uploads here.

PR #3423, opened 2020-12-19
docs: fix link in readme

bastimeyer:
Just a small correction in the readme. The section/header in the docs was renamed in #2861 and the readme didn't get updated accordingly...


gravyboat:
Good catch.

PR #3425, opened 2020-12-20
add --niconico-timeshift-offset option to the nicolive plugin

mark-ignacio:
`--niconico-timeshift-offset` allows the nicolive Streamlink plugin to start at a particular timestamp in the HLS stream for [timeshift streams](http://help.nicovideo.jp/live_en/2013/06/305what-is-a-timeshiftwatch-period.html).

This proves particularly useful because some broadcasts' HLS streams start with a long "waiting room" during a live broadcast that the web player skips past by default.

### DIFF
--- a/src/streamlink/plugins/nicolive.py
+++ b/src/streamlink/plugins/nicolive.py
@@ -11,6 +11,7 @@ from streamlink.plugin import Plugin, PluginArgument, PluginArguments
 from streamlink.plugin.api import useragents
 from streamlink.stream import HLSStream
 from streamlink.utils.times import hours_minutes_seconds
+from streamlink.utils.url import update_qsd
 
 _log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/nicolive.py
```diff
@@ -241,6 +250,14 @@ def handle_api_message(self, message):
         if message_parsed["type"] == "stream":
             data = message_parsed["data"]
             self.hls_stream_url = data["uri"]
+            # load in the offset for timeshift live videos
+            offset = self.get_option("timeshift-offset")
+            if offset:
+                if 'timeshift' in self.wss_api_url:
+                    self.hls_stream_url += '&start={0}'.format(offset)
```
back-to, line 257: use `update_qsd`
https://github.com/streamlink/streamlink/blob/master/tests/test_utils_url.py#L33-L45

src/streamlink/plugins/nicolive.py
```diff
@@ -241,6 +250,14 @@ def handle_api_message(self, message):
...
+                    _log.debug("Timeshift offset stream URL is now {0}".format(self.hls_stream_url))
+                else:
+                    _log.debug("Not a timeshift video, ignoring offset")
```
back-to, line 260: you don't need this

src/streamlink/plugins/nicolive.py
```diff
@@ -241,6 +250,14 @@ def handle_api_message(self, message):
         if message_parsed["type"] == "stream":
             data = message_parsed["data"]
             self.hls_stream_url = data["uri"]
+            # load in the offset for timeshift live videos
+            offset = self.get_option("timeshift-offset")
+            if offset:
```
back-to, line 255:
```py
            if offset and 'timeshift' ...
```

PR #3442, opened 2020-12-25
tools: force LF line endings via .gitattributes

bastimeyer:
https://stackoverflow.com/questions/29435156

- add additional linting steps to Github actions for checking the
  line endings and file permissions
- fix line endings and file permissions


gravyboat:
Good addition. Thanks @bastimeyer.

Issue #2055, opened 2018-09-05
gst-player example doesn't appear to work on macOS


jlsalmon:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description

I tried running `examples/gst-player.py` and it doesn't produce any output or video playback for me, just a blank Python icon in the dock.

OS: macOS 10.13.6
gst-python version: 1.14.2 (good and bad plugins installed)
gstreamer version: 1.14.2

### Expected / Actual behavior

I expected a window to pop up with the stream playing.

### Reproduction steps / Explicit stream URLs to test

1. Run the `gst-player.py` example on macOS with any stream.

### Log output

No log output is produced, even when modifying the script to set `streamling.set_loglevel("debug")`.

### Additional comments, screenshots, etc.

Running the script through my debugger I see that the `StreamlinkPlayer.on_source_setup` method gets called, but the `StreamlinkPlayer.on_source_need_data` method is never called.

back-to:
The `gst-player.py` **example is outdated**,
but it should at least start the player correctly.

This is the output on Linux for Youtube and Twitch,
the player starts as expected.

```
$ ./examples/gst-player.py URL best
./examples/gst-player.py:9: PyGIWarning: Gst was imported without specifying a version first. Use gi.require_version('Gst', '1.0') before import to ensure that the right version gets loaded.
  from gi.repository import GObject as gobject, Gst as gst
./examples/gst-player.py:100: PyGIDeprecationWarning: Since version 3.11, calling threads_init is no longer needed. See: https://wiki.gnome.org/PyGObject/Threading
  gobject.threads_init()
./examples/gst-player.py:21: PyGIDeprecationWarning: GObject.MainLoop is deprecated; use GLib.MainLoop instead
  self.mainloop = gobject.MainLoop()
```

`GStreamer 1.14.2`

```py
>>> import gi
>>> gi.__version__
'3.28.3'
```

PR #3445, opened 2020-12-25
Examples: use python3 interpreter instead of python

amurzeau:
Change python interpreter of `gst-player.py` to python3 as now only python 3 is supported, and to match the other example `opencv-face.py`.


back-to:
`gst-player.py` is totally outdated and broken, a small update like this is not really needed.

https://github.com/streamlink/streamlink/blob/master/examples/gst-player.py

https://github.com/streamlink/streamlink/issues/2055

PR #3459, opened 2020-12-28
build: include man page in wheels

### LINKED ISSUE #1753: No man page with pip install
### Checklist

- [ ] This is a bug report.
- [x] I have read the contribution guidelines.

### Description

When installing streamlink with pip, no man page gets installed

a man page gets installed during installation of streamlink with pip

1. ``pip install --user streamlink``
2. ``man streamlink``
3. ``No manual entry for streamlink``
4. I get the same results when using ``pip install streamlink``

### Logs

```
[cli][debug] OS:         Linux-4.13.0-43-generic-x86_64-with-Ubuntu-17.10-artful
[cli][debug] Python:     3.6.3
[cli][debug] Streamlink: 0.12.1

```

bastimeyer:
Resolves #1753 

This adds the man page, if it exists, to wheels via the `data_files` keyword:
https://setuptools.readthedocs.io/en/latest/references/keywords.html

The `data_files` keyword is marked as deprecated, but there doesn't seem to be another way of mapping files from one path to another in the distribution file. What it says about not being supported by wheels is wrong. `package_data` doesn't seem to work here without changing build scripts. At least I wasn't able to make it work.

`sdist` does not include any `data_files`.

Regarding the manpages themselves, it might be necessary on some systems to extend the `MANPATH` env var with `MANPATH="${XDG_DATA_HOME:-$HOME/.local/share}/man:$MANPATH"` so that it looks up man pages in `$XDG_DATA_HOME/man` (which defaults to `$HOME/.local/share/man`) too.

To validate the changes:
```
$ make --directory=docs clean
$ python setuptools.py bdist_wheel
$ unzip -l dist/streamlink-2.0.0+10.gc6a1806-py3-none-any.whl | grep share/man/man1/streamlink.1

$ make --directory=docs clean man
$ python setuptools.py bdist_wheel
$ unzip -l dist/streamlink-2.0.0+10.gc6a1806-py3-none-any.whl | grep share/man/man1/streamlink.1
    49409  2020-12-28 12:05   streamlink-2.0.0+10.gc6a1806.data/data/share/man/man1/streamlink.1
$ pip install ./dist/streamlink-2.0.0+10.gc6a1806-py3-none-any.whl
$ ls ~/.local/share/man/man1/streamlink.1 
/home/basti/.local/share/man/man1/streamlink.1
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -150,7 +150,10 @@ jobs:
       - name: Install dependencies
         run: |
           ./script/install-dependencies.sh
+          python -m pip install -r docs-requirements.txt
           python -m pip install --upgrade wheel twine
+      - name: Build man page
+        run: make --directory=docs man
       - name: Installer file name
         id: installer
         run: echo ::set-output name=filename::streamlink-$(python setup.py --version | sed 's/+/_/')
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,7 @@ from setuptools import find_packages, setup
 import versioneer
 
 
+data_files = []
 deps = [
     "requests>=2.21.0,<3.0",
     "isodate",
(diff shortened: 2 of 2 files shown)


bastimeyer:
Btw, the man page itself should probably be changed, as it currently includes the entire `cli.rst` docs page, which is technically not a man page:
https://github.com/streamlink/streamlink/blob/009bbd82e2e40bc141f85125eeb5e0569f52e750/docs/conf.py#L184-L186

PR #3483, opened 2021-01-09
session: implement --interface, --ipv4 and --ipv6

### LINKED ISSUE #2192: Specify manually the network interface to use in a multi-homed setup?
## Feature Request
Add the possibility to manually specify the network interface to use. Which can be very useful for multi-homed networks with multiple network interfaces.

### Description

In Streamlink you can specify the proxy to use for streaming. Similar to that it would be also very useful if it would be possible to specify which network interface to use for streaming.

Be able to specify for example "--interface eth1" to use the eth1 interface for streaming.

### LINKED ISSUE #3474: param for source address or IPv4/IPv6
## Feature Request

- [x] I am using the latest development version from the master branch.

### Description

Is it possible to add parameters to select which ip to use on recording?

Like `curl`'s `-4 -6 --interface`, `youtube-dl`'s `-4 -6 --source-address`

If it can be implemented, we can prevent 429 errors at many cases.

I get 429 error on some cases because of IPv4 NAT, although a dedicated IPv6 address is available. 

### Additional comments, etc.

bastimeyer:
Resolves #3474 
Resolves #2192

Implements `--interface`, `-4`/`--ipv4` and `-6`/`--ipv6`, similar to curl.

**Concerns**
- What I don't like about this implementation is that it overrides the address family resolving in `requests.packages.urllib3.util.connection` globally, as it's not session based. This means that when someone creates two sessions from the API and sets the `ipv4` or `ipv6` option, all other sessions will be affected as well.
- This does only affect HTTP/HTTPS connections. Web sockets for example are currently all managed individually in certain plugins. The implementation of a common WSSession similar to the HTTPSession would fix this.
- Not sure if the "general" arguments group is the correct choice. Technically it could also go into the "transport" arguments group.

----

**Examples**

Default interface:
```bash
$ streamlink -l none -o /dev/null twitch.tv/gamesdonequick best &
$ ss -npt | awk '/streamlink/ {print $4; exit}'
192.168.178.99:33253
```

Second interface:
```bash
$ streamlink -l none -o /dev/null twitch.tv/gamesdonequick best --interface 192.168.100.1 &
$ ss -npt | awk '/streamlink/ {print $4; exit}'
192.168.100.1:43101
```

Invalid interface:
```bash
$ streamlink -l none -o /dev/null twitch.tv/gamesdonequick best --interface 1.2.3.4
error: No playable streams found on this URL: twitch.tv/gamesdonequick
```

Automatic address family resolving:
```bash
$ streamlink -l none -o /dev/null 'youtube.com/watch?v=xcJtL7QggTI' best &
$ ss -npt | awk '/streamlink/ {print $5; exit}'
[2a00:1450:4001:819::200e]:443
```

IPv4 resolving:
```bash
$ streamlink -l none -o /dev/null 'youtube.com/watch?v=xcJtL7QggTI' best --ipv4 &
$ ss -npt | awk '/streamlink/ {print $5; exit}'
74.125.126.91:443
```

IPv6 resolving (IPv4-only example):
```bash
$ streamlink -l none -o /dev/null twitch.tv/gamesdonequick best --ipv6
error: No playable streams found on this URL: twitch.tv/gamesdonequick
```

### DIFF
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -2,8 +2,11 @@ import logging
 import pkgutil
 from collections import OrderedDict
 from functools import lru_cache
+from socket import AF_INET, AF_INET6
 
 import requests
+import requests.packages.urllib3.util.connection as urllib3_connection
+from requests.packages.urllib3.util.connection import allowed_gai_family
 
 from streamlink import __version__, plugins
 from streamlink.compat import is_win32
@@ -30,6 +33,9 @@ class Streamlink:
     def __init__(self, options=None):
         self.http = api.HTTPSession()
         self.options = Options({
+            "interface": None,
+            "ipv4": False,
+            "ipv6": False,
             "hds-live-edge": 10.0,
             "hds-segment-attempts": 3,
             "hds-segment-threads": 1,
@@ -83,7 +89,13 @@ class Streamlink:
         **Available options**:
 
         ======================== =========================================
-        hds-live-edge            ( float) Specify the time live HDS
+        interface                (str) Set the network interface,
+                                 default: ``None``
+        ipv4                     (bool) Resolve address names to IPv4 only.
+                                 This option overrides ipv6, default: ``False``
+        ipv6                     (bool) Resolve address names to IPv6 only.
+                                 This option overrides ipv4, default: ``False``
+        hds-live-edge            (float) Specify the time live HDS
                                  streams will start from the edge of
                                  stream, default: ``10.0``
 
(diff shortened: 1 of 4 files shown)


bastimeyer:
Rebased and fixed merge conflict which was introduced by #3486

gravyboat (approved):
I was hoping other people would weigh in but this has been open for a bit so I think we should get it merged. While the override is a bit annoying I don't think it's a problem in the grand scheme of things. If people are using this option they know what they're doing.

Issue #2988, opened 2020-05-25
List returned from `streamlink.streams` sometimes empty


theodoretliu:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.

### Description

Hi, I'm using `streamlinks` to pull frames from Twitch streams, and it's been generally great. However, I've been running into intermittent issues with it. I have a simple Flask app set up similar to below
```python
app = Flask(__name__)
session = Streamlink()

@app.route('/')
def index():
    user = request.args.get('user')
    streams = session.streams(f"https://twitch.tv/{user}")
    # streams is sometimes length 0 here ...
```
I'm positive that the stream that's being fetched is online. I recognize that `streamlink` will return 0 if the stream is offline, but that is definitely not the case.

Once I restart the web server, it usually goes back to normal and returns the correct number of streams.

### Expected / Actual behavior

I expect `streamlink` to return a list of streams exactly when the stream is online and empty otherwise.

### Reproduction steps / Explicit stream URLs to test

The code above is the best I can do. I've tried to recreate the behavior by running the `session.streams` function in an infinite loop while I manually have my own Twitch channel go online and offline, but it seems to work as expected. It's only in this long running web server that I see the issue. Some of the bread crumbs from Sentry say there are underlying 404s going on when requesting the HLS stream, but that's all I really know.

beardypig:
Hi @theodoretliu. There is some state stored in the streamlink sessions between invocations of the `index` call. I suggest you create a new streamlink session in `index` and see if that resolves the issue.

theodoretliu:
@beardypig thanks for the tip! I've actually tried that before, but that also caused a crazy memory leak on my web server; it seems like the session object (or something within it) is not getting garbage collected because of some circular references. Any tips on that?

theodoretliu:
@beardypig @back-to could I get a follow-up on the potential cause of a memory leak?

back-to:
@theodoretliu 

> could I get a follow-up on the potential cause of a memory leak?

well we don't know what you are doing, your code in the first post do not work.

theodoretliu:
@back-to  here's a really short reproducible example
```python
from streamlink import Streamlink

while True:
    session = Streamlink()

    stream = session.streams("https://www.twitch.tv/mrsavage")
```
If you run this program, the memory consumption will slowly but steadily rise.

back-to:
@theodoretliu 

seems like python doesn't remove all the garbage data fast enough

[...]

PR #3486, opened 2021-01-12
utils: remove custom memoize decorator

bastimeyer:
This replaces the custom `memoize` decorator in favor of the native [`functools.lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) decorator (available since py 3.2).

Since I didn't set `maxsize=None`, this also changes the cache size of each decorated method, which was infinite prior to these changes and is now set to 128 with the LRU algorithm.

Not sure why I had to force-push and explicitly set `maxsize=128` on py `<3.8`, as it's the default there as well:
https://github.com/python/cpython/blob/3.7/Lib/functools.py#L458
https://github.com/python/cpython/blob/3.8/Lib/functools.py#L487

### DIFF
--- a/src/streamlink/plugins/wwenetwork.py
+++ b/src/streamlink/plugins/wwenetwork.py
@@ -1,13 +1,13 @@
 import json
 import logging
 import re
+from functools import lru_cache
 from urllib.parse import parse_qsl, urlparse
 
 from streamlink import PluginError
 from streamlink.plugin import Plugin, PluginArgument, PluginArguments
 from streamlink.plugin.api import useragents
 from streamlink.stream import HLSStream
-from streamlink.utils import memoize
 from streamlink.utils.times import seconds_to_hhmmss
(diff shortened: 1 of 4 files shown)


bastimeyer:
Btw, could the infinite cache size of `session.resolve_url` be the cause of https://github.com/streamlink/streamlink/issues/2988#issuecomment-634229002 ?
This basically keeps all session instance references on the method cache, so the GC can't free it up.

beardypig (approved):

beardypig:
> Btw, could the infinite cache size of `session.resolve_url` be the cause of [#2988 (comment)](https://github.com/streamlink/streamlink/issues/2988#issuecomment-634229002) ?
> This basically keeps all session instance references on the method cache, so the GC can't free it up.

You might be on to something there. If it is that, then setting `maxsize` would solve it.

Issue #791, opened 2017-04-14
Auto-detect file extension for output file


nyuszika7h:
### Checklist

- [ ] This is a bug report.
- [ ] This is a plugin request.
- [x] This is a feature request.
- [x] I used the search function to find already opened/closed issues or pull requests.

### Description

I looked at the documentation and issues but I didn't find a way to automatically choose the correct file extension for the output file when using `-o`. Of course, for example I know that it will be `.ts` for HLS streams, but I might not always know. It would be nice to be able to use `-o filename.{ext}` or something like that. Or an option to remux, since if I use `.mkv` for example, it won't remux to a Matroska container. It will still save fine, but I need to rename the file afterwards for it to make sense.

### Environment details (operating system, python version, etc.)

* Windows 10 x64
* Python 3.6.1
* streamlink 0.5.0

bastimeyer:
IMHO, the video/audio container shouldn't be determined via the extension of the output file (and vice versa). File name extensions are just being used by your OS to link programs with this *potential* file format. It doesn't matter which extension a file has as long as your video player, editing software, etc. is able to correctly parse the container format and video/audio data.

> Or an option to remux

A `--mux` or `--remux` parameter would make more sense where you can choose the container format and properties. Then you can set the file name extension accordingly to your desires.

You can always remux the output on the fly with ffmpeg like this:
```bash
streamlink -O URL QUALITY | ffmpeg -i pipe:0 -c:v copy -c:a copy -f matroska -y file.mkv
```

nyuszika7h:
You're right, but file extensions are still widely used for audio and video files and even if players can handle it, it doesn't make sense to have an extension that doesn't match the container. Thanks for the ffmpeg tip, though.

imerr:
with #1576 merged would it be possible to see similar formatting for the output filename?

(I'm not sure how the internals work or how much work this would be, just me looking for a way to do this and finding the --title flag can do formatting)

GrayFace:
I'll mirror my proposal here then:

**Description**
Weirdly, metadata variables are supported only for rarely used "--title" parameter. They would be very useful for output parameter.

**Proposal:**

Add a new parameter, say "-ox", that would act like "-o", but with variables support.
Expose {title}, {author}, {category} etc. to it.
Also, at some point in the future having upload data/time as a variable(s) would be very useful. I suspect there was no incentive for adding them, because there's not much need for them in "--title" parameter.

flower1024:
[...]

PR #3495, opened 2021-01-19
docs: fix description of `--ffmpeg-fout`

bastimeyer:
as well as `--ffmpeg-{audio,video}-transcode`

The current description implies that this CLI argument remuxes the output, which is not the case. See https://github.com/streamlink/streamlink/issues/791#issuecomment-762567588


bastimeyer:
https://deploy-preview-3495--streamlink.netlify.app/cli.html#cmdoption-ffmpeg-fout

PR #3499, opened 2021-01-21
docs: remove CLI tutorial from man page

bastimeyer:
and add synopsis with examples

----

I'd like to embed some sections from the CLI page in the man page, but that would require rewriting the CLI page and making it modular, so that the sections can be included without duplicating them. At least the tutorial does not get included with this PR.

The examples could probably be improved, so if you have suggestions, please tell.

```
$ make --directory=docs clean man
$ man ./docs/_build/man/streamlink.1
```

### DIFF
--- a/docs/_man.rst
+++ b/docs/_man.rst
@@ -1,0 +1,29 @@
+:orphan:
+
+
+Streamlink
+==========
+
+Synopsis
+--------
+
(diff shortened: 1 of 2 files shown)


gravyboat:
Looks fine to me. Let's leave this open for a day and see if anyone else has feedback.

PR #3512, opened 2021-01-25
plugins.radiko: faster download api

iucario:
radiko: use faster download api

### DIFF
--- a/src/streamlink/plugins/radiko.py
+++ b/src/streamlink/plugins/radiko.py
@@ -49,7 +49,7 @@ class Radiko(Plugin):
         return url, token
 
     def _timefree(self, station_id, start_at):
-        m3u8_url = 'https://tf-rpaa.smartstream.ne.jp/tf/playlist.m3u8'
+        m3u8_url = 'https://radiko.jp/v2/api/ts/playlist.m3u8'
         token, area_id = self._authorize()
         lsid = hashlib.md5(str(random.random()).encode('utf-8')).hexdigest()
         end_at = self._get_xml(start_at, station_id)


bastimeyer:
This was just changed by #3394. Your PR would revert this.
Can you please explain your reasoning? If this keeps getting changed back and forth, then there needs to be a better solution.

iucario:
Because this api downloads much faster. 

The old one `tf-rpaa.smartstream.ne.jp`  takes 2 hours to download a 2 hour stream, while this one takes minutes.

Also some stations on radiko.jp only use `radiko.jp/v2/api/ts` for timefree, e.g. HOUSOU-DAIGAKU. So the current version of streamlink won't work on those station's timefree stream.

gravyboat:
This has been hanging for a while now. I'm going to need some data that the changes here actually improve the plugin since #3394 was recently merged. Can you please show how this is faster with some benchmarks?

iucario:
OK I'll compare the 2 APIs in speed and availability.

1. **Speed**
- `tf-rpaa.smartstream.ne.jp` takes N minutes to download a N-minute audio. It's like live streaming. Every 5 seconds you get a 5-second-long piece. This API is used both in `live` and `timefree`, maybe that's why they stream in the same way.
- `radiko.jp/v2/api/ts`
I tested it on a 2-hour audio, and got result `Written 41.2 MB (52s @ 809.9 KB/s)`.

    Apparently `radiko.jp/v2/api/ts` is much faster.

2. **Availability**
    There are 15 stations on radiko.jp. `tf-rpaa.smartstream.ne.jp` doesn't work on RN1, JOAK-FM, JOAK, HOUSOU-DAIGAKU the 4 stations. While `radiko.jp/v2/api/ts` works on all 15 stations.

Conclusion. I think we can safely merge this change.

gravyboat:
Is `tf-rpaa.smartstream.ne.jp` only serving live content or can you get the same live content off of `radiko.jp/v2/api/ts`? I don't really care about people downloading video/audio in a timely manner if it's been recorded previously. Streamlink is designed for watching/listening to LIVE streams, not pre-recorded content so the performance of that content isn't something that matters to me.

iucario:
`radiko.jp/v2/api/ts` can only be used for `timefree` not `live`.

There are 2 functions for `live` and `timefree` respectively in the code with different URLs. I only suggest changing the URL of `timefree`. No influence on `live`.

bastimeyer:
I don't mind changing it back to the old URL, but according to this comment

> [Old m3u8 url has been invaild when I used steamlink for radiko site on 12/6 first.](https://github.com/streamlink/streamlink/pull/3394#issuecomment-741604793)

the old URL was invalid at some point, which is the reason why it got changed in #3394.

Your only reasoning is that it downloads faster, which is legitimate, but considering that it already has been "invalid"/offline/inaccessible once doesn't make it sound very future-proof. As said, changing it back-and-forth is not a solution here, so if it's indeed "unstable", then there should at least be a check for availability of the faster endpoint before falling back to the current one, if it's possible.

iucario:
Asked someone who has been using radiko.jp for years, he said that `radiko.jp/v2/api/ts` had been available all the time. 
I guess @smallbomb couldn't access the URL was because of geo-blocking instead of invalid URL.

smallbomb:
@iucario
My situation:
I can access `tf-rpaa.smartstream.ne.jp`  instead of `radiko.jp/v2/api/ts`  using the same IP  at the time. :(

I have an idea.
When `radiko.jp/v2/api/ts` can't be access first, change to try `tf-rpaa.smartstream.ne.jp` 
Keep speed and availability.

These are my idea. what do you think?

iucario:
@smallbomb 
As long as we can keep `radiko.jp/v2/api/ts` I'm OK with that.

kikuyan:
Sorry for my sudden intrusion.

I'm facing just the same problems as @iucario and expecting them to be solved, but the situation seems to have not changed recently. I did some research and found that `radiko.jp/v2/api/ts` should be available to all 110 radio stations. `tf-rpaa.smartstream.ne.jp` is available to 39, only about 1/3, radio stations.

May I ask @smallbomb how you run streamlink to listen to radiko streams? Using VPN or proxy? You don't seem to be in Japan from your profile.
Do you remember which radio station did error occur? What specific error?

smallbomb:
@kikuyan 

> May I ask @smallbomb how you run streamlink to listen to radiko streams?

VPN (sever list from VPN Gate)

> Do you remember which radio station did error occur? What specific error?

step1: 
I used streamlink tool to download audio by 2~3 vpn server but still failed.
step2:
I open browser debug mode, then I observed it access `tf-rpaa.smartstream.ne.jp`  (I could listen radio using gerneral browser.)
step3:
I tried to get `radiko.jp/v2/api/ts` playlist data by browser but the response was `404 not found`.
However, I could get  `tf-rpaa.smartstream.ne.jp` playlist data by browser.
Finally:
I changed to `tf-rpaa.smartstream.ne.jp` from `radiko.jp/v2/api/ts`, then I could normally download audio by tool.

kikuyan:
Thank you for your quick reply and sorry for my late response.

I know VPN Gate, but when I use one, I still can use `radiko.jp/v2/api/ts` without problems for the streams which `tf-rpaa.smartstream.ne.jp` is used by the web browser. (I don't need vpn and usually don't use it, I live in Japan.)
For example, for these streams:
https://radiko.jp/#!/ts/QRR/20210327090000
https://radiko.jp/#!/ts/YFM/20210327090000

I'm wondering what's the difference, where the cause resides. `radiko.jp/v2/api/ts` should  be accessible to everyone.

Could you please try the stream below with your web browser? This is the stream for which only `radiko.jp/v2/api/ts` is supposed to be provided. 
https://radiko.jp/#!/ts/RN1/20210327090000

Do you see `radiko.jp/v2/api/ts` is used by the web browser when played? If so, could you please try it with streamlink with radiko plugin modified to use `radiko.jp/v2/api/ts`? Can it be played?

smallbomb:
@kikuyan 

> Do you see radiko.jp/v2/api/ts is used by the web browser when played? If so, could you please try it with streamlink with radiko plugin modified to use radiko.jp/v2/api/ts? Can it be played?

Thank you for providing urls.

Now, I can download audio using `radiko.jp/v2/api/ts`. 
I can't reproduce the same problem... 
Sorry. :(

kikuyan:
Thank you very much!

Do you mean that you can now download audio from QRR and/or YFM, too?
If yes, that's strange but a good news.
If no, that's still strange because it means that you can use it for some radio stations but cannot for other ones. While we, @iucario and I, can use it for every radio stations.

smallbomb:
> Do you mean that you can now download audio from QRR and/or YFM, too?

Yes. :)

kikuyan:
Thank you again!
But I should have waited a little longer... #3655
Sorry to have troubled you.

gravyboat:
I'm going to close this since https://github.com/streamlink/streamlink/pull/3655 exists and will address this issue going forward.

PR #3533, opened 2021-02-02
tests.plugins: parametrize can_handle_url tests 

bastimeyer:
1. **Fix invalid plugin class names**
   Class names should always be written in camel-case with the first letter being uppercase.
2. **Use `PluginCanHandleUrl` base class for `can_handle_url` plugin tests**
   - The plugin URL tests are currently a total mess with several different test styles. This commit therefore adds a common base class that runs the test assertions on declarative URL lists, namely `should_match` and `should_not_match`, which was the most common style. This simplifies the plugin tests and makes them much more legible.
   - The URL tests get parametrized via `pytest_generate_tests` in `tests/plugins/conftest.py`, so that each URL gets turned into its own test case.
   - This commit also removes lots of unnecessary and unrelated negative URL tests. There might still be some left for certain plugins where I wasn't 100% sure that the URLs are indeed unrelated. Most unrelated URL tests were youtube and tvcatchup, probably due to copy-pasting from other test modules.
   - Instead of having inconsistent negative tests, a list of generic negative URL matches has been added.

### DIFF
--- a/src/streamlink/plugins/ard_live.py
+++ b/src/streamlink/plugins/ard_live.py
@@ -10,7 +10,7 @@ from streamlink.utils import parse_json, verifyjson
 log = logging.getLogger(__name__)
 
 
-class ard_live(Plugin):
+class ARDLive(Plugin):
     _url_re = re.compile(r"https?://((www|live)\.)?daserste\.de/")
     _player_re = re.compile(r'''data-ctrl-player\s*=\s*"(?P<jsondata>.*?)"''')
     _player_url_schema = validate.Schema(
@@ -78,4 +78,4 @@ class ard_live(Plugin):
                         log.error("Unexpected stream type: '{0}'".format(stream_))
 
 
-__plugin__ = ard_live
+__plugin__ = ARDLive
--- a/src/streamlink/plugins/ard_mediathek.py
+++ b/src/streamlink/plugins/ard_mediathek.py
@@ -31,7 +31,7 @@ _media_schema = validate.Schema({
 log = logging.getLogger(__name__)
 
 
-class ard_mediathek(Plugin):
+class ARDMediathek(Plugin):
     @classmethod
     def can_handle_url(cls, url):
         return _url_re.match(url) is not None
@@ -86,4 +86,4 @@ class ard_mediathek(Plugin):
                     log.error("Failed to extract {0} streams: {1}".format(parser_name, err))
 
 
-__plugin__ = ard_mediathek
+__plugin__ = ARDMediathek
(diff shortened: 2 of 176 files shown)


beardypig:
The unrelated URLs are important for the tests, to ensure that some error has not been made to match unrelated URLs.

beardypig:
The `pytest` method of generating tests is superior to iterating through lists and asserting in turn. Using this method you will get a failure at the first failed URL in each test. Each URL should be tested in its own test so that you can see all the failed cases in one go.

bastimeyer:
> The unrelated URLs are important for the tests, to ensure that some error has not been made to match unrelated URLs.

That's why there is one generic list of negative matches. I didn't just remove everything blindly.

> Each URL should be tested in its own test so that you can see all the failed cases in one go.

True. This wasn't the case for the vast majority of test cases so far though.
I'll take a look at this some time later with the necessary improvements.

beardypig:
> That's why there is one generic list of negative matches. I didn't just remove everything blindly.

Didn't see that, my bad. The change log is long :)

bastimeyer:
Github is having issues right now. The CI builds are a bit delayed. (already fixed the currently failing stuff which I forgot to remove before pushing)

----

I've removed the mixin and instead created a `PluginCanHandleUrl` base class with the `test_can_handle_url_{positive,negative}()` tests in `tests/plugins/__init__.py`.

The tests then get parametrized via `pytest_generate_tests` in `tests/plugins/conftest.py` by checking the base class and test names and passing the appropriate URL lists as parameters.

Since test classes can't inherit from `unittest.TestCase` in order for `pytest_generate_tests` to work, I had to split up those plugin tests which were previously testing URLs and other plugin behavior. This means that URL tests are now all in separate `TestPluginCanHandleURL{PluginName}` classes.

bastimeyer:
Since all URL tests are now parametrized, the number of tests has changed from

> 926 passed, 17 skipped

to

> 2388 passed, 17 skipped

bastimeyer:
Btw, this also reduces the overall code coverage by 1.58% due to 884 previously covered lines being removed from the tests:
https://app.codecov.io/gh/streamlink/streamlink/compare/3533/tree/tests

The codecov PR status checks are for some reason still missing after the hiccups yesterday, so it's possible that if this gets merged, it'll fail on the merge commit because of the 1% test threshold set up in the codecov config.

gravyboat:
Are we good to merge this or is there other feedback?

bastimeyer:
Btw, the latest push was just a rebase onto the new master HEAD with additional test changes for the newly merged plugin.

If you're having trouble reviewing the second commit, just ignore the individual plugin URL tests. All I did with these tests was:

1. translate the positive and negative URL lists into the new declarative format
2. let the test class inherit from the new test base class where the tests are defined
3. split up those plugin test classes which were testing URLs and plugin logic (due to a pytest limitation when inheriting from `unittest.TestCase`)
4. remove the generic URL negative tests, as these are already set via the base class for every plugin

And as requested, the URLs all get dynamically parametrized via pytest's `pytest_generate_tests(metafunc)` in `tests/plugins/conftest.py`.

- https://docs.pytest.org/en/stable/parametrize.html#pytest-generate-tests
- https://docs.pytest.org/en/stable/pythonpath.html#test-modules-conftest-py-files-inside-packages

PR #3542, opened 2021-02-06
plugins.nasa: add plugin for NasaTV

### LINKED ISSUE #3541: Plugin request: NASATV
## Plugin Request
- [x] I am using the latest development version from the master branch.

### Description
A free variety online TV channel, run by NASA.

### Example stream URLs
1. https://www.nasa.gov/multimedia/nasatv/#public

bastimeyer:
Resolves #3541 

I don't know how much sense it makes adding this plugin, because the main "public" stream is also being broadcasted on Youtube AFAICT. I'm also not sure how often the API version string and "vendorID/streamID" strings which are required for getting the JSON data will change. They are hardcoded into the plugin, because it's almost impossible to extract it from the minified JS code, which could also change at any point and break the extraction code. There's also an "education" stream according to the code, but I couldn't find it on their site and it also has a different format which I didn't spent any time on figuring out.

Also, should the plugin be renamed to `NasaTV`, or `NASA` instead of `Nasa`?

```
$ streamlink -l debug 'https://www.nasa.gov/multimedia/nasatv/#public'
[cli][debug] OS:         Linux-5.11.0-rc6-2-git-x86_64-with-glibc2.32
[cli][debug] Python:     3.9.1
[cli][debug] Streamlink: 2.0.0+35.g1d33f6e
[cli][info] Found matching plugin nasa for URL https://www.nasa.gov/multimedia/nasatv/#public
Available streams: 720p_549k (worst), 720p_812k, 720p_2059k (best)
```

```
$ streamlink -l debug 'https://www.nasa.gov/multimedia/nasatv/#media'
[cli][debug] OS:         Linux-5.11.0-rc6-2-git-x86_64-with-glibc2.32
[cli][debug] Python:     3.9.1
[cli][debug] Streamlink: 2.0.0+35.g1d33f6e
[cli][info] Found matching plugin nasa for URL https://www.nasa.gov/multimedia/nasatv/#media
Available streams: 720p_549k (worst), 720p_812k, 720p_2059k (best)
```

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -111,6 +111,7 @@ mitele                  mitele.es            Yes   No    Streams may be geo-rest
 mjunoon                 mjunoon.tv           Yes   Yes
 mrtmk                   play.mrt.com.mk      Yes   Yes   Streams may be geo-restricted to North Macedonia.
 n13tv                   13tv.co.il           Yes   Yes   Streams may be geo-restricted to Israel.
+nasa                    nasa.gov             Yes   No
 nbc                     nbc.com              No    Yes   Streams are geo-restricted to USA. Authentication is not supported.
 nbcnews                 nbcnews.com          Yes   No
 nbcsports               nbcsports.com        No    Yes   Streams maybe be geo-restricted to USA. Authentication is not supported.
--- a/src/streamlink/plugins/nasa.py
+++ b/src/streamlink/plugins/nasa.py
@@ -1,0 +1,94 @@
+import re
+
+from streamlink.plugin import Plugin, PluginError
+from streamlink.plugin.api import validate
+from streamlink.stream import HLSStream
+from streamlink.utils import parse_json
+
+
+class Nasa(Plugin):
+    # hardcoded API version, extracted from player.min.js
+    _api_version = "201510271133"
+    # hardcoded "vendor-id/stream-id", extracted from nasa.js (EmberJS component properties)
+    _vendorandstreamids = {
+        "public": "3117/108566",
+        "media": "3117/108567"
+    }
+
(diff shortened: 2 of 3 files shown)


gravyboat:
Is this worth adding? Isn't it just a bunch of prerecorded content which means it violates rule 6? There's on explanation as to why we should add this in the original issue either, where is the live content? How often is it running live content? I don't know, seems pretty pointless for us.

skulblakka:
At least the main channel shows a mix of live and prerecorded content. You can find a schedule of their live content [here](https://www.nasa.gov/multimedia/nasatv/schedule.html). However just like @bastimeyer said both channels found on the page are also available through YouTube making an extra plugin probably superfluous.

gravyboat:
I don't have an issue so much with supporting the original source since there are other problematic plugins like this (many of the news related ones) but it's the low amount of live content. They're only live a couple of times a week for an hour or two, so the total live time per week is a couple hours. My main concern here is I don't want to start getting other requests for plugins like this that go live for a very small amount of time and it's recorded content the rest of the time. We've already had a few requests like that which we closed specifically for that reason.

Billy2011:
For the Nasa channels there is already the **ustreamtv plugin** with many channels on [https://video.ibm.com/channel/nasa-media-channel#](https://video.ibm.com/channel/nasa-media-channel#)

back-to:
There urls are static or at least won't change that often.
A plugin is not really needed.

https://ntv1.akamaized.net/hls/live/2014075/NASA-NTV1-HLS/master.m3u8
https://ntv2.akamaized.net/hls/live/2013923/NASA-NTV2-HLS/master.m3u8

bastimeyer:
Let's close this. As said, the content is available on youtube anyway

PR #3544, opened 2021-02-07
plugins.tvtoya: minor fixes

mkbloke:
- update url_re for www. and literal dot escaping
- remove unnecessary update_scheme()


gravyboat:
Looks good, thanks @mkbloke.

Issue #3538, opened 2021-02-05
afreecatv.com 'Could not open stream' message


tdi2020:
## Bug Report

- [x] This is a bug report and I have read the contribution guidelines.
- [x] I am using the latest development version from the master branch.

### Description

Hi Guys,
It's my first time writing a post.I read the notice well, but I might be a little clumsy. Please understand.

Until now, I have created a bat file and downloaded afreecatv without any problems. 
However, from February 5, 2021, the error message as below appeared and it is no longer working. 
May I know what kind of problem this is and how to solve it?

I'm most curious about why it's not working well without any problems.
I swear I've never modified or touched anything since it worked.

The account I tested for this download is afreecatv.com/ayoona
But she doesn't have a constant broadcasting time.

### Reproduction steps / Explicit stream URLs to test

1. afreecatv.com/ayoona
2. ...
3. ...

### Log output

```
Checking for ayoona every 1.0 seconds. Record with best quality.
ayoona online. Stream recording in session.
[cli][info] Found matching plugin afreeca for URL play.afreecatv.com/ayoona
[cli][info] Available streams: sd (worst), hd, original (best)
[cli][info] Opening stream: original (hls)
[...]

PR #3549, opened 2021-02-08
plugins.afreeca: use 'gs_cdn_pc_web' and 'common'

back-to:
gs_cdn --> gs_cdn_pc_web
flash  --> common

the api returns `gs_cdn` but it doesn't work with every region.
use `gs_cdn_pc_web` instead.

closes https://github.com/streamlink/streamlink/issues/3538


gravyboat:
Looks good @back-to, thanks.

Issue #3569, opened 2021-02-16
TLC.py not work


worldmediaman:
## Error Report

- [x] This is a bug report and I have read the contribution guidelines.
- [x] I am using the latest development version from the main branch.

### Description

I would like to play the TLC Turkey live stream via Streamlink in Germany

### Expected / actual behavior

<! - What do you expect and what will actually happen? ->

When I open the page via Firefox I can see the live stream
Unfortunately not via Streamlink.

### Replication Steps / Explicit Stream URLs to Test

<! - How can we reproduce that? Please note the exact steps in the specified list format. If you need further steps please add them. ->

Open https: //www.tlctv.com.tr/canli-izle

Please look through a browser, I can see the live stream.

Then via Streamlink

### Log output

streamlink -l debug https://www.tlctv.com.tr/canli-izle
[cli] [debug] OS: Windows 10
[cli] [debug] Python: 3.8.6
[cli] [debug] Streamlink: 2.0.0
[cli] [debug] Requests (2.25.1), Socks (1.7.1), Websocket (0.57.0)
[cli] [info] Found matching plugin tlctr for URL https://www.tlctv.com.tr/canli-izle
[plugins.tlctr] [debug] URL = https: //dygvideo.dygdigital.com/live/hls/tlctv? token = VGcxK0ptK2dQbURrNVpxS3JtVDNtSnNqYkhvVjdrL25zTlJ4dUUzcGNiND0 = & toslND0e =
[utils.l10n] [debug] Language code: de_DE
Available streams: 180p (worst), 270p, 576p, 720p (best)

streamlink -l debug https://www.tlctv.com.tr/canli-izle best
[cli] [debug] OS: Windows 10
[cli] [debug] Python: 3.8.6
[cli] [debug] Streamlink: 2.0.0
[cli] [debug] Requests (2.25.1), Socks (1.7.1), Websocket (0.57.0)
[cli] [info] Found matching plugin tlctr for URL https://www.tlctv.com.tr/canli-izle
[plugins.tlctr] [debug] URL = https: //dygvideo.dygdigital.com/live/hls/tlctv? token = Vk9QZHhaWjBRN0NmVDJxZG96Vm1YUkVrVG95ZjFrSm1MOE1QSHF0ZHhQlcz0e = & opens token_type = & opens token_type =
[utils.l10n] [debug] Language code: de_DE
[cli] [info] Available streams: 180p (worst), 270p, 576p, 720p (best)
[...]

PR #3570, opened 2021-02-17
plugins.tlctr: update/fix

mkbloke:
- update to use API URL
- add API schema

Fixes: #3569

### DIFF
--- a/src/streamlink/plugins/tlctr.py
+++ b/src/streamlink/plugins/tlctr.py
@@ -2,16 +2,22 @@ import logging
 import re
 
 from streamlink.plugin import Plugin
+from streamlink.plugin.api import validate
 from streamlink.stream import HLSStream
 
 log = logging.getLogger(__name__)
 
 
 class TLCtr(Plugin):
-
     _url_re = re.compile(r'https?://(?:www\.)?tlctv\.com\.tr/canli-izle')
-    _hls_re = re.compile(
-        r'''["'](?P<url>https?://[^/]+/live/hls/[^"']+)["']''')
+    _token_params_re = re.compile(r'var liveUrl = ".*?tlctv(.*?)";')
+    _api_url = 'https://dygvideo.dygdigital.com/live/hls/tlctvdai{0}&json=true'
+
+    _api_schema = validate.Schema({
+        'xtra': {
+            'url': validate.url(),
+        },
+    }, validate.get('xtra'), validate.get('url'))
 
     @classmethod
     def can_handle_url(cls, url):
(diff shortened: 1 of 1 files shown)


worldmediaman:
/usr/sbin/streamlinksrv manualstart debug
pidfile /var/run/streamlink.pid does not exist. Daemon not running?
[streamlinksrv][info] Wed Feb 17 16:42:36 2021 Server (1.5.0) started
[streamlinksrv][debug] Host:            dm920
[streamlinksrv][debug] Port:            8088
[streamlinksrv][debug] OS:              Linux-3.14-1.17-dm920-armv7l-with-glibc2.4
[streamlinksrv][debug] Python:          2.7.18
[streamlinksrv][info] Streamlink:      1.27.1.0 / 2020-12-22
[streamlinksrv][debug] youtube-dl:     2021.02.04.1
[streamlinksrv][debug] Requests(2.23.0), Socks(1.7.1), Websocket(0.57.0)
Failed to load plugin tlctr:
  **

- File "/usr/lib/python2.7/site-packages/streamlink/plugins/tlctr.py", line 36
-     log.debug(f'API URL={api_url}')

**
                                 ^
SyntaxError: invalid syntax

mkbloke:
> [streamlinksrv][debug] Python: 2.7.18

The minimum Python version required since the release of Streamlink 2.0.0 is 3.6.  You'll have to upgrade it, I'm afraid.

mkbloke:
For info: it appears not to be geo-restricted.  I should have mentioned that in the initial post.

worldmediaman:
vielen Dank für die Info

tests/plugins/test_tlctr.py
```diff
@@ -8,3 +8,7 @@ class TestPluginCanHandleUrlTLCtr(PluginCanHandleUrl):
     should_match = [
         'https://www.tlctv.com.tr/canli-izle',
     ]
+
+    should_not_match = [
+        'https://example.com/index.html',
+    ]
```
bastimeyer: Generic negative URL tests are already added by the `PluginCanHandleUrl` base class.

src/streamlink/plugins/tlctr.py
```diff
@@ -2,16 +2,22 @@
...
 log = logging.getLogger(__name__)
 
 
 class TLCtr(Plugin):
-
     _url_re = re.compile(r'https?://(?:www\.)?tlctv\.com\.tr/canli-izle')
-    _hls_re = re.compile(
-        r'''["'](?P<url>https?://[^/]+/live/hls/[^"']+)["']''')
+    _token_params_re = re.compile(r'var liveUrl = ".*?tlctv(.*?)";')
+    _api_url = 'https://dygvideo.dygdigital.com/live/hls/tlctvdai{0}&json=true'
```
bastimeyer, line 14: Since you're trying to match the `_api_url` here except for the query parameters, you should set the `_api_url` value to the API base URL (`https://dygvideo.dygdigital.com/live/hls/tlctvdai`) without any query parameters and try to match it in the regex here instead (also make the whitespace optional). Then parse the regex match as a URL, add the `json=true` parameter and serialize it again for making the API call.

bastimeyer (changes requested):

src/streamlink/plugins/tlctr.py
mkbloke: I'm not sure if I'm misunderstanding things, but the API base URL is not quite the same as the base URL extracted from the page:
```
https://dygvideo.dygdigital.com/live/hls/tlctv # base URL from page
https://dygvideo.dygdigital.com/live/hls/tlctvdai # base for API URL
```
There could be a better way to do this though.  The only changes necessary are to substitute `tlctv` for `tlctvdai` and append the `json=true` parameter.

Perhaps it would better to extract the entire URL from the page, then: `extracted_url.replace('tlctv', 'tlctvdai')` and finally, append the extra parameter get the required API URL?

mkbloke: Just to add: Strictly speaking, appending `json=true` does not appear to be necessary, I see.  Without it, the request directly returns the M3U8 file...

mkbloke:
Ah, ******, I think this has been a waste of time.  Whereas before, content would play via the web player but not the plugin, now the web player also seems to show the region lock video and so does this updated plugin...

AFAIK, the existing plugin works fine, so I'm just going to close this for now...

Issue #3334, opened 2020-11-11
plugin for http://yoda.az/


mike8032:
## Plugin Request

- [x ] This is a plugin request and I have read the contribution guidelines and plugin request requirements.

### Description

this is an official page for channels from azerbaijcan. there is no plugin yet for this country on streamlink.

### Example stream URLs

http://yoda.az/

### Additional comments, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

worldmediaman:
I also wanted to express the wish to be able to play these channels via Streamlink. Have seen that a request has already been made on this.

Explained again in a little more detail.

Azerbaycan channels can be watched via this page http://yoda.az. These are original livestreams.
the live streams can be played via the windows 10 web browser without geoblocking.

Example:
http://yoda.az/idmanlive/index.php

Unfortunately, Streamlink does not support it:

streamlink -l debug http://yoda.az/idmanlive/index.php
[cli] [debug] OS: Windows 10
[cli] [debug] Python: 3.8.6
[cli] [debug] Streamlink: 2.0.0
[cli] [debug] Requests (2.25.1), Socks (1.7.1), Websocket (0.57.0)
error: No plugin can handle URL: http://yoda.az/idmanlive/index.php

would be really great if this would be playable via streamlink

mkbloke:
This looks like it should be simple enough.  As far as I can see, it meets all the plugin request guidelines except for the dreaded number 10...

Devs, please assign an appropriate label to this; if approved, I'll make a start on it shortly.

gravyboat:
@mkbloke It doesn't meet requirements 3 or 6 either. There's a full blown schedule of TV shows on the left hand side of the page. The provided link for `idmanlive` is streaming a sports match from 2019. None of this is live.

mkbloke:
@gravyboat, thanks for checking it out.  I was a bit confused about the programme listing on the left hand side - it appears you can go back, but not forward beyond live time.  That suggested to me that the TS data is kept on the CDN to allow for that, but it's still live.  Clicking CANLI, I think, takes you to live and from there the +5:00 button doesn't advance beyond current time in AZ.  It appears to me that it's live, but the web player allows for playing content broadcast at least a few hours before.

I might well be wrong on the above, of course, but that's how it looked to me.  Cheers.

gravyboat:
[...]

PR #3609, opened 2021-03-06
plugins.auftanken: new plugin

### LINKED ISSUE #3602: https://www.auftanken.tv/
## Plugin Request

### Example stream URLs

1. https://tvhosa-pez0302.sctv.ch/http_adaptive_streaming/GEMBFJAABFBBBEAG.m3u8
2. ...
3. ...

### Additional comments, etc.
site uses tokens, so example url will prolly change very soon.
swiss channel, works with non .ch ip too

Billy2011:
closes #3602

### DIFF
--- a/src/streamlink/plugins/auftanken.py
+++ b/src/streamlink/plugins/auftanken.py
@@ -1,0 +1,31 @@
+import logging
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.stream import HLSStream
+
+log = logging.getLogger(__name__)
+
+
+class AuftankenTV(Plugin):
+    _url_re = re.compile(r"https?://(?:www\.)?auftanken\.tv/livestream/?")
+    _hls_url_re = re.compile(r"(https://.+?/http_adaptive_streaming/\w+\.m3u8)")
+
+    PLAYER_URL = "https://webplayer.sbctv.ch/auftanken/"
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return cls._url_re.match(url) is not None
+
+    def get_title(self):
+        return "auftanken.TV Livestream"
+
+    def _get_streams(self):
+        res = self.session.http.get(self.PLAYER_URL)
+        m = self._hls_url_re.search(res.text)
+        if m:
+            for s in HLSStream.parse_variant_playlist(self.session, m.group(1)).items():
+                yield s
+
+
+__plugin__ = AuftankenTV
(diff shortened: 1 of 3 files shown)


src/streamlink/plugins/auftanken.py
```diff
@@ -0,0 +1,31 @@
...
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.stream import HLSStream
+
+log = logging.getLogger(__name__)
+
+
+class AuftankenTV(Plugin):
+    _url_re = re.compile(r"https?://(?:www\.)?auftanken.tv/livestream/?")
```
bastimeyer, line 11:
```suggestion
    _url_re = re.compile(r"https?://(?:www\.)?auftanken\.tv/livestream/?")
```

bastimeyer (commented):
The plugin itself looks fine, and it's rather simple, but, and this has just been discussed earlier today / yesterday in another thread, this site is the pure definition of niche. Just take a look at their target audience, the program being broadcasted and their social media presence, eg. on Twitter with a whopping follower number of one. Which value does a plugin for this site bring to Streamlink? The plugin requests also doesn't give any additional information about it.

Billy2011:
> Which value does a plugin for this site bring to Streamlink?

I don’t think you can answer that in a general way. This channel offers a varied program for anyone who is not only interested in gaming, news, etc. Should streamlink only be there for the broad masses?

beardypig:
@Billy2011 well said. But each new plugin slows down the loading process (which really only affects slower machines) and adds maintenance burden. 

<offtopic>
Loading times could be improved by changing a few things though. All of the plugins compile the regexs they might only use once, and if the plugin is used - but they are still compiled on import. So of the URL regexs could be replaced by simple substring matches, etc. We could introduce a cache of previously used URLs and which plugins they matched. etc. etc.
</offtopic>

Billy2011:
> But each new plugin slows down the loading process (which really only affects slower machines) and adds maintenance burden.

Then you should limit the number of plugins, because this is the only way to avoid subjective reviews. For me, social "media presence" is also not an objective criterion, unless only the supposed mass taste counts, which is propagated somewhere.

bastimeyer:
My point was that this kind of niche-content that's being shown there in addition to the state of their social media presence is simply an indicator for their viewer numbers, because the site itself doesn't show it. I don't care about the actual content.

Adding a plugin to Streamlink which will probably only be used by at most one person from time to time does not bring any value to the overall project, because as @beardypig already said, it would unnecessarily slow down the entire application due to its current plugin resolving logic. This might not be much on machines that are not slow, but it also adds maintenance burden for something that is probably not being used anyway.

And on top of that, the future of smaller channels / sites like that is always unknown and if they go offline / bankrupt, then we have to remove the plugin again. This has already been the case dozens of times for similar sites and the offline-report came years later after they went down, because nobody noticed it, which is a good indicator of how often niche-plugins are used in this project.

As @gravyboat said in https://github.com/streamlink/streamlink/issues/3334#issuecomment-792076010, we need to find a solution for this, so that niche-plugins can exist but don't disrupt the overall project.

PR #3611, opened 2021-03-08
docs: add appimages section to install docs

### LINKED ISSUE #669: Feature request: Linux binaries
I did this previously in a different application, see:
https://github.com/projectatomic/atomicapp/commit/925fba1b60de59c99b2a6be2ea75e7f8e74298eb

Shouldn't be too hard to do.

Similar to a project such as https://docs.docker.com/compose/install/ it'd be nice to be able to have a single binary to install instead of having to install Python and have the correct version of it running, etc.

Since we already have `pip`, `.exe`'s, it'd be nice to have the appropriate binaries for both Linux + Mac OS X as well.

### LINKED ISSUE #3522: Flatpak package
Let's create a Flatpak package for Streamlink.
https://docs.flatpak.org/en/latest/index.html

There have been lots of questions and issues regarding installing Streamlink on Ubuntu and Ubuntu-based Linux distros over the past couple of days/weeks, as the native packages are always outdated and the available PPA doesn't help on older distro releases. Also issues with py2/py3, which is confusing to some users, and `pip --user` is as well. A flatpak package could help here, as everything would be integrated into the package's environment/container.

----

I've already been working on a flatpak package build config and an initial version can be found here:
https://github.com/bastimeyer/io.github.streamlink.streamlink

#### Todo

- fix package issues
- bundle ffmpeg
- add CI build config
- add sign+deploy config

#### Issues

- Streamlink loses its version information when the package gets installed, but is working correctly when run from the build directory.
- Not sure how to work around the `/usr/bin/` filesystem portal limitation, so that system-wide players can be run by Streamlink from within the flatpak container. Can the system's `/usr/bin/` path be mapped to a different one and `PATH` extended?

Since I don't have much (or basically any) experience with Flatpak, I'd appreciate any kind of help.

### LINKED ISSUE #3572: Streamlink AppImage
Opening a new thread for this instead of posting in #669...

My initial idea to solve the packaging problems on Ubuntu(-based) systems (and other Linux distros) was creating a Flatpak image for Streamlink (#3522), but due to it running in an isolated container, this opens tons of new issues regarding launching a player or other runtime user dependencies.

This means that the better approach for having a "universal Linux package" is creating an AppImage for Streamlink, which is a simple executable with the packed application inside it, which is run on an automatically mounted squash-fs on the regular host system. All that needs to be done by the user is download the AppImage file, make it executable (`chmod +x`) and run it. There are also tools available for automating this and for better integrating the appimages into the system, like AppImageLauncher for example.
- https://appimage.org/
- https://github.com/TheAssassin/AppImageLauncher

I've already created an AppImage for Streamlink Twitch GUI a couple of months ago which has been published in the latest release in December, and I haven't seen any issues with it so far.

The only downside of AppImages is that it's a manual process and it's not handled by the system's package manager. AFAIK, AppImages can also have update information embedded into them, but I don't like this idea.

----

So over the past couple of days, I've been working on an AppImage build config for Streamlink.

At first I've tried using [`python-appimage`](https://github.com/niess/python-appimage), but ran into some packaging issues with it (while also having working builds). What I didn't like about it was that it is pulling the appimagetool and docker images for building the AppImage carelessly from continuous builds without any signature or integrity checks. This isn't only potentially dangerous, but also generates appimages which are not reproducible. That's why I wrote my own build scripts that can build reproducible/deterministic appimages.

Similar to `python-appimage`, the python environment is copied from the [pypa/manylinux](https://github.com/pypa/manylinux) docker images, binaries get patched for their new package location, the python application gets installed in this environment and the appimage's AppDir gets prepared accordingly, so that `appimagetool` (which is also patched with an updated `mksquashfs`) can build the AppImage.

This is all very much unfinished and a couple of issues are still left, but the built appimages should be working fine.

----

### Issues left

- [x] A script for generating the pypi dependencies list needs to be added. Since the appimages are supposed to be deterministic, the package dependencies need to be listed [with their versions and sha256 hashes](https://pip.pypa.io/en/stable/reference/pip_install/#hash-checking-mode) in the build JSON-config, as Streamlink doesn't use a lock file. This is currently being done manually via `cd $(mktemp -d) && pip download streamlink=={{version}} && pip hash ./*` in the containers and copying/translating the data into the right JSON format.
- [x] Bytecode cache generation is currently disabled via [`--no-compile`](https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-no-compile), because pip does for some reason ignore the [`PYTHONHASHSEED=0`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONHASHSEED) env var when installing dependencies and compiling the bytecode, and it always randomizes it.
- [x] AppImage appstream data needs to be fixed (still using an older revision from the python-appimage experiments)
- [x] License data is currently missing.
- [ ] AppImages are currently unsigned.
- [x] Currently only `x86_64` and `i686` are available. `aarch64` will be added later.
- [x] A proper deploy config needs to be written.

----

The repo can currently be found here with the built appimages stored as build artifacts on Github actions:
- https://github.com/bastimeyer/streamlink-appimage
- https://github.com/bastimeyer/streamlink-appimage/actions/runs/605939714

bastimeyer:
Resolves #3572 
Resolves #669 
Closes #3522

----

I've just moved my streamlink-appimage repo to streamlink/streamlink-appimage and pushed the `2.0.0-1` release:
- https://github.com/streamlink/streamlink-appimage
- https://github.com/streamlink/streamlink-appimage/releases

See #3572 for all the additional infos about the appimage.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -67,6 +67,8 @@ Linux and BSD
 ==================================== ===========================================
 Distribution                         Installing
 ==================================== ===========================================
+AppImage                             See the `AppImages`_ section below
+
 `Arch Linux`_                        .. code-block:: bash
 
                                         sudo pacman -S streamlink
@@ -131,6 +133,9 @@ Distribution                         Installing
 
                                         sudo xbps-install streamlink
 ==================================== ===========================================
+
+Please see the `PyPI package and source code`_ or `AppImages`_ sections down below
+if a package is not available for your distro or platform, or if it's out of date.
 
 .. _Arch Linux: https://www.archlinux.org/packages/community/any/streamlink/
 .. _Arch Linux (aur, git): https://aur.archlinux.org/packages/streamlink-git/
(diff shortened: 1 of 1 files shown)


bastimeyer:
Just wondering, maybe we should add the "Python pip" entry to all tables, not just macOS:
https://deploy-preview-3611--streamlink.netlify.app/install.html

gravyboat:
@bastimeyer Yeah that seems worth it just to show that it is possible.

PR #3627, opened 2021-03-17
plugins.rtpplay: add schema and fix HLS URL

mvaladas:
RTP Play Plugin was failing for a couple of weeks now. The website changed structure and the URL parsing needed updating.

### DIFF
--- a/src/streamlink/plugins/rtpplay.py
+++ b/src/streamlink/plugins/rtpplay.py
@@ -9,7 +9,11 @@ from streamlink.utils import parse_json
 
 class RTPPlay(Plugin):
     _url_re = re.compile(r"https?://www\.rtp\.pt/play/")
-    _m3u8_re = re.compile(r"""hls:\s*(?:(["'])(?P<string>.+)\1|decodeURIComponent\((?P<obfuscated>\[.*?])\.join\()""")
+    _m3u8_re = re.compile(r"""
+        hls:\s*(?:(["'])(?P<string>[^"']+)\1
+        |
+        decodeURIComponent\((?P<obfuscated>\[.*?])\.join\()
+    """, re.VERBOSE)
 
     _schema_hls = validate.Schema(
         validate.transform(_m3u8_re.search),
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/rtpplay.py
```diff
@@ -25,7 +26,8 @@ def _get_streams(self):
             log.error("Could not find _m3u8_re")
             return
 
-        hls_url = m.group("url")
+        hls_url = eval("\"\".join((" + m.group(1) + "))")
```
back-to, line 29: don't use `eval`

back-to (changes requested):

src/streamlink/plugins/rtpplay.py
bastimeyer: Get the array and its content without the `.join()` call and parse it as JSON. Then do the array to string conversion and decoding afterwards in the plugin logic.

bastimeyer: This does obviously only work as long as the site doesn't add more "code obfuscation" which would require a JS runtime, like for example dynamic string concatenation or function calls, etc. Some sites do this so their sources can't be statically parsed, but this seems simple enough.

bastimeyer:
```
$ streamlink -l debug https://www.rtp.pt/play/direto/rtp1
[cli][debug] OS:         Linux-5.11.6-2-git-x86_64-with-glibc2.33
[cli][debug] Python:     3.9.2
[cli][debug] Streamlink: 2.0.0+57.gd8e980e
[cli][info] Found matching plugin rtpplay for URL https://www.rtp.pt/play/direto/rtp1
Available streams: 640k (worst), 1290k, 2540k (best)
```

```
$ streamlink -l debug https://www.rtp.pt/play/direto/antena1
[cli][debug] OS:         Linux-5.11.6-2-git-x86_64-with-glibc2.33
[cli][debug] Python:     3.9.2
[cli][debug] Streamlink: 2.0.0+57.gd8e980e
[cli][info] Found matching plugin rtpplay for URL https://www.rtp.pt/play/direto/antena1
Available streams: 170k (worst, best)
```

Issue #3633, opened 2021-03-20
Potential issue with websocket-client >= 0.58


mkbloke:
See: https://github.com/streamlink/streamlink/issues/3631#issuecomment-803231572

Could also be relevant to:
- twitcasting
- ustreamtv

I've not looked further into this yet and have so far never used WS in any plugins.  Just putting this here for info and for further investigation, either by fixing the affected plugins if possible or limiting websocket-client to < 0.58 perhaps (for a while anyway)...

mkbloke:
Hmm, seems like it could be relevant: https://github.com/websocket-client/websocket-client/issues/669

Issue #3631, opened 2021-03-19
plugins.nicolive: handle_api_message() takes 2 positional arguments but 3 were given


railannad:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.
- [ ] I am using the latest development version from the master branch.

### Description

I tried to record timeshift on Niconico , but it can not work on my machine

### Reproduction steps / Explicit stream URLs to test

My command was
`streamlink --niconico-email EMAIL --niconico-password PASSWORD --loglevel debug live.nicovideo.jp/watch/lv330886875`

### Log output

```
[cli][info] streamlink is running as root! Be careful!
[cli][debug] OS:         Linux-4.15.0-48-generic-x86_64-with-glibc2.27
[cli][debug] Python:     3.8.0
[cli][debug] Streamlink: 2.0.0
[cli][debug] Requests(2.25.1), Socks(1.7.1), Websocket(0.44.0)
[cli][info] Found matching plugin nicolive for URL live.nicovideo.jp/watch/lv330886875
[cli][debug] Plugin specific arguments:
[cli][debug]  --niconico-email=******** (email)
[cli][debug]  --niconico-password=******** (password)
[plugins.nicolive][debug] Getting video page: http://live.nicovideo.jp/watch/lv330886875
[plugins.nicolive][debug] Video page response code: 200
[plugins.nicolive][debug] Got wss_api_url: wss://a.live2.nicovideo.jp/unama/wsapi/v2/watch/36299718263388/timeshift?audience_token=36299718263388_anonymous-user-8f02406b-6ccf-4363-9b08-5e10ca81b46e_1616243986_824ff685839f5ddb9924cc56f4df569b0f190448&frontend_id=9
[plugins.nicolive][debug] Got frontend_id: 9
[plugins.nicolive][debug] Connecting: wss://a.live2.nicovideo.jp/unama/wsapi/v2/watch/36299718263388/timeshift?audience_token=36299718263388_anonymous-user-8f02406b-6ccf-4363-9b08-5e10ca81b46e_1616243986_824ff685839f5ddb9924cc56f4df569b0f190448&frontend_id=9
[plugins.nicolive][debug] Waiting for permit...
error from callback <bound method NicoLive.api_on_open of <streamlink.plugins.nicolive.NicoLive object at 0x7f86674cf940>>: api_on_open() takes 1 positional argument but 2 were given
[plugins.nicolive][debug] Waiting for permit...
[...]

PR #3634, opened 2021-03-20
plugins.nicolive: fixed websocket-client and use websocket-client>=0.58.0

back-to:
```
setup.py: require websocket-client>=0.58.0

Make sure all User use the same version where `callback(self, *args)` was changed.
```

```
plugins.nicolive: fixed websocket-client 
```

closes https://github.com/streamlink/streamlink/issues/3631
closes https://github.com/streamlink/streamlink/issues/3633

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -290,7 +290,7 @@ Name                                 Notes
 `iso3166`_                           Used for localization settings, provides country information
 `isodate`_                           Used for MPEG-DASH streams
 `PySocks`_                           Used for SOCKS Proxies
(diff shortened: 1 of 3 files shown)


bastimeyer:
So according to https://github.com/websocket-client/websocket-client/issues/669#issuecomment-797738080, the changes made in websocket-client 0.58.0 will stay now, even with the proposed 1.0.0 release, right?
See https://github.com/websocket-client/websocket-client/commit/3112b7d75b1e5d65cb8fdfca7801606649044ed1#commitcomment-47124551

Issue #3553, opened 2021-02-10
Zattoo: error: No playable streams found on this URL:


AdamNo:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.
- [x] I am using the latest development version from the master branch.

### Description

since yesterday evening 09.02.2021 ~19:00 i am getting for all streams i.e.: "error: No playable streams found on this URL: https://zattoo.com/watch/orf-2"

### Reproduction steps / Explicit stream URLs to test

[tvheadend@tvheadend-4-2 ~/tmp]$ streamlink --zattoo-email=xxxXXXxxx --zattoo-password=xxxXXXxxx https://zattoo.com/watch/rt-doc best -o out.ts 
[cli][info] Found matching plugin zattoo for URL https://zattoo.com/watch/rt-doc
error: No playable streams found on this URL: https://zattoo.com/watch/rt-doc

### Log output

[tvheadend@tvheadend-4-2 ~/tmp]$ streamlink --zattoo-email=xxxXXXxxx --zattoo-password=xxxXXXxxx https://zattoo.com/watch/rt-doc best -o out.ts --loglevel debug
[cli][debug] OS:         FreeBSD-11.3-RELEASE-p14-amd64-64bit-ELF
[cli][debug] Python:     3.6.9
[cli][debug] Streamlink: 2.0.0+40.gda0bfe9
[cli][debug] Requests(2.22.0), Socks(1.7.0), Websocket(0.56.0)
[cli][info] Found matching plugin zattoo for URL https://zattoo.com/watch/rt-doc
[cli][debug] Plugin specific arguments:
[cli][debug]  --zattoo-email=xxxXXXxxx (email)
[cli][debug]  --zattoo-password=******** (password)
[cli][debug]  --zattoo-stream-types=['hls'] (stream_types)
[plugins.zattoo][debug] _hello ...
[plugins.zattoo][debug] _login ... Attempting login as 5tu_sqoh9b6fp3kq@byom.de
[plugins.zattoo][debug] New Session Data
[plugins.zattoo][debug] Saved cookies: beaker.session.id, pzuid, zattoo.session
[plugins.zattoo][debug] _watch ...
[plugins.zattoo][debug] get channel ID for rt-doc
[...]

PR #3635, opened 2021-03-20
plugins.zattoo: use 'dash' as default stream

back-to:
'hls' was removed and can't be used anymore,
'hls5', 'hls7' or 'dash' is only available

closes https://github.com/streamlink/streamlink/issues/3553


gravyboat:
Looks good, thanks @back-to.

PR #3639, opened 2021-03-21
cli: debug-log arguments set by the user

bastimeyer:
We've had this discussion a while ago and I think we should log all arguments set by the user while debug logging, not just plugin specific arguments.

- The first commit therefore replaces the "Plugin specific arguments" logging with the new `streamlink_cli.main.log_current_arguments` method which logs all arguments set by the user, on the command line, from the main config file, from the plugin config file and from any custom config files. Sensitive plugin argument values get replaced (just like previously) and it prints double-dash argument names instead of the argument's `dest` name, if possible.
- The second commit additionally cleans up `streamlink_cli.main.log_current_versions` and adds tests, for good measure.

----

Example with a mixture of CLI arguments, main-config arguments and plugin-config arguments from my system:
```
$ streamlink -l debug --title foo twitch.tv/esl_csgo best
[cli][debug] OS:         Linux-5.11.7-1-git-x86_64-with-glibc2.33
[cli][debug] Python:     3.9.2
[cli][debug] Streamlink: 2.0.0+64.gce6a60c
[cli][debug] Arguments:
[cli][debug]  url=twitch.tv/esl_csgo
[cli][debug]  stream=['best']
[cli][debug]  --player=mpv
[cli][debug]  --player-args=--cache=yes --demuxer-max-bytes=750k --demuxer-max-back-bytes=2G
[cli][debug]  --title=foo
[cli][debug]  --hls-live-edge=2
[cli][debug]  --hls-segment-threads=2
[cli][debug]  --twitch-disable-hosting=True
[cli][debug]  --twitch-disable-ads=True
[cli][debug]  --twitch-disable-reruns=True
[cli][debug]  --twitch-low-latency=True
[cli][info] Found matching plugin twitch for URL twitch.tv/esl_csgo
^CInterrupted! Exiting...
```

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -551,21 +551,7 @@ def handle_url():
     try:
         plugin = streamlink.resolve_url(args.url)
         setup_plugin_options(streamlink, plugin)
-        log.info("Found matching plugin {0} for URL {1}".format(
-                 plugin.module, args.url))
-
-        plugin_args = []
-        for parg in plugin.arguments:
-            value = plugin.get_option(parg.dest)
-            if value:
-                plugin_args.append((parg, value))
-
-        if plugin_args:
-            log.debug("Plugin specific arguments:")
-            for parg, value in plugin_args:
-                log.debug(" {0}={1} ({2})".format(parg.argument_name(plugin.module),
-                                                  value if not parg.sensitive else ("*" * 8),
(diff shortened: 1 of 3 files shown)


back-to (approved):

gravyboat:
Agreed and it looks good. Thanks @bastimeyer.

PR #3600, opened 2021-03-03
plugins: fix and update removed plugins list

bastimeyer:
- add shell script for updating the removed plugins list from the git
  history of deleted and renamed plugin files, compared with the current
  list of plugins
- test that the removed plugins list does not include any active plugins
- update removing plugins section in developing docs

----

Resolves #3291 
See #2003

The new removed plugin entries come from here:

- `arconia` 3973bb9 (renamed)
- `atv` cad43f9 (renamed)
- `bnt` 8a66ee7 (renamed)
- `dplay` 2eb54a3 (deleted)
- `googledocs` 7eee543 (renamed)
- `kralmuzik` cad43f9 (deleted)
- `showtv` cad43f9 (renamed)
- `toya` 40e2bb0 (deleted)

PR #3644, opened 2021-03-23
build: include removed plugins file in sdist

### LINKED ISSUE #3643: ERROR: test_plugin_not_in_removed_list (tests.test_plugins_meta.TestPluginMeta) when building
## Bug Report

### Description

When building 2.1.0 on a clean chroot (Arch), I get the following build error:

ERROR: test_plugin_not_in_removed_list (tests.test_plugins_meta.TestPluginMeta)

The build to complete without errors

Try to build streamlink 2.1.0 on a chroot

### Log output

```
----------------------------------------------------------------------
  File "/build/streamlink/src/streamlink-2.1.0/tests/test_plugins_meta.py", line 64, in test_plugin_not_in_removed_list
    with open(os.path.abspath(os.path.join(streamlinkplugins.__path__[0], ".removed"))) as file:
FileNotFoundError: [Errno 2] No such file or directory: '/build/streamlink/src/streamlink-2.1.0/src/streamlink/plugins/.removed'
```

### Additional comments, etc.

I'm able to get past this issue by touching the file in question before running the actual tests:

https://github.com/archlinux/svntogit-community/blob/packages/streamlink/trunk/PKGBUILD#L31

bastimeyer:
Fixes #3643 

For some reason, the `src/streamlink/plugins/.removed` file is not part of the sdist tarball in `2.1.0`.
This PR forces the inclusion of that file via `MANIFEST.in`.

I don't know why, but when I build `2.1.0` locally, the file is part of the tarball, but the one built by the CI runner does not include it. Can someone confirm this?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,6 +4,7 @@ include README.md
 include LICENSE*
 include *requirements.txt
 include src/streamlink/_version.py
+include src/streamlink/plugins/.removed
 include versioneer.py
 
 recursive-include docs *


bastimeyer:
```bash
$ curl -sSL 'https://github.com/streamlink/streamlink/releases/download/2.1.0/streamlink-2.1.0.tar.gz' \
  | tar -tzf - \
  | grep src/streamlink/plugins/.removed

$ git checkout 2.1.0 && python setup.py sdist
$ tar -tzf dist/streamlink-2.1.0.tar.gz \
  | grep src/streamlink/plugins/.removed
streamlink-2.1.0/src/streamlink/plugins/.removed
```

bastimeyer:
According to

- https://docs.python.org/3/distutils/sourcedist.html
- https://packaging.python.org/guides/using-manifest-in/

the file shouldn't be in the tarball unless explicitly specified, which totally makes sense, but I don't understand why it's included when I build it locally on my system using the same python version and using a clean venv. Weird.

Just for the record, #3643 has only become a problem due to #3600 and the added test, which made the file a dependency for running tests.

beardypig (approved):
LGTM

PR #3655, opened 2021-03-28
plugins.radiko: use API instead of hardcoded URLs

back-to:
closes https://github.com/streamlink/streamlink/pull/3512

### DIFF
--- a/src/streamlink/plugins/radiko.py
+++ b/src/streamlink/plugins/radiko.py
@@ -1,13 +1,17 @@
 import base64
 import datetime
 import hashlib
+import logging
 import random
 import re
 import xml.etree.ElementTree as ET
-from urllib.parse import urlencode
 
 from streamlink.plugin import Plugin
+from streamlink.plugin.api import validate
 from streamlink.stream import HLSStream
+from streamlink.utils import parse_xml
+
+log = logging.getLogger(__name__)
 
 
 class Radiko(Plugin):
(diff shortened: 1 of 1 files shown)


kikuyan:
May I ask:
- Are areafree urls excluded? Areafree is a paid service.
- For timefree urls, could you please give `radiko.jp/v2/api/ts` priority, so that it be the same as to streamlink version 1.7.0?
Or, would it be possible to look for timefree url from the bottom to up in `radiko.jp/v3/station/stream/pc_html5/{station_id}.xml`, so that `radiko.jp/v2/api/ts` get picked up as a result?

As you probably know, there are cases where `radiko.jp/v3/station/stream/pc_html5/{station_id}.xml` contains several urls for each services according to radio stations. For timefree, `tf-rpaa.smartstream.ne.jp` and `radiko.jp/v2/api/ts` for some radio stations, `radiko.jp/v2/api/ts` only for others.
I'd been trying to find enough reasons to revert timefree url in #3512, just the moment I'm going to ask the dev team to do so, I noticed your PR.

kikuyan:
By the way, would it be possible for you to include modifications below in this PR? I've noticed these while tackling timefree in #3512.

- In `_get_streams()`, just before `yield`, http headers should be set by `update()`, not by `=`.
Currently, values already set are wiped out.
- In `_authorize()`, http headers should be set per request in `get()`, not by `update()`.
No accumulation should be needed.
- In `_get_xml()`, `http` url should be changed to `https`.

Issue #3626, opened 2021-03-16
plugins.bigo: not working


shiro789:
## Plugin Issue

- [ ] This is a plugin issue and I have read the contribution guidelines.
- [ ] I am using the latest development version from the master branch.

### Reproduction steps / Explicit stream URLs to test

1. https://www.bigo.tv/swxytx
2. https://bigo.tv/20566029

### Log output

```
[cli][info] Found matching plugin bigo for URL https://www.bigo.tv/swxytx
error: No playable streams found on this URL: https://www.bigo.tv/swxytx
```

### Additional comments, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

bastimeyer:
Neither of these two channels is online and it also seems that the plugin does only support input URLs with the "bigo ID", and not profile names, which means you'll have to find the right URL for the stream. The other streams on this site are working fine.

mkbloke:
@shiro789, how/from where are you getting a username?  I do not see anything other than IDs used on the website (while not logged in, as I do not have an account).

unnutricious:
I can confirm I saw this problem as well, but it appears to be on the bigo.tv side.  You can replicate this issue by going to https://bigo.tv/show and clicking on streamers randomly. Some will work and some won't, showing offline. The site seems to still be in transition from flash to html5, so that might be related. If you try sending a query to the API url  `https://www.bigo.tv/OInterface/getVideoParam?bigoId={0}`, the data section is empty, no wsUrl. It seems random, not tied to particular broadcasters, just each live stream it may or may not work.

The apps still work correctly, but they do not use HLS, they use the proprietary bigo protocol, which is beyond my comprehension to reverse-engineer. This plugin used to support it, and I removed it after it stopped working correctly a few years ago. I don't know if it's still the same or not.

back-to:
closing `https://bigo.tv/<ID>` works normally

unnutricious:
This is not the issue of id vs profile name, the site is still broken when viewing in a browser as of today. There's nothing that can be done on the streamlink side to fix it though, other than changing it to support the streaming protocol used in-app.

shiro789:
is there any way to fix manually?

trodiz:
> Neither of these two channels is online and it also seems that the plugin does only support input URLs with the "bigo ID", and not profile names, which means you'll have to find the right URL for the stream. The other streams on this site are working fine.

[...]

PR #3657, opened 2021-03-29
plugins.bigo: reflect recent changes in server API

trodiz:
Bigo Live's web api, which this plugin relies on, has changed some of its server endpoints and terminologies. This is the underlying reason for this issue: https://github.com/streamlink/streamlink/issues/3626. This patch solves it.

### DIFF
--- a/src/streamlink/plugins/bigo.py
+++ b/src/streamlink/plugins/bigo.py
@@ -7,13 +7,13 @@ from streamlink.stream import HLSStream
 
 class Bigo(Plugin):
     _url_re = re.compile(r"https?://(?:www\.)?bigo\.tv/([^/]+)$")
-    _api_url = "https://www.bigo.tv/OInterface/getVideoParam?bigoId={0}"
+    _api_url = "https://bigo.tv/studio/getInternalStudioInfo"
 
     _video_info_schema = validate.Schema({
         "code": 0,
         "msg": "success",
         "data": {
-            "videoSrc": validate.any(None, "", validate.url())
+            "hls_src": validate.any(None, "", validate.url())
         }
     })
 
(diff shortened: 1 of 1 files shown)


unnutricious:
This change itself looks fine, but I don't think it has any real effect. Both the old and new API still work, but from my quick tests as of today, both still do not work for many rooms. The hls_src field is missing in the new API just as the videoSrc field is missing in the old. I did observe that bigo.tv now seems to only show a small number of rooms compared to the app, so it seems the site/API is still just broken on the bigo server side.

back-to:
@unnutricious 

> both still do not work for many rooms. 

share the links

unnutricious:
> > 
> 
> share the links

It's not specific rooms, it's random per broadcast, so I can't share a link that would be of any use unless we are debugging this in real-time. To easily see the problem with the website now:
1. Go to bigo.tv and click LIVE and select a country.
2. Compare the result to the Android/Apple app for the same country. Only a small number of streams will show up. The ones that do show up will work in browser and if you query the API endpoint, hls_src will be present and populated.
2. Select a broadcast randomly from the app that does not show up on bigo.tv and query the API with siteId={bigo_id}. You will see the metadata is returned correctly, but most of the time (but not all the time), hls_src will be missing.

I observed that hls_src seems to be present when a stream first starts and then disappears after a few minutes in some cases. In other cases, I don't think it shows up at all, but it's hard to tell since I have no way to trigger on a broadcast.

back-to:
> Select a broadcast randomly from the app that does not show up on bigo.tv

If it doesn't work on there website, it won't work with Streamlink.

---

closing

The old API still works as expected, this PR is currently not needed.

eccccco:
My recent observation about this issue:

 I successfully used streamlink to open a BIGO stream (with username not ID). The Stream was live in the App.
 after 15 minutes i made the following tests (the stream was still playing over streamlink):

- stream was offline on the bigo website
- stream was online in the app and listed in the country list in the app (not private/premium or something like that)
- opening a new streamlink instance did NOT open the stream a second time (the initial stream was still working with streamlink)
- using the API request "...OInterface/getVideoParam?bigoId= " returned "code=0" msg "success" but NO values for wsUrl and videoSrc
- using the newer API request ".../studio/getInternalStudioInfo?siteId= " did NOT return the hlsUrl

trodiz:
> * using the newer API request ".../studio/getInternalStudioInfo?siteId= " did NOT return the hlsUrl

Do make sure that this new format of request is using POST method and not GET.

unnutricious:
> > * using the newer API request ".../studio/getInternalStudioInfo?siteId= " did NOT return the hlsUrl
> 
> Do make sure that this new format of request is using POST method and not GET.

I did try both POST and GET, I get the same issue and everything else shows as if the user is offline, 'alive' is 0. If you browse to the bigo.tv url, it will show the room as offline, and it is querying the API itself. Let me know if you want to discuss more details of if you are seeing different behavior. I was looking at supporting the bigo app protocol, but my progress has been slow, as it is all binary. This module used to support it, but I removed it a few years ago because the video was coming out corrupted. Since then, there is no way to get the ip/port info for the app stream from the website now that flash support has been removed.

PR #3661, opened 2021-03-31
plugins.picarto: rewrite/fix

### LINKED ISSUE #3647: plugins.picarto: 405 Client Error: Not Allowed for url: https://picarto.tv/process/channel
## Plugin Issue

1. when running the command; # streamlink picarto.tv/<streamname> best

### Log output

```
[cli][info] Found matching plugin picarto for URL picarto.tv/LukeTUGS
error: Unable to open URL: https://picarto.tv/process/channel (405 Client Error: Not Allowed for url: https://picarto.tv/process/channel)
```

### Additional comments, etc.

Picarto.tv recently updated their API.
Link to their API documents page [https://api.picarto.tv](https://api.picarto.tv)

mkbloke:
closes #3647

### DIFF
--- a/src/streamlink/plugins/picarto.py
+++ b/src/streamlink/plugins/picarto.py
@@ -1,14 +1,9 @@
 import logging
 import re
-import time
-from urllib.parse import unquote_plus, urlparse
-
-import websocket
 
 from streamlink.plugin import Plugin
-from streamlink.plugin.api import useragents, validate
+from streamlink.plugin.api import validate
 from streamlink.stream import HLSStream
-from streamlink.utils import parse_json
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


bastimeyer (commented):
I don't have the time to review this now, but the websocket logic needs to be refactored at some point. For anything HTTP related we have the HTTPSession, but the websocket stuff gets duplicated here for the fourth time, and this is a problem.

mkbloke:
OK.  I guess the WS stuff is just something to be added to the TODO list at some point in the future...

src/streamlink/plugins/picarto.py
```diff
@@ -1,128 +1,214 @@
...
+                    (unquote_plus(p.username), unquote_plus(p.password or ''))
+        return proxy_options
+
+    def get_stream(self, data, live=True, server=None):
+        # Proxy support adapted from the UStreamTV plugin (ustreamtv.py)
+        proxy_url = self.session.get_option('https-proxy')
+        if proxy_url is None:
+            proxy_url = self.session.get_option('http-proxy')
+        proxy_options = self.parse_proxy_url(proxy_url)
+        if proxy_options.get('http_proxy_host'):
```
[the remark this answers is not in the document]
mkbloke: Argh, thank you!

That's what happens when you add the proxy code in later.  I even tested it, but I didn't check to see if the WS traffic was *acutally* going through the proxy.  :anguished:

src/streamlink/plugins/picarto.py
```diff
@@ -1,128 +1,222 @@
...
+            log.debug('Using proxy ({0}://{1}:{2})'.format(
+                proxy_options.get('proxy_type') or 'http',
+                proxy_options.get('http_proxy_host'),
+                proxy_options.get('http_proxy_port') or 80))
+
+        if live and server:
+            wss_url = self.wss_url.format(
```
back-to, line 115:
```
    wss_url = 'wss://{0}/stream/json_{1}'
    live_wss_path = 'golive%2B{0}.js'
    vod_wss_server = 'recording-eu-1.picarto.tv'
    vod_wss_path = '{0}.js'
```

seems unnecessarily complicated

just use something like `wss_url = f'wss://{server}/stream/json_golive%{data}.js'`

mkbloke: Hmm, it appears that with an f-string I'd have to take the live and VOD URLs out of the class space and put them directly into the code.  I don't know if that's a good thing or better to leave them defined at the top with the other URLs?

How about defining one for each of live and VOD:
```
live_wss_url = 'wss://{server}/stream/json_golive%{data}.js'
vod_wss_url = 'wss://{server}/stream/json_{data}.js'
```
and then just use:
```
        if live and server:
            wss_url = self.live_wss_url.format(server=server, data=data)
        else:
            server = self.vod_wss_server
            wss_url = self.vod_wss_url(server=server, data=data)
```
instead?  It seems clean, while keeping the URL defs at the top with the others.

Issue #3668, opened 2021-04-01
plugins.youtube: video_id not found by regexp


etam:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.
- [x] I am using the latest development version from the master branch: current master is c7bd7ec6099192d2b3c7bed1bbcd472cb8949f95

### Description

Streamlink fails to get stream url of a live stream. It used to do it fine few days ago.

### Reproduction steps / Explicit stream URLs to test

1. Try to play `https://www.youtube.com/c/Urz%C4%85dTransportuKolejowego/live`

### Log output

```
> ./bin/streamlink --loglevel debug "https://www.youtube.com/c/Urz%C4%85dTransportuKolejowego/live" 720p --stream-url
Traceback (most recent call last):
  File "./bin/streamlink", line 33, in <module>
    sys.exit(load_entry_point('streamlink==2.1.1+1.gc7bd7ec', 'console_scripts', 'streamlink')())
  File "/home/etam/tmp/streamlink_master/lib64/python3.8/site-packages/streamlink_cli/main.py", line 1075, in main
    handle_url()
  File "/home/etam/tmp/streamlink_master/lib64/python3.8/site-packages/streamlink_cli/main.py", line 566, in handle_url
    streams = fetch_streams(plugin)
  File "/home/etam/tmp/streamlink_master/lib64/python3.8/site-packages/streamlink_cli/main.py", line 459, in fetch_streams
    return plugin.streams(stream_types=args.stream_types,
  File "/home/etam/tmp/streamlink_master/lib64/python3.8/site-packages/streamlink/plugin/plugin.py", line 317, in streams
    ostreams = self._get_streams()
  File "/home/etam/tmp/streamlink_master/lib64/python3.8/site-packages/streamlink/plugins/youtube.py", line 301, in _get_streams
    self.video_id = self._find_video_id(self.url)
  File "/home/etam/tmp/streamlink_master/lib64/python3.8/site-packages/streamlink/plugins/youtube.py", line 262, in _find_video_id
    return self._find_video_id(canon_link)
  File "/home/etam/tmp/streamlink_master/lib64/python3.8/site-packages/streamlink/plugins/youtube.py", line 227, in _find_video_id
    if m.group("video_id"):
AttributeError: 'NoneType' object has no attribute 'group'
```

### Additional comments, etc.
If in 
[...]

PR #3672, opened 2021-04-02
plugins.youtube: fix consent dialog

back-to:
closes https://github.com/streamlink/streamlink/issues/3668

### DIFF
--- a/src/streamlink/plugins/youtube.py
+++ b/src/streamlink/plugins/youtube.py
@@ -234,8 +234,7 @@ class YouTube(Plugin):
                 if _i.attributes.get("type") == "hidden":
                     c_data[_i.attributes.get("name")] = _i.attributes.get("value")
             log.debug(f"c_data_keys: {', '.join(c_data.keys())}")
-            self.session.http.post("https://consent.youtube.com/s", data=c_data)
-            res = self.session.http.get(url)
+            res = self.session.http.post("https://consent.youtube.com/s", data=c_data)
 
         datam = _ytdata_re.search(res.text)
         if datam:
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/youtube.py
```diff
@@ -296,6 +296,15 @@ def _get_stream_info(self, video_id):
         return info_parsed
 
     def _get_streams(self):
+        res = self.session.http.get(self.url)
+        if "consent.youtube" in res.url:
```
bastimeyer, line 300:
```suggestion
        if urlparse(res.url).netloc == "consent.youtube.com":
```

bastimeyer (commented):
This always sends a `GET` request to the input URL twice, first for the redirection check (and the `POST` afterwards) and then regularly in `_find_video_id`, even if there's no redirection to the consent page. If there's always a consent confirmation which I'm not sure about, then it's not a problem, but if not then it's redundant and unnecessarily deplays the resolving of the stream data.

src/streamlink/plugins/youtube.py
```diff
@@ -222,13 +222,21 @@ def _create_adaptive_streams(self, info, streams):
...
 
         res = self.session.http.get(url)
+        if urlparse(res.url).netloc == "consent.youtube.com":
+            c_data = {}
+            for _i in itertags(res.text, "input"):
+                if _i.attributes.get("type") == "hidden":
+                    c_data[_i.attributes.get("name")] = _i.attributes.get("value")
+            log.debug(f"c_data_keys: {', '.join(c_data.keys())}")
+            self.session.http.post("https://consent.youtube.com/s", data=c_data)
+            res = self.session.http.get(url)
```
bastimeyer, line 238:
```suggestion
            res = self.session.http.post("https://consent.youtube.com/s", data=c_data)
```

The `POST` automatically redirects to the intended source, so you don't have to unnecessarily re-request it.

bastimeyer (changes requested):

Issue #3568, opened 2021-02-15
Unable to use --player-fifo with MPV


Kein:
Streamlink 2.0.0
mpv 0.33.0-76-g93066ff12f Copyright © 2000-2020 mpv/MPlayer/mplayer2 projects

```
streamlink --player-fifo https://www.twitch.tv/channelName best
[cli][info] Found matching plugin twitch for URL https://www.twitch.tv/channelName 
[cli][info] Available streams: audio_only, 160p (worst), 360p, 480p, 720p, 720p60, 1080p60 (best)
[cli][info] Opening stream: 1080p60 (hls)
[cli][info] Creating pipe streamlinkpipe-1140
[cli][info] Starting player: "c:\Programs\MPV\mpv.exe"
error: Failed to start player: "c:\Programs\MPV\mpv.exe" (Process exited prematurely)
[cli][info] Closing currently open stream...
```

Works normally without FIFO

beardypig:
Does it work with VLC?

back-to:
https://raw.githubusercontent.com/streamlink/streamlink/master/.github/ISSUE_TEMPLATE/bug_report.md

https://streamlink.github.io/cli.html#cmdoption-loglevel
https://streamlink.github.io/cli.html#cmdoption-verbose-player

Kein:
> Does it work with VLC?
I dont have VLC but I've tried mplayer and it wont work too.

Provided log **IS** a verbose output

mkbloke:
> Provided log IS a verbose output, it is exactly the same as normal, there is no extra info. It just dies,

That's not what the command line you have provided above seems to indicate.

Kein:
> it is exactly the same as normal, there is no extra info. It just dies,

Kein:
Stop being so immature and dense. Your documentation clearly states MPV works with Named Pipe when in reality it isnt, which is a misleading lie.

```
streamlink --verbose-player https://www.twitch.tv/yuushi_1234 best
[cli][info] Found matching plugin twitch for URL https://www.twitch.tv/yuushi_1234
[cli][info] Available streams: audio_only, 160p (worst), 360p, 480p, 720p60, 1080p60 (best)
[cli][info] Opening stream: 1080p60 (hls)
[cli][info] Starting player: "c:\Programs\MPV\mpv.exe"
[cli][info] Player closed
[cli][info] Stream ended
[cli][info] Closing currently open stream...

[...]

PR #3709, opened 2021-04-27
utils.named_pipe: rewrite named pipes

bastimeyer:
- split into two classes with an abstract base class
- remove name parameter from class and generate proper names
- remove mode parameter from open() as it's always binary mode
- change type of NamedPipe.path from str to pathlib.Path
- add tests

This is just a code refactor and doesn't fix the other issues mentioned here, as it would require refactoring the cli.output classes and stream.ffmpegmux in addition to the current changes:
https://github.com/streamlink/streamlink/issues/3568#issuecomment-783130671

There are also a couple of other issues I found while rewriting the Windows specific stuff and it seems like it doesn't close the named pipe gracefully and instead just swallows/supresses errors on the last (invalid) write attempt. I couldn't find a solution for this now, hence the lack of `TestNamedPipeWindows.test_write_before_open`, which would also cover this issue.

### DIFF
--- a/src/streamlink/stream/ffmpegmux.py
+++ b/src/streamlink/stream/ffmpegmux.py
@@ -1,6 +1,4 @@
 import logging
-import os
-import random
 import subprocess
 import sys
 import threading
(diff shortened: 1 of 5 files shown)


gravyboat:
Looks great @bastimeyer! Sorry about the delay, been a busy week.

Issue #3708, opened 2021-04-27
plugins.mitele: API quality problems


tvfj:
## Plugin Issue

- [X] This is a plugin issue and I have read the contribution guidelines.

### Description

After the implementation of the new API, there are two packages of different qualities on the same link, with different video and audio outputs.

Some channels (like Cuatro) are available at 720p50, about 4000k, but on VLC it appears at lower quality.

### Log output

```
[cli][debug] OS:         Linux-5.4.0-72-generic-x86_64-with-glibc2.31
[cli][debug] Python:     3.9.2
[cli][debug] Streamlink: 2.1.1
[cli][debug] Requests(2.25.1), Socks(1.7.1), Websocket(0.58.0)
[cli][debug] Arguments:
[cli][debug]  url=https://mitele.es/directo/cuatro
[cli][debug]  stream=['720p_alt2']
[cli][debug]  --loglevel=debug
[cli][debug]  --player=vlc
[cli][info] Found matching plugin mitele for URL https://mitele.es/directo/cuatro
[plugins.mitele][debug] Stream: https://live.mitele.es/hls/live/2004102/cuatro/master.m3u8 (hls)
[utils.l10n][debug] Language code: es_ES
[plugins.mitele][debug] Stream: https://live.mitele.es/hls/live/2004102/cuatro/master.m3u8 (hls)
[utils.l10n][debug] Language code: es_ES
[plugins.mitele][debug] Stream: https://live.mitele.es/hls/live/2004102/cuatro/playlist.m3u8 (hls)
[utils.l10n][debug] Language code: es_ES
[cli][info] Available streams: 360p_alt (worst), 360p, 480p_alt, 480p, 576p, 720p_alt2, 720p_alt, 720p (best)
[cli][info] Opening stream: 720p_alt2 (hls)
[stream.hls][debug] Reloading playlist
[stream.hls][debug] First Sequence: 161956505; Last Sequence: 161956509
[stream.hls][debug] Start offset: 0; Duration: None; Start Sequence: 161956507; End Sequence: None
[stream.hls][debug] Adding segment 161956507 to queue
[cli][debug] Pre-buffering 8192 bytes
[stream.hls][debug] Adding segment 161956508 to queue
[stream.hls][debug] Adding segment 161956509 to queue
[cli][info] Starting player: vlc
[cli.output][debug] Opening subprocess: vlc --input-title-format https://mitele.es/directo/cuatro -
[stream.hls][debug] Download of segment 161956507 complete
[...]

PR #3721, opened 2021-05-08
stream.hls_playlist: round BANDWIDTH and parse as int

back-to:
same as DASH Streams https://github.com/streamlink/streamlink/pull/1995

---

before

```
[cli][info] Available streams: 360p_364k (worst), 480p_764k, 720p_1564k, 720p_2564k, 576p_5008k, 720p_6891k, 720p_9726k (best)
```

now

```
[cli][info] Available streams: 360p_360k (worst), 480p_760k, 720p_1600k, 720p_2600k, 576p_5000k, 720p_6900k, 720p_9700k (best)
```

---

useful for https://github.com/streamlink/streamlink/issues/3708


src/streamlink/stream/hls_playlist.py
```diff
@@ -94,7 +95,7 @@ def create_stream_info(self, streaminf, cls=None):
 
         bandwidth = streaminf.get("BANDWIDTH")
         if bandwidth:
-            bandwidth = float(bandwidth)
+            bandwidth = round(float(bandwidth), 1 - int(math.log10(float(bandwidth))))
```
bastimeyer, line 98: Why is bandwidth parsed as a float?

https://tools.ietf.org/html/rfc8216#section-4.3.4.2

> The following attributes are defined:
>
>      BANDWIDTH
>
>      The value is a decimal-integer of bits per second.  It represents
>      the peak segment bit rate of the Variant Stream.

bastimeyer:
Btw, the entire stream name+weights selection needs a major rewrite. We're currently guessing the stream weight from the yielded stream name in the streams dict instead of properly calculating the stream weights from various (additional) stream properties, like FPS for example. This would also fix issues with duplicates, but it'd probably be a breaking change, as this is part of the public API.

src/streamlink/stream/hls_playlist.py
back-to: seems like it was a fix for the twitch plugin

https://github.com/streamlink/streamlink/commit/4305c8c7ca90fb328daad537bd6f4bb5dd4e20b5

https://github.com/chrippa/livestreamer/pull/1397

Issue #3587, opened 2021-02-26
plugins.onetv: All channels except 1tv don't work


mgibert-bmat:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.
- [ ] I am using the latest development version from the master branch.

### Description

Want to get stream from onetv channels in Russia. With a Russian IP, so geobocked issue is discarded.
The only channel I manage to record is 1tv.ru. The rest don't work.

### Reproduction steps / Explicit stream URLs to test

1. streamlink "https://chetv.ru/online/"
2. streamlink "https://ctc.ru/online/"
3. streamlink "https://domashniy.ru/online/"
4. 

### Log output

```
[cli][info] streamlink is running as root! Be careful!
[cli][debug] OS:         Linux-4.15.0-136-generic-x86_64-with-Ubuntu-18.04-bionic
[cli][debug] Python:     3.6.9
[cli][debug] Streamlink: 2.0.0
[cli][debug] Requests(2.25.1), Socks(1.7.1), Websocket(0.57.0)
[cli][info] Found matching plugin onetv for URL https://chetv.ru/online/
[plugins.onetv][debug] Loading live stream for ctc-che...
error: Unable to open URL: https://media.1tv.ru/api/v1/ctc/playlist/ctc-che_as_array.json (404 Client Error: Not Found for url: https://media.1tv.ru/api/v1/ctc/playlist/ctc-che_as_array.json)
```

### Additional comments, etc.

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

bastimeyer:
Post a debug log
https://github.com/streamlink/streamlink/blame/master/.github/ISSUE_TEMPLATE/1_plugin_issue.md#L47-L58

mgibert-bmat:
Hi @bastimeyer , 
Debug log is updated now. 
Thanks!

PR #3743, opened 2021-05-22
plugins.onetv: split plugin

back-to:
**plugins.onetv**
- uses **API v1** which is only 1tv.ru
- remove broken VOD support

**plugins.mediavitrina**
- uses **API v2**

closes https://github.com/streamlink/streamlink/issues/3587
closes https://github.com/streamlink/streamlink/pull/3669

### DIFF
--- a/src/streamlink/plugins/mediavitrina.py
+++ b/src/streamlink/plugins/mediavitrina.py
@@ -1,0 +1,85 @@
+import logging
+import re
+
+from streamlink.plugin import Plugin
+from streamlink.plugin.api import validate
+from streamlink.stream import HLSStream
+from streamlink.utils import parse_json
+from streamlink.utils.url import update_qsd
+
+log = logging.getLogger(__name__)
+
+
+class MediaVitrina(Plugin):
+    _re_url_1 = re.compile(r'https?://(?P<channel>ctc(?:love)?|chetv|domashniy|5-tv)\.ru/(?:online|live)')
+    _re_url_2 = re.compile(r'https?://(?P<channel>ren)\.tv/live')
+    _re_url_3 = re.compile(r'https?://player\.mediavitrina\.ru/(?P<channel>[^/?]+.)(?:/[^/]+)?/[\w_]+/player\.html')
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return (
+            cls._re_url_1.match(url) is not None
+            or cls._re_url_2.match(url) is not None
+            or cls._re_url_3.match(url) is not None
+        )
+
+    def _get_streams(self):
+        channel = (self._re_url_1.match(self.url)
+                   or self._re_url_2.match(self.url)
+                   or self._re_url_3.match(self.url)
+                   ).group("channel")
+
+        channels = [
+            # ((channels), (path, channel))
+            (("5-tv", "tv-5", "5tv"), ("tv5", "tv-5")),
+            (("chetv", "ctc-che", "che_ext"), ("ctc", "ctc-che")),
+            (("ctc"), ("ctc", "ctc")),
+            (("ctclove", "ctc-love", "ctc_love_ext"), ("ctc", "ctc-love")),
+            (("domashniy", "ctc-dom", "domashniy_ext"), ("ctc", "ctc-dom")),
+            (("iz"), ("iz", "iz")),
+            (("mir"), ("mtrkmir", "mir")),
+            (("muztv"), ("muztv", "muztv")),
+            (("ren", "ren-tv", "rentv"), ("nmg", "ren-tv")),
+            (("russia1"), ("vgtrk", "russia1")),
+            (("russia24"), ("vgtrk", "russia24")),
+            (("russiak", "kultura"), ("vgtrk", "russiak")),
+            (("spas"), ("spas", "spas")),
+            (("tvc"), ("tvc", "tvc")),
+            (("tvzvezda", "zvezda"), ("zvezda", "zvezda")),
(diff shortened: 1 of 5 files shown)


src/streamlink/plugins/mediavitrina.py
```diff
@@ -0,0 +1,94 @@
...
+            path, channel = ("spas", "spas")
+        elif channel == "tvc":
+            path, channel = ("tvc", "tvc")
+        elif channel in ("tvzvezda", "zvezda"):
+            path, channel = ("zvezda", "zvezda")
+        elif channel in ("u", "u_ott"):
+            path, channel = ("utv", "u_ott")
+        else:
+            log.error(f"Unsupported channel: {channel}")
+            return
```
bastimeyer, line 66: It would be better if you'd add a dictionary here that maps the channel input names to path+channel properties.
This makes it much more legible and easier to maintain. And you avoid having lots of unnecessary code branches.

src/streamlink/plugins/mediavitrina.py
```diff
@@ -0,0 +1,94 @@
...
+    _re_url_2 = re.compile(r'https?://(?P<channel>ren)\.tv/live')
+    _re_url_3 = re.compile(r'https?://player\.mediavitrina\.ru/(?P<channel>[^/?]+.)(?:/[^/]+)?/[\w_]+/player\.html')
+
+    @classmethod
+    def can_handle_url(cls, url):
+        return (
+            cls._re_url_1.match(url) is not None
+            or cls._re_url_2.match(url) is not None
+            or cls._re_url_3.match(url) is not None
+        )
```
bastimeyer, line 24: Not sure if this should be changed to a single regex.

If we ever switch the plugin matching system and load plugins on demand with a list of regexes extracted from the plugin modules during build time, then we need to implement a system that can take multiple regexes into account for a single plugin. There are only a handful of plugins currently (5 or so) that use "extended" logic in the `can_handle_url` method.

Shouldn't be too much of an issue though, so feel free to ignore.

src/streamlink/plugins/onetv.py
```diff
@@ -31,72 +23,32 @@ def can_handle_url(cls, url):
...
+            ))
+
+        if not url:
+            return
+
+        if "georestrictions" in url:
+            log.error("Stream is geo-restricted")
+            return
+
+        url = url + "&" + urlencode(self.hls_session())
```
bastimeyer, line 50: Use `update_qsd` instead of concatenating the URL string

bastimeyer (commented):
Just commenting on the code, not the actual implementation, as it's geo-blocked and I don't have access.

src/streamlink/plugins/onetv.py
back-to: > 400 Client Error

`update_qsd` won't work here, `urlencode` will make the url invalid
https://github.com/streamlink/streamlink/blob/44d4afaa931e0e3f525ae9ec7adcb91f1020acb2/src/streamlink/utils/url.py#L103

bastimeyer: > `urlencode` will make the url invalid

How? Can you show an example URL?

`update_qsd` doesn't have test cases for proper URL encoding, but I don't see any issues with the current implementation.
Example tests for correctly encoded and for incorrecly encoded input URLs:

```diff
diff --git a/tests/test_utils_url.py b/tests/test_utils_url.py
index eb87e6df..2e0e14f4 100644
--- a/tests/test_utils_url.py
+++ b/tests/test_utils_url.py
@@ -43,3 +43,5 @@ def test_update_qsd():
     assert update_qsd("http://test.se?&two=", {"one": ''}, keep_blank_values=False) == "http://test.se?one=", \
         "should set one blank"
     assert update_qsd("http://test.se?one=", {"two": 2}) == "http://test.se?one=&two=2"
+    assert update_qsd("http://test.se?foo=%20&bar=%3F", {"foo": "!"}) == "http://test.se?foo=%21&bar=%3F"
+    assert update_qsd("http://test.se?foo= &bar=?", {"foo": "!"}) == "http://test.se?foo=%21&bar=%3F"
```

back-to: > How? Can you show an example URL?

`update_qsd` makes `cdn` invalid and the server doesn't accept it.

```
--- valid
<Response [200]>

https://mobdrm.1tv.ru/hls-live10/streams/1tv/1tv.m3u8?cdn=https://edge1.1internet.tv&netprav=/hls-vod11/streams/1tv/netprav.m3u8&e=1621770289&s=%2FmI3ltgPPHFCpQ


--- invalid
<Response [400]>
ERROR: Not valid CDN

https://mobdrm.1tv.ru/hls-live10/streams/1tv/1tv.m3u8?cdn=https%3A%2F%2Fedge1.1internet.tv&netprav=%2Fhls-vod11%2Fstreams%2F1tv%2Fnetprav.m3u8&e=1621770289&s=%2FmI3ltgPPHFCpQ
```

Billy2011: `urlencode` has an option `safe=''` with which it is possible that individual characters are not escaped:
```
urlencode({"cdn": "http://test.com"}, safe="/:")
'cdn=http://test.com'

```
`update_qsd` would have to be adapted accordingly.

bastimeyer: > `update_qsd` would have to be adapted accordingly.

#3746

bastimeyer (approved):
Code looks good now, hence the approval. Can't test the sites though due to geolocking

PR #3752, opened 2021-05-25
plugins.funimationnow: fix subtitle language

FaceHiddenInsideTheDark:
This fixes a common error with the FunimationNow failing due to an inability to handle content with these select subtitle formats attached to them, such as every single show released past the year 2021 that isn't dubbed.
To reproduce this issue, simply run for yourself 
`streamlink --funimation-email="xxx" --funimation-password="xxx" --http-cookie "incap_ses_xxxx_xxxxxx=xxxxxxxxxxxx" "https://www.funimation.com/en/shows/any-simulcast-show-not-listed-dubbed/"`

### DIFF
--- a/src/streamlink/plugins/funimationnow.py
+++ b/src/streamlink/plugins/funimationnow.py
@@ -238,7 +238,7 @@ class FunimationNow(Plugin):
                 for subtitle in exp.subtitles():
                     log.debug(f"Subtitles: {subtitle['src']}")
                     if subtitle["src"].endswith(".vtt") or subtitle["src"].endswith(".srt"):
-                        sub_lang = {"en": "eng", "ja": "jpn"}[subtitle["language"]]
+                        sub_lang = {"en": "eng", "ja": "jpn", "es": "esp", "pt": "por"}[subtitle["language"]]
                         # pick the first suitable subtitle stream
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/funimationnow.py
```diff
@@ -238,7 +238,7 @@ def _get_streams(self):
                 for subtitle in exp.subtitles():
                     log.debug(f"Subtitles: {subtitle['src']}")
                     if subtitle["src"].endswith(".vtt") or subtitle["src"].endswith(".srt"):
-                        sub_lang = {"en": "eng", "ja": "jpn"}[subtitle["language"]]
+                        sub_lang = {"en": "eng", "ja": "jpn", "es": "esp", "pt": "por"}[subtitle["language"]]
```
bastimeyer, line 241: Thanks for the PR, but could you please fix the key lookup as well, so that `KeyError`s won't get raised when there's a subtitle language that doesn't exist in the dictionary?

I'd suggest turning the `sub_lang` dictionary into a class member (either on the `Experience` class or the `FunimationNow` class) and changing the lookup to:

```suggestion
                        sub_lang = Experience.sub_langs.get(subtitle["language"])
                        if not sub_lang:
                            continue
```

Thanks

bastimeyer (dismissed):

PR #3757, opened 2021-05-28
plugins.mediaklikk: improve _url_re

### LINKED ISSUE #3210: Twitch.tv embedded ads (2020-09-29)

Billy2011:
add m4sport.hu domain

closes #3210

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -103,7 +103,8 @@ liveme                  liveme.com           Yes   --
 livestream              livestream.com       Yes   --
 lrt                     lrt.lt               Yes   No
 ltv_lsm_lv              ltv.lsm.lv           Yes   No    Streams may be geo-restricted to Latvia.
-mediaklikk              mediaklikk.hu        Yes   No    Streams may be geo-restricted to Hungary.
+mediaklikk              - mediaklikk.hu      Yes   No    Streams may be geo-restricted to Hungary.
+                        - m4sport.hu
 mediavitrina            mediavitrina.ru      Yes   No    Streams may be geo-restricted to Russia.
 mitele                  mitele.es            Yes   No    Streams may be geo-restricted to Spain.
 mjunoon                 mjunoon.tv           Yes   Yes   Streams may be geo-restricted to Pakistan.
--- a/src/streamlink/plugins/mediaklikk.py
+++ b/src/streamlink/plugins/mediaklikk.py
@@ -12,7 +12,7 @@ log = logging.getLogger(__name__)
 class Mediaklikk(Plugin):
     PLAYER_URL = "https://player.mediaklikk.hu/playernew/player.php"
 
-    _url_re = re.compile(r"https?://(?:www\.)?mediaklikk\.hu/[\w\-]+\-elo/?")
+    _url_re = re.compile(r"https?://(?:www\.)?(?:mediaklikk|m4sport)\.hu/(?:[\w\-]+\-)?elo/?")
     _id_re = re.compile(r'"streamId":"(\w+)"')
     _file_re = re.compile(r'"file":\s*"([\w\./\\=:\-\?]+)"')
 
--- a/tests/plugins/test_mediaklikk.py
+++ b/tests/plugins/test_mediaklikk.py
@@ -10,8 +10,12 @@ class TestPluginCanHandleUrlMediaklikk(PluginCanHandleUrl):
         'https://www.mediaklikk.hu/duna-world-radio-elo',
         'https://www.mediaklikk.hu/m1-elo',
         'https://www.mediaklikk.hu/m2-elo',
+        'https://m4sport.hu/elo/',
+        'https://m4sport.hu/elo/?channelId=m4sport+',
+        'https://m4sport.hu/elo/?showchannel=mtv4plus',
     ]
 
     should_not_match = [
         'https://www.mediaklikk.hu',
+        'https://m4sport.hu',
     ]


bastimeyer:
Thanks

PR #3760, opened 2021-05-30
plugins.olympicchannel: fix / rewrite

Billy2011:
add olympics.com

### DIFF
--- a/src/streamlink/plugins/olympicchannel.py
+++ b/src/streamlink/plugins/olympicchannel.py
@@ -1,58 +1,67 @@
-import json
 import logging
 import re
+from html import unescape as html_unescape
 from time import time
 from urllib.parse import urljoin, urlparse
 
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
 from streamlink.stream import HLSStream
+from streamlink.utils import parse_json
 
 log = logging.getLogger(__name__)
 
 
 class OlympicChannel(Plugin):
-    _url_re = re.compile(r"https?://(\w+\.)olympicchannel.com/../(?P<type>live|video|original-series|films)/?(?:\w?|[-\w]+)")
-    _tokenizationApiDomainUrl = """"tokenizationApiDomainUrl" content="/OcsTokenization/api/v1/tokenizedUrl">"""
-    _live_api_path = "/OcsTokenization/api/v1/tokenizedUrl?url={url}&domain={netloc}&_ts={time}"
-
+    _url_re = re.compile(r"https?://(\w+\.)?(?:olympics|olympicchannel)\.com/(?:[\w-]+/)?../.+")
+    _token_api_path = "/tokenGenerator?url={url}&domain={netloc}&_ts={time}"
     _api_schema = validate.Schema(
-        validate.text,
-        validate.transform(lambda v: json.loads(v)),
-        validate.url()
+        validate.transform(parse_json),
+        [{
+            validate.optional("src"): validate.url(),
+            validate.optional("srcType"): "HLS",
+        }],
+        validate.transform(lambda v: v[0].get("src")),
     )
-    _video_url_re = re.compile(r""""video_url"\scontent\s*=\s*"(?P<value>[^"]+)""")
-    _video_url_schema = validate.Schema(
-        validate.contains(_tokenizationApiDomainUrl),
-        validate.transform(_video_url_re.search),
-        validate.any(None, validate.get("value")),
-        validate.url()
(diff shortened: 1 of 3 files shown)


src/streamlink/plugins/olympicchannel.py
```diff
@@ -1,58 +1,68 @@
...
+            validate.all(
+                validate.transform(_data_url_re.search),
+                validate.get(1),
+            ),
+            validate.all(
+                validate.transform(_data_content_re.search),
+                validate.get(1),
+            ),
+        ),
+        validate.transform(html_unescape),
```
bastimeyer, line 39: It might be better to use `itertags` here instead of having a regex that depends on the attribute order, but it's not really important.

Be aware though that the validation schema does not handle invalid/missing data and will raise an error if none of the regexes match.

src/streamlink/plugins/olympicchannel.py
```diff
@@ -1,58 +1,68 @@
...
+    def _get_streams(self):
+        api_url = self.session.http.get(self.url, schema=self._data_content_schema)
+        if api_url.startswith("/") or api_url.startswith("http"):
+            api_url = urljoin(self.url, api_url)
+            stream_url = self.session.http.get(api_url, schema=self._api_schema, headers={"Referer": self.url})
+        elif api_url.startswith("[{"):
+            stream_url = self._api_schema.validate(api_url)
+        else:
+            log.debug(api_url)
+            return
```
bastimeyer, line 59: Raised errors from the validation schemas should ideally be caught here. Then you also don't need the startwith calls.

The log.debug call doesn't really belong here if the data is invalid and no streams will be returned.

bastimeyer (commented):
Except for some minor details, this is looking fine and the plugins seems to be working correctly.

src/streamlink/plugins/olympicchannel.py
Billy2011: > Be aware though that the validation schema does not handle invalid/missing data and will raise an error if none of the regexes match.

`validate.get (1)` causes a `PluginError` in this case and that is desirable:

`error: Unable to validate response text: 'NoneType' object has no attribute '__getitem__' or 'NoneType' object has no attribute '__getitem__'`

> Raised errors from the validation schemas should ideally be caught here. 

I don't know what you mean `PluginError` doesn't need to be dealt with here.

> Then you also don't need the startwith calls.

The startswith calls are required for the differentiation and further processing of the data.

bastimeyer: > and that is desirable

Some input URLs that match the regex should not raise an error, eg. this:
https://olympics.com/en/athletes/

Streamlink should return `error: No playable streams found on this URL: ...`, which means the plugin has to handle this in the validation schema and then return None.

Errors which are related to processing actual video content are okay.

> The startswith calls are required

What I meant was that it's redundant to check when you just attempt to parse it as JSON if you already catch errors, but it can also be kept.

src/streamlink/plugins/olympicchannel.py
```diff
@@ -1,58 +1,68 @@
...
-        return HLSStream.parse_variant_playlist(self.session, stream_url)
+    def _get_streams(self):
+        api_url = self.session.http.get(self.url, schema=self._data_content_schema)
+        if api_url and (api_url.startswith("/") or api_url.startswith("http")):
+            api_url = urljoin(self.url, api_url)
+            stream_url = self.session.http.get(api_url, schema=self._api_schema, headers={"Referer": self.url})
+        elif api_url and api_url.startswith("[{"):
+            stream_url = self._api_schema.validate(api_url)
+        else:
+            raise PluginError("No playable streams found on this URL: {0}".format(self.url))
```
bastimeyer, line 59: Don't raise a custom PluginError and just return None. This is not what a PluginError is for.

- https://github.com/streamlink/streamlink/blob/e747226d2917a611cae9594722434376968fab00/src/streamlink/plugin/plugin.py#L324-L330
- https://github.com/streamlink/streamlink/blob/e747226d2917a611cae9594722434376968fab00/src/streamlink_cli/main.py#L569-L570

bastimeyer (changes requested):

src/streamlink/plugins/olympicchannel.py
Billy2011: For me it's a `ValueError`, but if you think that's better, I'll change that.

bastimeyer (approved):

gravyboat:
Thanks @Billy2011.

PR #3766, opened 2021-06-01
cli: update config file paths and plugin dirs on all OSes

bastimeyer:
According to the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), only configuration files are supposed to be in the `XDG_CONFIG_HOME` directory. This means that Streamlink's custom plugin files directory should default to `XDG_DATA_HOME` and not to `XDG_CONFIG_HOME`, as plugin files are not configuration files.

- The first commit therefore refactors the CLI's path constants and enables having a list of default plugin dirs.
- The second commit adds `XDG_DATA_HOME` to the `PLUGIN_DIRS` on POSIX in addition to `XDG_CONFIG_HOME` for backwards compatibility.
- I've also reformatted the CLI docs a bit for better readability (check the preview).

----

This also raises the question whether we should add `%APPDATA%\streamlink\config` to the `CONFIG_FILES` list on Windows, in addition to `%APPDATA%\streamlink\streamlinkrc`. The current config file name comes from the time when Livestreamer was still using the `~/.livestreamerrc` config file, which is deprecated anyway (not officially flagged as deprecated, but we could do that too). Yes, these are technically "run command files", hence the "rc" suffix, but the primary/intended config file name on non-Windows systems is already "config" and this helps solving confusion, especially for the average Windows user.

Also not really sure about macOS. It has been treated as non-Windows the whole time (of course), but the XDG base dir spec doesn't really apply to it and they have other conventions (see the new log dir for example). This could also be corrected in another commit. Config path and plugin dir should be `${HOME}/Library/streamlink/config` and `${HOME}/Library/streamlink/plugins/` respectively if I'm not mistaken.

----

#### Question 1:

Add `%APPDATA%\streamlink\config` in addition to `%APPDATA%\streamlink\streamlinkrc` (and update config file path in the installer)?

#### Question 2:

Use correct config and plugin paths on macOS?

#### Question 3:

Document `~/.streamlinkrc` as deprecated in the docs and remove it in the next major release `3.0.0`?


gravyboat:
Changes look good, in regards to your questions:

1. Yes.
2. Yes.
3. Yes.

bastimeyer:
## Summary

1. **POSIX systems**
   Add `${XDG_DATA_HOME:-${HOME}/.local/share}/streamlink/plugins` as first plugins dir,
   but keep `${XDG_CONFIG_HOME:-${HOME}/.config}/streamlink/plugins` as secondary plugins dir.
2. **Windows**
   Add `%APPDATA%\streamlink\config` as first config file path,
   but keep `%APPDATA%\streamlink\streamlinkrc` as secondary config file path.
3. **macOS**
   - Add `${HOME}/Library/Application Support/streamlink/config` as first config file path,
     but keep config file paths from POSIX systems as secondary and third config file paths.
   - Add `${HOME}/Library/Application Support/streamlink/plugins` as first plugins dir,
     but keep plugin dirs from POSIX systems as secondary and third plugins dirs.


## Officially deprecate

- **Windows**
  - `%APPDATA%\streamlink\streamlinkrc`
- **macOS**
  - `${XDG_CONFIG_HOME:-${HOME}/.config}/streamlink/config`
  - `${HOME}/.streamlinkrc`
  - `${XDG_DATA_HOME:-${HOME}/.local/share}/streamlink/plugins`
  - `${XDG_CONFIG_HOME:-${HOME}/.config}/streamlink/plugins`
- **Linux/BSD**
  - `${HOME}/.streamlinkrc`
  - `${XDG_CONFIG_HOME:-${HOME}/.config}/streamlink/plugins`

bastimeyer:
Not sure if the OSes should be properly split up in the docs. I've added macOS as another row to the three tables, but kept the "Unix-like (POSIX)" rows, which implies that macOS is included. This avoids having redundant paths, but also doesn't show that the regular POSIX paths are deprecated on macOS. So should the "Unix-like (POSIX)" rows be renamed to "Linux/BSD" and those paths added to the macOS rows (as deprecated)?

gravyboat:
> So should the "Unix-like (POSIX)" rows be renamed to "Linux/BSD" and those paths added to the macOS rows (as deprecated)?

I'd say this is a good plan. It clarifies things further and leads to less confusion which is always a plus.

bastimeyer:
Updated the docs (with additional stylesheet updates) and properly split up the path definitions between macOS and Linux/BSD, which also means that the new `XDG_DATA_HOME` plugins dir is **not** part of the macOS plugins dir list, only the old/deprecated `XDG_CONFIG_HOME` one.

bastimeyer:
Do we need deprecation log messages when loading stuff from those paths?

gravyboat:
@bastimeyer I feel like it's overkill, but I've also learned people don't read the release notes and like to complain later. Will people read the CLI output noting as much? I'm not sure. If it's not too much work it's probably worth adding so we can point at release notes and CLI output when someone inevitably complains.

bastimeyer:
I've added the deprecation log messages. See my updated commit and its message.

- For some reason, logging a warning message still prints the message even while in json output mode, where loglevel should be none. That's why it's logging to info instead.
- It only logs deprecated plugin paths when at least one plugin was successfully loaded. Load failures are being logged separately in all cases.
- No tests, as there weren't any tests for `load_plugins`, `setup_args` and `setup_config_args` prior to my changes and I don't want to write them. The changes are simple enough though.

bastimeyer:
If you want, we can deprecate these paths in another PR. That would probably be better for transparency reasons.

gravyboat:
@bastimeyer Should the deprecation message include information about where the new config directory should live? Seems like just telling people the dir is deprecated without saying where to move things to might lead to some confusion.

> If you want, we can deprecate these paths in another PR. That would probably be better for transparency reasons.

Agreed.

bastimeyer:
I'll add a docs-reference to the log messages in the deprecation PR once this one has been merged. As you can see, I removed the deprecation commit, so this PR here should be good unless someone spots a mistake.

PR #3784, opened 2021-06-06
cli: deprecate old config files and plugin dirs

bastimeyer:
# Deprecated config file paths

- **Windows**
  - `%APPDATA%\streamlink\streamlinkrc`
- **macOS**
  - `${XDG_CONFIG_HOME:-${HOME}/.config}/streamlink/config`
  - `${HOME}/.streamlinkrc`
- **Linux/BSD**
  - `${HOME}/.streamlinkrc`

# Deprecated plugin directories

- **macOS**
  - `${XDG_CONFIG_HOME:-${HOME}/.config}/streamlink/plugins`
- **Linux/BSD**
  - `${XDG_CONFIG_HOME:-${HOME}/.config}/streamlink/plugins`

----

## Updated primary config file paths

- **Windows**
  `%APPDATA%\streamlink\config`
- **macOS**
  `${HOME}/Library/Application Support/streamlink/config`
- **Linux/BSD**
  `${XDG_CONFIG_HOME:-${HOME}/.config}/streamlink/config` (no changes)

## Updated primary plugin directories

- **Windows**
  `%APPDATA%\streamlink\plugins` (no changes)
- **macOS**
  `${HOME}/Library/Application Support/streamlink/plugins`
- **Linux/BSD**
  `${XDG_DATA_HOME:-${HOME}/.local/share}/streamlink/plugins`

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -28,7 +28,7 @@ from streamlink.plugin import PluginOptions
 from streamlink.stream import StreamProcess
 from streamlink.utils import LazyFormatter, NamedPipe
 from streamlink_cli.argparser import build_parser
-from streamlink_cli.compat import is_win32, stdout
+from streamlink_cli.compat import DeprecatedPath, is_win32, stdout
 from streamlink_cli.console import ConsoleOutput, ConsoleUserInputRequester
 from streamlink_cli.constants import CONFIG_FILES, DEFAULT_STREAM_METADATA, LOG_DIR, PLUGIN_DIRS, STREAM_SYNONYMS
 from streamlink_cli.output import FileOutput, PlayerOutput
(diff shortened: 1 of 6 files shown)


src/streamlink_cli/main.py
```diff
@@ -650,20 +651,32 @@ def setup_args(parser: argparse.ArgumentParser, config_files: List[Path] = None,
...
+            plugin = streamlink.resolve_url(args.url)
+            for config_file in CONFIG_FILES:
+                config_file = config_file.with_name(f"{config_file.name}.{plugin.module}")
+                if not config_file.is_file():
+                    continue
+                if type(config_file) is DeprecatedPath:
+                    log.info(f"Loaded plugin config from deprecated path, see CLI docs for how to migrate: {config_file}")
+                config_files.append(config_file)
+                break
+
```
bastimeyer, line 679: There is a small change in logic here. Previously it was trying to load multiple plugin specific config files, eg. `~/.config/streamlink/config.twitch` and `~/.streamlinkrc.twitch`, but since only the first non-plugin-specific default config file gets loaded (see above), plugin configs should have the same logic.

I'm also not sure if we should disable plugin config loading when `--config` is set or if we should add something like `--no-plugin-config` for disabling loading plugin configs.

PR #3798, opened 2021-06-18
plugins.playtv: removed - SEC_ERROR_EXPIRED_CERTIFICATE

back-to:
> 9. Sites which are unmaintained

All there websites have expired certificates.


bastimeyer:
> Issued On Monday, 20 April 2020 at 02:00:00
> Expires On Tuesday, 18 May 2021 at 01:59:59

That was a month ago, but the site also seems very broken otherwise (`--http-no-ssl-verify` in Streamlink and a plugin fix thus won't work) and those few streams which are not broken are loaded from other sites.

I agree, this can be removed.

PR #3803, opened 2021-06-19
ci.netlify: fix ignore command and fetch git tags

bastimeyer:
Lately, the builds on netlify did not get ignored properly and always ran regardless whether the files of the diff-check were modified or not. This attempts to fix it, but I'm not sure if it will.

I've intentionally based this PR on the commit before the release and it should not detect any changes to the changelog.md file and thus ignore the build. Let's see.

It also now fetches all git tags from the last 300 commits before building, similar to the GH workflow.


bastimeyer:
Let me quickly test if it correctly picks up documentation changes and then runs a build when I add another commit. If it does, then everything is working and I will remove the commit again.

PR #3811, opened 2021-06-21
plugins.rtvs: fixed livestream

back-to:
before

```
$ streamlink https://www.rtvs.sk/televizia/live-1
[cli][info] Found matching plugin rtvs for URL https://www.rtvs.sk/televizia/live-1
error: No playable streams found on this URL: https://www.rtvs.sk/televizia/live-1
```

after

```
$ streamlink https://www.rtvs.sk/televizia/live-1
[cli][info] Found matching plugin rtvs for URL https://www.rtvs.sk/televizia/live-1
```

### DIFF
--- a/src/streamlink/plugins/rtvs.py
+++ b/src/streamlink/plugins/rtvs.py
@@ -2,65 +2,46 @@ import re
 
 from streamlink.plugin import Plugin
 from streamlink.plugin.api import validate
-from streamlink.stream import HLSStream, RTMPStream
-
-RUURL = "b=chrome&p=win&v=56&f=0&d=1"
-
-_url_re = re.compile(r"https?://www.rtvs.sk/televizia/live-[\w-]+")
-_playlist_url_re = re.compile(r'"playlist": "([^"]+)"')
-
-_playlist_schema = validate.Schema(
-    [
-        {
-            "sources": [
-                validate.any(
-                    {
-                        "type": "dash",
-                        "file": validate.url(scheme="http")
-                    }, {
-                        "type": "hls",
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/rtvs.py
```diff
@@ -2,65 +2,46 @@
...
+        videos = parse_json(res.text, schema=validate.Schema({
+            "clip": {
+                "sources": [{
+                    "src": validate.url(),
+                    "type": str,
+                }],
+            }},
+            validate.get(("clip", "sources")),
+            validate.filter(lambda n: n["type"] == "application/x-mpegurl"),
+        ))
```
bastimeyer, line 42: Just a bit of nit-picking, but if you replace `"type": str` with `"type": "application/x-mpegurl"`, then you don't need `validate.filter` (unless you don't want to raise an error if no list item matches the dict schema).

bastimeyer (approved):
Plugin is looking good, but I'm not going to merge right away in case you want to change the validation schema (which is not necessary though).

src/streamlink/plugins/rtvs.py
back-to: @bastimeyer 
I don't want the error `Unable to validate JSON:`, if there are other `types` than `application/x-mpegurl`

bastimeyer: Yes, you are right, I misread `validate_sequence`
https://github.com/streamlink/streamlink/blob/5d1227e5977d3c1e98459799769ac6b207aab3dc/src/streamlink/plugin/api/validate.py#L352

Issue #3814, opened 2021-06-22
Improved plugins URL matching API


bastimeyer:
# Motivation

In order for Streamlink's `Session` to be able to find a matching plugin for the given input URL, plugins must implement the `can_handle_url` classmethod. For historical reasons and due to the work of many different people, every plugin is doing its own thing with how it handles the method's return value and how the URL regex(es) is/are defined, which is not only inconsistent, but also makes it difficult to read, adds complexity as well as maintenance burden and makes it difficult for new users to learn writing plugins.

The current API also doesn't allow for static code analysis of the URL regex(es), which would be a requirement if we want to eventually replace the inefficient plugin loading logic in the `Session` with a pre-build step that generates a JSON object with an array of built-in plugins and their regexes + matching priorities, that can then be used by the `Session` to find only a single plugin and load it instead of everything at once.

# Proposal

A better solution would therefore be having strict definitions of each plugin's URL regex(es) and their matching priorities, in a declarative way.

`Plugin.can_handle_url(url)` and `Plugin.priority(url)` could then be removed from built-in plugins and deprecated for third-party plugins.

## The `Plugin.matchers` attribute

Let's define a list of URL regexes and priorities.

```py
from typing import ClassVar, List, NamedTuple, Pattern

class Matcher(NamedTuple):
    pattern: Pattern
    priority: int

class Plugin:
    matchers: ClassVar[List[Matcher]] = None
```

A plugin could implement it like this (more elegant solution down below):

```py
from streamlink.plugin import LOW_PRIORITY, Matcher, Plugin

class MyPlugin(Plugin):
    matchers = [
        Matcher(re.compile("primary")),
        Matcher(re.compile("secondary", FLAGS), priority=LOW_PRIORITY)
    ]
```

And the `Session` could resolve an input URL like this:

```py
from streamlink.plugin.plugin import Matcher, NO_PRIORITY, Plugin

class Session:
    def resolve_url(self, url: str, follow_redirects: bool = False) -> Plugin
        url = update_scheme("http://", url)

        matcher: Matcher
        candidate = None
        priority = NO_PRIORITY
        for plugin in self.plugins.values():
            for matcher in plugin.matchers or []:
                if matcher.priority > priority and matcher.pattern.match(url) is not None:
                    candidate = plugin
                    priority = matcher.priority

        if candidate:
            return candidate(url)

        # `follow_redirects` logic...

        raise NoPluginError
```

### The `pluginmatcher` class decorator

[...]

PR #3821, opened 2021-06-23
plugin: new matchers API

bastimeyer:
Opening this as a draft, see #3814. 

The interesting commits are the first two and the last one, the rest are plugin updates, which are not that interesting and bloat up this diff a lot.

I've split up the plugin changes into multiple commits, based on the type of changes. Since the URL regexes had to be moved, I immediately applied fixes, as it would be duplicate work otherwise.

I'm not sure if I'm convinced with the regex pattern compilation in the decorator and the language injection annotations. After thinking about it again, it's probably better to just call `re.compile` in the decorator patameter.

Apart from that, the only thing that's missing is documentation.

### DIFF
--- a/src/streamlink/plugin/__init__.py
+++ b/src/streamlink/plugin/__init__.py
@@ -1,5 +1,9 @@
 from streamlink.exceptions import PluginError
 from streamlink.options import Argument as PluginArgument, Arguments as PluginArguments, Options as PluginOptions
-from streamlink.plugin.plugin import Plugin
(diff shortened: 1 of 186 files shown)


bastimeyer:
I've replaced `@pluginmatcher(pattern, flags, priority)` with `@pluginmatcher(re.compile(pattern, flags), priority)` now, as language injection annotations will cause issues in editors/IDEs other than PyCharm and are too verbose. This is much better I think.

Documentation still needs to be written, but that can be done later on to make things not more complicated here.

PR #3825, opened 2021-06-24
plugins.mediaklikk: rewrite plugin

bastimeyer:
Closes #3824

This rewrites the plugin with proper data extraction and validation schemas, and adds support for these sites, including VODs:
- Mediaklikk.hu
- M4sport.hu
- Hirado.hu
- Petofilive.hu

https://dunamsz.hu
https://en.wikipedia.org/wiki/MTVA_(Hungary)#Internet

The sports channels are geo-locked, but it should find the correct HLS streams on all URLs listed in the tests nevertheless.

### DIFF
--- a/src/streamlink/plugins/mediaklikk.py
+++ b/src/streamlink/plugins/mediaklikk.py
@@ -1,10 +1,12 @@
 import logging
 import re
+from urllib.parse import unquote, urlparse
 
-from streamlink.exceptions import PluginError
 from streamlink.plugin import Plugin
+from streamlink.plugin.api import validate
 from streamlink.stream import HLSStream
-from streamlink.utils.url import update_scheme
+from streamlink.utils import parse_json, update_scheme
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


back-to:
> The sports channels are geo-locked, but it should find the correct HLS streams on all URLs listed in the tests nevertheless.

```
https://c202-node61-cdn.connectmedia.hu/4501/vod/_definst_/r/mtva/2021/06/10/euro2020_bumper/index.smil/playlist.m3u8
https://c201-node61-cdn.connectmedia.hu/1104/xxx/xxx/index.m3u8?v=5iip:xxx
https://c202-node61-cdn.connectmedia.hu/4501/vod/_definst_/r/mtva/2021/06/10/euro2020_bumper/index.smil/playlist.m3u8

[cli][info] Available streams: 400k_alt (worst), 400k, 180p, 270p, 800k_alt, 800k, 360p, 1200k_alt, 1200k, 480p, 1600k_alt, 1600k, 720p, 3000k_alt, 3000k (best)
```

you need to skip vod urls for `https://m4sport.hu/elo/mtv4live/`

PR #3850, opened 2021-07-06
plugins.openrectv: update HLS URLs

vinyl-umbrella:
plugins.openrectv was unable to play the stream in media player due to an update to openrec.tv a few days ago.
So, I have made it possible and use low latency streaming video data.
I also added the feature to download uploaded videos.


bastimeyer (commented):
Thanks for the PR.

Could you please also update the validation schema (`_info_schema`) and add the missing fields there, so that there's no ugly `KeyError` when the property lookup fails and a validation error message is shown instead?

vinyl-umbrella:
@bastimeyer 
I updated validation schema. Thank you for let me know.

PR #3854, opened 2021-07-10
chore: replace issue templates with forms

bastimeyer:
GitHub's issue forms have been available in public beta for a bit now after being in closed beta for half a year.
- https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms
- https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema

This PR replaces the currently used issue templates with issue forms, which means users can't simply ignore the required details anymore and it is less confusing for people who don't know markdown syntax and HTML comments that were previously used in the issue templates to add help and description texts.

Similar to the issue templates, I've tried using the same style/scheme throughout all issue forms. I have merged a couple of sections for the sake of simplicity, eg. descriptions and reproduction steps, and also removed the generic issue template, as questions and meta threads should be opened on the discussion forum instead. If you think that this is a bad idea, I can revert this, but I'm not sure of the generic issue template can be an issue form and whether it'll be abused as the forms are more strict by design.

The issue forms can be tested on my fork repo, where I temporarily made the PR's branch the default one:
https://github.com/bastimeyer/streamlink/issues/new/choose

Please check for any issues and/or suggest improvements.


.github/ISSUE_TEMPLATE/1_plugin_issue.yml
```diff
@@ -0,0 +1,74 @@
...
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ### Thanks for reporting a plugin issue!
+
+        Plugin issues describe broken functionality within a plugin's code base, eg. the streaming site has made breaking changes, streams don't get resolved correctly, or authentication has stopped working, etc.
+
+        PLEASE DON'T IGNORE this template and fill in all the required details.
+        Issues that don't adhere to our request will be closed and ignored.
```
gravyboat, line 15: I think you should bold this whole line **Issues that don't adhere to our request will be closed and ignored.** for all of the templates.

gravyboat (approved):
Looks awesome @bastimeyer, I have one minor suggestion for a change, but if you don't agree that's fine. Merge when you're ready.

bastimeyer:
Btw, if you want to wait until issue forms leave beta status, then we can do that, but I don't think there will be any major changes. And if there are, it can be fixed immediately.

gravyboat:
@bastimeyer I don't see any reason to wait. We have access to them, let's use them. If we have to we can make changes later.

PR #3862, opened 2021-07-13
docs: fix headline anchors on deprecations page

bastimeyer:
https://streamlink.github.io/latest/deprecations.html#id1 (2.3.0)
and
https://streamlink.github.io/latest/deprecations.html#id2 (2.2.0)
are terrible anchor names and need to get fixed. It probably also makes sense distinguishing between streamlink and streamlink-cli on the deprecations page.

----

Let's also prepare the 2.3.0 release. We've had a lot of big changes, the new plugin matching API, fragmented HLS streams, a couple of fixes/improvements for high priority plugins (youtube), regular plugin fixes and a couple of plugin removals. Even if the commit log isn't that long yet, a minor release should be justified.

I'll be submitting the changelog tomorrow, so if there are further changes which should get included, post them.

### DIFF
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -1,8 +1,8 @@
 Deprecations
 ============
 
-2.3.0
------
+streamlink 2.3.0
+----------------
 
 Plugin.can_handle_url() and Plugin.priority()
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -84,8 +84,8 @@ message will be written to the info log output for the first plugin that gets re
    ``NoStreamsError`` when there are no streams to be found on that particular URL.
 
 
-2.2.0
------
+streamlink 2.2.0
+----------------
 
 Config file paths
 ^^^^^^^^^^^^^^^^^


back-to:
> It probably also makes sense distinguishing between streamlink and streamlink-cli on the deprecations page.

seems unnecessarily complicated, as we only use `streamlink` versions

bastimeyer:
Ok, let's simplify it.
But before I force-push, what is going on with the tests on Windows with py>=3.8 and with the Windows installer CI job?

back-to:
> what is going on with the tests on Windows with py>=3.8 and with

seems temporary

> Windows installer CI job?

`chardet` got removed from `requests`

https://github.com/psf/requests/pull/5797/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7

https://github.com/streamlink/streamlink/pull/3863/checks?check_run_id=3059726716#step:7:380

bastimeyer:
Sorry for the delay. I didn't have internet access for the past week due to the floods in my city/town. Rebased and ready for merging.

PR #3864, opened 2021-07-13
vendor: set requests to >=2.26.0,<3

bastimeyer:
- Bump requests to latest version:
  This replaces transitive dependency chardet with charset-normalizer
- Fix Windows installer dependencies:
  List dependencies as wheels with strict versions where possible.
  pynsist unfortunately doesn't support additional wheel checksums.

----

This pinpoints the dependency versions to strict values (all up2date), similar to the appimages, except for wheel checksums ("dependency lockfile"), as it's not supported by pynsist:
https://github.com/streamlink/streamlink-appimage/blob/2.2.0-1/config.json

The `iso639` dependency is not a wheel, so it has to be kept in the `packages` directive.

Tested a locally built installer in my Win10 VM...

----

#3862, #3863

### DIFF
--- a/script/makeinstaller.sh
+++ b/script/makeinstaller.sh
@@ -89,36 +89,19 @@ version=${STREAMLINK_PYTHON_VERSION}
 format=bundled
 
 [Include]
-; dep tree
-;   streamlink+streamlink_cli
-;       - pkg-resources (indirect)
-;           - pyparsing
-;           - packaging
-;           - six
-;       - iso639
-;       - iso3166
-;       - pycryptodome
-;       - requests
-;           - certifi
-;           - idna
-;           - urllib3
-;           - socks / sockshandler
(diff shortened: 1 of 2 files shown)


bastimeyer:
Why are the tests on Windows randomly failing all of a sudden?
`TestHLSStreamEncrypted.test_hls_encrypted_aes128_incorrect_padding_length` again...

Could this be caused by a side effect of #3856?

bastimeyer:
We've got two separate issues here. While this fixes the Windows installer CI job, tests on Windows are still randomly failing. The PR can't be merged due to branch protection rules. I'll restart the CI jobs now in hope that the required jobs pass and it can be merged. Otherwise we'll have to disable the protection rules, or I'll have to open a PR with both fixes (which would not be ideal).

Issue #3868, opened 2021-07-14
HLS tests randomly failing on Windows due to threading issues


bastimeyer:
### Checklist

- [X] This is a bug report and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest build from the master branch

### Description

This only started happening yesterday, and I still haven't understood why. I'm unable to reproduce this locally and it's only failing in the CI jobs. There haven't been any python or dependency upgrades prior to this, so this must be caused by an upgrade of the Windows CI environment, I think.
https://github.com/streamlink/streamlink/actions

----

For some reason, the following tests (so far) have been randomly failing due to the HLS stream's writer thread not writing the segments correctly. What is weird though is that these tests are supposed to be deterministic with lots of thread locks for writing and reading each segment, so there shouldn't be any timing issues between the threads.
- [`tests/streams/test_hls_filtered.py::TestFilteredHLSStream::test_filtered_closed`](https://github.com/streamlink/streamlink/blob/d04767fa5b812a4e971e454226cecb25b4d975ae/tests/streams/test_hls_filtered.py#L133-L167)
- [`tests/streams/test_hls_filtered.py::TestFilteredHLSStream::test_filtered_logging`](https://github.com/streamlink/streamlink/blob/d04767fa5b812a4e971e454226cecb25b4d975ae/tests/streams/test_hls_filtered.py#L57-L88)
- [`tests/streams/test_hls.py::TestHLSStreamEncrypted::test_hls_encrypted_aes128_incorrect_padding_length`](https://github.com/streamlink/streamlink/blob/d04767fa5b812a4e971e454226cecb25b4d975ae/tests/streams/test_hls.py#L243-L253)

----

Take a look at the [`test_filtered_closed`](https://github.com/streamlink/streamlink/blob/d04767fa5b812a4e971e454226cecb25b4d975ae/tests/streams/test_hls_filtered.py#L133-L167) integration test for example, which is the most simple one to understand. It sets up an HLS stream with only two segments which should all get filtered out. Ignore the additional `wait` mock, as it is not important for the issue.

[...]

PR #3869, opened 2021-07-15
tests: fix unnecessary hostname lookup in cli_main

bastimeyer:
I currently have limited internet access due to massive floods in my region :cloud_with_rain:. I've tried fixing #3868, but it's more complicated than I thought. We'll probably have to disable these affected tests until this is fixed.

This PR however fixes the tests while being offline.

### DIFF
--- a/tests/test_cli_main.py
+++ b/tests/test_cli_main.py
@@ -475,14 +475,14 @@ class TestCLIMainLogging(_TestCLIMainLogging):
             "-p", "custom",
             "--testplugin-bool",
             "--testplugin-password=secret",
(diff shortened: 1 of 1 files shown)


gravyboat:
> I currently have limited internet access due to massive floods in my region 🌧️.

Glad to hear that you are safe! I'll get this PR and the other one merged in. Let me know if there's anything I can do to help you with the flood situation.

bastimeyer:
Thanks. It's all good, at least where I live. Just no cable internet connection atm and weak mobile connectivity (as usual) :de: :-1:

Issue #3875, opened 2021-07-17
plugins.nimotv: Unable to validate response text


Trung0246:
### Checklist

- [X] This is a plugin issue and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

Hi, upon running streamlink 2.2.0 version on a nimo.tv livestream link I receive this:

![image](https://user-images.githubusercontent.com/11626920/126048360-4cc1a31e-96fa-4b52-8d7c-cb9e2c0e6df1.png)

### Debug log

```text
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.8.7
[cli][debug] Streamlink: 2.2.0
[cli][debug] Requests(2.25.1), Socks(1.7.1), Websocket(1.1.0)
[cli][debug] Arguments:
[cli][debug]  url=https://www.nimo.tv/live/3749744096
[cli][debug]  --loglevel=debug
[cli][debug]  --player="C:\Program Files\VideoLAN\VLC\vlc-gpu.exe" --file-caching=0 --network-caching=150 --clock-jitter=0 --clock-synchro=0 --repeat
[cli][debug]  --rtmp-rtmpdump=rtmpdump.exe
[cli][debug]  --rtmpdump=C:\Program Files (Portable & CMD)\streamlink\rtmpdump\rtmpdump.exe
[cli][debug]  --ffmpeg-ffmpeg=C:\Program Files (Portable & CMD)\ffmpeg\bin\ffmpeg.exe
[cli][info] Found matching plugin nimotv for URL https://www.nimo.tv/live/3749744096
[...]

PR #3882, opened 2021-07-24
plugins.nimotv: use 'mStreamPkg'

back-to:
using regex seems easier than https://github.com/streamlink/streamlink/issues/3875#issuecomment-882002731

might break at some point, but should work for now

closes https://github.com/streamlink/streamlink/issues/3875

---

@Trung0246 could you test it

### DIFF
--- a/src/streamlink/plugins/nimotv.py
+++ b/src/streamlink/plugins/nimotv.py
@@ -10,7 +10,7 @@ log = logging.getLogger(__name__)
 
 
 @pluginmatcher(re.compile(
-    r'https?://(?:www\.)?nimo\.tv/(?P<username>.*)'
+    r'https?://(?:www\.|m\.)?nimo\.tv/(?P<username>.*)'
 ))
 class NimoTV(Plugin):
     data_url = 'https://m.nimo.tv/{0}'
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/nimotv.py
```diff
@@ -46,21 +38,23 @@ class NimoTV(Plugin):
...
         1000: '480p',
-        2000: '720p',
+        2500: '720p',
         6000: '1080p',
     }
 
+    _re_appid = re.compile(br'appid=(\d+)')
+    _re_domain = re.compile(br'(https?:\/\/[A-Za-z]{2,3}.hls[A-Za-z\.\/]+)(?:V|&)')
+    _re_id = re.compile(br'id=([^|\\]+)')
+    _re_tp = re.compile(br'tp=(\d+)')
```
bastimeyer, line 48: These regexes should be good enough, even if it's not really possible to set proper anchors for extracting the data, as the anchors depend on the type of the preceding and following item in the JceStruct from the mStreamPkg data.

Since we're extracting strings from a JceStruct here, the two (?) bytes prior to the matching string value could be read and then decoded for the actual data type and length, so that the strings can be read precisely. That would make this approach a bit more proof without having to decode everything, but that's probably unnecessary and too complex.

Unless there's something wrong which I didn't notice, I'd say we should get this merged.

bastimeyer (approved):

gravyboat:
Thanks @back-to.

PR #3885, opened 2021-07-25
streams.hls: fix playlist_reload_time

### LINKED ISSUE #3884: hls.py: Exception in _playlist_reload_time() when hls-playlist-reload-time is "segment" or "live-edge" and sequences is empty
### Checklist

### Streamlink version

### Description

An exception of `TypeError: '>' not supported between instances of 'str' and 'int'` occurs here on line 273:
https://github.com/streamlink/streamlink/blob/193c5c6eb9147a56df25de160f4d35d6f230733a/src/streamlink/stream/hls.py#L268-L280

The first 2 `if` statements depend on `sequences` to not be empty, and the third `if` statement makes the assumption that `self.playlist_reload_time_override` is a numeric type.

https://github.com/streamlink/streamlink/blob/193c5c6eb9147a56df25de160f4d35d6f230733a/src/streamlink/stream/hls.py#L229-L232
We can see here `self.playlist_reload_time_override` is only guaranteed to be a numeric type when `hls-playlist-reload-time` is neither `segment` nor `live-edge`.

So when `hls-playlist-reload-time` is either `segment` or `live-edge`, and `sequences` is empty, `self.playlist_reload_time_override` will be a non-numeric type and an exception will occur when trying to do a comparison against an `int` in the third `if` statement.

A fix for this would be to verify `self.playlist_reload_time_override` is a numeric type in the third `if` statement.
```diff
-        if self.playlist_reload_time_override > 0:
+        if isinstance(self.playlist_reload_time_override, (float, int)) and self.playlist_reload_time_override > 0:
```

### Debug log

```text
[cli][debug] OS:         Linux-5.13.4-arch2-1-x86_64-with-glibc2.33
[cli][debug] Python:     3.9.6
[cli][debug] Streamlink: 2.2.0+46.g193c5c6
[cli][debug] Requests(2.26.0), Socks(1.7.1), Websocket(0.59.0)
[cli][debug] Arguments:
[cli][debug]  url=https://hakk.in/crash.m3u8
[cli][debug]  stream=['best']
[cli][debug]  --hls-playlist-reload-time=segment
[cli][info] Found matching plugin hls for URL https://hakk.in/crash.m3u8
[plugins.hls][debug] URL=https://hakk.in/crash.m3u8; params={}
[cli][info] Opening stream: live (hls)
Exception in thread Thread-HLSStreamWorker:
  File "/usr/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "/home/hakkin/.local/lib/python3.9/site-packages/streamlink/stream/segmented.py", line 84, in run
    for segment in self.iter_segments():
  File "/home/hakkin/.local/lib/python3.9/site-packages/streamlink/stream/hls.py", line 325, in iter_segments
    self.reload_playlist()
  File "/home/hakkin/.local/lib/python3.9/site-packages/streamlink/stream/hls.py", line 263, in reload_playlist
    self.playlist_reload_time = self._playlist_reload_time(playlist, sequences)
  File "/home/hakkin/.local/lib/python3.9/site-packages/streamlink/stream/hls.py", line 273, in _playlist_reload_time
    if self.playlist_reload_time_override > 0:
TypeError: '>' not supported between instances of 'str' and 'int'
[cli][error] Try 1/1: Could not open stream <HLSStream('https://hakk.in/crash.m3u8', None)> (Failed to read data from stream: Read timeout)
error: Could not open stream <HLSStream('https://hakk.in/crash.m3u8', None)>, tried 1 times, exiting
```

bastimeyer:
If hls-playlist-reload-time is set to "segment" or "live-edge" and
no segments are available, return the playlist's targetduration
instead of continuing with checking other input values.

Make immediately starting the HLS stream and reader thread optional
in TestMixinStreamHLS and fix subclasses of HLSStream (eg. Twitch)
which have custom open() logic due to the change of starting logic,
so that certain methods on the HLS classes can be mocked.

----

Fixes #3884 

```
$ streamlink -l debug --hls-playlist-reload-time=segment 'https://hakk.in/crash.m3u8' best
[cli][debug] OS:         Linux-5.13.1-2-git-x86_64-with-glibc2.33
[cli][debug] Python:     3.9.6
[cli][debug] Streamlink: 2.2.0+47.gf57fea8
[cli][debug] Requests(2.26.0), Socks(1.7.1), Websocket(1.1.0)
[cli][debug] Arguments:
[cli][debug]  url=https://hakk.in/crash.m3u8
[cli][debug]  stream=['best']
[cli][debug]  --player=mpv
[cli][debug]  --hls-playlist-reload-time=segment
[cli][info] Found matching plugin hls for URL https://hakk.in/crash.m3u8
[plugins.hls][debug] URL=https://hakk.in/crash.m3u8; params={}
[cli][info] Opening stream: live (hls)
^CInterrupted! Exiting...
```

```
$ streamlink -l debug --hls-playlist-reload-time=live-edge 'https://hakk.in/crash.m3u8' best
[cli][debug] OS:         Linux-5.13.1-2-git-x86_64-with-glibc2.33
[cli][debug] Python:     3.9.6
[cli][debug] Streamlink: 2.2.0+47.gf57fea8
[cli][debug] Requests(2.26.0), Socks(1.7.1), Websocket(1.1.0)
[cli][debug] Arguments:
[cli][debug]  url=https://hakk.in/crash.m3u8
[cli][debug]  stream=['best']
[cli][debug]  --player=mpv
[cli][debug]  --hls-playlist-reload-time=live-edge
[cli][info] Found matching plugin hls for URL https://hakk.in/crash.m3u8
[plugins.hls][debug] URL=https://hakk.in/crash.m3u8; params={}
[cli][info] Opening stream: live (hls)
^CInterrupted! Exiting...
```

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -270,7 +270,7 @@ class HLSStreamWorker(SegmentedStreamWorker):
             return sequences[-1].segment.duration
         if self.playlist_reload_time_override == "live-edge" and sequences:
             return sum([s.segment.duration for s in sequences[-max(1, self.live_edge - 1):]])
-        if self.playlist_reload_time_override > 0:
+        if type(self.playlist_reload_time_override) is float and self.playlist_reload_time_override > 0:
             return self.playlist_reload_time_override
         if playlist.target_duration:
             return playlist.target_duration
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -118,6 +118,16 @@ class TwitchHLSStreamReader(HLSStreamReader):
     __worker__ = TwitchHLSStreamWorker
     __writer__ = TwitchHLSStreamWriter
 
+    def __init__(self, stream):
+        if stream.disable_ads:
+            log.info("Will skip ad segments")
+        if stream.low_latency:
+            live_edge = max(1, min(LOW_LATENCY_MAX_LIVE_EDGE, stream.session.options.get("hls-live-edge")))
+            stream.session.options.set("hls-live-edge", live_edge)
+            stream.session.options.set("hls-segment-stream-data", True)
+            log.info(f"Low latency streaming (HLS live edge: {live_edge})")
+        super().__init__(stream)
+
 
 class TwitchHLSStream(HLSStream):
     __reader__ = TwitchHLSStreamReader
@@ -126,17 +136,6 @@ class TwitchHLSStream(HLSStream):
         super().__init__(*args, **kwargs)
         self.disable_ads = self.session.get_plugin_option("twitch", "disable-ads")
         self.low_latency = self.session.get_plugin_option("twitch", "low-latency")
-
-    def open(self):
-        if self.disable_ads:
-            log.info("Will skip ad segments")
-        if self.low_latency:
-            live_edge = max(1, min(LOW_LATENCY_MAX_LIVE_EDGE, self.session.options.get("hls-live-edge")))
-            self.session.options.set("hls-live-edge", live_edge)
-            self.session.options.set("hls-segment-stream-data", True)
-            log.info("Low latency streaming (HLS live edge: {0})".format(live_edge))
-
-        return super().open()
 
     @classmethod
     def _get_variant_playlist(cls, res):
--- a/tests/mixins/stream_hls.py
+++ b/tests/mixins/stream_hls.py
@@ -137,7 +137,7 @@ class HLSStreamReadThread(Thread):
 
         self.session = session
         self.stream = stream
-        self.reader = stream.open()
+        self.reader = stream.__reader__(stream)
 
         # ensure that at least one read was attempted before closing the writer thread early
         # otherwise, the writer will close the reader's buffer, making it not block on read and yielding empty results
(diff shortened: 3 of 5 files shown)


src/streamlink/stream/hls.py
```diff
@@ -266,9 +266,13 @@ def reload_playlist(self):
             self.process_sequences(playlist, sequences)
 
     def _playlist_reload_time(self, playlist: M3U8, sequences: List[Sequence]) -> float:
-        if self.playlist_reload_time_override == "segment" and sequences:
+        if self.playlist_reload_time_override == "segment":
+            if not sequences:
+                return playlist.target_duration
```
back-to, line 271: shouldn't we check if `playlist.target_duration` is not **None** and use `self.playlist_reload_time` if it is None
like it is under this

---

also `return sum([s.segment.duration for s in sequences[-max(1, self.live_edge - 1):]])` from `sequences` is a the same as `"live-edge"` basically a duplicate

bastimeyer: I was trying to avoid the is-float type check which would be required without the `and sequences` condition.

> `playlist.target_duration` is not **None**

This is technically a greater than zero comparison as [the target duration can't be None](https://github.com/streamlink/streamlink/blob/193c5c6eb9147a56df25de160f4d35d6f230733a/src/streamlink/stream/hls_playlist.py#L264) unless the HLS playlist is invalid. Even checking for greater than zero is rather unnecessary as a value of zero would imply that every segment must be at most zero seconds long, which doesn't make sense.

https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.3.1

----

> duplicate

That can't be avoided because of the different priorities of these checks. The second statement basically means "use the `live-edge` method if everything else fails". And the static value of 15, which is the default value (not sure why it was defined like this), is returned when there are no segments available.

PR #3893, opened 2021-07-30
session: deprecate options for spec. stream types

bastimeyer:
Having multiple session options and CLI arguments for different stream
types which are already covered by generic options/arguments is not only
redundant, but also confusing.

A distinction between different stream types does only make sense when
multiple different stream types are available and the user needs to
explicitly set different values for each of them, but since it's not
always clear which stream type is returned when a stream is selected
via the "best" stream name synonym for example, this makes it even more
confusing. And it's redundant as well, since only one stream can be
selected anyway.

- deprecate `{dash,hds,hls}-segment-attempts`
  in favor of `stream-segment-attempts`
- deprecate `{dash,hds,hls}-segment-threads`
  in favor of `stream-segment-threads`
- deprecate `{dash,hds,hls}-segment-timeout`
  in favor of `stream-segment-timeout`
- deprecate `{dash,hds,hls,rtmp,http-stream}-timeout`
  in favor of `stream-timeout` (dash/http-stream were never used)
- fix `HTTPStream` and use `stream-timeout` instead of `http-timeout`
- suppress deprecated CLI arguments
- update help texts of generic CLI arguments
- fix docs and add entry to deprecations page

----

1. Do we need tests for the option/argument name translations?
2. The "Stream transport options" section in the docs needs to be reordered/improved. In a separate PR.
3. `hls-segment-stream-data` should be removed as well and just set to `True`. In a separate PR.

### DIFF
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -39,24 +39,14 @@ class Streamlink:
             "ipv4": False,
             "ipv6": False,
             "hds-live-edge": 10.0,
-            "hds-segment-attempts": 3,
-            "hds-segment-threads": 1,
-            "hds-segment-timeout": 10.0,
-            "hds-timeout": 60.0,
             "hls-live-edge": 3,
-            "hls-segment-attempts": 3,
             "hls-segment-ignore-names": [],
-            "hls-segment-threads": 1,
-            "hls-segment-timeout": 10.0,
             "hls-segment-stream-data": False,
-            "hls-timeout": 60.0,
             "hls-playlist-reload-attempts": 3,
             "hls-playlist-reload-time": "default",
             "hls-start-offset": 0,
             "hls-duration": None,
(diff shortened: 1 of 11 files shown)


src/streamlink/session.py
```diff
@@ -317,6 +269,22 @@ def set_option(self, key, value):
             self.http.cert = value
         elif key == "http-timeout":
             self.http.timeout = value
+
+        # deprecated: {dash,hds,hls}-segment-attempts
+        elif key == "dash-segment-attempts" or key == "hds-segment-attempts" or key == "hls-segment-attempts":
```
back-to, line 274:
```python
elif key in ("dash-segment-attempts", "hds-segment-attempts", "hls-segment-attempts"):
```

might be more readable

back-to (approved):

PR #3894, opened 2021-07-31
stream.hls: remove hls-segment-stream-data option

bastimeyer:
- always stream data from unencrypted segment downloads
- never stream data from cached initialization sections, as the
  response content can only be consumed once when stream=True
- update --twitch-low-latency plugin argument and its description

----

This also fixes a bug with the previous media initialization sections (segment maps) implementation, which raises an exception when stream is set to True, as the content can only be consumed once when iterating and writing chunks to the buffer.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -781,13 +781,7 @@ def build_parser():
         Default is 3.
         """
     )
-    transport.add_argument(
-        "--hls-segment-stream-data",
-        action="store_true",
-        help="""
-        Immediately write segment data into output buffer while downloading.
-        """
-    )
+    transport.add_argument("--hls-segment-stream-data", action="store_true", help=argparse.SUPPRESS)
     transport.add_argument(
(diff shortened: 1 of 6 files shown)


src/streamlink_cli/argparser.py
```diff
@@ -781,13 +781,7 @@ def build_parser():
...
         """
     )
-    transport.add_argument(
-        "--hls-segment-stream-data",
-        action="store_true",
-        help="""
-        Immediately write segment data into output buffer while downloading.
-        """
-    )
+    transport.add_argument("--hls-segment-stream-data", help=argparse.SUPPRESS)
```
back-to, line 784: you still need `action="store_true"`
otherwiser it will break userscripts/configs

---

cli

```
streamlink: error: argument --hls-segment-stream-data: expected one argument
```
config file

```
[cli][debug]  --hls-segment-stream-data=--ffmpeg-ffmpeg=/usr/bin/ffmpeg
```

PR #3906, opened 2021-08-07
plugins.zattoo: fix HLS stream, added more debug details

back-to:
- use hls7 instead of hls5
- added more log details with validate.Schema


gravyboat:
Thanks @back-to.

PR #3908, opened 2021-08-08
tests.hls: fix playlist reload time tests

bastimeyer:
This should fix https://github.com/streamlink/streamlink/runs/3270358311#step:8:229 by waiting for the worker thread to terminate first after the entire stream has ended.

The problem with the previous test setup (introduced by be40dd4 in #3883) was that `await_write` didn't wait in any of the empty playlist tests, which meant that the main thread could continue while the worker hadn't processed the reload time yet.

### DIFF
--- a/tests/streams/test_hls.py
+++ b/tests/streams/test_hls.py
@@ -274,8 +274,6 @@ class TestHLSStreamEncrypted(TestMixinStreamHLS, unittest.TestCase):
 
 @patch("streamlink.stream.hls.HLSStreamWorker.wait", Mock(return_value=True))
 class TestHlsPlaylistReloadTime(TestMixinStreamHLS, unittest.TestCase):
-    __stream__ = EventedHLSStream
-
     segments = [
         Segment(0, duration=11),
         Segment(1, duration=7),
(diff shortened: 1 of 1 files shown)


back-to:
`requests>=2.26.0` could be used again or?

https://github.com/streamlink/streamlink/commit/a1fd4edee6c1f07f2ac7f4f1b010e16604abb0fd

---

_also a  side effect of this PR_

I got a VM with Windows which is **very slow**
and the test will always fail with a timeout because the machines is just to slow.

```
FAILED tests/streams/test_hls.py::TestHLSStreamEncrypted::test_hls_encrypted_aes128 - RuntimeError: Await write timeout: write_call=1
FAILED tests/streams/test_hls.py::TestHLSStreamEncrypted::test_hls_encrypted_aes128_incorrect_padding_content - RuntimeError: Await write time
FAILED tests/streams/test_hls.py::TestHLSStreamEncrypted::test_hls_encrypted_aes128_key_uri_override - RuntimeError: Await write timeout: writ
FAILED tests/streams/test_hls.py::TestHLSStreamEncrypted::test_hls_encrypted_aes128_with_map - RuntimeError: Await write timeout: write_call=5
```

bastimeyer:
> `requests>=2.26.0` could be used again or?

No, this is still not resolved.

> also a side effect of this PR

You've posted different, unrelated tests

bastimeyer:
> which is **very slow**
> and the test will always fail with a timeout

There is still a concurrency issue left somewhere in those tests which are using the EventedHLSStreamWriter, but I haven't managed to find it yet.

The entire HLS/Segmented stream implementation is very difficult to properly test, because there are multiple threads which depend on each other's results while one of them is polling the segment queue. The polled segment queue is not a particularly good design, and rewriting it while keeping compatibility and not breaking other stream type implementations is rather difficult.

PR #3909, opened 2021-08-09
plugins.nbcnews: fix stream URL extraction

Grabien:
The NBCNews plugin stopped working recently after updates of the website. This is a fix to this issue.

### DIFF
--- a/src/streamlink/plugins/nbcnews.py
+++ b/src/streamlink/plugins/nbcnews.py
@@ -13,8 +13,9 @@ log = logging.getLogger(__name__)
     r'https?://(?:www\.)?nbcnews\.com/now'
 ))
 class NBCNews(Plugin):
+    url_re = re.compile(r'https?://(?:www\.)?nbcnews\.com/now')
     json_data_re = re.compile(
-        r'<script id="__NEXT_DATA__" type="application/json">({.*})</script>'
+        r'<script type="application/ld+json">({.*?})</script>'
     )
     api_url = 'https://stream.nbcnews.com/data/live_sources_{}.json'
     token_url = 'https://tokens.playmakerservices.com/'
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/nbcnews.py
```diff
@@ -67,8 +44,16 @@ class NBCNews(Plugin):
     def get_title(self):
         return 'NBC News Now'
 
+    def _get_video_id(self, site_url):
+        maininfo_res = self.session.http.get(site_url)
+        video_info_json = re.findall(r'<script type="application/ld\+json">(.*?)</script>', maininfo_res.text)[0]
+        video_info_json = parse_json(video_info_json)
+        embedUrl = video_info_json["embedUrl"]
+        video_id = embedUrl.split("/")[-1]
+        return video_id
```
bastimeyer, line 53: This should be done in the validation schema, just like before, because it properly handles non-matching content via `validate.any(None, ...)` and thus doesn't raise any exceptions. You probably also meant to use the regex from the plugin's class attribute.

```py
    json_data_schema = validate.Schema(
        validate.transform(json_data_re.search),
        validate.any(None, validate.all(
            validate.get(1),
            validate.transform(parse_json),
            {"embedUrl": validate.url()},
            validate.get("embedUrl"),
            validation.transform(lambda url: url.split("/")[-1])
        ))
    )
```

The plugin's `_get_streams` should ideally also check if `video_id is None` and return in this case, so that it doesn't continue with missing data.

```py
    video_id = self.session.http.get(self.url, schema=self.json_data_schema)
    if video_id is None:
        return
```

bastimeyer (changes requested):
Thanks for the PR.
There are a couple of changes though which should be applied before this can be merged and the plugin be fixed.

PR #3910, opened 2021-08-10
stream.hls: close stream on initial parsing error

bastimeyer:
- close stream immediately when the first playlist fails to parse
- add tests for all playlist parsing errors in the worker thread

----

Follow up of #3903


gravyboat:

PR #3913, opened 2021-08-13
Support -t/--title formatting vars in -o/--output

### LINKED ISSUE #3898: Option to add formatting variables for output filename
### Checklist

### Description

**What is it?** 

mkbloke:
I thought I'd have a go at this, as it's been requested/mentioned a few times...

Things to note:

This adds a new dependency: [pathvalidate](https://github.com/thombashi/pathvalidate).  I think this is probably a worthwhile dep to add to ensure compatible and consistent filenames across supported platforms and there's little point in trying to reimplement it locally, I would have thought.  The pathvalidate module seems to be reasonably well maintained.

I've chosen to use pathvalidate in universal mode as this will result in consistent filenames across platforms (although that does impose undue limitations - e.g. on Linux), which seems a reasonable way to go.  One result of this is that filename length is truncated beyond 255 characters.

I've chosen to use underscore (`_`) as a replacement character rather than the default of just removing invalid characters, as it seems like it might be a bit more obvious as to what is going on.

closes #3898

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -588,6 +588,10 @@ def build_parser():
         Write stream data to FILENAME instead of playing it.
 
         You will be prompted if the file already exists.
+
+        The formatting variables available for the -t/--title option may be used.
+        Invalid characters (cross platform) are replaced with an underscore.
+        Filenames are truncated to 255 characters (cross platform limit).
         """
     )
     output.add_argument(
@@ -619,6 +623,10 @@ def build_parser():
         Open the stream in the player, while at the same time writing it to FILENAME.
 
         You will be prompted if the file already exists.
+
+        The formatting variables available for the -t/--title option may be used.
+        Invalid characters (cross platform) are replaced with an underscore.
+        Filenames are truncated to 255 characters (cross platform limit).
         """
     )
     output.add_argument(
@@ -628,6 +636,10 @@ def build_parser():
         Write stream data to stdout, while at the same time writing it to FILENAME.
 
         You will be prompted if the file already exists.
+
+        The formatting variables available for the -t/--title option may be used.
+        Invalid characters (cross platform) are replaced with an underscore.
+        Filenames are truncated to 255 characters (cross platform limit).
         """
     )
 
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -16,6 +16,7 @@ from pathlib import Path
 from time import sleep
 from typing import List
 
+import pathvalidate
 import requests
 from socks import __version__ as socks_version
 from websocket import __version__ as websocket_version
@@ -84,10 +85,12 @@ def create_output(plugin):
         if args.output == "-":
             out = FileOutput(fd=stdout)
         else:
+            args.output = create_filename(plugin, args.output)
             out = check_file_output(args.output, args.force)
     elif args.stdout:
         out = FileOutput(fd=stdout)
     elif args.record_and_pipe:
+        args.record = args.record_and_pipe = create_filename(plugin, args.record_and_pipe)
         record = check_file_output(args.record_and_pipe, args.force)
         out = FileOutput(fd=stdout, record=record)
     else:
@@ -109,6 +112,7 @@ def create_output(plugin):
         title = create_title(plugin)
 
         if args.record:
+            args.record = create_filename(plugin, args.record)
             record = check_file_output(args.record, args.force)
 
         log.info("Starting player: {0}".format(args.player))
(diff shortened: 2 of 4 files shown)


src/streamlink_cli/main.py
```diff
@@ -138,19 +142,35 @@ def create_http_server(*_args, **_kwargs):
...
+
+def create_filename(plugin=None, unformatted_str=None):
+    if plugin and unformatted_str and unformatted_str != "-":
+        try:
+            return pathvalidate.sanitize_filename(
+                apply_format_vars(plugin, unformatted_str),
+                replacement_text="_"
+            )
+        except pathvalidate.ValidationError as err:
+            console.exit("Failed to sanitize output filename: {0}", err)
```
bastimeyer, line 164: > pathvalidate.sanitize_filename

According to the docs, this also replaces `/` with the replacement character, which means it's not possible anymore to choose different directories. This needs to be changed to `sanitize_filepath`:
https://github.com/thombashi/pathvalidate#sanitize-a-filename
https://github.com/thombashi/pathvalidate#sanitize-a-filepath

This does keep characters like `:` on non-Windows OSes, right?
https://github.com/thombashi/pathvalidate/blob/v2.4.1/pathvalidate/_filename.py#L85-L89
https://github.com/thombashi/pathvalidate/blob/v2.4.1/pathvalidate/_base.py#L14-L17

src/streamlink_cli/argparser.py
```diff
@@ -588,6 +588,10 @@ def build_parser():
         Write stream data to FILENAME instead of playing it.
 
         You will be prompted if the file already exists.
+
+        The formatting variables available for the -t/--title option may be used.
+        Invalid characters (cross platform) are replaced with an underscore.
+        Filenames are truncated to 255 characters (cross platform limit).
```
bastimeyer, line 594: > `-t/--title`

There need to be spaces around the slash, otherwise this won't be turned into a link/anchor in the rendered HTML docs.

> 255 characters (cross platform limit)

If `max_len` is set to `None`, then it'll use platform specific defaults instead of 255:
- https://github.com/thombashi/pathvalidate/blob/v2.4.1/pathvalidate/_base.py#L52-L60
- https://github.com/thombashi/pathvalidate/blob/v2.4.1/pathvalidate/_base.py#L85-L95

bastimeyer (commented):
The pathvalidate lib seems reasonable for the purpose of these changes.

Before this can be merged, tests need to be added.

File name extensions are not covered by this. It may be worth adding, but I'm not sure yet what the best way of doing that would be. But let's ignore this for now.

src/streamlink_cli/argparser.py
bastimeyer: Removing the shorthand `-t` parameter is probably better.

back-to:
same as https://github.com/streamlink/streamlink/pull/2534

unnecessary dependencies are stupid

src/streamlink_cli/main.py
```diff
@@ -84,10 +85,12 @@ def create_output(plugin):
         if args.output == "-":
             out = FileOutput(fd=stdout)
         else:
+            args.output = create_filename(plugin, args.output)
```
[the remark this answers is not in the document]
mkbloke: I did wonder if doing that was a good idea or not...

I'll put the formatted filenames in new vars and change things around.

src/streamlink_cli/argparser.py
mkbloke: > If max_len is set to None, then it'll use platform specific defaults instead of 255:

That's good to know, I had not seen that in the docs.

mkbloke:
> unnecessary dependencies are stupid

So, abandon the feature or do it without a dep?  If the latter, I'm open to suggestions as to how to go about that without reimplementing pathvalidate within Streamlink.  I guess a simple way would be to try and open a file with the given format variable substitutions and catch any errors.  If it fails, tough luck for the user?

src/streamlink_cli/main.py
mkbloke: > According to the docs, this also replaces / with the replacement character, which means it's not possible anymore to choose different directories. This needs to be changed to sanitize_filepath:

Ah, yes, you're absolutely right!

> This does keep characters like : on non-Windows OSes, right?

Not in universal mode.  If you think it's better to keep characters like : for an OS that can support them in filenames, then that be done by running pathvalidate in auto mode instead.  The side effect of that would be different filenames for the same substitution vars depending on the OS used.  I don't know if that matters or if consistent filenames are better?

bastimeyer:
Just add a simple method to `streamlink_cli.compat` which replaces characters in a path string, similar to https://github.com/thombashi/pathvalidate/blob/v2.4.1/pathvalidate/_base.py#L14-L17 and add unit tests. Errors when trying to write need to be caught either way.

Btw, most of these characters like colons for example are supported on NTFS file systems, it's just Windows (or rather Windows explorer) which doesn't let the user access those files. The only thing that definitely needs to be replaced are path separators, but only in the substituted parts, not the entire string.

mkbloke:
I've had a re-think on the implementation of this feature,  I'm going to close this PR and open a new one to keep the discussion clean.

PR #3917, opened 2021-08-16
installer: upgrade to python 3.8.10

bastimeyer:
https://github.com/python/cpython/tags


bastimeyer:
Didn't notice that it failed to build... :confused: 

For some reason, there's no pre-built embed-version of 3.8.11 available:
- https://www.python.org/downloads/release/python-3811/
- https://www.python.org/downloads/release/python-3810/

bastimeyer:
> Python 3.8 isn't receiving regular bugfixes anymore, and binary installers are no longer provided for it.

A good reason to switch from 3.8 to 3.9 (or 3.10 in October when it gets released).

gravyboat:
@bastimeyer Do you want to deprecate 3.8 instead?

bastimeyer:
beardypig already suggested that we should drop 3.8 when it leaves its bugfix release cycle:
https://github.com/streamlink/streamlink/issues/3323#issuecomment-724100817

We have three options now:
1. bump to 3.8.10 and wait for the next release before switching to 3.9
2. bump to 3.8.10 and wait for 3.10 to be released in October
3. close this PR and switch to 3.9 now (requires documentation)

bastimeyer:
Closing in favor of #3918

SekiBetu:
many packages does not support python3.9, I think you guys are a bit too eager

bastimeyer:
Wrong thread. This PR is about 3.8 and has been closed because the build failed due to 3.8.11 not being available as an official pre-built download from python.org (source-only security release cycle).
https://github.com/streamlink/streamlink/runs/3342182827?check_suite_focus=true#step:7:377

**Read this here:**
https://github.com/streamlink/streamlink/pull/3918#issue-714354603

> many packages does not support python3.9, I think you guys are a bit too eager

Streamlink's Windows installer is a standalone application. Anything third party is irrelevant.

**Installations via setup.py and the published wheel on PyPI are still supported on 3.6, as you can clearly see here:**
https://github.com/streamlink/streamlink/blob/2.4.0/setup.py#L107

PR #3932, opened 2021-08-22
plugins.huomao: plugin removal

mkbloke:
This seems to be dead.  There are pingable (different) hosts for huomao.com and huomao.tv, but neither seem to respond on ports 80 or 443.  It has been like this for at least a week, but likely much longer.


bastimeyer:
https://en.wikipedia.org/wiki/List_of_streaming_media_services#Former_services

bastimeyer:
Thanks

PR #3933, opened 2021-08-22
cli: support -t formatting vars in -o, -r and -R

### LINKED ISSUE #3898: Option to add formatting variables for output filename

mkbloke:
Second try...

closes #3898

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -588,6 +588,9 @@ def build_parser():
         Write stream data to FILENAME instead of playing it.
 
         You will be prompted if the file already exists.
+
+        The formatting variables available for the --title option may be used.
+        Invalid characters in formatting variables are replaced with an underscore.
         """
     )
     output.add_argument(
@@ -619,6 +622,9 @@ def build_parser():
         Open the stream in the player, while at the same time writing it to FILENAME.
 
         You will be prompted if the file already exists.
+
+        The formatting variables available for the --title option may be used.
+        Invalid characters in formatting variables are replaced with an underscore.
         """
     )
     output.add_argument(
@@ -628,6 +634,28 @@ def build_parser():
         Write stream data to stdout, while at the same time writing it to FILENAME.
 
         You will be prompted if the file already exists.
+
+        The formatting variables available for the --title option may be used.
+        Invalid characters in formatting variables are replaced with an underscore.
+        """
+    )
+    output.add_argument(
+        "--fs-safe-rules",
+        choices=["windows", "unix"],
+        type=str,
+        help="""
+        The rules used to make formatting variables filesystem-safe are chosen
+        automatically according to the type of system in use.  This overrides
+        the automatic detection.
+
+        Intended for use when Streamlink is running on a UNIX like OS but writing
+        to Windows filesystems such as NTFS; USB devices using VFAT or exFAT; CIFS
+        shares that are enforcing Windows filename limitations, etc.
+
+        These characters are replaced with an underscore for the rules in use:
+
+          "windows": 0x00-0x1F 0x7F " * / : < > ? \\ |
+          "unix":    0x00 /
         """
     )
 
--- a/src/streamlink_cli/compat.py
+++ b/src/streamlink_cli/compat.py
@@ -1,16 +1,40 @@
 import os
+import re
 import sys
 from pathlib import Path
-
 
 is_darwin = sys.platform == "darwin"
 is_win32 = os.name == "nt"
 
 stdout = sys.stdout.buffer
 
+unprintable_ascii_chars = (chr(c) for c in range(32))
+
+FS_SAFE_REPLACEMENT_CHAR = "_"
+INVALID_FILENAME_CHARS = {
+    "windows": "".join(unprintable_ascii_chars) + "\x7f\"*/:<>?\\|",
+    "unix": "\x00/"
+}
+
+
+def make_fs_safe(unsafe_str, rules=None):
+    if rules and rules in INVALID_FILENAME_CHARS:
+        current_rules = rules
+    else:
+        if is_win32:
+            current_rules = "windows"
+        else:
+            current_rules = "unix"
+
+    sub_re = re.compile(
+        "[{:s}]".format(re.escape(INVALID_FILENAME_CHARS[current_rules])), re.UNICODE
+    )
+
+    return sub_re.sub(FS_SAFE_REPLACEMENT_CHAR, unsafe_str)
+
 
 class DeprecatedPath(type(Path())):
     pass
 
 
-__all__ = ["is_darwin", "is_win32", "stdout", "DeprecatedPath"]
+__all__ = ["is_darwin", "is_win32", "stdout", "make_fs_safe", "DeprecatedPath"]
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -28,7 +28,7 @@ from streamlink.plugin import PluginOptions
 from streamlink.stream import StreamProcess
 from streamlink.utils import LazyFormatter, NamedPipe
 from streamlink_cli.argparser import build_parser
-from streamlink_cli.compat import DeprecatedPath, is_win32, stdout
+from streamlink_cli.compat import DeprecatedPath, is_win32, make_fs_safe, stdout
 from streamlink_cli.console import ConsoleOutput, ConsoleUserInputRequester
 from streamlink_cli.constants import CONFIG_FILES, DEFAULT_STREAM_METADATA, LOG_DIR, PLUGIN_DIRS, STREAM_SYNONYMS
 from streamlink_cli.output import FileOutput, PlayerOutput
@@ -41,7 +41,7 @@ except AttributeError:
     pass  # Not windows
 QUIET_OPTIONS = ("json", "stream_url", "subprocess_cmdline", "quiet")
 
-args = console = streamlink = plugin = stream_fd = output = None
+args = console = formatted_filename = streamlink = plugin = stream_fd = output = None
 
 log = logging.getLogger("streamlink.cli")
 
@@ -76,19 +76,25 @@ def create_output(plugin):
      - A regular file
 
     """
+    global formatted_filename
 
     if (args.output or args.stdout) and (args.record or args.record_and_pipe):
         console.exit("Cannot use record options with other file output options.")
+
+    for unformatted_filename in args.output, args.record, args.record_and_pipe:
+        if unformatted_filename and isinstance(unformatted_filename, str):
+            formatted_filename = create_filename(plugin, unformatted_filename)
+            break
 
     if args.output:
         if args.output == "-":
             out = FileOutput(fd=stdout)
         else:
-            out = check_file_output(args.output, args.force)
+            out = check_file_output(formatted_filename, args.force)
     elif args.stdout:
         out = FileOutput(fd=stdout)
     elif args.record_and_pipe:
-        record = check_file_output(args.record_and_pipe, args.force)
+        record = check_file_output(formatted_filename, args.force)
         out = FileOutput(fd=stdout, record=record)
     else:
         http = namedpipe = record = None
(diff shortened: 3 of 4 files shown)


src/streamlink_cli/argparser.py
```diff
@@ -628,6 +634,28 @@ def build_parser():
...
 
         You will be prompted if the file already exists.
+
+        The formatting variables available for the --title option may be used.
+        Invalid characters in formatting variables are replaced with an underscore.
+        """
+    )
+    output.add_argument(
+        "--fs-safe-rules",
+        choices=["windows", "unix"],
```
bastimeyer, line 644: `unix` should be renamed to `POSIX`

src/streamlink_cli/compat.py
```diff
@@ -1,16 +1,40 @@
...
+            current_rules = "windows"
+        else:
+            current_rules = "unix"
+
+    sub_re = re.compile(
+        "[{:s}]".format(re.escape(INVALID_FILENAME_CHARS[current_rules])), re.UNICODE
+    )
+
+    return sub_re.sub(FS_SAFE_REPLACEMENT_CHAR, unsafe_str)
+
```
bastimeyer, line 34: The `.compat` modules are only used for importing constants, methods and classes with a common name that depend on the OS. Implementation logic should be moved to a utils module.

src/streamlink_cli/main.py
```diff
@@ -76,19 +76,25 @@ def create_output(plugin):
...
+    global formatted_filename
 
     if (args.output or args.stdout) and (args.record or args.record_and_pipe):
         console.exit("Cannot use record options with other file output options.")
 
+    for unformatted_filename in args.output, args.record, args.record_and_pipe:
+        if unformatted_filename and isinstance(unformatted_filename, str):
+            formatted_filename = create_filename(plugin, unformatted_filename)
+            break
+
```
bastimeyer, line 88: What's the reason for this? Even if `args.output`, `args.record` and `args.record_and_pipe` are mutually exclusive, taking the name from the first one that exists doesn't make sense. Is it because of the global variable and the other usages of `args.output` and `args.record` down below?

src/streamlink_cli/main.py
```diff
@@ -138,19 +144,43 @@ def create_http_server(*_args, **_kwargs):
     return http
 
 
+def apply_format_vars(plugin, unformatted_str, fs_safe=False):
+    title = plugin.get_title() or DEFAULT_STREAM_METADATA["title"]
+    author = plugin.get_author() or DEFAULT_STREAM_METADATA["author"]
+    category = plugin.get_category() or DEFAULT_STREAM_METADATA["category"]
+    game = plugin.get_category() or DEFAULT_STREAM_METADATA["game"]
```
bastimeyer, line 151: You are now always computing the title, author and category metadata, category even twice, which was done lazily before. This will unnecessarily delay the stream start in some cases.

src/streamlink_cli/main.py
```diff
@@ -318,7 +348,7 @@ def output_stream(plugin, stream):
                          args.player, err)
         else:
             console.exit("Failed to open output: {0} ({1})",
-                         args.output, err)
+                         formatted_filename, err)
```
bastimeyer, line 351: Can be turned into an f-string.

tests/test_cli_main.py
```diff
@@ -202,6 +214,142 @@ def test_handle_url_with_json_and_stream_url(self, console, args):
...
             console.exit.mock_calls.clear()
 
+    def test_create_filename_invalid_chars_native_rules_auto(self):
+        streamlink_cli.main.args = args = Mock()
+        args.fs_safe_rules = None
+
+        if is_win32:
+            test_chars = INVALID_FILENAME_CHARS["windows"]
+        else:
+            test_chars = INVALID_FILENAME_CHARS["unix"]
```
bastimeyer, line 224: Don't branch in each individual test and create OS-specific tests instead with `skip` decorators. Tests should include as few branches as possible, hence the 100% coverage requirement. It's probably also a good idea to create special test classes for this instead of adding more methods to `TestCLIMain`.

tests/test_cli_main.py
```diff
@@ -202,6 +214,142 @@ def test_handle_url_with_json_and_stream_url(self, console, args):
             self.assertEqual(console.exit.mock_calls, [call("The stream specified cannot be translated to a URL")])
             console.exit.mock_calls.clear()
 
+    def test_create_filename_invalid_chars_native_rules_auto(self):
+        streamlink_cli.main.args = args = Mock()
```
bastimeyer, line 218: Please properly mock this with the `@patch` decorator, so that it gets cleaned up after the test. Otherwise you're polluting the module's namespace.

bastimeyer (changes requested):
This is just a quick diff review to give early feedback. I haven't looked at the whole code yet.

src/streamlink_cli/main.py
mkbloke: My thinking was that as `args.output`, `args.record` and `args.record_and_pipe` are mutually exclusive and also used separately in a way that appeared to be unnecessary where the formatted name is substituted instead, it would be better to create just the one global variable for a formatted name as a solution, rather than one for each and leave the other usage of any of those three attributes in the code alone.

src/streamlink_cli/main.py
mkbloke: Oh yes,

src/streamlink_cli/main.py
mkbloke: What are your thoughts on replacing `args.output`, `args.record` and `args.record_and_pipe` here?  Do you think there should be separate format processed versions for each?  I'm wondering if a better solution might be to write a class that has these as properties that are either formatted or otherwise set to whatever they were originally.  That class could also then contain a property with the formatted title without the filesystem-safe processing for use where `title` is used currently.  That would mean `create_title()` would be removed and replaced with the class instance, although this would have to be done earlier in various places in the code to make the filename properties available when needed.  The class instance would be assigned to a global.  Does that seem like a reasonable way to go?

bastimeyer: > What are your thoughts on replacing args.output, args.record and args.record_and_pipe here?

The problem with the global var is that you're setting it in a deeply nested function call while reading it afterwards up the call stack. I don't think this is good. Just check the `args` attributes individually and override their values if a value exists. Ideally this should be done as early as possible, eg. in `setup_args`, but I don't think it's possible in this situation.

As I've mentioned multiple times, the `streamlink_cli` module is in desperate need of a complete rewrite that cleans up the usage of global variables. And it's way too bloated for a single module as well. This is why adding stuff like this makes it unnecessarily difficult :(

> Do you think there should be separate format-processed versions for each?

You mean LazyFormatter instances or different sets of available vars?

mkbloke: > The problem with the global var is that you're setting it in a deeply nested function call while reading it afterwards up the call stack. I don't think this is good. Just check the args attributes individually and override their values if a value exists. Ideally this should be done as early as possible, eg. in setup_args, but I don't think it's possible in this situation.

I think maybe I misunderstood what you meant in: https://github.com/streamlink/streamlink/pull/3913#discussion_r688263277.  I was under the impression you were objecting to overriding the three global file args at all and meant separate global vars should be used instead?  Is it the case that you meant overriding is OK, but just not at that point in the code?  It would appear that the earliest point it can be done is in `handle_url()`.

bastimeyer: Hm. After thinking about it again, setting the `args` attributes like you did in #3913 might be the best solution for now. When I wrote the comment I wasn't 100% aware of how the attributes were read in the following methods. Let's not make it unnecessarily difficult here. The problem of relying on side-effects from other methods when reading the `args` attributes remains though, but that shouldn't be too big of a deal for this feature. If I start arguing about that then I need to argue about a lot of other stuff in the CLI module, which I don't want now. Sorry about the confusion.

mkbloke: OK.  I did have another idea, otherwise, if setting the args in `handle_url` made sense.  This would have involved adding a `title` keyword arg to the `output_*()` and `create_output()` methods.  The reason for this being that I'd generate the formatted versions of the file args and the title at the same time.  Each of the `plugin.get_*()` methods would only be called once here and only if callable, otherwise taking the default value string.  I don't know if that seems reasonable or you'd rather stick with how it was done in #3913?

If you still prefer the idea of what was done in #3913, I can go back to something like that, although I still like the idea of integrating some of what I've said above, such that the `plugin.get_*()` methods are only called once - if callable - to eventually generate the formatted versions of both the file arg and title.

Either way, I'm guessing that limiting the number of calls to these plugin methods would be considered beneficial to potentially limit any delay in the stream start, as per your comment in this PR (https://github.com/streamlink/streamlink/pull/3933#discussion_r693544988)?

Sorry to throw more confusion into the mix!

bastimeyer: > `handle_url`

You can't read the stream metadata before the actual stream has been chosen, and this does happen quite late in `output_stream` when `success_open` is `True`, so passing another parameter to the various `output_` methods is not possible.

What looks like a better solution is reading the `filename` attribute of the `FileOutput` instance (if the file descriptor for stdout is not set) instead of overriding the `args` attributes. `args.output` and `args.record` are read in `output_stream` itself and `read_stream`, which is called by `output_stream` and the FileOutput instance is already passed as a parameter.

src/streamlink_cli/main.py
mkbloke: > You can't read the stream metadata before the actual stream has been chosen, and this does happen quite late in output_stream when success_open is True, so passing another parameter to the various output_ methods is not possible.

I'm trying to understand why that's the case.  The `_get_streams` method for the plugin has already been called in `fetch_streams()` via `plugin.streams()`, so it seems like there should be valid available metadata here?  Testing seems to indicate that this is the case?

src/streamlink_cli/main.py
mkbloke: > You can't read the stream metadata before the actual stream has been chosen, and this does happen quite late in output_stream when success_open is True, so passing another parameter to the various output_ methods is not possible.

I'm trying to understand why that's the case.  The `_get_streams` method for the plugin has already been called in `fetch_streams()` via `plugin.streams()`, so it seems like there should be valid available metadata here?  Testing seems to indicate that this is the case?

bastimeyer: Sorry, I worded this wrong and I should've been more precise. I meant the output type choice.

> The `_get_streams` method for the plugin has already been called

It's up to the plugin's `get_{author,category,title}` methods to return the metadata, which means there could be fetching logic in these methods instead of `_get_streams`, and fetching metadata when it's not necessary should be avoided to prevent unnecessary stream launch delays. Most plugins currently fetch metadata in `_get_streams` and return the cached data in these metadata getter methods, but that's not a guaranteed pattern.

What's important is the output type (passthrough, http, json, stream-url, player/file) that gets decided in `handle_stream`, which means it would only make sense reading metadata in `output_stream`, as this method is responsible for creating the output (player or file) where we need the substituted file name. The `output_stream` call is however part of a for-loop, which depends on its return value, which again depends on its inner `success_open` variable, and this must be `True`. Since you need the substituted file name in the `console.exit` and `progress` calls after the `create_output()` call, it probably makes more sense getting the metadata in `output_stream` rather than `create_output`, or you'd have to return a tuple in `create_output`.

Hope that makes sense and I hope I haven't missed something myself. It's very difficult to keep all potential cases in your head at the same time when the CLI design isn't that great.

mkbloke: Many thanks for the clarification.  I think what I have in the third iteration is something quite a bit nicer.  It lazy loads and caches title/author/category/game/url for both title and filename creation, so for example, plugin.get_title() will only be called if there's a title formatting variable in the title and/or filename spec, but it'll only be called once, even if used in both specs.  I know the URL doesn't really need lazy loading/caching, but it's a harmless side effect of the implementation.  Let me put it all together and check that it works as aspected, then I'll push it for you to take a look at it.  I won't bother with the tests until the basic methodology meets with approval.

PR #3941, opened 2021-08-26
cli: add title to json

odnar-dev:
image

image

used as mpv backend :
image


src/streamlink_cli/main.py
```diff
@@ -407,7 +407,8 @@ def handle_stream(plugin, streams, stream_name):
 
     # Print JSON representation of the stream
     elif console.json:
-        console.msg_json(stream)
+        title = create_title(plugin)
```
bastimeyer, line 410: `create_title` is no more. This has just been replaced in #3962.

Unfortunately I had this PR here off the radar and moved the new title/filename formatter initialization to a different method which won't work for these changes. The formatter (see down below) needs to be moved to the `handle_url` method and passed as an argument to `handle_stream`, so both `console.msg_json()` calls can read the title from the formatter instance via

```py
formatter.title(args.title, defaults=DEFAULT_STREAM_METADATA) if args.title else args.url
```

bastimeyer (changes requested):

bastimeyer:
After thinking about it again (I only had the diff in mind in my previous comment, sorry), do we want the interpolated `--title` parameter value in the JSON output, or the individual stream metadata properties?

The `--title` parameter is just a utility for formatting the window title of the player launched by Streamlink, which is not useful for JSON data, where individual metadata properties make much more sense.

odnar-dev:
> After thinking about it again (I only had the diff in mind in my previous comment, sorry), do we want the interpolated `--title` parameter value in the JSON output, or the individual stream metadata properties?

both

> The `--title` parameter is just a utility for formatting the window title of the player launched by Streamlink, which is not useful for JSON data, where individual metadata properties make much more sense.

i have a script that use streamlink as mpv backend instead of youtube-dl for some specific sites like twitch... [$](https://gist.github.com/ChrisK2/8701184fe3ea7701c9cc) 
just by typing "mpv %url%"
it run 'streamlink -j url quality' then extract streamurl from the response and passe it to mpv
this is needed so they are no need to call streamlink two times to get the title

odnar-dev:
i am not sure if this is worth adding right now ;
if user wanted only a specific metadata like the author name he can alwayse call `streamlink -j --title "{author}" %url% best`
if he wanted multiple metadata properties, he probably gonna combine them; so instead of just calling streamlink with the formate he wanted, he gonna need to combine them manually 

also this way respect the formate in streamlink config file if i called `streamlink -j %url% best`

PR #3950, opened 2021-08-28
plugins.pluto: geo-restricted to USA

vstavrinov:
This plugin is working for me only with the USA proxy. BTW for a long time I am using static links without proxy. This may be a hint for the workaround.

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -126,7 +126,7 @@ orf_tvthek              tvthek.orf.at        Yes   Yes
 picarto                 picarto.tv           Yes   Yes
 piczel                  piczel.tv            Yes   No
 pixiv                   sketch.pixiv.net     Yes   --
-pluto                   pluto.tv             Yes   Yes   geo-restricted to USA
+pluto                   pluto.tv             Yes   Yes   geo restricted to the Americas and Europe
 pluzz                   - france.tv          Yes   Yes   Streams may be geo-restricted to France, Andorra and Monaco.
                         - ludo.fr
                         - zouzous.fr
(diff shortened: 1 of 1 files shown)


bastimeyer:
The site's wikipedia page says that content is available in europe and other countries as well, but limited:
https://en.wikipedia.org/wiki/Pluto_TV#Availability

Is this a recent change you've observed? Which addresses from which countries did you try?

vstavrinov:
From Russia for a long time

bastimeyer:
Well, it is working just fine with a German IP address, so "geo-restricted to USA" is not correct. Since there is geo-restriction, it's probably best to just copy what Wikipedia says: "geo restricted to the Americas and Europe" (USA + Latin America + Canada, and Europe)

The plugin's URL matcher also needs a fix, but that's not relevant to this PR.

vstavrinov:
For Norway, it doesn't work too, but in a different way.

bastimeyer:
> but in a different way

Different in which way? If some European countries are blocked as well, then just say "Americans and parts of Europe".

Please also keep the same format which other plugin matrix entries are using, for consistency reasons

> Streams are geo-restricted to ...

vstavrinov:
URL redirected to is different, while all countries where it is not available are directed to the same URL.

If it is available not for all Europe, why do You sure It is available for all Americas countries? Would You like to check all of them?  It may be better to output an appropriate error message if there is a reasonable way to detect geo-restriction?  This is a common question for all plugins.

bastimeyer:
I was merely referring to what the Wikipedia page says because it's unfeasible to check every potential country. If you think that it's better to add a geo-restriction error message when content is not available, then please go for it. Then the docs can say something like "Geo-restrictions may apply", so we don't have to specify which countries.

PR #3952, opened 2021-08-28
vendor: add lxml dependency

### LINKED ISSUE #3944: Add lxml dependency
### Checklist

### Description

Streamlink should finally switch to a proper HTML/XML parser for extracting data instead of using cheap regex workarounds which don't work properly. I've already commented on this issue last year:
https://github.com/streamlink/streamlink/issues/3241#issuecomment-706486239

The reason why I'm suggesting this again right now is that I was trying to fix the deutschewelle plugin (https://dw.com) yesterday and ran into issues with the `itertags` utility method, which is based on simple regexes for iterating HTML nodes and their attributes+body. `itertags` for example does not work with nested nodes, which makes adding ridiculous custom regexes necessary. Just take a look at this madness:
https://github.com/streamlink/streamlink/blob/3668770d608f0fab54d40a46acd6720a97f63775/src/streamlink/plugins/deutschewelle.py#L18-L29

With `lxml` (https://lxml.de/), HTML page contents can be parsed and the data extracted via XPath queries and/or the respective API methods. The methods are similar to python's native `xml.etree.ElementTree`, which itself is considered too slow and unsafe in certain cases. I am by no means an expert regarding python's standard library though, so if someone has better insight here, please share. In regards to packaging, this lib is available on basically every packaging system and adding it as a dependency here only has benefits.

I'd suggest that we add `lxml` as a dependency now and start using it for extracting data from HTML documents. The validation schema methods could be improved for this as well. There's also the `parse_xml` utility method, which is currently based on the native module.

Comments?

bastimeyer:
Resolves #3944 

This adds the `lxml` dependency for parsing XML and HTML to Streamlink's `install_requires`.

There's also the `lxml-stubs` package available on pypi for adding type definitions which are lacking in the `lxml` package, but I don't know how this external package works. I'm using pycharm professional and it's generating types from the docs automatically (not that well though). `lxml-stubs` is supposed to go into `dev-requirements.txt` I think. Should this be added?

----

The validation methods and `parse_xml` utility method have not been modified yet and are still using the standard library methods for parsing XML. DASH streams and the validation methods (which are only called one single plugin - rtve) are therefore not affected.

I've tried to update the methods, but ran into encoding issues while reading XML from strings (utf-8) with a custom encoding set in the XML declaration (`<?xml version="..." encoding="..."?>`). This needs some refactoring first.

----

As an example on how XPath queries massively simplify plugins in a safe way, take a look at this diff:
https://github.com/bastimeyer/streamlink/compare/requirements/lxml...bastimeyer:plugins/deutschewelle/rewrite-xpath

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -289,6 +289,7 @@ Name                                 Notes
 `iso-639`_                           Used for localization settings, provides language information
 `iso3166`_                           Used for localization settings, provides country information
 `isodate`_                           Used for MPEG-DASH streams
+`lxml`_                              Used for processing HTML and XML data
 `PySocks`_                           Used for SOCKS Proxies
 `websocket-client`_                  At least version **0.58.0**. (used for some plugins)
 
@@ -321,6 +322,7 @@ With these two environment variables it is possible to use `pycrypto`_ instead o
 .. _iso-639: https://pypi.org/project/iso-639/
 .. _iso3166: https://pypi.org/project/iso3166/
 .. _isodate: https://pypi.org/project/isodate/
+.. _lxml: https://lxml.de/
 .. _PySocks: https://github.com/Anorov/PySocks
 .. _websocket-client: https://pypi.org/project/websocket-client/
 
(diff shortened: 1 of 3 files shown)


gravyboat:
`lxml` is well maintained and actively supported, I'm fine with adding this.

PR #3954, opened 2021-08-28
cli: support -t formatting vars in -o, -r and -R

### LINKED ISSUE #3898: Option to add formatting variables for output filename

mkbloke:
Sorry for opening yet another PR, it wasn't my original intention.

Tests to come if this code looks reasonable...

closes #3898

### DIFF
--- a/src/streamlink/utils/formatter.py
+++ b/src/streamlink/utils/formatter.py
@@ -1,0 +1,64 @@
+import re
+
+from streamlink_cli.compat import is_win32
+from streamlink_cli.constants import DEFAULT_STREAM_METADATA, FS_SAFE_REPLACEMENT_CHAR, INVALID_FILENAME_CHARS
+
+
+class Formatter(dict):
+    def __init__(self, plugin, args):
+        self.plugin = plugin
+        self.args = args
+        self.cache = self.Cache(self.plugin)
+
+    def __missing__(self, k):
+        if k in ("title", "author", "category", "game", "url"):
+            self[k] = self._make_fs_safe(self.cache[k])
+        else:
+            return f"{{{k}}}"
+
+        return self[k]
+
+    def _make_fs_safe(self, unsafe_str):
+        if self.args.fs_safe_rules in INVALID_FILENAME_CHARS:
+            rules = self.args.fs_safe_rules
+        else:
+            if is_win32:
+                rules = "Windows"
+            else:
+                rules = "POSIX"
+
+        sub_re = re.compile(
+            "[{:s}]".format(re.escape(INVALID_FILENAME_CHARS[rules])), re.UNICODE
+        )
+
+        return sub_re.sub(FS_SAFE_REPLACEMENT_CHAR, unsafe_str)
+
+    def get_formatted_filename(self, unformatted_filename):
+        return unformatted_filename.format_map(self)
+
+    def get_formatted_title(self, unformatted_title):
+        if self.args.title:
+            return unformatted_title.format_map(self.cache)
+        else:
+            return self.args.url
+
+    class Cache(dict):
+        def __init__(self, plugin):
+            self.plugin = plugin
+
+        def __missing__(self, k):
+            if k == "title":
+                v = self.plugin.get_title() or DEFAULT_STREAM_METADATA["title"]
+            elif k == "author":
+                v = self.plugin.get_author() or DEFAULT_STREAM_METADATA["author"]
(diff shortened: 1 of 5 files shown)


mkbloke:
I'm not sure if I've put the Formatter class file in the right place.  Would it be better in `src/streamlink_cli/utils/`, or is `src/streamlink/utils/` the right place?

bastimeyer:
I will take a look at this tomorrow as it's a bit late here. I can already tell you though that this belongs into streamlink_cli, if it makes sense having another formatter class, which I don't know yet.

mkbloke:
Cool, I'll move it if the rest of it looks OK to you when you can properly look at it.  I think it's worthwhile having the extra formatter class to do both title and filename formatting in a lazy/caching manner.  If it wasn't for the use of LazyFormatter in `hls.py` I'd have been tempted to get rid of LazyFormatter completely (in other commits, I guess), but that's assuming this code looks acceptable to you overall...

src/streamlink/utils/formatter.py
```diff
@@ -0,0 +1,64 @@
...
+        else:
+            return self.args.url
+
+    class Cache(dict):
+        def __init__(self, plugin):
+            self.plugin = plugin
+
+        def __missing__(self, k):
+            if k == "title":
+                v = self.plugin.get_title() or DEFAULT_STREAM_METADATA["title"]
```
back-to, line 51: all `DEFAULT_STREAM_METADATA` are pretty stupid, when it get used for a filename
for a player title it is like whatever, but when used for a file it will try  to create files with the same name.

https://github.com/streamlink/streamlink/blob/6cdc3ebb141f83c8fe34f2426685443fa1e1b7de/src/streamlink_cli/constants.py#L11-L16

src/streamlink/utils/formatter.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class Formatter(dict):
+    def __init__(self, plugin, args):
+        self.plugin = plugin
+        self.args = args
+        self.cache = self.Cache(self.plugin)
+
+    def __missing__(self, k):
+        if k in ("title", "author", "category", "game", "url"):
```
back-to, line 14: seems hard coded, `Formatter` is not really reusable for other stuff

src/streamlink/utils/formatter.py
mkbloke: I don't really see how it's that much of a problem using these default strings in filenames when title/author/category/game cannot be retrieved via the plugin.  They seem as good as any to me.

Potentially, yes, if a plugin does not implement any of the metadata methods then it's going to fall back to a filename that will possibly be a duplicate if Streamlink has been called before with a plugin where this has also happened, but the file will not be overwritten unless `--force` has been specified.  I don't think this code should have to do anything to avoid that situation.  Why should it need to do so?

src/streamlink/utils/formatter.py
mkbloke: That is true, but it was never intended to be a general reusable formatter.  Its purpose is only to supply formatted title and filesystem-safe filename strings in a lazy manner and without multiple calls to the `plugin.get_*()` metadata methods.  Perhaps I should rename it to better reflect that.

bastimeyer:
Here's a cleaned up version:
- Diff from master:
  https://github.com/streamlink/streamlink/compare/master...bastimeyer:cli/output-variables
- Diff from this PR:
  https://github.com/mkbloke/streamlink/compare/file-output-format-vars-3...bastimeyer:cli/output-variables

Changes from this PR:
- no hardcoded variables in the Formatter class
- no dependencies on the resolved plugin and CLI args in the Formatter class
- no unnecessary nested Cache class
- streamlink does not depend on streamlink_cli (the Formatter could instead also replace LazyFormatter and get moved back to streamlink with a slight modification... needs a decision)
- default values are only applied to titles, but not file names
- path character replacement has its own module that can be tested independently
- unprintable chars get replaced on POSIX as well (why not?)
- formatter initialization has been moved from `handle_url` to `handle_stream`

Tests are obviously still missing.

I'm not sure if I want to open a PR for this myself, because I don't want to shut down your efforts of implementing this feature, @mkbloke. What do you suggest should we do?

mkbloke:
@bastimeyer, it's fine if you want to open the PR yourself with what you have done.  It's substantially different from what I implemented, so I think it makes sense for you to do so.  Cheers.

PR #3960, opened 2021-08-28
plugins.pluto: fix URL match for 2 letter language codes

mkbloke:
fixes #3959


bastimeyer:
Please add tests for the updated URL matcher.

mkbloke:
Oops, I forgot about updating the tests... should be OK now, cheers.

mkbloke:
Actually, I might just add some extra checks for the language letter stuff on the 'should not match' side.

mkbloke:
Ah, never mind!

PR #3961, opened 2021-08-29
GitHub Action to lint Python code

cclauss:

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -1,0 +1,25 @@
+name: lint_python
+on: [pull_request, push]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip wheel
+      - run: pip install bandit black codespell flake8 flake8-bugbear
+                         flake8-comprehensions isort mypy pytest pyupgrade safety
+      - run: bandit --recursive --skip B101 . || true  # B101 is assert statements
+      - run: black --check . || true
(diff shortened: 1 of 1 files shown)


bastimeyer (commented):
There's already a linting config for flake8+plugins here which uses the configs defined in setup.cfg:
- https://github.com/streamlink/streamlink/blob/05bdf991b367ce8a5f150ec1fd7239c4a7aaff89/.github/workflows/main.yml#L44-L52
- https://github.com/streamlink/streamlink/blob/05bdf991b367ce8a5f150ec1fd7239c4a7aaff89/setup.cfg#L12-L53

This linting config is run in the test job, which is a requirement for some of the other jobs to start. Adding another workflow makes it run in parallel without any effects on a failure, especially if you add `|| true` to each shell command.

- Bandit and Safety are an alternative to GH's native CodeQL, which we haven't merged after an earlier attempt that was dismissed because of packaged legacy code that's still included and won't get touched anymore
- Black is very much opinionated and considered "beta". While I don't see much wrong with its style and appreciate a modernization of the code style in general, Streamlink still has a lot of old code with weird styles which was added years ago and hasn't been touched yet.
  Adding "aggressive" linting rules is good for the project maintainers, but it ultimately makes it unwelcoming for new contributors, same with pre-commit hooks (which I hate for this very reason).
- The rules of bugbear are rather meh but it could be added to the already existing linting config, as it's a flake8 plugin.
- Codespell seems unnecessary.
- Pyupgrade is unnecessary and doesn't work well with its default config. I already made this observation when I ported Streamlink's code to 3.6 and dropped 2.7 last year.
- isort is unnecessary because we already have an import order linting config for flake8.

Streamlink also doesn't define its requirements via a `requirements.txt` and running tests again in the linting config doesn't make much sense.

PR #3962, opened 2021-08-29
cli: player title and file output metadata vars

### LINKED ISSUE #3898: Option to add formatting variables for output filename

bastimeyer:
- support the same plugin metadata variables in `--output`, `--record`
  and `--record-and-pipe` like they already exist in `--title`:
  `{url}`, `{author}`, `{category}`/`{game}` and `{title}`
- substitute unsupported file name characters on specific OS-types
- add `--fs-safe-rules` CLI argument for overriding substitution logic
- add new Formatter class with caching and parameter translation
- create formatter instance in `handle_stream` and pass it to output
  methods where player titles or file names need to be formatted
- rewrite test_cli_main tests

Co-Authored-By: Ian Cameron <1661072+mkbloke@users.noreply.github.com>

----

Resolves #3898
Related #3954, #3933, #3913 

----

- ~~The critical code path where the formatter gets initialized is currently not tested because there weren't any test prior to these changes. I will try to add some tests tomorrow or so. This requires setting up ton of mocks, which is ugly.~~
- The new Formatter could be moved to the streamlink package and replace the LazyFormatter which is currently only used in the HLSWriter for the HLS key URI override stuff.
- There's currently no variable for the output stream's file container, which makes these variables not that useful IMO, but whatever.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -588,6 +588,9 @@ def build_parser():
         Write stream data to FILENAME instead of playing it.
 
         You will be prompted if the file already exists.
+
+        The formatting variables available for the --title option may be used.
+        Unsupported characters in substituted variables will be replaced with an underscore.
         """
     )
     output.add_argument(
@@ -619,6 +622,9 @@ def build_parser():
         Open the stream in the player, while at the same time writing it to FILENAME.
 
         You will be prompted if the file already exists.
+
+        The formatting variables available for the --title option may be used.
+        Unsupported characters in substituted variables will be replaced with an underscore.
         """
     )
     output.add_argument(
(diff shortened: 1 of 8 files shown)


bastimeyer:
I've added a new test which checks the formatter's initialization and the correct metadata variable defintions. This still doesn't test every change in the main CLI module, but I don't care, because adding more and more tests with 100s of mocks for this aged module doesn't make sense when the design isn't good to begin with. As said, a complete rewrite is needed.
https://app.codecov.io/gh/streamlink/streamlink/compare/3962/diff

Also just noticed that my test changes fix a mistake which was present since several years where it always created an HTTPServer instance when running file output tests:
- https://app.codecov.io/gh/streamlink/streamlink/compare/3962/changes#D1L104
- https://github.com/streamlink/streamlink/blob/05bdf991b367ce8a5f150ec1fd7239c4a7aaff89/tests/test_cli_main.py#L251-L260
- https://github.com/streamlink/streamlink/blob/05bdf991b367ce8a5f150ec1fd7239c4a7aaff89/src/streamlink_cli/main.py#L101-L107

back-to (approved):

PR #3963, opened 2021-08-30
pyupgrade --py36-plus **/*.py

cclauss:
@bastimeyer Please help me understand your comments at #3961 that pyupgrade _doesn't work well with its default config._  Removing `"r"` from `open()` calls bugs me but beyond that the other modifications seem acceptable.

This pull request was created with the command [`pyupgrade --py36-plus **/*.py`](https://github.com/asottile/pyupgrade#readme)

Python f-strings simplify the code and [should speed up execution](https://www.scivision.dev/python-f-string-speed).

### DIFF
--- a/docs/ext_argparse.py
+++ b/docs/ext_argparse.py
@@ -59,7 +59,7 @@ class ArgparseDirective(Directive):
             lambda m: (
                 _option_re.sub(
                     lambda m2: (
-                        ":option:`{0}`".format(m2.group(1))
+                        f":option:`{m2.group(1)}`"
                         if m2.group(1) in self._available_options
                         else m2.group(0)
                     ),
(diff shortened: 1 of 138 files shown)


back-to:
there is no point in reviewing or merging on big commit

some files should not be changed in bulk changes such as `versioneer.py` and other

bastimeyer:
The f-string changes were left out in #3368 intentionally, because they caused the diff of the py36 upgrade last year to bloat up massively for no reason. Some maintainers also disliked f-strings (at least at the time), so these changes were ignored partly because of this as well. Code-mod tools like pyupgrade are also not able to understand certain styles, and blindly commiting changes can make thing worse.

If you want to help improve the code style, that's fine, but please do it in separate commits for each logical change. Pyupgrade doesn't support this, so you'll have to dig into its code and comment out certain code-modification calls, which is stupid. This is what I was referring to.

Also please don't modify files and directories which are not supposed to be modified, which are

- `./versioneer.py`
- `./docs`
- `./examples`
- `./src/**/packages`

----

https://github.com/streamlink/streamlink/pulls?q=label%3Apython-2+is%3Apr+is%3Amerged
https://github.com/streamlink/streamlink/pulls?q=label%3Apython-3+is%3Apr+is%3Amerged

cclauss:
Thanks for your explanation.

PR #3967, opened 2021-08-31
plugin.api.validate: switch to lxml.etree

bastimeyer:
- replace xml.etree from standard library with lxml.etree in
  plugin.api.validate, utils.parse_xml, plugins and tests
- add validate.xml_xpath() and validate.xml_xpath_string() methods
- fix validate.get() and return attribute value for given attribute key
- refactor validate.xml_element() validation method
- fix and add tests

----

#3952


bastimeyer:
This is not part of this PR, but I think it's time to add `parse_{xml,html,json}` utility methods to the validation module. It's a bit silly that almost every plugin has to import this on its own and define a validate.transform() call for it.

PR #3969, opened 2021-08-31
plugins.booyah: add support for source stream

lbnascimento:
The current booyah plugin does not support the source quality. This PR adds support for it.

### DIFF
--- a/src/streamlink/plugins/booyah.py
+++ b/src/streamlink/plugins/booyah.py
@@ -67,7 +67,7 @@ class Booyah(Plugin):
 
     def get_title(self):
         return self.title
-        
+
     @classmethod
     def stream_weight(cls, stream):
         if stream == "source":
--- a/src/streamlink/plugins/booyah.py
+++ b/src/streamlink/plugins/booyah.py
@@ -1,5 +1,6 @@
 import logging
 import re
+import sys
 from urllib.parse import urljoin
 
 from streamlink.plugin import Plugin, pluginmatcher
@@ -17,13 +18,6 @@ class Booyah(Plugin):
     vod_api_url = 'https://booyah.live/api/v3/playbacks/{0}'
     live_api_url = 'https://booyah.live/api/v3/channels/{0}'
     streams_api_url = 'https://booyah.live/api/v3/channels/{0}/streams'
-    
-    @classmethod
-    def stream_weight(cls, key):
-        if key.startswith("source"):
-            return 65535, "source"
-
-        return Plugin.stream_weight(key)
 
     auth_schema = validate.Schema({
         'expiry_time': int,
@@ -61,17 +55,6 @@ class Booyah(Plugin):
         },
     })
 
-    streams_schema = validate.Schema({
-        'stream_addr_list': [{
-            'resolution': str,
-            'url_path': str,
-        }],
-        'mirror_list': [{
-            'url_domain': validate.url(),
-        }],
-        'source_stream_url_path': str,
-    })
-
     author = None
     category = None
     title = None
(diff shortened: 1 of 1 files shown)


bastimeyer (commented):
Thanks for the PR.

There's a problem with the stream names. The same stream name gets yielded multiple times, which overrides it on every consecutive call, making the mirror iteration pointless. Just return the first one or modify the stream name depending on the available mirror server. The resolution dependent streams from above also need to be fixed in this regard.

The validation schema also needs to be updated if you want to access additional properties, so it is obvious which data is needed/accessed.

----

However, I'd say that the mirror selection is unnecessary as a whole and a plugin shouldn't yield/return the same streams from different sources. The mirror selection should get replaced in favor of the `default_mirror` value, which is part of the JSON response. The `"Auto"` stream resolution also is a variant playlist which contains every other stream resolution except `"source"`, so the whole for-loop of the `stream_addr_list` is useless as well.

This means that all that should be done here to yield the right streams is select the mirror depending on the `default_mirror` value, get its URL from the matching `mirror_list` item (or the first one if none matches) and join the chosen mirror server URL with the path of the "Auto" (variant) and "source" stream resolutions.

```diff
diff --git a/src/streamlink/plugins/booyah.py b/src/streamlink/plugins/booyah.py
index ea729274..099977b4 100644
--- a/src/streamlink/plugins/booyah.py
+++ b/src/streamlink/plugins/booyah.py
@@ -1,5 +1,6 @@
 import logging
 import re
+import sys
 from urllib.parse import urljoin
 
 from streamlink.plugin import Plugin, pluginmatcher
@@ -54,16 +55,6 @@ class Booyah(Plugin):
         },
     })
 
-    streams_schema = validate.Schema({
-        'stream_addr_list': [{
-            'resolution': str,
-            'url_path': str,
-        }],
-        'mirror_list': [{
-            'url_domain': validate.url(),
-        }],
-    })
-
     author = None
     category = None
     title = None
@@ -77,6 +68,12 @@ class Booyah(Plugin):
     def get_title(self):
         return self.title
 
+    @classmethod
+    def stream_weight(cls, stream):
+        if stream == "source":
+            return sys.maxsize, "source"
+        return super().stream_weight(stream)
+
     def do_auth(self):
         res = self.session.http.post(self.auth_api_url)
         self.session.http.json(res, self.auth_schema)
@@ -119,15 +116,32 @@ class Booyah(Plugin):
         self.title = user_data['channel']['name']
 
         res = self.session.http.get(self.streams_api_url.format(stream_id))
-        streams = self.session.http.json(res, schema=self.streams_schema)
-
-        for stream in streams['stream_addr_list']:
-            if stream['resolution'] != 'Auto':
-                for mirror in streams['mirror_list']:
-                    yield stream['resolution'], HLSStream(
-                        self.session,
-                        urljoin(mirror['url_domain'], stream['url_path']),
-                    )
+        streams = self.session.http.json(res, schema=validate.Schema({
+            "default_mirror": str,
+            "mirror_list": [{
+                "name": str,
+                "url_domain": validate.url(),
+            }],
+            "source_stream_url_path": str,
+            "stream_addr_list": [{
+                "resolution": str,
+                "url_path": str,
+            }],
+        }))
+
+        mirror = (
+            next(filter(lambda item: item["name"] == streams["default_mirror"], streams["mirror_list"]), None)
+            or next(iter(streams["mirror_list"]), None)
+        )
+        if not mirror:
+            return
+
+        auto = next(filter(lambda item: item["resolution"] == "Auto", streams["stream_addr_list"]), None)
+        if auto:
+            yield from HLSStream.parse_variant_playlist(self.session, urljoin(mirror["url_domain"], auto["url_path"])).items()
+
+        if streams["source_stream_url_path"]:
+            yield "source", HLSStream(self.session, urljoin(mirror["url_domain"], streams["source_stream_url_path"]))
 
     def _get_streams(self):
         self.do_auth()
```

PR #3970, opened 2021-08-31
plugin: metadata attributes

bastimeyer:
- add author, category and title attributes to base Plugin class
- make getters return attributes instead of None
- remove unnecessary duplicate attributes and getters from plugins
- keep getters in plugins which use custom logic

----

This change makes it much easier to set the plugin metadata by simply setting the attributes instead of overriding the `get_{author,category,title}` methods every single time with different return values.

The `author`, `category` and `title` attributes are already a well established pattern, so making them part of the Plugin base class is the right choice.

### DIFF
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -202,6 +202,11 @@ class Plugin:
     # a reference to the `re.Match` result of the first matching matcher
     match: Match
 
+    # plugin metadata attributes
+    author: Optional[str] = None
+    category: Optional[str] = None
+    title: Optional[str] = None
+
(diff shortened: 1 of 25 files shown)


odnar-dev:
just add my [pr](https://github.com/streamlink/streamlink/pull/3941) to this, i couldn't fix the tests anyway

bastimeyer:
@odnar-dev 
I will take a look at this feature once I've finished some other stuff. Sorry for wasting your time with #3941. The changes should be simple though. Just a bit difficult when stuff is changing the whole time on the master branch, I understand.

gravyboat (approved):

PR #3972, opened 2021-08-31
plugin.api.validate: add parse_{json,html,xml,qsd}

bastimeyer:
1. plugin.api.validate: add args+kwargs to transform
2. plugin.api.validate: add parse_{json,html,xml,qsd}
   - add parse_{json,html,xml,qsd} validate.transform wrapper methods
   - refactor utils module and move parse* methods into utils.parse
   - refactor parse* methods and use common parsing + exception handling
   - reformat all-export list in utils

### DIFF
--- a/src/streamlink/plugin/api/validate.py
+++ b/src/streamlink/plugin/api/validate.py
@@ -24,11 +24,19 @@ from urllib.parse import urlparse
 from lxml.etree import Element, iselement
 
 from streamlink.exceptions import PluginError
+from streamlink.utils.parse import (
+    parse_html as _parse_html,
+    parse_json as _parse_json,
+    parse_qsd as _parse_qsd,
+    parse_xml as _parse_xml
+)
+
 
 __all__ = [
     "any", "all", "filter", "get", "getattr", "hasattr", "length", "optional",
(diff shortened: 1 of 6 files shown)


bastimeyer:
???
https://codecov.io/gh/streamlink/streamlink/src/3d101763688f919863d11170d8b620aba194b216/tests/test_api_validate.py#L74

bastimeyer:
`streamlink.plugin.api.utils` still re-exports `parse_json`, `parse_xml` and `parse_qsd` (as `parse_query`):
https://github.com/streamlink/streamlink/blob/5c035fceb8f50e7cae548e34786ca61b1a3630a7/src/streamlink/plugin/api/utils.py#L5-L7

I didn't add the new methods here, because the goal is to remove this module eventually, as `itertags` is not needed with properly parsed HTML pages via lxml and XPath queries. The new methods make it especially easy to use.

I have already finished refactoring a couple of plugins where I removed the itertags imports and switched to XPath queries, but I don't think I want to refactor every plugin (at least not any time soon). Once the remaining plugins have switched to XPath queries though, we can deprecate itertags and the entire api.utils module, and then remove it in a major version bump.

gravyboat:
Sounds like a solid plan to me and this looks good. Thanks @bastimeyer.

PR #3993, opened 2021-09-05
cli: add {time:format} var to --output / --title

bastimeyer:
This adds support for the `{time}` / `{time:format}` variable to `--title` and `--output`, which has been a much requested feature in the past. The answer has always been "use your command line shell", which is not that great and Streamlink should support something very basic like that.

The `:format` part is optional and based on the standard string format specifiers described in PEP3101:
https://www.python.org/dev/peps/pep-3101/#format-specifiers
The logic for this is however much simplified and doesn't support nesting, etc, because we don't need it here.

Format strings on variables where there's no formatting callback registered get ignored. Conversion flags like `!r`, `!s`, etc, get ignored as well.
https://www.python.org/dev/peps/pep-3101/#explicit-conversion-flag

----

- In order to re-format the variables' values, the Formatter class had to be rewritten, based on string.Formatter's parse() method. It doesn't inherit from string.Formatter because of the formatting simplification. This is the first commit, which also fixes/improves the tests.
- The second commit then adds support for the parsed `format_spec`. Instead of adding a second parameter to the Formatter's constructor with formatting functions, the mapping functions could've also received a parameter with the optional formatting string, but that would've prevented caching, so this is the better choice (I think).
- The third commit adds the `{time}` var to the CLI module, fixes the default string value of datetime.now() (for file name reasons), and adds documentation. Linking to the python docs ([`datetime.strftime()`](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes)) via reST is unfortunately not possible in argparse arguments because of the `--help` output and man page content, which can't have reST.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -569,6 +569,11 @@ def build_parser():
         {{url}}
             URL of the stream.
 
+        {{time}}
+            The current timestamp, which can optionally be formatted via {{time:format}}.
+            This format parameter string is passed to Python's datetime.strftime() method,
+            so all usual time directives are available. The default format is "%Y-%m-%d_%H-%M-%S".
+
         Examples:
 
             %(prog)s -p vlc --title "{{title}} -!- {{author}} -!- {{category}} \\$A" <url> [stream]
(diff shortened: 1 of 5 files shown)


gravyboat:
Great addition @bastimeyer, thanks!

PR #4000, opened 2021-09-08
docs: bump furo docs req to 2021.09.08

bastimeyer:
- upgrade furo to 2021.08.31
- fix/restore custom theme overrides

----

The furo theme had its first stable release last week.
https://github.com/pradyunsg/furo/releases

Most of the changes are only bugfixes, but there's also a custom theme variant selection now which makes it worth upgrading. The only noticeable difference should be a short scroll animation for link anchors, the rest should look and behave the same except for very minor spacing changes.

This version of furo also requires sphinx 4 now, but that shouldn't be a problem anymore after 5e65d2d9227814ea303300b4dd7269c215fa1865 fixed a build option which caused problems between sphinx 3 and 4.

Please carefully compare the docs preview with the live docs before merging. There's a chance that I missed something that needs fixing because I didn't check every change in the changelog. Only tested in latest Chromium and Firefox.

### DIFF
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,3 +1,3 @@
 sphinx>=3.0
-furo==2020.12.28.beta23
+furo==2021.09.08
 recommonmark>=0.5.0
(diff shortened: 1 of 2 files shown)


docs-requirements.txt
```diff
@@ -1,3 +1,3 @@
 sphinx>=3.0
-furo==2020.12.28.beta23
+furo==2021.08.31
```
back-to, line 2: could be updated

https://github.com/pradyunsg/furo/releases/tag/2021.09.08

bastimeyer: This got released two hours after I made this PR. The only relevant change here is markup related (theme button at the top).

Issue #4002, opened 2021-09-08
vidio plugin error [https://www.vidio.com/live/204-sctv]


rollykaunang:
### Checklist

- [X] This is a plugin issue and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

[cli][error] Try 1/1: Could not open stream <HLSStream

### Debug log

```text
C:\WINDOWS\system32>streamlink https://www.vidio.com/live/204-sctv best --logleve debug
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.9.6
[cli][debug] Streamlink: 2.4.0
[cli][debug] Requests(2.26.0), Socks(1.7.1), Websocket(1.2.1)
[cli][debug] Arguments:
[cli][debug]  url=https://www.vidio.com/live/204-sctv
[cli][debug]  stream=['best']
[cli][debug]  --loglevel=debug
[cli][debug]  --rtmp-rtmpdump=rtmpdump.exe
[cli][debug]  --rtmpdump=C:\Program Files (x86)\Streamlink\rtmpdump\rtmpdump.exe
[cli][debug]  --ffmpeg-ffmpeg=C:\Program Files (x86)\Streamlink\ffmpeg\ffmpeg.exe
[cli][info] Found matching plugin vidio for URL https://www.vidio.com/live/204-sctv
[plugins.vidio][debug] Getting stream tokens
[plugins.vidio][debug] HLS URL: https://app-etslive-2.vidio.com/live/204/master.m3u8
[...]

PR #4003, opened 2021-09-10
http_session: override urllib3 percent-encoding

bastimeyer:
According to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1), URLs are considered equal even if they differ in the case of the hexadecimal digits of percent-encoded characters, eg. `%FF` == `%ff`. URL normalizers **should** always use uppercase characters, which is why `urllib3` correctly transforms those lowercase digits to uppercase ones.

This is a problem though for sites which compare the request URL (or parts of it) byte for byte and return different responses depending on that. See https://github.com/streamlink/streamlink/issues/4002#issuecomment-916038530

This PR therefore adds overrides for `urllib3` which make it not transform hexadecimal digits of percent-encoded characters to uppercase. Since URLs are still considered equal, this shouldn't break anything and fix the problem mentioned above. If the original behavior is required, then request URLs can still be transformed manually. HTTP redirections could be a problem, but the risk is rather low.

Streamlink currently requires `urllib3>=1.21.1,<1.27`, as defined by [`requests>=2.26.0`](https://github.com/streamlink/streamlink/blob/2.4.0/setup.py#L13) here:
https://github.com/psf/requests/blob/v2.26.0/setup.py#L48

There are two different overrides for `urllib3>=1.25.8` and `urllib3>=1.25.4,<1.25.8`, as noted in the code comments, because `urllib3==1.25.8` had a minor revision of the string substitution. Versions of urllib3 prior to `1.25.4` unfortunately are using external libs for modifying the percent-encoding stuff and I'm not sure if the problem can be fixed there.

The added tests are working fine in both version ranges when the overrides are applied and will only fail in the lowercase test fixture when no override is applied, which is expected.

Should I add a test runner for `urllib3==1.25.4` so that it can test both version ranges?

### DIFF
--- a/src/streamlink/plugin/api/http_session.py
+++ b/src/streamlink/plugin/api/http_session.py
@@ -1,4 +1,5 @@
 import time
+from typing import Any, Callable, List, Pattern, Tuple
 
 import requests.adapters
 import urllib3
@@ -8,6 +9,9 @@ from streamlink.exceptions import PluginError
 from streamlink.packages.requests_file import FileAdapter
 from streamlink.plugin.api import useragents
 from streamlink.utils import parse_json, parse_xml
+
+
+urllib3_version = tuple(map(int, urllib3.__version__.split(".")[:3]))
 
 
 try:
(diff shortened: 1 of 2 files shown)


rollykaunang:
Thank You Bastimeyer and Back-to, The vidio plugin is working again... Good Job... For your hard working off course I will donate again.. Thank You So Much!

PR #4008, opened 2021-09-12
utils: replace LazyFormatter with new Formatter

bastimeyer:
- Replace utils.lazy_formatter with cli.utils.formatter, add the generic
  `format` method and create a Formatter subclass in cli.utils.formatter
  with additional `title` and `filename` methods for usage in cli.main.
- Update HLSStreamWriter.create_decryptor, which was the last remaining
  code using the old LazyFormatter.

----

By moving the code, `streamlink.utils.formatter` is now being tested via the subclass in `streamlink_cli.utils.formatter`, but that's alright because of the two additional public methods in the subclass we need to test anyway. No need to add two separate test files.

The changes in `HLSStreamWriter.create_decryptor` are [fully tested](https://github.com/streamlink/streamlink/blob/5d9964d9dc4f1dc6586e7fd2fa330ed087d3fec1/tests/streams/test_hls.py#L207-L224) and [covered](https://codecov.io/gh/streamlink/streamlink/src/5d9964d9dc4f1dc6586e7fd2fa330ed087d3fec1/src/streamlink/stream/hls.py#L61).


src/streamlink_cli/utils/formatter.py
```diff
@@ -1,63 +1,12 @@
...
-
-            if field_name is None:
-                continue
-
-            value = self._get_value(field_name, format_spec, defaults)
-            result.append(mapper(str(value)))
-
-        return "".join(result)
-
+class Formatter(_BaseFormatter):
```
back-to, line 7: two `Formatter` with the same name seems suboptimal,
might just remove this file and use only `src/streamlink/utils/formatter.py`

bastimeyer: The reason for this are the `title` and `filename` methods on the subclass which keep the original interface for the CLI stuff. `filename` depends on `streamlink_cli.utils.path` which has the logic for file name substitutions and belongs into the CLI module.

PR #4020, opened 2021-09-15
plugins.pluzz: rewrite plugin

### LINKED ISSUE #4017: plugins.pluzz: doesn't work anymore
### Checklist

### Streamlink version

Latest stable release

### Description

This plugin stopped working last week. The workflow is the same as current plugin, but API_URL and the json response format have changed, leading to errors and failure to get the streams

I've fixed the code and it's working so far on live streams and on the few vod videos i've checked. I can do a PR if you want, but i won't be able to perform current functionalities coverage evaluation before a long time.

### Debug log

```text
[cli][debug] OS:         Linux-5.10.0-8-amd64-x86_64-with-glibc2.31
[cli][debug] Python:     3.9.7
[cli][debug] Streamlink: 2.4.0
[cli][debug] Arguments:
[cli][debug]  url=https://www.france.tv/france-3/direct.html
[cli][debug]  stream=['best']
[cli][info] Found matching plugin pluzz for URL https://www.france.tv/france-3/direct.html
[plugins.pluzz][debug] Country: FR
[plugins.pluzz][debug] Video ID: 29bdf749-7082-4426-a4f3-595cc436aa0d
error: Unable to open URL: http://sivideo.webservices.francetelevisions.fr/tools/getInfosOeuvre/v2/?idDiffusion=29bdf749-7082-4426-a4f3-595cc436aa0d (404 Client Error: Not Found for url: http://sivideo.webservices.francetelevisions.fr/tools/getInfosOeuvre/v2/?idDiffusion=29bdf749-7082-4426-a4f3-595cc436aa0d)
```

bastimeyer:
Resolves #4017 

This is a complete plugin rewrite which was necessary because of the API changes and video-id extraction.

This also removes the unused HDS streams, subtitle muxing and specific error messages for geo-blocked or expired content. Since ludo.fr and zouzous.fr were both redirecting to france.tv, I removed them as well.

I've tested all of the URLs listed in the test file and removed those which returned a 4xx.

According to some API URL tests via curl, it looked like all those query parameters were required to get valid JSON responses. This is reconstructed from the site's JS code, but since some parameters are optional and since there's also a timezone parameter, I can not guarantee that this is working 100% correcly for all the content.

It's possible that subtitle muxing could be re-implemented at some point, but I couldn't manage to find any useful data in the API responses apart from an empty `captions` array.

### DIFF
--- a/src/streamlink/plugins/pluzz.py
+++ b/src/streamlink/plugins/pluzz.py
@@ -1,216 +1,132 @@
 import logging
 import re
-import sys
-import time
+from datetime import datetime
+from urllib.parse import urlparse
 
-from streamlink.plugin import Plugin, PluginArgument, PluginArguments, pluginmatcher
-from streamlink.plugin.api import validate
-from streamlink.stream import DASHStream, HDSStream, HLSStream, HTTPStream
-from streamlink.stream.ffmpegmux import MuxedStream
+from isodate import LOCAL as LOCALTIMEZONE
+
+from streamlink.plugin import Plugin, PluginError, pluginmatcher
+from streamlink.plugin.api import useragents, validate
+from streamlink.stream import DASHStream, HLSStream
+from streamlink.utils.url import update_qsd
 
 log = logging.getLogger(__name__)
 
 
-@pluginmatcher(re.compile(r'''
-    https?://(
-        (?:www\.)?france\.tv/.+\.html
+@pluginmatcher(re.compile(r"""
+    https?://(?:
+        (?:www\.)?france\.tv/
         |
-        www\.(ludo|zouzous)\.fr/heros/[\w-]+
-        |
-        (.+\.)?francetvinfo\.fr
+        (?:.+\.)?francetvinfo\.fr/
     )
-''', re.VERBOSE))
+""", re.VERBOSE))
 class Pluzz(Plugin):
-    GEO_URL = 'http://geo.francetv.fr/ws/edgescape.json'
-    API_URL = 'http://sivideo.webservices.francetelevisions.fr/tools/getInfosOeuvre/v2/?idDiffusion={0}'
-    TOKEN_URL = 'http://hdfauthftv-a.akamaihd.net/esi/TA?url={0}'
-    SWF_PLAYER_URL = 'https://staticftv-a.akamaihd.net/player/bower_components/player_flash/dist/' \
-                     'FranceTVNVPVFlashPlayer.akamai-7301b6035a43c4e29b7935c9c36771d2.swf'
+    PLAYER_VERSION = "5.51.35"
+    GEO_URL = "https://geoftv-a.akamaihd.net/ws/edgescape.json"
+    API_URL = "https://player.webservices.francetelevisions.fr/v1/videos/{video_id}"
 
-    _pluzz_video_id_re = re.compile(r'''(?P<q>["']*)videoId(?P=q):\s*["'](?P<video_id>[^"']+)["']''')
-    _jeunesse_video_id_re = re.compile(r'playlist: \[{.*?,"identity":"(?P<video_id>.+?)@(?P<catalogue>Ludo|Zouzous)"')
-    _sport_video_id_re = re.compile(r'data-video="(?P<video_id>.+?)"')
-    _embed_video_id_re = re.compile(r'href="http://videos\.francetv\.fr/video/(?P<video_id>.+?)(?:@.+?)?"')
-    _hds_pv_data_re = re.compile(r"~data=.+?!")
-    _mp4_bitrate_re = re.compile(r'.*-(?P<bitrate>[0-9]+k)\.mp4')
-
-    _geo_schema = validate.Schema({
-        'reponse': {
-            'geo_info': {
-                'country_code': validate.text
-            }
-        }
-    })
-
-    _api_schema = validate.Schema({
-        'videos': validate.all(
-            [{
-                'format': validate.any(
-                    None,
-                    validate.text
-                ),
-                'url': validate.any(
-                    None,
-                    validate.url(),
-                ),
-                'statut': validate.text,
-                'drm': bool,
-                'geoblocage': validate.any(
-                    None,
-                    [validate.all(validate.text)]
-                ),
-                'plages_ouverture': validate.all(
-                    [{
-                        'debut': validate.any(
-                            None,
-                            int
-                        ),
-                        'fin': validate.any(
-                            None,
-                            int
-                        )
-                    }]
-                )
-            }]
-        ),
-        'subtitles': validate.any(
-            [],
-            validate.all(
-                [{
-                    'type': validate.text,
-                    'url': validate.url(),
-                    'format': validate.text
-                }]
-            )
-        )
-    })
-
-    _player_schema = validate.Schema({'result': validate.url()})
-
-    arguments = PluginArguments(
-        PluginArgument("mux-subtitles", is_global=True)
-    )
+    _re_ftv_player_videos = re.compile(r"window\.FTVPlayerVideos\s*=\s*(?P<json>\[{.+?}])\s*;\s*(?:$|var)", re.DOTALL)
+    _re_player_load = re.compile(r"""player\.load\s*\(\s*{\s*src\s*:\s*(['"])(?P<video_id>.+?)\1\s*}\s*\)\s*;""")
 
     def _get_streams(self):
+        self.session.http.headers.update({
+            "User-Agent": useragents.CHROME
+        })
+        CHROME_VERSION = re.compile(r"Chrome/(\d+)").search(useragents.CHROME).group(1)
+
         # Retrieve geolocation data
-        res = self.session.http.get(self.GEO_URL)
-        geo = self.session.http.json(res, schema=self._geo_schema)
-        country_code = geo['reponse']['geo_info']['country_code']
-        log.debug('Country: {0}'.format(country_code))
+        country_code = self.session.http.get(self.GEO_URL, schema=validate.Schema(
+            validate.parse_json(),
+            {"reponse": {"geo_info": {
+                "country_code": str
+            }}},
+            validate.get(("reponse", "geo_info", "country_code"))
+        ))
+        log.debug(f"Country: {country_code}")
 
         # Retrieve URL page and search for video ID
-        res = self.session.http.get(self.url)
-        if 'france.tv' in self.url:
-            match = self._pluzz_video_id_re.search(res.text)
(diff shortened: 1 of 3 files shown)


bastimeyer:
@FlorianIragne could you please check this PR and see if it's working for you?

FlorianIragne:
Hi,

yes it's working. Thanks for your prompt action.
FYI, I've found out that the only required query string parameters for API_URL are device_type and browser (tested on all your updated test urls).
Thanks!

melmorabity:
> FYI, I've found out that the only required query string parameters for API_URL are device_type and browser (tested on all your updated test urls).

Actually `country_code` is still mandatory, as well as `os`. By the way:
* `os=android` will retrieve DASH streams (or HLS for a few old videos)
* `os=ios` will return HLS streams only
This may be useful to provide those 2 kind of streams.

bastimeyer:
> These two parameters are enough to query the FTV API

I'd prefer keeping all of them, because these were the ones which were not optional in the JS code which I debugged, even if some might not be necessary to get a proper response from the API server (see `constructGatewayUrl` down below - they include sourcemaps with their minified code... thanks, I guess). I definitely also had mixed results while testing via curl and selectively removing some of the parameters. Can't remember any specifics though.

> * `os=ios` will return HLS streams only

That would also require setting `device_type` to `mobile`, as they depend on each other. Mixing these values and generating queries which can't be built via the JS can be figured out on the server side, and the goal here is to blend in with legitimate requests.

Not sure how much sense it makes forcing HLS (if it's indeed correct what you say), but it might be the better choice. I'll have a look later today, need to go now.

Btw, there are three different types of API responses
- `workflow: "token-akamai"` (the one expected by this PR - `workflow` JSON property not checked)
- `workflow: "dai"` (ads? external/dailymotion?)
- `workflow: "token-vualto"` (DRM)

```js
// webpack://magnetoscope/./src/core/media/Media.js
import { of } from 'rxjs';
import { map, switchMap } from 'rxjs/operators';
import { playerInfo, requestAPI, systemInfo, videoUnlocker, deepMerge } from '../../utils';
import { formatGatewayError, formatGeolocError, formatTokenError, formatDRMError } from '../../error/formats';
import { geoloc, drm } from '../../configuration';

export default class Media {
  constructor(video) {
    this.video = video;
  }

  static addSystemInfoQueryString(sysInfo) {
    const sysInfoToSend = ['browser', 'browserVersion', 'os', 'osVersion', 'isMobile'];
    const camelCaseToUnderscore = s => s.replace(/([a-z])([A-Z])/g, '$1_$2').toLowerCase();

    /* iPad must have isMobile set to true */
    const isMobile = sysInfo.isMobile || (!sysInfo.isMobile && sysInfo.isTablet);
    const format = (k, v) => (k === 'isMobile' ? `device_type=${isMobile ? 'mobile' : 'desktop'}` : `${camelCaseToUnderscore(k)}=${v}`);

    return Object.entries(sysInfo)
      .filter(([key]) => sysInfoToSend.includes(key))
      .reduce((qs, [key, value]) => `${qs}&${format(key, value)}`, '');
  }

  static addDomainQueryString({ hostname } = {}, originUrl) {
    let domain = hostname;
    /* If an origin url is given, we want to track as if we were on the given product */
    if (originUrl) domain = new URL(originUrl).hostname;
    return `&domain=${domain || 'unknown'}`;
  }

  static constructGatewayUrl(gatewayUrl, {
    src, country, width, height, mode, gmt, originUrl, startOver, videoProductId
  }) {
    const { version } = playerInfo();
    const stdParams = `country_code=${country}&w=${width}&h=${height}&player_version=${version}`;
    const diffusionMode = mode ? `&diffusion_mode=${mode}` : '';
    const domain = Media.addDomainQueryString(window.location, originUrl);
    const startOverQS = startOver ? '&startover=true' : '';
    const sysInfo = Media.addSystemInfoQueryString(systemInfo);
    const gmtQS = gmt ? `&gmt=${encodeURIComponent(gmt.replace('GMT', ''))}` : '';
    const embed = originUrl ? '&embed=true' : '';
    const videoProductIdQS = videoProductId ? `&video_product_id=${videoProductId}` : '';
    return `${gatewayUrl}${src}?${stdParams}${domain}${sysInfo}${diffusionMode}${gmtQS}${embed}${startOverQS}${videoProductIdQS}`;
  }

  static isLive(media) {
    return Boolean(media.video.is_live
      && typeof media.video.is_live !== 'undefined'
      && media.video.is_live);
  }

  static handleGeoLoc(url) {
    return requestAPI({
      url,
      selector: ({ json: { reponse: { geo_info: { country_code: country } } } }) => ({ country }),
      formatError: formatGeolocError
    });
  }

  static handleGMT() {
    const offset = new Date().getTimezoneOffset();
    const sign = offset > 0 ? '-' : '+';
    return sign + `${parseInt(Math.abs(offset) / 60, 10)}`.padStart(2, '0') + `${Math.abs(offset) % 60}`.padStart(2, '0');
  }

  static handleWorkflow(workflows, handler) {
    return switchMap(media =>
      (workflows.includes(media.video.workflow) && media.video.token ?
        handler(media) :
        of(media)));
  }

  static handleToken(media) {
    const encodedUrl = encodeURIComponent(media.video.url);
    return requestAPI({
      url: `${media.video.token}&url=${encodedUrl}`,
      selector: ({ json: { url: tokenizedUrl } }) =>
        Object.assign({}, media, {
          video: Object.assign({}, media.video, {
            url: tokenizedUrl
          })
        }),
      formatError: formatTokenError
    });
  }

  static handleDrm(media) {
    return requestAPI({
      url: media.video.token,
      method: 'POST',
      body: {
        id: media.meta.id,
        drm_type: media.video.drm_type,
        license_type: 'online'
      },
      selector: ({ json: { token } }) => ({
        ...media,
        video: {
          ...media.video,
          drm: {
            token,
            playReadyLaUrl: drm.playReadyLaUrl,
            widevineLaUrl: drm.widevineLaUrl,
            fairplayLaUrl: drm.fairplayLaUrl
          }
        }
      }),
      formatError: formatDRMError
    });
  }

  static handleDai(media) {
    if (!media.markers.pub.mediaTailorUrl) return of(media);

    return requestAPI({
      url: `${media.markers.pub.mediaTailorUrl}?User-Agent=ios`,
      method: 'POST',
      body: {
        reportingMode: 'server'
      },
      selector: Media.daiSelector(media)
    });
  }

  static daiSelector(media) {
    return ({ json: { manifestUrl, trackingUrl } }) => {
      const awsId = manifestUrl.split('aws.sessionId=')[1];
      const amazonDomain = new URL(media.markers.pub.mediaTailorUrl).origin;
      const mediaUrl = `${media.video.url}?aws.sessionId=${awsId}`;
      const pollingUrl = `${amazonDomain}${trackingUrl}`;
      return {
        ...media,
        isDAI: true,
        video: {
          ...media.video,
          url: mediaUrl
        },
        markers: {
          ...media.markers,
          pub: {
            ...media.markers.pub,
            pollingUrl
          }
        }
      };
    };
  }

  static sanitizeMedia({ media, playerConfig, config }) {
    const mergeConfig = deepMerge(
      playerConfig,
      config,
      ((config.index >= 1 && !Object.prototype.hasOwnProperty.call(config, 'autostart')) ? { autostart: true } : {})
    );

    const { preload, autostart, diffusion } = mergeConfig;
    const isLive = Media.isLive(media);

    return (autostart ?
      videoUnlocker.getAutoplay(playerConfig) :
      of(false)
    ).pipe(map((autostartEnabled) => {
      /**
       * Autostart check :
       * - check autostart video config - if absent, it'll default to autostart: true
       * - check if autoplay is permitted on the current browser
       * - autoplay is usually blocked until the first user gesture, if we're in a "playlist"
       *   (france.tv tunnel) we can allow autostart on new video loads (video tags will be unlocked)
       *   by checking media.config.index
       */
      const shouldAutostart = mergeConfig.autostart && (autostartEnabled || config.index >= 1);
      return ({
        ...media,
        isLive,

        config: {
          ...mergeConfig,
          ...(diffusion ? ({
            diffusion: {
              ...diffusion,
              length: (diffusion.length ? diffusion.length : null) /* normalize tunnel length */
            }
          }) : {}),
          autostart: shouldAutostart,
          /* We preload if autostart true, or on desktop if preload true */
          preload: (!media.video.timeshiftable && autostart) || (!systemInfo.isMobile && !isLive && preload)
        }
      });
    }));
  }

  get stream$() {
    const {
      src, config, playerConfig, width, height
    } = this.video;

    const {
      diffusion: { mode } = {},
      originUrl,
      startOver,
      videoProductId
    } = { ...playerConfig, ...config };
    const gmt = Media.handleGMT();

    return Media.handleGeoLoc(geoloc.url)
      .pipe(switchMap(({ country }) => requestAPI({
        url: Media.constructGatewayUrl(
          playerConfig.webservices.gateway,
          {
            src: encodeURIComponent(src), country, width, height, mode, gmt, originUrl, startOver, videoProductId
          }
        ),
        formatError: formatGatewayError,
        selector: ({ json }) => json
      }).pipe(
        Media.handleWorkflow(['dai'], Media.handleDai),
        Media.handleWorkflow(['token-akamai', 'dai'], Media.handleToken),
        Media.handleWorkflow(['token-vualto'], Media.handleDrm),
        switchMap(media => Media.sanitizeMedia({ media, playerConfig, config }))
      )));
  }
}
```

marmeladema:
Thank you very much for working on this!
About subtitles, they never worked for me before, so I don't think that's a big deal.

bastimeyer:
Ok, I had another look at it and it does indeed force HLS with `os=ios`, so let's do it. However, it's a muxed HLS stream, so there's still the FFmpeg dependency here.

Also changed the API validation schema to include the `workflow` property and changed the output tuple.

Since this pretty much fixes the plugin, I think this can be merged now. If there are further improvements, they can be added later.

> About subtitles, they never worked for me before

Yeah, I noticed that when I implemented the `--mux-subtitles` global CLI argument that replaced the plugin-specific `*-mux-subtitles` ones.

src/streamlink/plugins/pluzz.py
```diff
@@ -1,216 +1,132 @@
...
 
 
-@pluginmatcher(re.compile(r'''
-    https?://(
-        (?:www\.)?france\.tv/.+\.html
-        |
-        www\.(ludo|zouzous)\.fr/heros/[\w-]+
+@pluginmatcher(re.compile(r"""
+    https?://(?:
+        (?P<francetv>(?:www\.)?france\.tv)/
```
back-to, line 18: 1. you don't use this groups in `pluginmatcher`, they could be removed
2. https://github.com/streamlink/streamlink/blob/master/docs/plugin_matrix.rst needs an update

Issue #4026, opened 2021-09-17
artetv: de/fr Livestreams aren't playable anymore


synaestic:
### Checklist

- [X] This is a plugin issue and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest build from the master branch

### Description

Since about a week the live channels aren't playable anymore. However VODs working fine.

### Debug log

```text
streamlink https://www.arte.tv/de/live/ worst -l debug
[cli][debug] OS:         Linux-5.14.3-arch1-1-x86_64-with-glibc2.33
[cli][debug] Python:     3.9.7
[cli][debug] Streamlink: 2.4.0+17.g24c59a2
[cli][debug] Requests(2.26.0), Socks(1.7.1), Websocket(0.59.0)
[cli][debug] Arguments:
[cli][debug]  url=https://www.arte.tv/de/live/
[cli][debug]  stream=['worst']
[cli][debug]  --loglevel=debug
[cli][info] Found matching plugin artetv for URL https://www.arte.tv/de/live/
error: No playable streams found on this URL: https://www.arte.tv/de/live/

streamlink https://www.arte.tv/fr/direct/ best -l debug
[cli][debug] OS:         Linux-5.14.3-arch1-1-x86_64-with-glibc2.33
[cli][debug] Python:     3.9.7
[cli][debug] Streamlink: 2.4.0+17.g24c59a2
[cli][debug] Requests(2.26.0), Socks(1.7.1), Websocket(0.59.0)
[cli][debug] Arguments:
[cli][debug]  url=https://www.arte.tv/fr/direct/
[cli][debug]  stream=['best']
[cli][debug]  --loglevel=debug
[cli][info] Found matching plugin artetv for URL https://www.arte.tv/fr/direct/
error: No playable streams found on this URL: https://www.arte.tv/fr/direct/
```

PR #4027, opened 2021-09-17
plugins.arte: switch to arte.tv v2 API

melmorabity:
The Arte.tv V1 API doens't seem to work anymore for live streams (see #4026).

Both web site and mobile app use the V2 API, which requires an authentication token. The one from the website is used here for this fix.

### DIFF
--- a/src/streamlink/plugins/artetv.py
+++ b/src/streamlink/plugins/artetv.py
@@ -9,24 +9,29 @@ from streamlink.plugin.api import validate
 from streamlink.stream import HLSStream
 
 log = logging.getLogger(__name__)
-JSON_VOD_URL = "https://api.arte.tv/api/player/v1/config/{0}/{1}?platform=ARTE_NEXT"
-JSON_LIVE_URL = "https://api.arte.tv/api/player/v1/livestream/{0}"
+JSON_VOD_URL = "https://api.arte.tv/api/player/v2/config/{0}/{1}"
+JSON_LIVE_URL = "https://api.arte.tv/api/player/v2/config/{0}/LIVE"
+V2_API_BEARER = "MzYyZDYyYmM1Y2Q3ZWRlZWFjMmIyZjZjNTRiMGY4MzY4NzBhOWQ5YjE4MGQ1NGFiODJmOTFlZDQwN2FkOTZjMQ"
 
 _video_schema = validate.Schema({
-    "videoJsonPlayer": {
-        "VSR": validate.any(
-            [],
-            {
-                validate.text: {
-                    "height": int,
-                    "mediaType": validate.text,
-                    "url": validate.text,
-                    "versionProg": int,
-                    "versionLibelle": validate.text
-                },
-            },
-        )
-    }
+    "data": {
+        "attributes": {
+            "streams": validate.any(
+                [],
+                [{
+                    "url": validate.url(),
+                    "slot": int,
+                    "protocol": validate.text,
+                    "versions": validate.all(
+                        [{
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/artetv.py
```diff
@@ -9,24 +9,29 @@
 from streamlink.stream import HLSStream
 
 log = logging.getLogger(__name__)
-JSON_VOD_URL = "https://api.arte.tv/api/player/v1/config/{0}/{1}?platform=ARTE_NEXT"
-JSON_LIVE_URL = "https://api.arte.tv/api/player/v1/livestream/{0}"
+JSON_VOD_URL = "https://api.arte.tv/api/player/v2/config/{0}/{1}"
+JSON_LIVE_URL = "https://api.arte.tv/api/player/v2/config/{0}/LIVE"
+V2_API_BEARER = "MzYyZDYyYmM1Y2Q3ZWRlZWFjMmIyZjZjNTRiMGY4MzY4NzBhOWQ5YjE4MGQ1NGFiODJmOTFlZDQwN2FkOTZjMQ"
```
bastimeyer, line 14: Please move the constants from the global scope to the class as class attributes

src/streamlink/plugins/artetv.py
```diff
@@ -9,24 +9,29 @@
...
+                    "versions": validate.all(
+                        [{
+                            "label": validate.text
+                        }],
+                        validate.transform(lambda x: x[0])
+                    )
+                }]
+            )
+        }
+    },
```
bastimeyer, line 34: Please move the validation schema as well. Preferably directly to the `schema` argument of the `session.http.get` call, so that the whole schema does only get built when actually using the Arte plugin, as Streamlink currently loads all plugins at once.

You can add `validate.parse_json()` as the first argument of `validate.Schema()` there to avoid the `session.http.json(res, schema)` call, which is much cleaner. `validate.text` should also be replaced with a simple `str`, as it's superfluous since 2.0.0. `validate.transform(lambda x: x[0])` can be replaced with a simple `validate.get(0)`.

src/streamlink/plugins/artetv.py
```diff
@@ -60,14 +65,14 @@ def _get_streams(self):
             json_url = JSON_LIVE_URL.format(language)
         else:
             json_url = JSON_VOD_URL.format(language, video_id)
-        res = self.session.http.get(json_url)
+        res = self.session.http.get(json_url, headers={"Authorization": "Bearer {}".format(V2_API_BEARER)})
```
bastimeyer, line 68: Use f-strings instead of `str.format()`.

src/streamlink/plugins/artetv.py
```diff
@@ -60,14 +65,14 @@ def _get_streams(self):
...
         video = self.session.http.json(res, schema=_video_schema)
 
-        if not video["videoJsonPlayer"]["VSR"]:
+        if not (vsr := video["data"]["attributes"]["streams"]):
```
bastimeyer, line 71: Assignment expressions in if-statements are only supported on py38 and above

src/streamlink/plugins/artetv.py
```diff
@@ -60,14 +65,14 @@ def _get_streams(self):
...
             return
 
-        vsr = video["videoJsonPlayer"]["VSR"]
         return self._create_stream(vsr)
 
 
 __plugin__ = ArteTV
+
```
bastimeyer, line 78: flake8: W391 blank line at end of file

Issue #4047, opened 2021-09-25
Make scheme-less input URLs use HTTPS instead of HTTP


bastimeyer:
### Checklist

- [X] This is a feature request and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22feature+request%22)

### Description

Streamlink has been treating scheme-less input URLs as `http` instead of `https` so far, but this is problematic:

1. For sites which redirect from http to https, this still requires an unencrypted initial request. HSTS (HTTP Strict-Transport-Security) doesn't have any effect, since this information doesn't get stored anywhere, so every time Streamlink is run with the same input URL, it'll always make this initial unencrypted http request.
2. For sites which don't redirect to https and support both unencrypted and encrypted connections for their content, this makes it unnecessarily insecure, as an encrypted connection could be used instead.
3. For sites which don't redirect to https and don't support unencrypted connections for their content, plugins have to explicitly force the https scheme for the input URL. Some plugins currently also implement the `update_scheme` utility method incorrectly, but that's a different issue.

The only concern here is for sites which don't support TLS/https at all. Those input URLs would need explicit `http://` schemes. That shouldn't be the problem of Streamlink though and encourage unencrypted connections everywhere else.

Another thing is inputs of unaware users which simply pass an input URL with the `http://` scheme prefix, even though the site is fully using TLS. Plugins should take care of this and always update the scheme of the input URL. That would require an update of said utility method or the addition of a new/simpler one.

Btw, Chromium already forces https for every input URL with its current beta version and will only fall back if it's not supported by the site. Maybe this could also be done here as well at a lower level, in the http session class, as an alternative to plugins having to do it on their own.

PR #4050, opened 2021-09-26
utils.url: make update_scheme always update target

bastimeyer:
The previous implementation of update_scheme only made it set schemes
on scheme-less URLs and it kept target URLs that include a scheme intact.

This is confusing for plugin implementers and it has already been used
incorrectly a lot of times when trying to force the https protocol on
unencrypted http URLs. Since the method is called update_scheme, it
should do exactly that, in all cases.

- always update target URL scheme with current URL scheme
- parametrize tests, re-order assertions and add new ones

----

ref #4047

It is possible that this could cause issues when this pattern is used where `self.url` is a `http://` URL and the target is a `https://` URL where `https` is required and the server doesn't redirect, meaning the target will be degraded from https to http:
```
update_scheme(self.url, target)
```

There are 13 matches when searching for `update_scheme(self.url` in `src/`:
```
$ grep -RHn 'update_scheme(self.url' src/ | sed -E 's/ +/ /g' | column -tl2
src/streamlink/plugins/nbc.py:20:          url = update_scheme(self.url, platform_url)
src/streamlink/plugins/nbcsports.py:20:    url = update_scheme(self.url, platform_url)
src/streamlink/plugins/albavision.py:89:   playlist_url = m and update_scheme(self.url, m.group(1))
src/streamlink/plugins/albavision.py:116:  playlist_url = m and update_scheme(self.url, m.group(1))
src/streamlink/plugins/cdnbg.py:65:        iframe_url = update_scheme(self.url, html_unescape(iframe_url))
src/streamlink/plugins/delfi.py:41:        src = update_scheme(self.url, x['src'])
src/streamlink/plugins/dogus.py:39:        mobile_url = mobile_url_m and update_scheme(self.url, mobile_url_m.group("url"))
src/streamlink/plugins/invintus.py:41:     return HLSStream.parse_variant_playlist(self.session, update_scheme(self.url, hls_url))
src/streamlink/plugins/mediaklikk.py:70:   validate.map(lambda p: update_scheme(self.url, p["file"]))
src/streamlink/plugins/streamable.py:33:   stream_url = update_scheme(self.url, info["url"])
src/streamlink/plugins/vk.py:56:           iframe_url = update_scheme(self.url, _i.attributes['src'])
src/streamlink/plugins/webtv.py:64:        url = update_scheme(self.url, source["src"])
src/streamlink/plugins/earthcam.py:61:     hls_url = update_scheme(self.url, f"{hls_domain}{hls_playpath}")
```

and 7 matches for `update_scheme("http://`:
```
$ grep -RHnE "update_scheme\([\"']http://" src/ | sed -E 's/ +/ /g' | column -tl2
src/streamlink/plugins/akamaihd.py:18:  url = update_scheme("http://", data.get("url"))
src/streamlink/plugins/hds.py:21:       url = update_scheme("http://", data.get("url"))
src/streamlink/plugins/hls.py:21:       url = update_scheme("http://", data.get("url"))
src/streamlink/plugins/http.py:18:      url = update_scheme("http://", data.get("url"))
src/streamlink/session.py:234:          self.http.proxies["http"] = update_scheme("http://", value)
src/streamlink/session.py:236:          self.http.proxies["https"] = update_scheme("http://", value)
src/streamlink/session.py:356:          url = update_scheme("http://", url)
```

### DIFF
--- a/src/streamlink/utils/url.py
+++ b/src/streamlink/utils/url.py
@@ -25,12 +25,12 @@ _re_uri_implicit_scheme = re.compile(r"""^[a-z0-9][a-z0-9.+-]*://""", re.IGNOREC
 
 def update_scheme(current: str, target: str) -> str:
     """
-    Take the scheme from the current URL and apply it to the
-    target URL if the target URL starts with // or is missing a scheme
+    Take the scheme from the current URL and apply it to the target URL
     :param current: current URL
     :param target: target URL
-    :return: target URL with the current URLs scheme
+    :return: target URL with the current URL's scheme
     """
+    current_p = urlparse(current)
     target_p = urlparse(target)
 
     if (
@@ -42,14 +42,10 @@ def update_scheme(current: str, target: str) -> str:
         # target URLs without scheme and netloc: ("http://", "foo.bar/foo") -> "http://foo.bar/foo"
         or not target_p.scheme and not target_p.netloc
     ):
-        return f"{urlparse(current).scheme}://{urlunparse(target_p)}"
-
-    # target URLs without scheme but with netloc: ("http://", "//foo.bar/foo") -> "http://foo.bar/foo"
-    if not target_p.scheme and target_p.netloc:
-        return f"{urlparse(current).scheme}:{urlunparse(target_p)}"
+        return f"{current_p.scheme}://{urlunparse(target_p)}"
 
     # target URLs with scheme
-    return target
+    return urlunparse(target_p._replace(scheme=current_p.scheme))
 
 
 def url_equal(first, second, ignore_scheme=False, ignore_netloc=False, ignore_path=False, ignore_params=False,
--- a/tests/utils/test_url.py
+++ b/tests/utils/test_url.py
@@ -23,14 +23,17 @@ def test_prepend_www(url, expected):
     assert expected == prepend_www(url)
 
 
-def test_update_scheme():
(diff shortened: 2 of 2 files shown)


bastimeyer:
Well, it does exactly that in the `tests/plugins/test_stream.py` tests when it tries to resolve the test stream URLs, because `Session.resolve_url()` calls `update_url("http://", url)`.
https://github.com/streamlink/streamlink/blob/c8f0e432368a11a43eaf4bef4bd9882760ae5318/src/streamlink/session.py#L356

This means that `Session.resolve_url()` needs to be updated first according to #4047 (or the failing tests updated)

bastimeyer:
Would adding a `force=False` keyword argument make sense, so that it'll only update the target when it has a scheme if force is True? Then we don't break the current usages and can set force=True where we actually want to override the scheme. That wouldn't solve the issue of the weird method name/implementation though.

bastimeyer:
I'm going to open a new PR with an update later

mkbloke:
> That wouldn't solve the issue of the weird method name/implementation though.

I was thinking perhaps the existing `update_scheme()` might better be called something like `add_scheme()` instead, then a new method could be added such as `upgrade_scheme()` or some such for scheme upgrades.  Swings and roundabouts, as they say - the `force` keyword also works...

PR #4065, opened 2021-10-03
plugins.showroom: cleanup

back-to:
- remove RTMP
- cleanup room_id
- error message for 403 eror on playlist

### DIFF
--- a/src/streamlink/plugins/showroom.py
+++ b/src/streamlink/plugins/showroom.py
@@ -20,12 +20,6 @@ class ShowroomHLSStreamReader(HLSStreamReader):
 class ShowroomHLSStream(HLSStream):
     __reader__ = ShowroomHLSStreamReader
 
-    @classmethod
-    def _get_variant_playlist(cls, res):
-        if res.headers["Content-Type"] != "application/x-mpegURL":  # content_region_permission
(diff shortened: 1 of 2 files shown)


bastimeyer:
```
[cli][info] Found matching plugin showroom for URL https://www.showroom-live.com/adbbc5792409
[plugins.showroom][debug] {'is_live': 1, 'room_id': 374348, 'room': {'content_region_permission': 1, 'is_free': 0}}
error: Failed to parse playlist: invalid Content-Type image/png
```

back-to:
> error: Failed to parse playlist: invalid Content-Type image/png

you are geo-blocked, works with an IP from HongKong

bastimeyer:
Yes, I am aware. The HLS playlist shows a PNG when geo-blocked. The plugin shouldn't return an error though.

If there's no other way of checking whether the user is geo-blocked, the plugin should send a `HEAD` request first to see what the content-type header is, so that `HLSStream.parse_variant_playlist` doesn't raise an error.

Btw, `validate.text` should be replaced with `str` in the validation schema.

back-to:
`HEAD` won't work, it is always `error 405`
only `GET` will work

PR #4068, opened 2021-10-03
session: move from http to https as default scheme

### LINKED ISSUE #4047: Make scheme-less input URLs use HTTPS instead of HTTP

bastimeyer:
Apply the https scheme to all input URLs if the scheme is missing:
- session:
  - Session.resolve_url()
  - session option `http-proxy` (`--http-proxy` CLI argument)
- plugins.akamaihd: `akamaihd://URL`
- plugins.dash: `dash://URL` or `URL.mpd`
- plugins.hds: `hds://URL` or `URL.f4m`
- plugins.hls: `hls://URL` or `URL.m3u8`
- plugins.http: `httpstream://URL`

Regular http URLs (non-https/TLS) will need to be set explicitly now.

Also
- update/fix test_session
- refactor/fix test_dash
- refactor/fix test_stream

----

Resolves #4047 
Follow-up of #4053

- **Does this introduce a breaking change?**
  This changes the way user input is handled. In almost all cases, this simply avoids the initial insecure http request when no scheme was defined, which then gets redirected by the server to a secure connection, so no breaking change here. For sites though which don't support https/TLS, setting the http scheme is now mandatory, which is the reason why I'm not sure if this is a breaking change. It probably is...
- **Does updating the `http-proxy` option make sense?**
  If we change the stream URL input, then we should change this too, IMO.
  Also, having two separate options for http and https is stupid and we should merge this at some point.

### DIFF
--- a/src/streamlink/plugins/akamaihd.py
+++ b/src/streamlink/plugins/akamaihd.py
@@ -15,7 +15,7 @@ log = logging.getLogger(__name__)
 class AkamaiHDPlugin(Plugin):
     def _get_streams(self):
         data = self.match.groupdict()
-        url = update_scheme("http://", data.get("url"), force=False)
+        url = update_scheme("https://", data.get("url"), force=False)
         params = parse_params(data.get("params"))
         log.debug(f"URL={url}; params={params}")
 
--- a/src/streamlink/plugins/dash.py
+++ b/src/streamlink/plugins/dash.py
@@ -4,6 +4,7 @@ import re
 from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.plugin import LOW_PRIORITY, stream_weight
 from streamlink.stream.dash import DASHStream
+from streamlink.utils.url import update_scheme
 
 log = logging.getLogger(__name__)
 
@@ -28,7 +29,7 @@ class MPEGDASH(Plugin):
             return stream_weight(stream)
 
     def _get_streams(self):
-        url = self.match.group(1)
+        url = update_scheme("https://", self.match.group(1), force=False)
         log.debug(f"Parsing MPD URL: {url}")
 
         return DASHStream.parse_manifest(self.session, url)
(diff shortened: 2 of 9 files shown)


gravyboat:
As much as I hate to say it I do think we should consider this a breaking change as stupid as that may sound. I don't see any reason not to merge this or to delay though, we have no obligation to avoid introducing breaking changes.

bastimeyer:
We're not breaking sites/plugins, we're breaking user inputs. Anything non-https does still work if `http://` gets set explicitly.

I would've preferred waiting a bit before merging, but it's too late now.

Since this is a breaking change, we should see what we can include in the next release that will be a breaking change as well, to avoid having lots of unnecessary major version bumps. It doesn't have to be everything marked as deprecated though (considering the time of the deprecation)

gravyboat:
@bastimeyer My bad I didn't realize from the tone of your comment that you wanted to wait and thought you were noting it was a breaking change that should be merged any way.

PR #4071, opened 2021-10-05
python 3.10

bastimeyer:
Python 3.10 has been released yesterday and just been pushed to [actions/python-versions](https://github.com/actions/python-versions/commit/2dce5239a363de4bf918b343ff8538b8a8670856), so actions/setup-python can install it.

I've also included a commit which adds 3.10 to the package classifiers list in setup.py, but I'm not sure if we should add it yet until wheels for all OSes of all dependencies are available ([lxml for example](https://pypi.org/project/lxml/#files)). It does only matter though if we decide to publish a new release without this being resolved first.


gravyboat:
This looks good. Do you want to merge it or wait? I don't want to merge early again like the other PR.

gravyboat (approved):

bastimeyer:
We're currently still using a workaround for installing lxml on cp310-win_amd64, because official wheels are not available and it's failing to install from source due to issues with libxml2 (#4009).

I've contacted the lxml devs about the lack of a pre-built wheels and I would like to see this being resolved first before we should include the second commit, because that advertises py 3.10 readiness, which isn't 100% true yet for all of Streamlink's dependencies.

I'm going to remove the second commit. The CI runners should still be updated (with the workaround included). Once the deps issues get resolved, we can update the package classifiers list and remove the lxml install workaround.

PR #4072, opened 2021-10-06
script: fix update-removed-plugins bash script

bastimeyer:
Take the current git status into consideration when updating the list of removed plugins, so that the script can be run before committing a plugin removal.


gravyboat:
Nice catch! This is a great change.

Issue #4040, opened 2021-09-23
Removal of Flash related streaming protocols


bastimeyer:
### Checklist

- [X] This is a feature request and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22feature+request%22)

### Description

I'm opening this thread as a proposal / checklist for the eventual removal of Flash related streaming protocols from Streamlink.

With the [official deprecation of Flash by Adobe in mid 2017](https://en.wikipedia.org/wiki/Adobe_Flash#End_of_life), streaming protocols based around Flash/RTMP have become obsolete. [Flash has already reached its end of life at the end of 2020 and has been removed from web browsers entirely](https://en.wikipedia.org/wiki/Adobe_Flash#End_of_life), meaning it has become irrelevant and so have the associated streaming protocols.

Streaming websites have switched to HLS or DASH a long time ago, and those few sites which are still using the obsolete streaming protocols are either outdated/unmaintained or located in regions where web technology doesn't move as fast (or at all).

Streamlink currently still includes `RTMP`, `HDS`, `AkamaiHD` and custom stream implementations. These either require third party binaries (RTMPdump) or libraries packaged with Streamlink ([`streamlink.packages.flashmedia`](https://github.com/streamlink/streamlink/commits/master/src/streamlink/packages/flashmedia)). Neither of these stream implementations or libraries are maintained.

At the time of writing this post, there are still a couple of plugins left which import those obsolete stream implementations. This is either because the plugins haven't been updated in the past several years and already include alternatives, or because some sites are still requiring them.

Then there are also the stream-protocol-plugins, which raises the question of how relevant these plugins are today if almost every site or application has switched to a more modern streaming protocol. For RTMP there's the exception that it's still being used for publishing streams to certain broadcasting sites, so I could see a potential use case here, but I'm not entirely sure how relevant it is for Streamlink.

----

## RTMPStream

- [x] [rtmp](https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugins/rtmp.py) - protocol-plugin
- [x] [app17](https://github.com/streamlink/streamlink/blob/master/src/streamlink/plugins/app17.py) - HLS implemented
[...]

PR #4076, opened 2021-10-07
plugins.tv5monde: remove plugin

bastimeyer:
- live TV stream is 404 for a while now
- VODs do not work on main site
- VODs do not work on "plus" site
- "afrique" domain is now a subdomain on the main site
- "tivi" stuff still works

If the live stream wasn't 404 I would've rewritten the plugin (#4040), but without it, it's just a VOD plugin, which we don't want to include. Once they fix it, this can be re-implemented if requested, but since nobody reported a plugin issue, I doubt it'll happen any time soon.


bastimeyer:
I used a French IP address of course. Looks like it's 404 on `https` but works fine on `http`. Their site is using the https link though, which made me believe it was offline.

```
$ streamlink --http-proxy socks5://localhost:1920 'https://live.tv5monde.com/fbs.html?'
[cli][info] Found matching plugin tv5monde for URL https://live.tv5monde.com/fbs.html?
error: Unable to open URL: https://live.tv5monde.com/fbs.html? (404 Client Error: Not Found for url: https://live.tv5monde.com/fbs.html)
```

```
$ streamlink --http-proxy socks5://localhost:1920 'http://live.tv5monde.com/fbs.html?'
[cli][info] Found matching plugin tv5monde for URL http://live.tv5monde.com/fbs.html?
Available streams: 450k_alt (worst), 450k, 650k_alt, 650k, 860k_alt, 860k, 1200k_alt, 1200k, 1500k_alt, 1500k, 2300k_alt, 2300k (best)
```

I'll have a look at a re-implementation as soon as I get the time (still have some plugin changed stashed from a couple of days ago). We could also revert this commit, but I don't really see the point since the other stuff is broken.

PR #4080, opened 2021-10-09
plugins.twitch: add support for passing auth credentials

ZipFile:
Recent events revealed that `X-Device-Id` is used not only for statistics and being authorized directly relates to whether you should be served server-side ads or not. Let's add this function back. It's no use for regular users, but for those who automate stream recording or build tools around streamlink - should be handy.

Alternatively, `http-headers` option could be used, but if I read the documentation right, it will inject header into all the requests (including hls stream), which is undesirable (this kind of auth is specific only to twitch gql api).

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -185,12 +185,33 @@ class UsherService:
 
 
 class TwitchAPI:
-    headers = {
-        "Client-ID": "kimne78kx3ncx6brgo4mv6wki5h1ko",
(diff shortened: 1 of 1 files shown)


ZipFile:
* Rebased with master
* Renamed `--twitch-auth-token` to `--twitch-oauth-token`
* Added caching for headers property (when `TwitchAPI` is initialized right now options are not loaded yet)
* Not touching documentation yet, awaiting feedback from other maintainers.

PR #4086, opened 2021-10-10
plugins.twitch: add device-id headers

back-to:
`server_ads` is `False` and should skip ADs for now,
also show some debug informations.

```
{
    'adblock': False,
    'geoblock_reason': '',
    'expires': 163300000,
    'hide_ads': False,
    'server_ads': False,
    'show_ads': True
}
```

Ref https://github.com/pixeltris/TwitchAdSolutions/blob/master/bypass/bypass-ublock-origin.js
Ref https://github.com/streamlink/streamlink/issues/3210

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -146,7 +146,7 @@ class UsherService:
         self.session = session
 
     def _create_url(self, endpoint, **extra_params):
-        url = "https://usher.ttvnw.net{0}".format(endpoint)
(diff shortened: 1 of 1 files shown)


bastimeyer:
Can someone confirm?
With the device headers, I get lots of short midroll ads but no preroll ads, but if I remove the headers again, I only get preroll ads but not midroll ads.

Midroll ads are far worse for Streamlink's output due to the stream discontinuities. Filtering out preroll ads just delays the stream start.

Issue #4102, opened 2021-10-19
ERROR: test_open_windows (tests.test_output.TestFileOutput) ... AttributeError: <module 'streamlink_cli.output' does not have the attribute 'msvcrt'


mjevans:
### Checklist

- [X] This is a bug report and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest build from the master branch

### Description

I use the archlinux package (streamlink-git) which pulls from the master / main branch on github.  By default the build script runs the unit tests too.

It looks like test_open_windows() is generating an AttributeError because of msvcrt, which only exists for Micorosft Windows platforms.

Example of working
https://github.com/streamlink/streamlink/blob/24c59a23103922977991acc28741a323d8efa7a1/tests/utils/test_named_pipe.py#L119

BUG
https://github.com/streamlink/streamlink/blob/e152c539ebcd64605cab01fe1bc18db86d98b269/tests/test_output.py#L77

Though it would be more future proof if is_Nix or similar existed rather than just is_win32 vs not is_win32.

### Debug log

```text
test_close_before_open (tests.utils.test_named_pipe.TestNamedPipeWindows) ... skipped 'test only applicable on Windows'
test_create (tests.utils.test_named_pipe.TestNamedPipeWindows) ... skipped 'test only applicable on Windows'
test_export (tests.utils.test_named_pipe.TestNamedPipeWindows) ... skipped 'test only applicable on Windows'
test_named_pipe (tests.utils.test_named_pipe.TestNamedPipeWindows) ... skipped 'test only applicable on Windows'
test_parse_json (tests.utils.test_parse.TestUtilsParse) ... ok
test_parse_qsd (tests.utils.test_parse.TestUtilsParse) ... ok
test_parse_xml (tests.utils.test_parse.TestUtilsParse) ... ok
test_parse_xml_entities (tests.utils.test_parse.TestUtilsParse) ... ok
test_parse_xml_entities_fail (tests.utils.test_parse.TestUtilsParse) ... ok
[...]

PR #4107, opened 2021-10-19
setup: disable test command

bastimeyer:
See #4102
Let's keep that issue thread open though, so that package maintainers who are running tests incorrectly in their package configs can see it.

1. Is there a better way to disable this other than checking `"test" in sys.argv`?
2. Is the error message worded well enough? We could also add links to the respective docs for example.

----

- https://setuptools.pypa.io/en/stable/userguide/commands.html#test-build-package-and-run-a-unittest-suite
- https://github.com/pytest-dev/pytest-runner/
- https://docs.pytest.org/en/6.2.x/pythonpath.html#invoking-pytest-versus-python-m-pytest

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,10 +3,15 @@ import codecs
 import sys
 from os import environ, path
 from sys import argv, path as sys_path
+from textwrap import dedent
 
 from setuptools import find_packages, setup
 
 import versioneer
+
+
+def format_msg(text, *args, **kwargs):
+    return dedent(text).strip(" \n").format(*args, **kwargs)
 
 
 CURRENT_PYTHON = sys.version_info[:2]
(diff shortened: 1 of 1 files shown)


bastimeyer:
I think the message is fine without any links.

----

I already mentioned earlier that I want to move most of the metadata defined imperatively in setup.py into the [declarative setup.cfg format](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html) and clean up everything. The package metadata is a bit messy right now and hasn't been touched in ages. There were a lot of changes in the past years in regards to python packaging standards, and we should stay up2date. It's a bit difficult though following and understanding all the changes here while having the supported python versions, dependencies and downstream system packages in mind.

After looking at the [setuptools keywords docs](https://setuptools.pypa.io/en/latest/userguide/keywords.html), I found the following about the `test_suite` keyword, which is relevant for this PR/issue:

> `test_suite`
> A string naming a unittest.TestCase subclass (or a package or module containing one or more of them, or a method of such a subclass), or naming a function that can be called with no arguments and returns a unittest.TestSuite. If the named suite is a module, and the module has an additional_tests() function, it is called and the results are added to the tests to be run. If the named suite is a package, any submodules and subpackages are recursively added to the overall test suite.
> 
> **Specifying this argument enables use of the test command to run the specified test suite, e.g. via setup.py test.** See the section on the test command below for more details.
> 
> New in `41.5.0`: Deprecated the test command.

Should we remove it in this PR, or should we do this when cleaning up the package metadata? Removing it instead of adding the sys.exit call doesn't have any effect though, despite what the documentation is saying. Either way, it should be removed at some point.

PR #4112, opened 2021-10-20
plugins.twitch: remove device-id headers

bastimeyer:
#4086 

`twitch-web-wall-mason` device-id headers are causing midroll ads now, which is bad for Streamlink's output.

The regular preroll ads may be annoying, but at least filtering out those via `--twitch-disable-ads` works without causing any issues.

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -187,8 +187,6 @@ class UsherService:
 class TwitchAPI:
     headers = {
         "Client-ID": "kimne78kx3ncx6brgo4mv6wki5h1ko",
-        "Device-ID": "twitch-web-wall-mason",
-        "X-Device-Id": "twitch-web-wall-mason",
     }
 
     def __init__(self, session):


back-to:
it still works, with a small change

```diff
diff --git a/src/streamlink/plugins/twitch.py b/src/streamlink/plugins/twitch.py
index 72ed9f40..b1daa108 100644
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -187,6 +187,8 @@ class UsherService:
 class TwitchAPI:
     headers = {
         "Client-ID": "kimne78kx3ncx6brgo4mv6wki5h1ko",
+        "Device-Id": "twitch-web-wall-mason",
+        "X-Device-Id": "twitch-web-wall-mason",
     }

     def __init__(self, session):
@@ -300,7 +302,7 @@ class TwitchAPI:
             login=channel_or_vod if is_live else "",
             isVod=not is_live,
             vodID=channel_or_vod if not is_live else "",
-            playerType="embed"
+            playerType="site"
         )
         subschema = validate.any(None, validate.all(
             {
```

```
{'adblock': False, 'geoblock_reason': '', 'hide_ads': False, 'server_ads': False, 'show_ads': True}
```

PR #4120, opened 2021-10-27
cli: remove the https-proxy option in favour of a single http-proxy option

### LINKED ISSUE #4094: Drop `https-proxy` option in favor of a single `http-proxy` option
### Checklist

### Description

How much sense would it make dropping `https-proxy` in favor of only having the `http-proxy` option?

The reason [why these two options were added](f9ebaa8b33eac95f88d796ae779500f4b00f3c95) is `requests`, which requires proxy configs for each protocol/scheme separately. It also can take scheme+netloc strings as proxy configs, but Streamlink doesn't support setting configs like this via CLI arguments.

- https://github.com/streamlink/streamlink/blame/1a1a3a06a2655f299fd718f5343799ec13ec7ce4/src/streamlink/session.py#L233-L239
- https://docs.python-requests.org/en/latest/user/advanced/#proxies
- https://github.com/psf/requests/blob/v2.26.0/requests/sessions.py#L373-L376

This protocol-specific configuration in `requests` is based on [`urllib.request.getproxies()`](https://docs.python.org/3/library/urllib.request.html#urllib.request.getproxies), which [similar to cURL](https://man.archlinux.org/man/curl.1#ENVIRONMENT) supports the `NO_PROXY`, `HTTP_PROXY`, `HTTPS_PROXY` and `ALL_PROXY` env vars (all-uppercase or all-lowercase), [unless `trust_env` is set to `False`](https://github.com/psf/requests/blob/v2.26.0/requests/sessions.py#L701-L713) (which can be done in Streamlink via the `http-trust-env` option).

Added in 2f76a1b98b7a862fd7a850a5971b7c7a0abb0f36 was the logic for setting the value for `https-proxy` via `http-proxy` when `https-proxy` was not set yet, to make defining proxy configs less redundant and confusing.

But how much sense does it make having a separate option for HTTPS connections? Even though we have just switched to default HTTPS schemes for scheme-less input URLs in #4068, this doesn't mean that every request will be done via HTTPS if a plugin needs to make requests via HTTP for example or if the user has explicitly set the HTTP scheme in the input URL. **This allows for accidental connections via HTTP over the direct connection when only `https-proxy` is set, which is a problem.**

I also don't see any use case for having a configuration which sets one proxy for HTTP and another for HTTPS connections. It doesn't make any sense, especially when almost every connection is made via HTTPS/TLS anyway. [Streamlink CLI also doesn't support setting HTTPS proxies first](https://github.com/streamlink/streamlink/blob/1a1a3a06a2655f299fd718f5343799ec13ec7ce4/src/streamlink_cli/main.py#L707-L711), which makes the is-not-set condition in the Session's `http-proxy` option useless for Streamlink CLI, because it will always be true.

Another point is that every other HTTP option/argument does not have an HTTPS counterpart, which can be confusing for some users.

----

That's why I'd suggest:

- turning `https-proxy` into an alias for `http-proxy` and suppressing the CLI argument
- or dropping `https-proxy` entirely (breaking change)

`requests` would still pick up the env vars if set.

beardypig:
I have made the `--https-proxy` command line argument in to an alias for `--http-proxy` and suppressed the help output for `--https-proxy`. `https-proxy` has been removed from the `get/set_option` methods, and it will have no effect if set. An `info` log level message is auto emitted if the `--https-proxy` option is used. 

resolves #4094

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -367,10 +367,11 @@ Progressive HTTP, HTTPS, etc   httpstream:// [1]_
 Proxy Support
 -------------
 
-You can use the :option:`--http-proxy` or :option:`--https-proxy` options to
-change the proxy server that Streamlink will use for HTTP and HTTPS requests respectively.
-For convenience reasons, :option:`--http-proxy` will automatically set the
-value of :option:`--https-proxy` as well, if it has not been set by the user.
+You can use the :option:`--http-proxy` option to change the proxy server
+that Streamlink will use for HTTP and HTTPS requests. :option:`--http-proxy` sets
+the proxy for all HTTP and HTTPS requests, including WebSocket connections.
+If separate proxies for each protocol are required, they can be set using
+environment variables - see `Requests Proxies Documentation`_
 
 Both HTTP and SOCKS proxies are supported, as well as authentication in each of them.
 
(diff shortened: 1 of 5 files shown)


gravyboat (approved):
Looks good to me.

PR #4129, opened 2021-10-31
plugins.tga: remove plugin

bastimeyer:
I am not 100% sure if this site is still operating or if it's just an issue on my side because I can't access it. There seems to be content, but it could also be fake, which is not too uncommon.

See my comment from #4040:
https://github.com/streamlink/streamlink/issues/4040#issuecomment-953843253

- this plugin was once meant for "Tencent's Gaming Platform TGA", but it went offline in favor of longzhu.com
- according to the few results on google, longzhu.com is the orga of an esports team which was founded in 2015 and was then bought/dissolved in 2016
- "Tencent Gaming Platform" was discontinued in favor of "WeGame" in 2017 [according to wikipedia](https://en.wikipedia.org/wiki/WeGame)
- longzhu.com is failing to load any kinds of assets except images, on both http and https, regardless whether an IP from or outside of China is being used
- streams are not loading
- the plugin hasn't been maintained in years, has a broken URL matcher and would require a complete rewrite
- it has only been modified once (in 2018) with a miniscule change after the streamlink fork
- docs are outdated/incorrect

Full commit history:
https://github.com/streamlink/streamlink/commits/98428d663e1cac084c4597419445042e6a81f9fd/src/streamlink/plugins/tga.py

We could just remove the RTMPStream import and call it a day, but I don't see the point keeping the plugin if it's unmaintained.


gravyboat:
Let's get rid of the plugin. If someone wants it back we can discuss it at that point, it would need to be completely reworked any way.

PR #4138, opened 2021-10-31
plugins.ltv_lsm_lv: update the plugin for the new page layout

### LINKED ISSUE #4131: ltv_lsm_lv.py: Does a typo makes it non-functioning?
### Checklist

### Streamlink version

Latest stable release

### Description

This plugin doesn't work for me on an Enigma2 receiver. When I try to open
#SERVICE 4097:0:1:22B8:446:1:C00000:0:0:0:http%3a//127.0.0.1%3a8088/https%3a//ltv.lsm.lv/lv/tiesraide/ltv1:LTV 1
#DESCRIPTION LTV 1
I get an error message in the debug log (see below). I think this might be caused by a simple typo in the .py file. In line 13 it reads:

 r"https?://ltv\.lsm\.lv/lv/tieshraide"

I do think it should rather be 'tiesraide' instead of 'tieshraide' as the actual URLs of the TV stations are: 

[https://ltv.lsm.lv/lv/tiesraide/ltv1](url)
[https://ltv.lsm.lv/lv/tiesraide/ltv7](url)
[https://ltv.lsm.lv/lv/tiesraide/visiemltv](url)

### Debug log

```text
root@sf4008:~# /usr/sbin/streamlinksrv manualstart debug
pidfile /var/run/streamlink.pid does not exist. Daemon not running?
[streamlinksrv][info]  Sun Oct 31 08:22:04 2021 Server (1.8.2) started
[streamlinksrv][debug] Host:           sf4008
[streamlinksrv][debug] Port:           8088
[streamlinksrv][debug] OS:             Linux-4.1.20-1.9-armv7l-with-glibc2.4
[streamlinksrv][debug] Python:         2.7.18
[streamlinksrv][info]  Streamlink:     1.27.8.0.r1_mod / 2021-10-11
[streamlinksrv][debug] youtube-dl:     2021.06.06
[streamlinksrv][debug] Requests(2.26.0), Socks(1.7.1), Websocket(0.59.0)
[streamlinksrv][error] No plugin can handle URL: https://ltv.lsm.lv/lv/tiesraide/ltv1
[streamlinksrv][debug] Send Offline clip
```

beardypig:
fixes #4131

### DIFF
--- a/src/streamlink/plugins/ltv_lsm_lv.py
+++ b/src/streamlink/plugins/ltv_lsm_lv.py
@@ -1,45 +1,128 @@
 import logging
 import re
-from urllib.parse import urlparse
+from urllib.parse import urlsplit, urlunsplit
 
 from streamlink.plugin import Plugin, pluginmatcher
-from streamlink.plugin.api.utils import itertags
-from streamlink.stream.hls import HLSStream
+from streamlink.plugin.api import validate
+from streamlink.stream.hls import HLSStream, HLSStreamReader, HLSStreamWorker
 
 log = logging.getLogger(__name__)
 
 
+def copy_query_url(to, from_):
+    """
+    Replace the query string in one URL with the query string from another URL
+    """
+    return urlunsplit(urlsplit(to)._replace(query=urlsplit(from_).query))
+
+
+class LTVHLSStreamWorker(HLSStreamWorker):
+    def process_sequences(self, playlist, sequences):
+        super().process_sequences(playlist, sequences)
+        # update the segment URLs with the query string from the playlist URL
+        self.playlist_sequences = [
+            sequence._replace(
+                segment=sequence.segment._replace(
+                    uri=copy_query_url(sequence.segment.uri, self.stream.url)
+                )
+            )
+            for sequence in self.playlist_sequences
+        ]
+
+
+class LTVHLSStreamReader(HLSStreamReader):
+    __worker__ = LTVHLSStreamWorker
+
+
+class LTVHLSStream(HLSStream):
+    __reader__ = LTVHLSStreamReader
+
+    @classmethod
+    def parse_variant_playlist(cls, *args, **kwargs):
+        streams = super().parse_variant_playlist(*args, **kwargs)
+
+        for stream in streams.values():
+            stream.args["url"] = copy_query_url(stream.args["url"], stream.url_master)
+
+        return streams
+
+
 @pluginmatcher(re.compile(
-    r"https?://ltv\.lsm\.lv/lv/tieshraide"
+    r"https://ltv\.lsm\.lv/lv/tiesraide"
 ))
 class LtvLsmLv(Plugin):
     """
     Support for Latvian live channels streams on ltv.lsm.lv
     """
+
+    def __init__(self, url: str):
+        super().__init__(url)
+        self._json_data_re = re.compile(r'teliaPlayer\((\{.*?\})\);', re.DOTALL)
+
+        self.main_page_schema = validate.Schema(
+            validate.parse_html(),
+            validate.xml_xpath_string(".//iframe[contains(@src, 'ltv.lsm.lv/embed')][1]/@src"),
+            validate.url()
+        )
+
+        self.embed_code_schema = validate.Schema(
+            validate.parse_html(),
+            validate.xml_xpath_string(".//live[1]/@*[name()=':embed-data']"),
+            str,
+            validate.parse_json(),
+            {"source": {"embed_code": str}},
+            validate.get(("source", "embed_code")),
+            validate.parse_html(),
+            validate.xml_xpath_string(".//iframe[@src][1]/@src"),
+        )
+
+        self.player_apicall_schema = validate.Schema(
+            validate.transform(self._json_data_re.search),
+            validate.any(
+                None,
+                validate.all(
+                    validate.get(1),
+                    validate.transform(lambda s: s.replace("'", '"')),
+                    validate.transform(lambda s: re.sub(r",\s*\}", "}", s, flags=re.DOTALL)),
+                    validate.parse_json(),
+                    {"channel": str},
+                    validate.get("channel")
+                )
+            )
+        )
+
+        self.sources_schema = validate.Schema(
+            validate.parse_json(),
+            {
+                "source": {
+                    "sources": validate.all(
+                        [{"type": str, "src": validate.url()}],
+                        validate.filter(lambda src: src["type"] == "application/x-mpegURL"),
+                        validate.map(lambda src: src.get("src"))
+                    ),
+                }},
+            validate.get(("source", "sources"))
+        )
+
     def _get_streams(self):
+        api_url = "https://player.cloudycdn.services/player/ltvlive/channel/{channel_id}/"
+
         self.session.http.headers.update({"Referer": self.url})
 
-        iframe_url = None
-        res = self.session.http.get(self.url)
-        for iframe in itertags(res.text, "iframe"):
-            if "embed.lsm.lv" in iframe.attributes.get("src"):
-                iframe_url = iframe.attributes.get("src")
-                break
+        iframe_url = self.session.http.get(self.url, schema=self.main_page_schema)
+        log.debug(f"Found embed iframe: {iframe_url}")
+        player_iframe_url = self.session.http.get(iframe_url, schema=self.embed_code_schema)
+        log.debug(f"Found player iframe: {player_iframe_url}")
+        channel_id = self.session.http.get(player_iframe_url, schema=self.player_apicall_schema)
+        log.debug(f"Found channel ID: {channel_id}")
 
-        if not iframe_url:
-            log.error("Could not find player iframe")
-            return
-
-        log.debug("Found iframe: {0}".format(iframe_url))
-        res = self.session.http.get(iframe_url)
-        for source in itertags(res.text, "source"):
-            if source.attributes.get("src"):
-                stream_url = source.attributes.get("src")
-                url_path = urlparse(stream_url).path
-                if url_path.endswith(".m3u8"):
-                    yield from HLSStream.parse_variant_playlist(self.session, stream_url).items()
-                else:
-                    log.debug("Not used URL path: {0}".format(url_path))
+        stream_sources = self.session.http.post(api_url.format(channel_id=channel_id),
+                                                data=dict(refer="ltv.lsm.lv",
+                                                          playertype="regular",
+                                                          protocol="hls"),
+                                                schema=self.sources_schema)
+        for surl in stream_sources:
+            yield from LTVHLSStream.parse_variant_playlist(self.session, surl).items()
 
 
 __plugin__ = LtvLsmLv
--- a/tests/plugins/test_ltv_lsm_lv.py
+++ b/tests/plugins/test_ltv_lsm_lv.py
@@ -6,10 +6,10 @@ class TestPluginCanHandleUrlLtvLsmLv(PluginCanHandleUrl):
     __plugin__ = LtvLsmLv
 
     should_match = [
-        "https://ltv.lsm.lv/lv/tieshraide/example/",
-        "http://ltv.lsm.lv/lv/tieshraide/example/",
-        "https://ltv.lsm.lv/lv/tieshraide/example/live.123/",
-        "http://ltv.lsm.lv/lv/tieshraide/example/live.123/",
+        "https://ltv.lsm.lv/lv/tiesraide/example/",
+        "https://ltv.lsm.lv/lv/tiesraide/example/",
+        "https://ltv.lsm.lv/lv/tiesraide/example/live.123/",
+        "https://ltv.lsm.lv/lv/tiesraide/example/live.123/",
     ]
 
     should_not_match = [
@@ -21,4 +21,5 @@ class TestPluginCanHandleUrlLtvLsmLv(PluginCanHandleUrl):
         "http://ltv.lsm.lv/other-site/",
         "https://ltv.lsm.lv/lv/other-site/",
         "http://ltv.lsm.lv/lv/other-site/",
+        "https://ltv.lsm.lv/lv/tieshraide/example/",
     ]


beardypig:
The streams appear to be geo-locked, so I cannot test them.

src/streamlink/plugins/ltv_lsm_lv.py
```diff
@@ -1,45 +1,83 @@
...
+
+    main_page_schema = validate.Schema(
+        validate.parse_html(),
+        validate.xml_xpath_string(".//iframe[contains(@src, 'ltv.lsm.lv/embed')][1]/@src"),
+        validate.url()
+    )
+
+    embed_code_schema = validate.Schema(
+        validate.parse_html(),
+        validate.xml_xpath_string(".//live/@*"),
```
bastimeyer, line 30: `validate.xml_xpath_string()` can return both `str` and `None`. If it's `None`, then `validate.parse_json()` will raise a TypeError instead of a PluginError, which means you'll have to add a `str` validation after `validate.xml_xpath_string()`.

src/streamlink/plugins/ltv_lsm_lv.py
```diff
@@ -1,45 +1,83 @@
...
+        validate.parse_json(),
+        {"source": {"embed_code": validate.text}},
```
bastimeyer, line 32: `validate.text` should be replaced with `str` since the removal of py2. Same down below...

src/streamlink/plugins/ltv_lsm_lv.py
```diff
@@ -1,45 +1,83 @@
...
+        validate.get("source"), validate.get("embed_code"),
```
bastimeyer, line 33: `validate.get` shouldn't be chained, as it can receive nested keys via a tuple parameter nowadays:
`validate.get(("source", "embed_code"))`
Same down below...

src/streamlink/plugins/ltv_lsm_lv.py
```diff
@@ -1,45 +1,83 @@
...
+        player_iframe_url = self.session.http.get(iframe_url, schema=self.embed_code_schema)
+        log.debug(f"Found player iframe: {player_iframe_url}")
+        channel_id = self.session.http.get(player_iframe_url, schema=self.player_apicall_schema)
+        log.debug(f"Found channel ID: {channel_id}")
+
+        stream_sources = self.session.http.post(self.API_URL.format(channel_id=channel_id),
+                                                data=dict(refer="ltv.lsm.lv",
+                                                          playertype="regular",
+                                                          protocol="hls"))
+        for surl in self.session.http.json(stream_sources, schema=self.sources_schema):
```
bastimeyer, line 79: Instead of calling `self.session.http.json`, you could add `validate.parse_json()` to the used schema, which fits the style of the other schemas and avoids the secondary method call on the http session.

src/streamlink/plugins/ltv_lsm_lv.py
```diff
@@ -1,45 +1,83 @@
...
+    )
+
+    embed_code_schema = validate.Schema(
+        validate.parse_html(),
+        validate.xml_xpath_string(".//live/@*"),
+        validate.parse_json(),
+        {"source": {"embed_code": validate.text}},
+        validate.get("source"), validate.get("embed_code"),
+        validate.parse_html(),
+        validate.xml_xpath_string(".//iframe/@src"),
```
bastimeyer, line 35: XPath queries always return lists of nodes. The `xml_xpath_string` wrapper simply turns the query into `str(query)`, but that doesn't change the fact that a query can return multiple nodes. Without limiting the number of nodes to one via the `[1]` predicate, the query could return nonsense data.

```py
validate.xml_xpath_string(".//iframe[@src][1]/@src"),
```

The `.//live/@*` query from above has the same issue and needs to be fixed with a more strict query.

bastimeyer (commented):
Just a couple of comments in regards to tidying up the validation schemas. I didn't check the actual stream outputs either.

Another (optional) change which you should make is moving the schema definitions from the plugin class into the `_get_streams()` or subsequent methods. We've already done this in most of the recent plugin updates / re-implementations, so that the schemas don't have to be unnecessarily initialized when parsing the plugin module, which saves resources. This will be fixed eventually once the new plugin resolver has been implemented that doesn't parse all plugin modules at once, but for now it's better to initialize them on demand, especially when it's lots of them.

src/streamlink/plugins/ltv_lsm_lv.py
beardypig: I felt this was more readable.

src/streamlink/plugins/ltv_lsm_lv.py
beardypig: The issue with the `live` tag, and why the xpath is the way it is, is because the attribute name is `:embed-data` which I could not get to work any other way.

src/streamlink/plugins/ltv_lsm_lv.py
beardypig: I disagree with this suggestion, I don't think `str` is a good name as it can be conflated with the actual type. Also, every other instance (that I could see) is `validate.text`.

bastimeyer: Well, it's no big deal, but the reason why I suggested the change is that `validate.text` is just an alias for the actual `str` type and it executes the same singledispatch validation method `validate_type` as other base types like `int`. `text` a left-over module export of the py2 compatibility era, and removing it would've required changing all validation schemas at that time, which didn't make much sense, so it was left.
- https://github.com/streamlink/streamlink/commit/c94a8107916a284e69c8c14dda3894d9f8d81495#diff-c415fd588e79ab1a5ef3a0fc4fbf7a90769778435b75a80b589e9c7b02afcbd7R37
- https://github.com/streamlink/streamlink/blob/37ee5347508980ee768ac0695714f01c54600933/src/streamlink/plugin/api/validate.py#L427-L436

All newly added or edited plugins are using the base types in their validation schemas.
```sh
$ grep ': str' src/streamlink/plugins/*.py | wc -l
168
$ grep 'validate.text' src/streamlink/plugins/*.py | wc -l
149
```

src/streamlink/plugins/ltv_lsm_lv.py
bastimeyer: Does `@*[name()=':embed-data']` work? This selects all attributes and then filters them by name.

bastimeyer: On which page can I find the `live` tag on their site?

beardypig: yeah, `@*[name()=':embed-data']` works, nice.

src/streamlink/plugins/ltv_lsm_lv.py
beardypig: fair enough :)

beardypig:
I moved the schema definitions in to the `__init__` method.

bastimeyer:
By the way, the streams don't work because the HLS playlist requests require an additional query parameter. This parameter is set on the variant playlist URL but not on the actual stream playlist URLs, and it needs to be applied to the individual playlist URLs.

Variant playlist:
```
https://ltvlive2167.cloudycdn.services/ltvlive/_definst_/ltvlive_ltv06_ltv1_g_x44_43186_default_1326_hls.smil/playlist.m3u8?Khi0MBGCgHNV_Ks4lsahc1u-NlhwL3za7BHYtmzF4dzMDSEBtZTJ7djOpPO4FmqpejtiS7vvap1SW5tzV2d5UwW_Mr2C7nJziPXTVMbQ_8IIu00gen4-6hvP3g69VQDb8fC4ei0uI4Gu7xYLRi3Pqas2nZPeliTb3VACfMCoKQvNX6vajt_hCd7ombJndrfuMyzlV-Z2fQJc0yS7gys
```
```m3u
#EXTM3U
#EXT-X-VERSION:4
#EXT-X-STREAM-INF:BANDWIDTH=5128000,NAME="1080p",RESOLUTION=1920x1080
chunklist_b5128000_t64MTA4MHA=.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=2128000,NAME="720p",RESOLUTION=1280x720
chunklist_b2128000_t64NzIwcA==.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=1128000,NAME="360p",RESOLUTION=640x360
chunklist_b1128000_t64MzYwcA==.m3u8
```

Stream playlist without query param:
```
https://ltvlive2167.cloudycdn.services/ltvlive/_definst_/ltvlive_ltv06_ltv1_g_x44_43186_default_1326_hls.smil/chunklist_b1128000_t64MzYwcA==.m3u8
```
```html
<?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
         "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
        <head>
                <title>403 - Forbidden</title>
        </head>
        <body>
                <h1>403 - Forbidden</h1>
        </body>
</html>
```

Stream playlist with query param:
```
https://ltvlive2167.cloudycdn.services/ltvlive/_definst_/ltvlive_ltv06_ltv1_g_x44_43186_default_1326_hls.smil/chunklist_b1128000_t64MzYwcA==.m3u8?Khi0MBGCgHNV_Ks4lsahc1u-NlhwL3za7BHYtmzF4dzMDSEBtZTJ7djOpPO4FmqpejtiS7vvap1SW5tzV2d5UwW_Mr2C7nJziPXTVMbQ_8IIu00gen4-6hvP3g69VQDb8fC4ei0uI4Gu7xYLRi3Pqas2nZPeliTb3VACfMCoKQvNX6vajt_hCd7ombJndrfuMyzlV-Z2fQJc0yS7gys
```
```m3u
#EXTM3U
#EXT-X-VERSION:4
#EXT-X-TARGETDURATION:4
#EXT-X-MEDIA-SEQUENCE:931184
#EXT-X-DISCONTINUITY-SEQUENCE:0
#EXTINF:4.0,
media-uehq1zjur_b1128000_t64MzYwcA==_931184.ts
#EXTINF:4.0,
media-uehq1zjur_b1128000_t64MzYwcA==_931185.ts
#EXTINF:4.0,
media-uehq1zjur_b1128000_t64MzYwcA==_931186.ts
```

beardypig:
Ah, well spotted. I will take a look at that then.

bastimeyer:
You probably have already noticed, but the segments also require this query string, which means the entire HLSStream needs to be subclassed and the segment url logic replaced.

beardypig:
Yeah, I lacked the energy to do that last night ... I will probably take a look at it today though.

bastimeyer (commented):
The `NamedTuple`s can be cloned easily via `_replace` instead of initializing new ones. Then you don't have to import and be explicit about all the other params. Let me post a quick diff instead of commenting on the lines.

```diff
diff --git a/src/streamlink/plugins/ltv_lsm_lv.py b/src/streamlink/plugins/ltv_lsm_lv.py
index 9bd92c57..980e7aa9 100644
--- a/src/streamlink/plugins/ltv_lsm_lv.py
+++ b/src/streamlink/plugins/ltv_lsm_lv.py
@@ -4,8 +4,7 @@ from urllib.parse import urlsplit, urlunsplit
 
 from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
-from streamlink.stream.hls import HLSStream, HLSStreamReader, HLSStreamWorker, Sequence
-from streamlink.stream.hls_playlist import Segment
+from streamlink.stream.hls import HLSStream, HLSStreamReader, HLSStreamWorker
 
 log = logging.getLogger(__name__)
 
@@ -18,21 +17,14 @@ def copy_query_url(to, from_):
 
 
 class LTVHLSStreamWorker(HLSStreamWorker):
-    def process_sequences(self, playlist, sequences):
-        super(LTVHLSStreamWorker, self).process_sequences(playlist, sequences)
-        # update the segment URLs with the query string from the playlist URL
+    def process_sequences(self, *args, **kwargs):
+        super().process_sequences(*args, **kwargs)
         self.playlist_sequences = [
-            Sequence(sequence.num,
-                     Segment(
-                         copy_query_url(sequence.segment.uri, self.stream.url),
-                         sequence.segment.duration,
-                         sequence.segment.title,
-                         sequence.segment.key,
-                         sequence.segment.discontinuity,
-                         sequence.segment.byterange,
-                         sequence.segment.date,
-                         sequence.segment.map
-                     ))
+            sequence._replace(
+                segment=sequence.segment._replace(
+                    uri=copy_query_url(sequence.segment.uri, self.stream.url)
+                )
+            )
             for sequence in self.playlist_sequences
         ]
 
@@ -48,8 +40,8 @@ class LTVHLSStream(HLSStream):
     def parse_variant_playlist(cls, *args, **kwargs):
         streams = super().parse_variant_playlist(*args, **kwargs)
 
-        for name in streams.keys():
-            streams[name].args["url"] = copy_query_url(streams[name].args["url"], streams[name].url_master)
+        for stream in streams.values():
+            stream.args["url"] = copy_query_url(stream.args["url"], stream.url_master)
 
         return streams
 
```

It might also be worth exploring some extensibility in the HLS classes at some point, but for now this should be good enough.

Streams are working with these changes.

bastimeyer:
Btw, there's a typo in the commit message (and PR title).

src/streamlink/plugins/ltv_lsm_lv.py
```diff
@@ -1,45 +1,128 @@
...
+def copy_query_url(to, from_):
+    """
+    Replace the query string in one URL with the query string from another URL
+    """
+    return urlunsplit(urlsplit(to)._replace(query=urlsplit(from_).query))
+
+
+class LTVHLSStreamWorker(HLSStreamWorker):
+    def process_sequences(self, playlist, sequences):
+        super(LTVHLSStreamWorker, self).process_sequences(playlist, sequences)
```
bastimeyer, line 21:
```suggestion
        super().process_sequences(playlist, sequences)
```
Let's not go backwards in time... Every other super-call in the code base is in the implicit py3 style and we should keep it like that.
```
$ GIT_PAGER=cat git grep -E 'super\\([^\\)]'
src/streamlink/plugins/ltv_lsm_lv.py:        super(LTVHLSStreamWorker, self).process_sequences(playlist, sequences)
```

bastimeyer (commented):
Should be good to go after the super-call has been cleaned up.

src/streamlink/plugins/ltv_lsm_lv.py
beardypig: > Let's not go backwards in time...

beardypig: Overzealous IDE compatibility auto-correct.

PR #4141, opened 2021-11-02
ci.github: check for unicode bidi control chars

bastimeyer:
[CVE-2021-42574](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-42574) has been published a couple of days ago:

> An issue was discovered in the Bidirectional Algorithm in the Unicode Specification through 14.0. It permits the visual reordering of characters via control sequences, which can be used to craft source code that renders different logic than the logical ordering of tokens ingested by compilers and interpreters. Adversaries can leverage this to encode source code for compilers accepting Unicode such that targeted vulnerabilities are introduced invisibly to human reviewers.

Even though [Github has started showing a warning](https://github.blog/changelog/2021-10-31-warning-about-bidirectional-unicode-text/) about this vulnerability ([example](https://github.com/nickboucher/trojan-source/blob/be30965b7091c14bda65c1d82f582c8733779827/Python/commenting-out.py)), there's currently no linting config here to prevent unicode bidirectional control characters from being committed and merged. Projects with lots of different contributors should take care of this. We already have similar linting configs, so let's add one for this too, because why not, it's very simple...

The explicit control characters are listed here:
- https://www.unicode.org/reports/tr9/#Directional_Formatting_Characters
- https://developer.mozilla.org/en-US/docs/Web/Guide/Unicode_Bidrectional_Text_Algorithm

The regex of the grep command matches any of these characters in all of the tracked files by git and it returns status 1 if it finds one. This project doesn't need bidirectional text, neither in plugins for certain international sites or in the docs, so adding this global linting config is fine.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,6 +54,9 @@ jobs:
       - name: Lint (line terminators)
         if: startsWith(matrix.os, 'ubuntu')
         run: "! grep 'with CRLF line terminators' <(git ls-files | file -nNf-)"
+      - name: Lint (unicode bidirectional control characters)
+        if: startsWith(matrix.os, 'ubuntu')
+        run: "! git grep -EIn $'[\\u2066\\u2067\\u2068\\u2069\\u202A\\u202B\\u202C\\u202D\\u202E]'"
       - name: Lint (file permissions)
         if: startsWith(matrix.os, 'ubuntu')
         run: "! grep -Ev '^644' <(git ls-files src/ tests/ | xargs stat '--format=%a %n')"


beardypig (approved):

.github/workflows/main.yml
```diff
@@ -54,6 +54,9 @@ jobs:
       - name: Lint (line terminators)
         if: startsWith(matrix.os, 'ubuntu')
         run: "! grep 'with CRLF line terminators' <(git ls-files | file -nNf-)"
+      - name: Lint (unicode bidirectional control characters)
+        if: startsWith(matrix.os, 'ubuntu')
+        run: "! grep -E -- $'[\u2066-\u2069\u202A-\u202E]' <(git ls-files)"
```
bastimeyer, line 59: Wait a sec, this checks file names, not file contents.

Is there a better way to fix this other than this:

```suggestion
        run: "! grep -E -- $'[\u2066-\u2069\u202A-\u202E]' $(git ls-files)"
```

This might exceed the argument length. I don't want to pipe grep into xargs because I want grep to be first.

beardypig, line 59: Does this work?
```suggestion
        run: "! git grep -E -- $'[\u2066-\u2069\u202A-\u202E]'"
```

beardypig: Does this work?
```suggestion
        run: "! git grep -E -- $'[\u2066-\u2069\u202A-\u202E]'"
```

bastimeyer: Git doesn't like character ranges for some reason. That should work, I think. Can you confirm before I update?

```sh
git grep -EIn $'[\u2066\u2067\u2068\u2069\u202A\u202B\u202C\u202D\u202E]'
```

beardypig: I also couldn't get the ranges to work, I thought it might be some console option or user-error. 

Those characters seem to cover all those listed in the mozilla page. 👍

bastimeyer: Hm, this confuses Github's output.
https://github.com/streamlink/streamlink/runs/4081108757?check_suite_focus=true#step:7:1

Is there a different way to write the unicode characters for the bash script, so that it doesn't mess with the log output?

bastimeyer: Works:
https://github.com/streamlink/streamlink/runs/4081244407?check_suite_focus=true#step:7:2

beardypig (approved):

.github/workflows/main.yml
beardypig: LGTM

PR #4156, opened 2021-11-06
plugins.twitch: new plugin command --twitch-api-header

back-to:
- can be used for `Authorization` or other headers
- change PlaybackAccessToken playerType to `site` as `embed` will show ads
- remove some unsed HLS and headers code

---

```
streamlink https://twitch.tv/twitch best \
    --twitch-api-header "X-Device-Id=twitch-web-wall-mason" \
    --twitch-api-header "Device-Id=twitch-web-wall-mason"
```

closes https://github.com/streamlink/streamlink/pull/4080

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -2,6 +2,7 @@ import json
 import logging
 import re
 from collections import namedtuple
+from functools import lru_cache
 from random import random
 from urllib.parse import urlparse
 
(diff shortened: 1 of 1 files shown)


bastimeyer (commented):
It's GQL (Graph Query Language), not GPL.

Adding a generic parameter for adjusting the API requests headers makes more sense than adding multiple ones for specific authentication/identification methods, so I'm in favor of that. The problem here though is that it's not particularly user-friendly, but neither is extracting headers from the web browser, so it's fine.

bastimeyer:
It might be better to just call it `twitch-api-header`. This is less confusing, people don't have to memorize specific names (gql) and it's shorter (no http-).

bastimeyer:
`playerType="site"` also causes (preroll-)ads. There's no difference. Did you get midroll ads with `embed` or why did you change it?

back-to:
server_ads will be False with `site` but True with `embed`

`[plugins.twitch][debug] {'adblock': False, 'geoblock_reason': '', 'hide_ads': False, 'server_ads': False, 'show_ads': True}`

when used with 

```
streamlink https://twitch.tv/twitch best \
    --twitch-api-header "X-Device-Id=twitch-web-wall-mason" \
    --twitch-api-header "Device-Id=twitch-web-wall-mason"
```

bastimeyer (approved):

PR #4163, opened 2021-11-10
session.resolve_url: return plugin class + URL 

### LINKED ISSUE #4091: Plugin options do not get set during plugin initialization
### Checklist

### Streamlink version

### Description

Plugin options currently do not get set during plugin initialization and are only available when `Plugin.streams` is called, which is too late when certain setups are required during init which depend on the provided options (see #4080).

The reason for this is how the plugins get resolved.

This is the current `Session.resolve_url()` method, which returns a new instance of a matching plugin which it then caches for the passed input via the `lru_cache` decorator:
https://github.com/streamlink/streamlink/blob/1a1a3a06a2655f299fd718f5343799ec13ec7ce4/src/streamlink/session.py#L344-L377

The first time the plugin gets resolved is in `streamlink_cli.main.setup_config_args()`, as it needs to find a plugin config file for the input URL, so that plugin specific configs can be set:
https://github.com/streamlink/streamlink/blob/1a1a3a06a2655f299fd718f5343799ec13ec7ce4/src/streamlink_cli/main.py#L683-L694
A plugin instance is not required here, because all that's needed is the plugin name, so only the plugin class should be resolved instead.

The second time the matching plugin gets resolved (from cache) is in `streamlink_cli.main.handle_url()`:
https://github.com/streamlink/streamlink/blob/1a1a3a06a2655f299fd718f5343799ec13ec7ce4/src/streamlink_cli/main.py#L563-L565
As you can see, it resolves the plugin instance and afterwards sets up the plugin options via `setup_plugin_options(streamlink, plugin)`, which means the options are set too late.

`setup_plugin_options` however doesn't need a plugin instance. All it needs is the plugin name and its arguments, which is a class attribute.
https://github.com/streamlink/streamlink/blob/1a1a3a06a2655f299fd718f5343799ec13ec7ce4/src/streamlink_cli/main.py#L881-L913

This means that the session's `resolve_url` needs to be changed (or extended with an initialize keyword), or its calls be replaced in the cli.main module with another method which resolves plugin classes instead. Then the resolved plugin could be initialized after the options have been set.

----

What's the best solution for this?

- changing `Session.resolve_url` is a significant breaking change
- extending it with an optional keyword doesn't fix the issue for third party python projects implementing streamlink if we keep backwards compatibility
- neither does adding another method for only resolving plugin classes to `Session`

### Debug log

```text
-
```

### LINKED ISSUE #4161: Plugin keep init if running too many sessions in the same time
### Checklist

### Streamlink version

Latest stable release

### Description

Plugin:
```py
# encoding: utf-8
import re
from streamlink.plugin import Plugin, PluginArgument, PluginArguments, pluginmatcher

@pluginmatcher(re.compile(
    r'(http)?(s)?(://)?(www\.)?test.com/(?P<channel>[^/]+)'
))
class Test(Plugin):
    def __init__(self, url: str) -> None:
        self.url = url
        
        try:
            self.load_cookies()
        except RuntimeError:
            pass  # unbound cannot load

        self.lock = self.session.get_option("lock")
        with self.lock:
            print("__init__", self.url)
            
    def _get_streams(self):
        with self.lock:
            print("_get_streams", self.url)

__plugin__ = Test
```

reproduced code:
```
import streamlink
import time
import threading

lock = threading.Lock()

def GetStream(url):
    session = streamlink.Streamlink()
    session.set_option('lock', lock)
    session.options.set('lock', lock)
    while True:
        streams = session.streams(url)
        time.sleep(10)

count = 200
threading_pool = []
for i in range(count):
    thread = threading.Thread(target = GetStream, args = ('test.com/{}'.format(i) , ))
    threading_pool.append(thread)

for thread in threading_pool:
    thread.start()
    
for thread in threading_pool:
    thread.join()
```

I just found out it's because of "@lru_cache(maxsize=128)" in session.py
After I change it to 1024, everything is fine.
Anyway, there should be a warning, queue is full or what
I just upgrade Streamlink then realize some feature is broken, LOL

### Debug log

```text
```

bastimeyer:
Breaking change:
Instead of resolving a plugin instance in `Streamlink.resolve_url(url)`
from the provided input URL (which can be redirected), resolve the
plugin class and the resulting URL, and cache the tuple. Also affects
`Streamlink.resolve_url_no_redirect(url)`.

The main reason for this change is streamlink_cli and how the plugin
options get set. The plugin options need to be set on the class before
it gets initialized, so that the instance can access the options in its
constructor method. This also fixes any kind of state stored on the
plugin instance.

- resolve and cache a tuple of the plugin class and resulting URL
- initialize plugin in streamlink_cli after applying its options
- remove plugin variable from global scope in streamlink_cli
- fix, rewrite and add tests

----

Resolves #4091 
Resolves #4161 

----

### Notes

Updating the URL in `streamlink_cli`'s argument parser setup and removing the redirect logic from `Streamlink.resolve_url()` (as suggested here: https://github.com/streamlink/streamlink/issues/4091#issuecomment-952391848) didn't seem to be trivial because of how the session is set up and how plugin config files are handled in `setup_config_args()`, which gets called twice before `handle_url()` gets called. The current solution should be good enough, and the logic of `streamlink.streams()` does not get affected because of that, which is good.

1. I didn't add a special test for checking if plugin options are available in the plugin constructor when initialized via `streamlink_cli`, but you can check for yourself that it's working by adding a breakpoint in one of the plugin constructors.
2. Since the potentially redirected URL gets returned now, another log message could be added in `handle_url()` if the input and resolved URL are different.
3. The session tests now can't make accidental HTTP requests due to mocking all requests
4. I've added a second commit for adding more type annotations (anything related to `handle_url()`)

### DIFF
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -3,7 +3,7 @@ import pkgutil
 from collections import OrderedDict
 from functools import lru_cache
 from socket import AF_INET, AF_INET6
-from typing import Dict, Optional, Type
+from typing import Dict, Optional, Tuple, Type
 
 import requests
 import requests.packages.urllib3.util.connection as urllib3_connection
@@ -341,10 +341,10 @@ class Streamlink:
             return plugin.get_option(key)
 
     @lru_cache(maxsize=128)
-    def resolve_url(self, url: str, follow_redirect: bool = True) -> Plugin:
+    def resolve_url(self, url: str, follow_redirect: bool = True) -> Tuple[Type[Plugin], str]:
         """Attempts to find a plugin that can use this URL.
 
-        The default protocol (http) will be prefixed to the URL if
+        The default protocol (https) will be prefixed to the URL if
         not specified.
 
         Raises :exc:`NoPluginError` on failure.
@@ -373,11 +373,12 @@ class Streamlink:
                     priority = prio
 
         if candidate:
-            return candidate(url)
+            return candidate, url
 
         if follow_redirect:
             # Attempt to handle a redirect URL
             try:
+                # noinspection PyArgumentList
                 res = self.http.head(url, allow_redirects=True, acceptable_status=[501])
 
                 # Fall back to GET request if server doesn't handle HEAD.
@@ -391,10 +392,10 @@ class Streamlink:
 
         raise NoPluginError
 
-    def resolve_url_no_redirect(self, url):
+    def resolve_url_no_redirect(self, url: str) -> Tuple[Type[Plugin], str]:
         """Attempts to find a plugin that can use this URL.
 
-        The default protocol (http) will be prefixed to the URL if
+        The default protocol (https) will be prefixed to the URL if
         not specified.
 
         Raises :exc:`NoPluginError` on failure.
@@ -404,7 +405,7 @@ class Streamlink:
         """
         return self.resolve_url(url, follow_redirect=False)
 
-    def streams(self, url, **params):
+    def streams(self, url: str, **params):
         """Attempts to find a plugin and extract streams from the *url*.
 
         *params* are passed to :func:`Plugin.streams`.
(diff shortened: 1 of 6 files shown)


beardypig (commented):
I haven't had chance to look at this in depth, but the diff looks good. I had done it in a similar way on another branch, but I didn't have time to update the tests. I had also started on a test for the plugin options being available in `__init__`.

PR #4167, opened 2021-11-12
plugins.twitch: refactor api-headers

bastimeyer:
With #4163 being merged, the API headers initialization can be simplified now


bastimeyer:
Btw, `--twitch-api-header "X-Device-Id=twitch-web-wall-mason" --twitch-api-header "Device-Id=twitch-web-wall-mason"` does not prevent preroll ads anymore...

PR #4168, opened 2021-11-12
streams: remove HDS/AkamaiHD and flashmedia pkg

bastimeyer:
- drop HDS and AkamaiHD stream implementations
- drop HDS and AkamaiHD plugins
- drop streamlink.packages.flashmedia
- remove stream.flvconcat, stream.playlist and plugins.common_swf
- remove "hds" from default stream types list
- remove all `--hds-*` CLI arguments and `hds-*` session options
- remove unneeded flashmedia license text files
- update tests
- update docs and CLI argument help texts

----

ref #4040

Should this be done in one commit or does this need to be split up into multiple commits?

This PR intentionally doesn't include RTMP stuff, because it's not 100% certain yet whether we want to remove it too, and it would also require some more changes related to rtmpdump, documentation, etc, which I think should definitely be done separately.


gravyboat (approved):
> Should this be done in one commit or does this need to be split up into multiple commits?

I'm fine with one. Let's see what others think.

back-to:
> This PR intentionally doesn't include RTMP stuff, because it's not 100% certain yet whether we want to remove it too

probably yes

PR #4169, opened 2021-11-13
stream: remove RTMP and RTMPDump dependency

bastimeyer:
- drop RTMP stream implementation
- drop RTMP plugin
- drop RTMPDump dependency
- remove stream.streamprocess and utils.{rtmp,swf}
- remove "rtmp" from default stream types list
- remove "rtmp" from player-passthrough options list
- remove all `--rtmp*` CLI args and `rtmp-*` session options
- remove all `--subprocess-*` CLI args and `subprocess` session options,
  as these were used only by StreamProcess which is no longer needed
- update tests
- update docs and CLI argument help texts

----

This will require all users on Windows (who are using the Windows installer) to update their config files, as `--rtmp-rtmpdump` and its old deprecated parameter `--rtmpdump` have been removed, and those were set by default in the config file previously.

I'm not 100% satisfied with the rewritten docs yet in regards to the optional parameters, but this needs to be improved later, not now. I also had to remove one of the example parameters which I wanted to include in the rewritten docs section because of a bug in HTTPStream which I will fix afterwards (`method=POST`).


back-to:
> This will require all users on Windows (who are using the Windows installer) to update their config files, as --rtmp-rtmpdump and its old deprecated parameter --rtmpdump have been removed, and those were set by default in the config file previously.

`help=argparse.SUPPRESS` might be better instead of completely removing it
or `ignore_unknown=True` could be used.

bastimeyer:
Suppressing could be done, but this will be a breaking change anyway, so I don't see a big issue with removing them now. Then there's also the issue with the related CLI args which were not set by the Windows installer config. Still a breaking change, but no intervention required.

I don't mind adding `--rtmp-rtmpdump` and `--rtmpdump` back and suppressing them (although it's kinda annoying). You decide.

PR #4177, opened 2021-11-15
release: 3.0.0

### LINKED ISSUE #4145: 3.0.0 checklist
Let's prepare a checklist for the next release. Since there's a breaking change on the master branch (#4068), the next release will require a major version bump. This raises the question if we should remove some of the previously deprecated stuff now or postpone it, and if there are more breaking changes that should be included now.

### Breaking changes

- [x] #4091
  Changing `Session.resolve_url` to return plugin classes instead of plugin class instances would fix that bug, but would be a breaking change. This should ideally be done now.
- [x] #4040
  All plugins which previously imported the old protocol implementations have already been changed / fixed / removed. Only ustreamtv is left with a custom protocol that requires `streamlink.packages.flashmedia`.
  Protocol plugins could be removed, but since flashmedia can't be removed yet due to ustreamtv, it doesn't make too much sense dropping them now while keeping the flashmedia package.
  - [x] RTMPStream
    Removing it now would enable the removal of rtmpdump.
  - [x] HDS / AkamaiHD
    These two are the most irrelevant protocols today.
  - [x] `streamlink.packages.flashmedia`
    I've been working on a ustreamtv re-implementation, but this requires a new generic websocket session implementation first if I want to do it right, and this still needs more time to finalize. Other plugins like twitcasting and nicolive which require websocket stuff will benefit from this as well, as there's currently a lot of duplicate boilerplate code. I've already finished rewriting the twitcasting plugin and am in the final stages of the nicolive plugin and both are using the new websocket session. ustreamtv will come afterwards. I need to do it this way so I can see if the design and implementation is good enough, because there were two different websocket-client APIs in use by those plugins prior to that.

### Dependencies

- [x] #4170 
  May be worth including this in 3.0.0 and cleaning up the optional dependencies

### Currently deprecated stuff that could be removed

We don't have any deadlines for the deprecated stuff, so let's take a quick look.
https://streamlink.github.io/latest/deprecations.html

- [ ] ~~Config and cache dirs/paths, and plugin sideloading paths~~
  Could be done because postponing the removal until another major release could be a bit overwhelming for the user if there are more breaking changes on that release. However, it still feels a bit early, so I would say that we should keep it for now.
- [ ] ~~`Plugin.can_handle_url()` and `Plugin.priority()`~~
  Should only be done when there's a new plugin resolver
- [ ] ~~Stream-type related CLI arguments~~
  Way too early as they only have been deprecated in the last release
- [x] Other suppressed CLI arguments
  - [x] `--rtmpdump`
    Deprecated alias for `--rtmp-rtmpdump`. Should be removed with RTMPStream.
  - [ ] ~~`--hls-stream-segment-data`~~
    CLI argument got restored (#4126).

### Open issues that should be resolved first

- [x] #4126
  As I mentioned there, this may be a bug in requests/urllib3. Reverting `stream=True` could be an option, but that would require adding back `--hls-stream-segment-data`.

bastimeyer:
Closes #4145 

Let's prepare the 3.0.0 release.

Writing this changelog took me longer than I anticipated. I haven't fully checked and proof-read it myself yet, so if there's anything wrong or needs to be improved, please comment. I had to put the breaking changes into their own list, because otherwise it would've been way too much. That's why I also didn't include some plugin changes which I would've included otherwise. It also already includes my PR from earlier today which hasn't been merged yet.

Even if the date of the changelog is set for today, nothing is set in stone yet. I'd suggest waiting a bit first to see if there are any last-minute changes. As always, I will update the changelog when new stuff gets merged.

- https://github.com/bastimeyer/streamlink/blob/releases/3.0.0/CHANGELOG.md
- https://deploy-preview-4177--streamlink.netlify.app/changelog.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,158 @@
 # Changelog
 
+## streamlink 3.0.0 (2021-11-17)
+
+Breaking changes:
+
+- BREAKING: dropped support for RTMP, HDS and AkamaiHD streams ([#4169](https://github.com/streamlink/streamlink/pull/4169), [#4168](https://github.com/streamlink/streamlink/pull/4168))
+  - removed the `rtmp://`, `hds://` and `akamaihd://` protocol plugins
+  - removed all Flash related code
+  - upgraded all plugins using these old streaming protocols
+  - dropped RTMPDump dependency
+- BREAKING: removed the following CLI arguments (and respective session options): ([#4169](https://github.com/streamlink/streamlink/pull/4169), [#4168](https://github.com/streamlink/streamlink/pull/4168))
+  - `--rtmp-rtmpdump`, `--rtmpdump`, `--rtmp-proxy`, `--rtmp-timeout`  
+    Users of Streamlink's Windows installer will need to update their [config file](https://streamlink.github.io/cli.html#configuration-file).
+  - `--subprocess-cmdline`, `--subprocess-errorlog`, `--subprocess-errorlog-path`
+  - `--hds-live-edge`, `--hds-segment-attempts`, `--hds-segment-threads`, `--hds-segment-timeout`, `--hds-timeout`
+- BREAKING: switched from HTTP to HTTPS for all kinds of scheme-less input URLs. If a site or http-proxy doesn't support HTTPS, then HTTP needs to be set explicitly. ([#4068](https://github.com/streamlink/streamlink/pull/4068), [#4053](https://github.com/streamlink/streamlink/pull/4053))
+- BREAKING/API: changed `Session.resolve_url()` and `Session.resolve_url_no_redirect()` to return a tuple of a plugin class and the resolved URL instead of an initialized plugin class instance. This fixes the availability of plugin options in a plugin's constructor. ([#4163](https://github.com/streamlink/streamlink/pull/4163))
+- BREAKING/requirements: dropped alternative dependency `pycrypto` and removed the `STREAMLINK_USE_PYCRYPTO` env var switch ([#4174](https://github.com/streamlink/streamlink/pull/4174))
+- BREAKING/requirements: switched from `iso-639`+`iso3166` to `pycountry` and removed the `STREAMLINK_USE_PYCOUNTRY` env var switch ([#4175](https://github.com/streamlink/streamlink/pull/4175))
+- BREAKING/setup: disabled unsupported Python versions, disabled the deprecated `test` setuptools command, removed the `NO_DEPS` env var, and switched to declarative package data via `setup.cfg` ([#4079](https://github.com/streamlink/streamlink/pull/4079), [#4107](https://github.com/streamlink/streamlink/pull/4107), [#4115](https://github.com/streamlink/streamlink/pull/4115) [#4113](https://github.com/streamlink/streamlink/pull/4113))
+
+Release highlights:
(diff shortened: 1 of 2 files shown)


bastimeyer:
There's an issue with the sdist tarballs. It includes everything and seems to ignore MANIFEST.in. I'll have a look at this later. Just noticed that because I wanted to include script/build-shell-completions.sh in the sdist (or is that a bad idea?).

bastimeyer:
Let's include the shell completion files in the sdist directly instead of the build script. We're already doing this with the man page that gets built before the sdist gets built.

The built files are currently stored in `build/shtab/`

https://packaging.python.org/guides/using-manifest-in/
> After processing the MANIFEST.in file, setuptools removes the build/ directory as well as any directories named RCS, CVS, or .svn from the sdist, and it adds a PKG-INFO file and an *.egg-info directory. This behavior cannot be changed with MANIFEST.in.

How intuitive... This means the directory needs to be changed. Where do we put them instead? I can only think of creating a new `completions` directory in the root dir and adding it to .gitignore. Is that okay?

gravyboat:
> How intuitive... This means the directory needs to be changed. Where do we put them instead? I can only think of creating a new completions directory in the root dir and adding it to .gitignore. Is that okay?

Yeah that's fine. I'll review the release notes when I have time later tonight.

gravyboat:
Forgot to submit my comment last night. Looks good to me @bastimeyer. We can wait on someone else if you want or merge when you're ready.

back-to:
https://deploy-preview-4177--streamlink.netlify.app/deprecations.html#streamlink-3-0-0

there is just `Removal of separate https-proxy option`

seems just outdated, if there is only one thing mentioned there.

bastimeyer:
The deprecations page (currently) doesn't list removals, just stuff that has been noted as deprecated but which is still working without any changes required by the user.

Do we want to list removals / breaking changes on the deprecations page as well? I've added the deprecations with the 2.2.0 release, which is why there aren't any removals included. This was previously done via the changelog.

gravyboat:
Do removals and breaking changes really fit on the deprecations page? I'm fine with just having them in the changelog.

bastimeyer:
Since none of the removals of 3.0.0 have been marked as deprecated, I don't think it makes too much sense having them there if we treat the page as a pure list of deprecations. This is of course different for stuff that has been marked as deprecated and those deprecations should always be updated with their removal release information.

If we want to include the breaking changes of 3.0.0, then the 2.0.0 release is missing its information, so it should be added there as well. Also, if we add removals to the page, then we have to rename it.

I don't mind updating the page and I could open a PR for this later today or tomorrow, but I don't think it's important.

gravyboat:
@bastimeyer Let's skip it then unless @back-to disagrees.

back-to:
it is not that important

bastimeyer:
Go ahead, unless you want to wait for @beardypig to review too

gravyboat:
@bastimeyer I'm on my work machine so I can't cut the release tag right now, I can do it later tonight and either merge this now or it can wait.

bastimeyer:
No pressure... I'll update the release date again in three hours if no one has merged+tagged it by then.

gravyboat:
Tagged: https://github.com/streamlink/streamlink/tree/3.0.0

PR #4185, opened 2021-11-17
plugins.crunchyroll: add metadata attributes

kyldery:
Currently, none of the metadata attributes is set, this PR adds them using 3 fields provided by the API

- `media.name` as _title_ is the title of the episode
- `media.series_name` as _author_ is the name of the anime/show
- `media.media_type` as _category_ is a more general category, like "anime" or "drama"

### DIFF
--- a/src/streamlink/plugins/crunchyroll.py
+++ b/src/streamlink/plugins/crunchyroll.py
@@ -298,11 +298,6 @@ class Crunchyroll(Plugin):
 
         return Plugin.stream_weight(key)
 
-    def get_category(self):
-        if self.category:
-            self.category = self.category.capitalize()
-        return self.category
-
     def _get_streams(self):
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/crunchyroll.py
```diff
@@ -39,6 +39,9 @@ def parse_timestamp(ts):
 })
 _media_schema = validate.Schema(
     {
+        "name": validate.text,
+        "series_name": validate.text,
+        "media_type": validate.text,
```
bastimeyer, line 44: Does the API response always include these properties when you add the additional fields to the request (down below)? Are they always strings or can there be null values? I'm asking, because metadata should be optional if possible, to not break anything.

bastimeyer:
Thanks

PR #4187, opened 2021-11-18
ci.github: add CodeQL code scanning workflow

bastimeyer:
Now that all of the unmaintained and third-party code has been removed, let's try adding a CodeQL workflow again
#3217

----

https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs

> The following query suites are built into CodeQL code scanning and are available for use.
> 
> Query suite | Description
> -- | --
> security-extended | Queries of lower severity and precision than the default queries
> security-and-quality | Queries from security-extended, plus maintainability and reliability queries
>
> When you specify a query suite, the CodeQL analysis engine will run the queries contained within the suite for you, in addition to the default set of queries.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,8 @@
+paths:
+  - src/
+  - tests/
+paths-ignore:
+  - 'src/**/packages/'
+  - src/streamlink/_version.py
+queries:
+  - uses: security-and-quality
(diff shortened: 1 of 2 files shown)


bastimeyer:
Why is the scanning results check still queued after over 30 mins?
https://github.com/streamlink/streamlink/pull/4187/checks?check_run_id=4254948834

I've pretty much used the default workflow config, so I'm a bit surprised here. The analysis also took over 9 minutes, which is annoying. Maybe because it first needs to build the initial database, but still...
https://github.com/streamlink/streamlink/runs/4254846236?check_suite_focus=true

PR #4190, opened 2021-11-20
cli: override default signal handlers

### LINKED ISSUE #4188: Brief exit message for KeyboardInterrupt
### Checklist

### Description

Thanks for maintaining streamlink! It entertains me a lot. :+1: :rocket:

I'd like an option to reduce the verbosity of `KeyboardInterrupt` errors, ideally to one line at most.

I run `streamlink 2.1.1+12.g3fdd208.dirty` on Ubuntu focal as the last step of a script. Sometimes a message from an early step of that script means that I need a manual prepare step before I should start watching, so I press Ctrl+c to abort. Most of the times, this causes streamlinks's python to print about 20 lines of Traceback before the message "KeyboardInterrupt", and a useless blank line after that. Thus I have to scroll up to see the messages from my script. It would be nice to have an easy way to opt-out of that stack trace and the blank line.

bastimeyer:
- Don't raise a KeyboardInterrupt until streamlink_cli has been fully
  initialized. This prevents a stack call from being printed to stderr
  when streamlink_cli gets interrupted early.
- Restore default signal handlers once streamlink_cli has finished its
  initialization, so that KeyboardInterrupt exceptions can be caught to
  perform clean up tasks.
- Restore default signal handlers in test immediately after importing
  streamlink_cli once as early as possible, to be able to cancel the
  test runners regularly via a KeyboardInterrupt.
- Refactor CLI logging tests, which test the log output of the
  initialization and need to stop execution at some point.

----

resolves #4188 

Both the regular entry script and invocation via `python -m streamlink{,_cli}` work.

```bash
$ /usr/bin/timeout --preserve-status --signal=SIGINT .1s streamlink twitch.tv/bobross
$ echo $?
130

$ /usr/bin/timeout --preserve-status --signal=SIGINT 1s streamlink twitch.tv/bobross
[cli][info] Found matching plugin twitch for URL twitch.tv/bobross
Interrupted! Exiting...
$ echo $?
130
```

This does not fix a `KeyboardInterrupt` from being raised and its call stack being printed to `stderr` if the user sends a `SIGINT` signal to the process even before the signal override happens. The signal overrides happen as early as possible, but python has to load parts of its standard library first, so there's a small time window here where we don't have any control.

There's also a secondary time window in the streamlink_cli.main.main code, as it restores the default signal handlers but then only catches KeyboardInterrupt exceptions in some cases. That shouldn't be too important though.

### DIFF
--- a/src/streamlink_cli/__init__.py
+++ b/src/streamlink_cli/__init__.py
@@ -1,0 +1,12 @@
+from signal import SIGINT, SIGTERM, signal
+from sys import exit
+
+
+def _exit(*_):
+    # don't raise a KeyboardInterrupt until streamlink_cli has been fully initialized
+    exit(128 | 2)
+
+
+# override default SIGINT handler (and set SIGTERM handler) as early as possible
+signal(SIGINT, _exit)
+signal(SIGTERM, _exit)
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -685,7 +685,9 @@ def setup_config_args(parser, ignore_unknown=False):
 
 
 def setup_signals():
-    # Handle SIGTERM just like SIGINT
+    # restore default behavior of raising a KeyboardInterrupt on SIGINT (and SIGTERM)
+    # so cleanup code can be run when the user stops execution
+    signal.signal(signal.SIGINT, signal.default_int_handler)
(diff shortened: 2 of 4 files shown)


mk-pmb:
Thank you for fixing it! :+1:

Issue #4205, opened 2021-11-23
Player variables in --title


bastimeyer:
### Checklist

- [X] This is a bug report and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest build from the master branch

### Description

I've noticed this issue while rewriting the [`--title` docs](https://streamlink.github.io/latest/cli.html#cmdoption-title) (mentioned in #4203 that this needs to be done).

## MPV

It looks like [MPV does only support its own format variables](https://mpv.io/manual/stable/#properties) in its [`--title`](https://mpv.io/manual/stable/#options-title) argument but not in [`--force-media-title`](https://mpv.io/manual/stable/#options-force-media-title). `--force-media-title` has replaced `--title` in #3405 last year, because `--title` does only change the window title and not the title of the stream inside MPV's pseudo GUI. `--force-media-title` does both.

Since we want to be able to change both the stream title and window title and therefore can't switch back, the MPV custom format variables are all unsupported. This means that the rather wonky logic for escaping the MPV title string should be removed and the docs also need to be fixed.

- https://github.com/streamlink/streamlink/blame/e4461933539fe7ae518f97ad0d22775bf24f8363/src/streamlink_cli/output.py#L211-L215
- https://github.com/streamlink/streamlink/blame/e4461933539fe7ae518f97ad0d22775bf24f8363/src/streamlink_cli/output.py#L150-L186

Btw, users still can set MPV's `--title` parameter with any property expansion they want, but not via Streamlink's `--title` parameter.  MPV's `{media-title}` variable gets the value of what Streamlink passes via `--force-media-title`. I don't think though that this should be added to the docs, as it's kinda ridiculous. Example:

```sh
$ streamlink \
  --title '{author} - {category} - {title}' \
  -p mpv \
  -a '"--title=${{media-title}} // ${{mpv-version}}" {playerinput}' \
  twitch.tv/gorgc best

$ xdotool search --class mpv getwindowname
Gorgc - Dota 2 - a little cold // mpv 0.34.0
```

## VLC

[...]

PR #4206, opened 2021-11-23
cli.output: remove MPV title variable escape logic

bastimeyer:
Variables are only supported by MPV in its `--title` parameter, and not
its `--force-media-title` parameter.

- remove MPV title variable escape logic
- update docs

----

#4205 

I will update the `--title` and metadata vars docs in a later PR. This will cause merge conflicts otherwise.


beardypig (approved):

bastimeyer:
Fixed the commit message...

PR #4207, opened 2021-11-24
docs: add dedicated metadata variables section

bastimeyer:
- add dedicated metadata variables section to the CLI docs
- rewrite --title help text
- update --output, --record and --record-and-pipe help texts

----

ref #4205 

This moves the metadata variables from the argparse help texts into the CLI docs. This unfortunately means that the argparse help texts can't link to the metadata variables section. At least I don't know how to do this.

It also means that the metadata vars are currently not part of the man page anymore, because the man page only includes the argparse stuff and not the entire CLI document (because we don't want the CLI tutorial in the man page). This could be fixed in the future by splitting up the sections into their own documents and including those in the CLI document and man page.

https://deploy-preview-4207--streamlink.netlify.app/cli.html#metadata-variables
https://deploy-preview-4207--streamlink.netlify.app/cli.html#cmdoption-title

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -10,7 +10,7 @@ from streamlink.utils.args import (
 )
 from streamlink.utils.times import hours_minutes_seconds
 from streamlink_cli.constants import (
-    DEFAULT_STREAM_METADATA, PLAYER_ARGS_INPUT_DEFAULT, PLAYER_ARGS_INPUT_FALLBACK, STREAM_PASSTHROUGH, SUPPORTED_PLAYERS
+    PLAYER_ARGS_INPUT_DEFAULT, PLAYER_ARGS_INPUT_FALLBACK, STREAM_PASSTHROUGH, SUPPORTED_PLAYERS
 )
 from streamlink_cli.utils import find_default_player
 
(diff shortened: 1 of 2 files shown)


gravyboat (approved):
Doc changes look good to me, let's see if anyone spots something I missed.

src/streamlink_cli/argparser.py
```diff
@@ -627,8 +598,13 @@ def build_parser():
...
         You will be prompted if the file already exists.
 
-        The formatting variables available for the --title option may be used.
+        Please see the "Metadata variables" section of Streamlink's CLI documentation for all available metadata variables.
+
         Unsupported characters in substituted variables will be replaced with an underscore.
+
+        Example:
+
+            %(prog)s --record-and-pipe "~/recordings/{author}/{category}/{id}-{time:%Y%m%d%H%M%S}.ts" <URL> [STREAM]
```
mkbloke, line 607: I don't know if it's a Netlify thing only, and it's certainly very minor, but I see that different CSS is applied to `and` in `--record-and-pipe` here.  I thought it was eyes at first!  Technically, that's nothing to do with the content of this PR though, but it's just something I noticed and thought I would mention.

bastimeyer: It's because of the custom argparse extension which tries to build ReST from certain text patterns used in the argparse help texts:
https://github.com/streamlink/streamlink/blob/3309ed0620fa4aedeee218a48996d8284a90bf4b/docs/ext_argparse.py#L22
https://github.com/streamlink/streamlink/blob/3309ed0620fa4aedeee218a48996d8284a90bf4b/docs/ext_argparse.py#L72-L73

Text indented by at least 2 spaces gets turned into a literal-block which then gets highlighted by Sphinx automatically when there's no block type specified:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

There are a couple of other examples in the docs where it's parsing it incorrectly, but we can't use full ReST syntax in the argparse help texts because of the `--help` output and the man page, so it's a bit difficult to fix this.

mkbloke: It's so trivial it was not worth worrying about anyway, but more so if the fix is tricky.  I just wasn't sure if it had been noticed before.

PR #4219, opened 2021-11-28
plugins.svtplay: fix live channel URL matching

zappepappe:
Update to the new plugin URL matching API removed a couple of characters in the matching URL which broke the live channel matching.

### DIFF
--- a/src/streamlink/plugins/svtplay.py
+++ b/src/streamlink/plugins/svtplay.py
@@ -12,7 +12,7 @@ log = logging.getLogger(__name__)
 
 
 @pluginmatcher(re.compile(
-    r'https?://(?:www\.)?(?:svtplay|oppetarkiv)\.se(/(kanaler/)?)'
+    r'https?://(?:www\.)?(?:svtplay|oppetarkiv)\.se(/(kanaler/)?.*)'
 ))
 class SVTPlay(Plugin):
     api_url = 'https://api.svt.se/videoplayer-api/video/{0}'


src/streamlink/plugins/svtplay.py
```diff
@@ -12,7 +12,7 @@
 
 
 @pluginmatcher(re.compile(
-    r'https?://(?:www\.)?(?:svtplay|oppetarkiv)\.se(/(kanaler/)?)'
+    r'https?://(?:www\.)?(?:svtplay|oppetarkiv)\.se(/(kanaler/)?.*)'
```
bastimeyer, line 15: Appending `.*` to the regular expression doesn't do anything, [as `re.match` is being used for matching the input URL](https://github.com/streamlink/streamlink/blob/aa63c4eda9f47805aa7424c3cbeec55b8fe2dc27/src/streamlink/session.py#L339-L344), so it doesn't matter which comes afterwards unless there's an end-of-string anchor (`$`).

https://docs.python.org/3/library/re.html#re.match
> **re.match(pattern, string, flags=0)**
> If zero or more characters at the beginning of *string* match the regular expression pattern, return a corresponding match object.

These are the plugin URL tests, and as you can see, it matches URLs with arbitrary data:
https://github.com/streamlink/streamlink/blob/aa63c4eda9f47805aa7424c3cbeec55b8fe2dc27/tests/plugins/test_svtplay.py#L8-L24

zappepappe:
Well, I'm not going to argue with you on if this is the right approach or not. But it clearly does something as without those characters live channel matching is broken and with them it works.

`streamlink.exe --loglevel debug https://www.svtplay.se/kanaler/svt1?start=auto
[cli][debug] Streamlink: 3.0.3
[cli][debug] Arguments:
[cli][debug]  url=https://www.svtplay.se/kanaler/svt1?start=auto
[cli][info] Found matching plugin svtplay for URL https://www.svtplay.se/kanaler/svt1?start=auto
[plugins.svtplay][debug] Path=/kanaler/
[plugins.svtplay][debug] Live ID=ch-kanaler/
error: Unable to open URL: https://api.svt.se/videoplayer-api/video/ch-kanaler/ (404 Client Error: Not Found for url: https://api.svt.se/videoplayer-api/video/ch-kanaler/)`

`streamlink.exe --loglevel debug https://www.svtplay.se/kanaler/svt1?start=auto
[session][debug] Plugin svtplay is being overridden by C:\Users\Zarper\AppData\Roaming\streamlink\plugins\svtplay.py
[cli][debug] Streamlink: 3.0.3
[cli][debug] Arguments:
[cli][debug]  url=https://www.svtplay.se/kanaler/svt1?start=auto
[cli][info] Found matching plugin svtplay for URL https://www.svtplay.se/kanaler/svt1?start=auto
[plugins.svtplay][debug] Path=/kanaler/svt1?start=auto
[plugins.svtplay][debug] Live ID=ch-svt1
[utils.l10n][debug] Language code: sv_SE
[stream.dash][debug] Available languages for DASH audio streams: sv-x-tal, sv (using: sv)
Available streams: 234p (worst), 360p, 540p, 720p_alt, 720p (best)`

bastimeyer:
> error: Unable to open URL

This has nothing to do with the plugin matcher. This is a different error that needs fixing.

The plugin matches correctly:
> [cli][info] Found matching plugin svtplay for URL https://www.svtplay.se/kanaler/svt1?start=auto

zappepappe:
If the quotes are from me the first part with the error is from the current default and the second part is from loading a plugin with my changes applied.

mkbloke:
@bastimeyer, do you know why `match.groups()` isn't matching the way it used to?  It looks like both `live` and `path` are assigned the inner group match?

zappepappe:
Should I close this pull request and open an issue instead?

bastimeyer:
Apologies, I got confused here for a moment, because the content is geo-blocked for me, so the plugin fails either way.
And yes, the `.*` which got removed in 98428d663e1cac084c4597419445042e6a81f9fd affects the capture group which isn't covered by the tests.

bastimeyer:
Thanks

PR #4232, opened 2021-12-04
plugins.pluto: rewrite/fix

### LINKED ISSUE #3462: PlutoTV stops streaming when switching to a commercial
## Bug Report

- [x ] This is a bug report and I have read the contribution guidelines.

### Description

Streaming stops whenever plutoTV switches to a local commercial.  I have seen this on various pluto news channels such as NEWSY and Newsmax.  After a news segment, the channel will switch to local commercial and a message from pluto states that they will be right back - but the stream hangs at that point, never to return unless, the stream is restarted.

I expect that if the channel switches segments that the stream will continue to play as normal, but it does not do this, the stream stops

1. ...Tune to plutotv- Newsy and watch for 5-10 minutes until the segment changes
2. ...for me at the change, streaming stops
3. ...

### Log output

```
REPLACE THIS TEXT WITH THE LOG OUTPUT
```

### Additional comments, etc.

mkbloke:
- use v4 API
- update URL path matching for "Watch Now" VoD URLs
- filter advert segments
- add more metadata support
- update URL tests for "Watch Now" VoD URLs

closes #3462

### DIFF
--- a/src/streamlink/plugins/pluto.py
+++ b/src/streamlink/plugins/pluto.py
@@ -1,99 +1,171 @@
 import logging
 import re
+from urllib.parse import parse_qs, urljoin
 from uuid import uuid4
 
 from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
-from streamlink.stream.ffmpegmux import MuxedStream
-from streamlink.stream.hls import HLSStream
+from streamlink.stream.hls import HLSStream, HLSStreamReader, HLSStreamWriter
 from streamlink.utils.url import update_qsd
 
 log = logging.getLogger(__name__)
 
 
-@pluginmatcher(re.compile(r'''
+class PlutoHLSStreamWriter(HLSStreamWriter):
+    ad_re = re.compile(r"_ad/creative/|dai\.google\.com")
+
+    def should_filter_sequence(self, sequence):
+        return self.ad_re.search(sequence.segment.uri) is not None or super().should_filter_sequence(sequence)
+
+
+class PlutoHLSStreamReader(HLSStreamReader):
+    __writer__ = PlutoHLSStreamWriter
+
+
+class PlutoHLSStream(HLSStream):
+    __shortname__ = "hls-pluto"
+    __reader__ = PlutoHLSStreamReader
+
+
+@pluginmatcher(re.compile(r"""
     https?://(?:www\.)?pluto\.tv/(?:\w{2}/)?(?:
-        live-tv/(?P<slug_live>[^/?]+)/?$
+        live-tv/(?P<slug_live>[^/]+)
         |
-        on-demand/series/(?P<slug_series>[^/]+)/season/\d+/episode/(?P<slug_episode>[^/]+)$
+        on-demand/series/(?P<slug_series>[^/]+)(?:/season/\d+)?/episode/(?P<slug_episode>[^/]+)
         |
-        on-demand/movies/(?P<slug_movies>[^/]+)$
-    )
-''', re.VERBOSE))
+        on-demand/movies/(?P<slug_movies>[^/]+)
+    )/?$
+""", re.VERBOSE))
 class Pluto(Plugin):
-    def _schema_media(self, slug):
-        return validate.Schema(
-            [{
-                'name': str,
-                'slug': str,
-                validate.optional('stitched'): {
-                    'urls': [
-                        {
-                            'type': str,
-                            'url': validate.url(),
-                        }
-                    ]
-                }
-            }],
-            validate.filter(lambda k: k['slug'].lower() == slug.lower()),
-            validate.get(0),
-        )
-
-    def _get_streams(self):
-        data = None
-
-        m = self.match.groupdict()
-        if m['slug_live']:
-            res = self.session.http.get('https://api.pluto.tv/v2/channels')
-            data = self.session.http.json(res,
-                                          schema=self._schema_media(m['slug_live']))
-        elif m['slug_series'] and m['slug_episode']:
-            res = self.session.http.get(f'http://api.pluto.tv/v3/vod/slugs/{m["slug_series"]}')
-            data = self.session.http.json(
-                res,
-                schema=validate.Schema(
-                    {'seasons': validate.all(
-                        [{'episodes': self._schema_media(m['slug_episode'])}],
-                        validate.filter(lambda k: k['episodes'] is not None))},
-                    validate.get('seasons'),
-                    validate.get(0),
-                    validate.any(None, validate.get('episodes'))
-                ),
-            )
-        elif m['slug_movies']:
-            res = self.session.http.get('https://api.pluto.tv/v3/vod/categories',
-                                        params={'includeItems': 'true', 'deviceType': 'web'})
-            data = self.session.http.json(
-                res,
-                schema=validate.Schema(
-                    {'categories': validate.all(
-                        [{'items': self._schema_media(m['slug_movies'])}],
-                        validate.filter(lambda k: k['items'] is not None))},
-                    validate.get('categories'),
-                    validate.get(0),
-                    validate.any(None, validate.get('items')),
-                ),
-            )
-
-        log.trace(f'{data!r}')
-        if data is None or not data.get('stitched'):
+    def _get_api_data(self, type, slug, filter=None):
+        log.debug(f"slug={slug}")
+        app_version = self.session.http.get(self.url, schema=validate.Schema(
+            validate.parse_html(),
+            validate.xml_xpath_string(".//head/meta[@name='appVersion']/@content"),
+            validate.any(None, str),
+        ))
+        if not app_version:
             return
 
-        self.title = data['name']
-        stream_url_no_sid = data['stitched']['urls'][0]['url']
-        device_id = str(uuid4())
-        stream_url = update_qsd(stream_url_no_sid, {
-            'deviceId': device_id,
-            'sid': device_id,
-            'deviceType': 'web',
-            'deviceMake': 'Firefox',
-            'deviceModel': 'Firefox',
-            'appName': 'web',
-        })
+        log.debug(f"app_version={app_version}")
 
-        self.session.set_option('ffmpeg-fout', 'mpegts')
-        for q, s in HLSStream.parse_variant_playlist(self.session, stream_url).items():
-            yield q, MuxedStream(self.session, s)
+        return self.session.http.get(
+            "https://boot.pluto.tv/v4/start",
+            params={
+                "appName": "web",
+                "appVersion": app_version,
+                "deviceVersion": "94.0.0",
+                "deviceModel": "web",
+                "deviceMake": "firefox",
+                "deviceType": "web",
+                "clientID": str(uuid4()),
+                "clientModelNumber": "1.0",
+                type: slug,
+            },
+            schema=validate.Schema(
+                validate.parse_json(), {
+                    "servers": {
+                        "stitcher": validate.url(),
+                    },
+                    validate.optional("EPG"): [{
+                        "name": str,
+                        "id": str,
+                        "slug": str,
+                        "stitched": {
+                            "path": str,
+                        },
+                    }],
+                    validate.optional("VOD"): [{
+                        "name": str,
+                        "id": str,
+                        "slug": str,
+                        "genre": str,
+                        "stitched": {
+                            "path": str,
+                        },
+                        validate.optional("seasons"): [{
+                            "episodes": validate.all([{
+                                "name": str,
+                                "_id": str,
+                                "slug": str,
+                                "stitched": {
+                                    "path": str,
+                                },
+                            }], validate.filter(lambda k: filter and k["slug"] == filter)),
+                        }],
+                    }],
+                    "sessionToken": str,
+                    "stitcherParams": str,
+                },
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/pluto.py
```diff
@@ -1,99 +1,160 @@
...
+        if data and path:
+            log.trace(f"data={data!r}")
+            log.debug(f"path={path}")
+
+            return self._get_playlist(
+                data["servers"]["stitcher"],
+                path,
+                data["stitcherParams"],
+                data["sessionToken"],
+            )
```
bastimeyer, line 157: This should get simplified. All these lookups are redundant and barely legible.

Since you require `data` and `path` to not be `None` and since `path` only gets set in these nested if-statements, just invert the conditions and return `None`, so that all this nesting can be avoided. Also, why is the `seasons` property optional in the schema but mandatory here for returning streams (otherwise `path` is `None`)?

Example (untested):
```py
    @staticmethod
    def _get_media_data(data, key, slug):
        media = data.get(key)
        if media and media[0]["slug"] == slug:
            return media[0]

    def _get_streams(self):
        m = self.match.groupdict()
        if m["slug_live"]:
            data = self._get_api_data("channelSlug", m["slug_live"])
            media = self._get_media_data(data, "EPG", m["slug_live"])
            if not media:
                return

            self.id = media["id"]
            self.title = media["name"]
            path = media["stitched"]["path"]

        elif m["slug_series"] and m["slug_episode"]:
            data = self._get_api_data("episodeSlugs", m["slug_series"], filter=m["slug_episode"])
            media = self._get_media_data(data, "VOD", m["slug_series"])
            if not media or "seasons" not in media:
                return
            episode = media["seasons"][0]["episodes"][0]
            if episode["slug"] != m["slug_episode"]:
                return

            self.author = media["name"]
            self.category = media["genre"]
            self.id = episode["_id"]
            self.title = episode["name"]
            path = episode["stitched"]["path"]

        elif m["slug_movies"]:
            data = self._get_api_data("episodeSlugs", m["slug_movies"])
            media = self._get_media_data(data, "VOD", m["slug_movies"])
            if not media:
                return

            self.category = media["genre"]
            self.id = media["id"]
            self.title = media["name"]
            path = media["stitched"]["path"]

        else:
            return

        log.trace(f"data={data!r}")
        log.debug(f"path={path}")

        return self._get_playlist(
            data["servers"]["stitcher"],
            path,
            data["stitcherParams"],
            data["sessionToken"],
        )
```

bastimeyer (commented):
Thanks for the PR.

Just a quick review of the diff... I didn't check the website and validity of the plugin code except for the first URL that was added to the regex tests, and that didn't work. I think it's because the site is returning different data. It's showing some German Star Trek Episode for me instead of the show from the URL.

emveepee:
> Thanks for the PR.
> 
> Just a quick review of the diff... I didn't check the website and validity of the plugin code except for the first URL that was added to the regex tests, and that didn't work. I think it's because the site is returning different data. It's showing some German Star Trek Episode for me instead of the show from the URL.

Does the release streamlink work for you for series VOD?  Perhaps in your region they haven't move to the v4 calls used in the US.  For v4 VOD it seems to work well here.

mkbloke:
@bastimeyer: was it: `http://pluto.tv/en/on-demand/series/great-british-menu-ptv1/season/5/episode/north-west-fish-2009-5-7-ptv1` you tried?

I'm guessing this is from an IP in Germany?  I'll take and look and try to see what's going on there.  It should be the case that content not available in a given country will not return any alternative.  That seemed to the case for both live and VoD content that isn't available in the UK, for example, for me when I was testing it.

> Does the release streamlink work for you for series VOD? Perhaps in your region they haven't move to the v4 calls used in the US. For v4 VOD it seems to work well here.

That's a possibility.  I'll check in a browser from a German IP with any known problematic VoD URLs to see what happens there verses what the plugin is doing.

bastimeyer:
No, the show from the link in the tests
https://pluto.tv/en/on-demand/series/great-british-menu-ptv1/season/5/episode/north-west-fish-2009-5-7-ptv1
redirects to
https://pluto.tv/en/on-demand
because it's not available here, so different data is returned.

This isn't a problem of the plugin though.

src/streamlink/plugins/pluto.py
```diff
@@ -1,99 +1,160 @@
...
 log = logging.getLogger(__name__)
 
 
-@pluginmatcher(re.compile(r'''
+class PlutoHLSStreamWriter(HLSStreamWriter):
+    ad_re = re.compile(r"_ad/creative/|dai\.google\.com")
+
+    def should_filter_sequence(self, sequence):
+        return (self.ignore_names and self.ignore_names.search(sequence.segment.uri) is not None) \
+            or self.ad_re.search(sequence.segment.uri) is not None
```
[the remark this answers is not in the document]
mkbloke: Ah, yes, that makes sense.

src/streamlink/plugins/pluto.py
```diff
@@ -1,99 +1,160 @@
...
         )
 
+    def _get_playlist(self, host, path, params, token):
+        qs = parse_qs(params)
+        qs["jwt"] = token
+        for q, s in HLSStream.parse_variant_playlist(
+            self.session,
+            update_qsd(urljoin(host, path), qs),
+        ).items():
+            yield q, PlutoHLSStream(self.session, s.url)
```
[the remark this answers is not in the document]
mkbloke: > Does yield from PlutoHLSStream.parse_variant_playlist(...).items() not work?

Yes, it does, of course!  Thanks.

> streamlink.utils.parse.update_qsd() already parses and adds query string parameters, so you don't have to parse it manually and update the dict manually. update_qsd(urljoin(host, path), {"jwt": token}) should work (untested).

https://github.com/streamlink/streamlink/blob/f3a1b77fe36b8ce47768e91ae89a0549d61dd05e/src/streamlink/plugins/pluto.py#L104-L106

The complete URL requires the stitcher parameters, passed via `params` as a string and then to also add `jwt` for a valid query string.  Is there a better way to combine the two query string elements together in order to build the required complete URL?

src/streamlink/plugins/pluto.py
mkbloke: That does look a lot cleaner, thanks.

> Also, why is the seasons property optional in the schema but mandatory here for returning streams (otherwise path is None)?

The `seasons` property does not appear in the returned API data for movies, so it's optional there, but when looking up series based VoD content it's mandatory to at least get the correct `path` property.  Does that seem reasonable?

src/streamlink/plugins/pluto.py
```diff
@@ -1,99 +1,160 @@
...
+                    validate.optional("VOD"): [{
+                        "name": str,
+                        "id": str,
+                        "slug": str,
+                        "genre": str,
+                        "stitched": {
+                            "path": str,
+                        },
+                        validate.optional("seasons"): [{
+                            "episodes": validate.Schema([{
```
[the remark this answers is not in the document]
mkbloke: Ah, thanks.

mkbloke:
I haven't updated the query string processing as yet, because I haven't heard back about alternatives.

I assume (I was confused by the word "showing"), that what was actually meant was that the `log.trace()` output shows the data that matches the redirected URL not the input URL, but that the plugin does the right thing by not returning a stream when the content is not available in certain countries.

src/streamlink/plugins/pluto.py
```diff
@@ -1,99 +1,166 @@
...
+            self.id = media["id"]
+            self.title = media["name"]
+            path = media["stitched"]["path"]
+
+        elif m["slug_series"] and m["slug_episode"]:
+            data = self._get_api_data("episodeSlugs", m["slug_series"], filter=m["slug_episode"])
+            media = self._get_media_data(data, "VOD", m["slug_series"])
+            if not media or "seasons" not in media:
+                return
+            episode = media["seasons"][0]["episodes"][0]
```
bastimeyer, line 131: `media["seasons"][0]["episodes"][0]` doesn't work if you try to watch a show that's not part of season 1 (index 0). Every `episodes` list will be empty other than the one of the current seasons object if it's not the first season.

```
$ streamlink -l debug 'https://pluto.tv/en/on-demand/series/south-park-de/season/13/episode/der-ring-2008-13-1301' best
[cli][debug] OS:         Linux-5.15.5-2-git-x86_64-with-glibc2.33
[cli][debug] Python:     3.9.9
[cli][debug] Streamlink: 3.0.3+3.g59e8669
[cli][debug] Arguments:
[cli][debug]  url=https://pluto.tv/en/on-demand/series/south-park-de/season/13/episode/der-ring-2008-13-1301
[cli][debug]  stream=['best']
[cli][debug]  --player=mpv
[cli][info] Found matching plugin pluto for URL https://pluto.tv/en/on-demand/series/south-park-de/season/13/episode/der-ring-2008-13-1301
[plugins.pluto][debug] slug=south-park-de
[plugins.pluto][debug] app_version=5.106.0-f3e2ac48d1dbe8189dc784777108b725b4be6be2
  File "/home/basti/.local/bin/streamlink", line 33, in <module>
    sys.exit(load_entry_point('streamlink', 'console_scripts', 'streamlink')())
  File "/home/basti/repos/streamlink/src/streamlink_cli/main.py", line 1056, in main
    handle_url()
  File "/home/basti/repos/streamlink/src/streamlink_cli/main.py", line 565, in handle_url
  File "/home/basti/repos/streamlink/src/streamlink_cli/main.py", line 459, in fetch_streams
    return plugin.streams(stream_types=args.stream_types,
  File "/home/basti/repos/streamlink/src/streamlink/plugin/plugin.py", line 336, in streams
  File "/home/basti/repos/streamlink/src/streamlink/plugins/pluto.py", line 131, in _get_streams
    episode = media["seasons"][0]["episodes"][0]
IndexError: list index out of range
```

mkbloke: Ah, well spotted, thanks.  I must have checked this only in cases where the season happened to be at index 0, which of course is unrelated to the actual season number, which is what I was going by.

Is there any way to filter the empty episodes within the schema also?  I'm having trouble trying to work out how that might be achieved, if it's possible.  Perhaps it might be better to loop through the episodes checking the slug of each for a match within the main plugin code instead?

bastimeyer: I'd say that the validation schema should be kept as simple as possible.

bastimeyer (approved):
I've tested live streams, VODs (series) and VODs (movies), and it seems to work fine now. However, I can't check some of the links added to the tests as they are not available to me. It does fail with an empty result though, so it should be fine.

emveepee:
> I've tested live streams, VODs (series) and VODs (movies), and it seems to work fine now. However, I can't check some of the links added to the tests as they are not available to me. It does fail with an empty result though, so it should be fine.

Perhaps add a second test for the non pvt1 variant of that stream,  https://pluto.tv/en/on-demand/series/great-british-menu/season/5/episode/north-west-fish-2009-5-7

I've never figured out why some streams have that added of ptv#

emveepee:
> Does it work differently for you and if so, from which country are you playing it?  

US via VPS  

```
[cli][info] Found matching plugin pluto for URL https://pluto.tv/en/on-demand/series/great-british-menu/season/5/episode/north-west-fish-2009-5-7
[plugins.pluto][debug] slug=great-british-menu
[plugins.pluto][debug] app_version=5.106.0-f3e2ac48d1dbe8189dc784777108b725b4be6be2
[plugins.pluto][debug] path=/stitch/hls/episode/60ede8997afdca00139e7d0b/master.m3u8
Available streams: 570k (worst), 1000k, 1500k, 2100k, 3100k (best)

mkbloke:
OK.  I'm not sure why that's the case.  My theory is that perhaps there can be different edits in different regions, so they have unique slugs for those edits and catalogue them separately.

I see that if I try to use the UK URL in the US, the website behaves in the same manner, i.e. it redirects to the main On Demand landing page.

My feeling is that the plugin as it is now replicates the experience you get on the website.  Am I right in thinking that you are proposing to try again with a slug that has the `ptv-[0-9]` removed?  I'm not sure that's the right way to go, although I'd implement it if the consensus was that it was worthwhile.

emveepee:
> My feeling is that the plugin as it is now replicates the experience you get on the website. Am I right in thinking that you are proposing to try again with a slug that has the `ptv-[0-9]` removed? 

No you can't just remove the ptv values they won't work.  I was suggesting having multiple test cases until this can be figured out.  In Canada there is little on pluto.tv and no vod so most tests fail anyway.

mkbloke:
I'm not quite following what you're proposing.  How do you envisage these test cases working?

emveepee:
> I'm not quite following what you're proposing. How do you envisage these test cases working?

I guess my thinking is don't put those pvt links in the test if they can be avoided, they are confusing.  I assume the only purpose is to test the regex and not that they provide an m3u8.

mkbloke:
> I guess my thinking is don't put those pvt links in the test if they can be avoided, they are confusing.

Ah, I see.  I was completely misunderstanding what you were talking about before now!

> I assume the only purpose is to test the regex and not that they provide an m3u8.

That's correct.  Any URL you put in the tests for any plugins have the possibility that they may not be accessible in countries outside of the one that the happened to be used during development.  I'm not sure if adding a non-pvt URL would help to reduce any confusion though?

PR #4244, opened 2021-12-11
plugins.steam: refactor plugin

bastimeyer:
- replace itertags with XPath
- simplify logic a bit
- move schemas from plugin class scope
- fix log messages


gravyboat (approved):
Forgot this plugin even existed. Thanks for the refactor.

PR #4258, opened 2021-12-19
logger: change NONE loglevel to sys.maxsize

### LINKED ISSUE #4235: output with --json flag is not parseable if handled error occurs
### Checklist

### Streamlink version

Latest stable release

### Steps to reproduce
1. Run streamlink against a YouTube channel that is not live, e.g. `streamlink youtube.com/c/LudwigAhgren -j`
2. Output is unparseable as JSON:

### Expected Behavior
All output on stdout should be parseable when the  `--json` is enabled (barring an unrecoverable error)
```json
{
  "error": "No playable streams found on this URL: youtube.com/c/LudwigAhgren"
}
```
### Actual Behavior
Plugin errors are printed to stdout along with the parseable JSON output
```json
[plugins.youtube][error] Could not find videoId on channel page
{
  "error": "No playable streams found on this URL: youtube.com/c/LudwigAhgren"
}
```

### Possible Solutions
* Write error messages to stderr instead of stdout when `--json` flag is enabled (or always)
* Suppress error messages when `--json` flag is enabled, since the content is duplicated in the JSON output

### Debug log

```text
[plugins.youtube][error] Could not find videoId on channel page
{
  "error": "No playable streams found on this URL: youtube.com/c/LudwigAhgren"
}
```

*Note*:   
the debug log is with the proper `--loglevel debug` param, it seems `--json` suppresses debug messages already. System info is as follows:
```
[cli][debug] Streamlink: 3.0.3
```

bastimeyer:
Resolves #4235 

The reason why there are still log messages being written to stdout/stderr when any "silent" CLI argument is set (`--json`, `--quiet`, `--stream-url` or even `--loglevel=none`) is because of how Streamlink's `NONE` log level gets interpreted in Python's logging module.

Each logger instance (subclasses of `logging.Logger` or `streamlink.logger.StreamlinkLogger`) can optionally have their own log level. When a logger instance needs to check whether a certain message can be written to the output or not, it first tries to read its own log level value, and if it is set to `logging.UNSET` (which is the default and its value is 0), then it continues with its parent logger instance until it reaches `logging.root` (`RootLogger`). This RootLogger does have a default log level of 30, aka. "warning".
https://github.com/python/cpython/blame/3.10/Lib/logging/__init__.py#L1926-L1927

`logging.Logger.getEffectiveLevel()`:
https://github.com/python/cpython/blame/3.10/Lib/logging/__init__.py#L1701-L1713

> ```py
>     def getEffectiveLevel(self):
>         """
>         Get the effective level for this logger.
>         Loop through this logger and its parents in the logger hierarchy,
>         looking for a non-zero logging level. Return the first one found.
>         """
>         logger = self
>         while logger:
>             if logger.level:
>                 return logger.level
>             logger = logger.parent
>         return NOTSET
> ```

Since child loggers usually don't have their own log level set, the loggers like the one in the YouTube plugin for example look up the level of Streamlink's own root logger `streamlink.logger.root`. If the level of Streamlink's root logger however is set to `NONE` (0), then the level of the global RootLogger will be used ("warning"), which means everything equal to or above "warning", like error log messages for example, will still be logged.

----

The fix is to change the `streamlink.logger.NONE` value from 0 to `sys.maxsize`, so that `getEffectiveLevel()` doesn't continue with the next parent logger if one of the logger instance's level is set to `NONE`. This has been broken since the reimplementation of the logger module in 2018.

Please review carefully before merging, because I don't want to unintentionally break stuff.

The other changes are all optional. See the commit messages for details.

----

https://github.com/streamlink/streamlink/issues/4235#issue-1071382500
```
$ streamlink youtube.com/c/LudwigAhgren -j
{
  "error": "No playable streams found on this URL: youtube.com/c/LudwigAhgren"
}
```

### DIFF
--- a/src/streamlink/logger.py
+++ b/src/streamlink/logger.py
@@ -1,24 +1,41 @@
 import logging
+import sys
 from datetime import datetime
-from logging import CRITICAL, DEBUG, ERROR, INFO, NOTSET, WARNING
+from logging import CRITICAL, DEBUG, ERROR, INFO, WARNING
+from pathlib import Path
 from threading import Lock
+from typing import IO, List, Optional, Union
 
+
+FORMAT_STYLE = "{"
+FORMAT_BASE = "[{name}][{levelname}] {message}"
+FORMAT_DATE = "%H:%M:%S"
+REMOVE_BASE = ["streamlink", "streamlink_cli"]
+
+# Make NONE ("none") the highest possible level that suppresses all log messages:
+#  `logging.NOTSET` (equal to 0) can't be used as the "none" level because of `logging.Logger.getEffectiveLevel()`, which
+#  loops through the logger instance's ancestor chain and checks whether the instance's level is NOTSET. If it is NOTSET,
+#  then it continues with the parent logger, which means that if the level of `streamlink.logger.root` was set to "none" and
+#  its value NOTSET, then it would continue with `logging.root` whose default level is `logging.WARNING` (equal to 30).
+NONE = sys.maxsize
+# Add "trace" to Streamlink's log levels
 TRACE = 5
-_levelToName = {
+
+# Define Streamlink's log levels (and register both lowercase and uppercase names)
+_levelToNames = {
+    NONE: "none",
     CRITICAL: "critical",
     ERROR: "error",
     WARNING: "warning",
     INFO: "info",
     DEBUG: "debug",
     TRACE: "trace",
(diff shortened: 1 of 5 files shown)


gravyboat (approved):
Looks good to me @bastimeyer, good investigation in to the root cause. I'll leave this open today to see if others have any concerns, if not let's merge this in.

PR #4261, opened 2021-12-20
plugins.yupptv: override encoding, set Origin header

chkuendig:
Not sure for how long this was broken - but for me YuppTV doesn't work without overriding the encoding and adding an addition header.

### DIFF
--- a/src/streamlink/plugins/yupptv.py
+++ b/src/streamlink/plugins/yupptv.py
@@ -4,27 +4,27 @@ import time
 
 from streamlink.plugin import Plugin, PluginArgument, PluginArguments, pluginmatcher
 from streamlink.plugin.api import useragents
-from streamlink.stream.hls import HLSStream, HLSStreamReader, HLSStreamWorker
+from streamlink.stream.hls import HLSStream, HLSStreamReader
 
 log = logging.getLogger(__name__)
 
-def override_encoding(resp, *args, **kwargs):
-    resp.encoding = "utf-8"
 
-class YuppTVHLSStreamWorker(HLSStreamWorker):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs);
-        self.reader.request_params.update(hooks={'response': override_encoding})
- 
 class YuppTVHLSStreamReader(HLSStreamReader):
-    __worker__ = YuppTVHLSStreamWorker
+    def __init__(self, stream):
+        super().__init__(stream)
+        self.request_params.update(hooks={"response": stream.override_encoding})
 
 
 class YuppTVHLSStream(HLSStream):
     __reader__ = YuppTVHLSStreamReader
+
+    @staticmethod
+    def override_encoding(res, **kwargs):
+        res.encoding = "utf-8"
+
     @classmethod
     def _get_variant_playlist(cls, res):
-        res.encoding = "UTF-8"
+        cls.override_encoding(res)
         return super()._get_variant_playlist(res)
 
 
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/yupptv.py
```diff
@@ -4,10 +4,29 @@
...
+    __worker__ = YuppTVHLSStreamWorker
+
+
+class YuppTVHLSStream(HLSStream):
+    __reader__ = YuppTVHLSStreamReader
+    @classmethod
+    def _get_variant_playlist(cls, res):
+        res.encoding = "UTF-8"
+        return super()._get_variant_playlist(res)
+
```
bastimeyer, line 29:
```suggestion
from streamlink.stream.hls import HLSStream, HLSStreamReader

log = logging.getLogger(__name__)


class YuppTVHLSStreamReader(HLSStreamReader):
    def __init__(self, stream):
        super().__init__(stream)
        self.request_params.update(hooks={"response": stream.override_encoding})


class YuppTVHLSStream(HLSStream):
    __reader__ = YuppTVHLSStreamReader

    @staticmethod
    def override_encoding(res, **kwargs):
        res.encoding = "utf-8"

    @classmethod
    def _get_variant_playlist(cls, res):
        cls.override_encoding(res)
        return super()._get_variant_playlist(res)

```

bastimeyer (commented):
Thanks for the PR.
Some code changes are required before this can be merged.

Regarding the encoding, I can't tell if it's indeed required. The response encoding is automatically detected by [charset_normalizer](https://github.com/ousret/charset_normalizer) if the `Response.encoding` attribute is not set, and I don't see anything weird in the HLS playlists of the freely available content on the site (I don't have an account and don't want to create one).

back-to:
> Regarding the encoding, I can't tell if it's indeed required.

you got **_chardet_** installed

---

it is an issue with `charset_normalizer` is doesnt work again

https://github.com/streamlink/streamlink/blob/0cc78e7db9c76a3c52297c7a32938534625d4110/src/streamlink/plugins/goltelevision.py
https://github.com/streamlink/streamlink/blob/a0437c2e7bd62bce744eb922951e2c7bf580fa1d/src/streamlink/plugins/raiplay.py

---

would work with chardet


https://github.com/chardet/chardet

with chardet installed

```
[cli][info] Available streams: 216p (worst), 270p, 288p, 396p, 406p (best)
[cli][info] Opening stream: 406p (hls)
[stream.hls][debug] First Sequence: 12075; Last Sequence: 12104
[stream.hls][debug] Start offset: 0; Duration: None; Start Sequence: 12102; End Sequence: None
[stream.hls][debug] Adding segment 12102 to queue
```

without

`[cli][error] Try 1/1: Could not open stream <HLSStream('https://yuppmslukmo1.akamaized.net/h`

chkuendig:
I can confirm that the encoding is required. I took the raiplay fix in 3dad9152790a3711e71a18f7f2cb43fab0442214 actually as inspiration for this PR (with the addition of the hook to also fix it in the `chunks.m3u8` playlists)

bastimeyer:
> you got **_chardet_** installed

```
$ pip list | grep char
charset-normalizer            2.0.9
```

I am aware that python-requests still tries to import chardet first, even though it requires charset_normalizer on py3:
https://github.com/psf/requests/blob/v2.26.0/requests/compat.py#L11-L14
https://github.com/psf/requests/blob/v2.26.0/setup.py#L44-L45

bastimeyer:
`HLSStream` doesn't need to be subclassed. Any additional keywords from `HLSStream.parse_variant_playlist()` will be passed to the subsequent `session.http.get()` calls, so adding
```py
    @staticmethod
    def _override_encoding(res, **kwargs):
        res.encoding = "utf-8"
```
to the plugin class and calling
```py
HLSStream.parse_variant_playlist(self.session, stream_url, hooks={"response": self._override_encoding})
```
should be enough.

PR #4267, opened 2021-12-26
plugins.vk: rewrite and remove itertags

bastimeyer:
- drop IPHONE6 user agent
- update matcher regexes
- rewrite redirection logic
- rewrite validation schemas, remove itertags
- add id, author and title metadata
- drop HTTPStream and add DASHStream, prefer HLSStream
- update tests

----

Tested various VODs and live streams, but not 100% sure if I've missed something...

### DIFF
--- a/src/streamlink/plugins/vk.py
+++ b/src/streamlink/plugins/vk.py
@@ -1,78 +1,102 @@
 import logging
 import re
-from html import unescape as html_unescape
 from urllib.parse import parse_qsl, unquote, urlparse
 
 from streamlink.exceptions import NoStreamsError
-from streamlink.plugin import Plugin, pluginmatcher
-from streamlink.plugin.api import useragents
-from streamlink.plugin.api.utils import itertags
+from streamlink.plugin import Plugin, PluginError, pluginmatcher
+from streamlink.plugin.api import validate
+from streamlink.stream.dash import DASHStream
 from streamlink.stream.hls import HLSStream
-from streamlink.stream.http import HTTPStream
-from streamlink.utils.url import update_scheme
 
 log = logging.getLogger(__name__)
 
 
 @pluginmatcher(re.compile(
-    r"https?://(?:\w+\.)?vk\.com/video(?:\?z=video)?(?P<video_id>-?\d*_\d*)"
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/vk.py
```diff
@@ -1,78 +1,97 @@
...
 log = logging.getLogger(__name__)
 
 
 @pluginmatcher(re.compile(
-    r"https?://(?:\w+\.)?vk\.com/video(?:\?z=video)?(?P<video_id>-?\d*_\d*)"
+    r"https?://(?:\w+\.)?vk\.com/videos?(?:\?z=video)?(?P<video_id>-?\d+_\d+)"
 ))
 @pluginmatcher(re.compile(
-    r"https?://(\w+\.)?vk\.com/videos-?\d*"
+    r"https?://(\w+\.)?vk\.com/.+"
```
back-to, line 18: > https://vk.com/video/for_kids?z=video%2Fpl_-22277933_51053000

some playlist url might result in an IndexError

```
    video_id = self.match.group("video_id")
IndexError: no such group
```

src/streamlink/plugins/vk.py
```diff
@@ -1,78 +1,97 @@
...
+                    validate.transform(lambda text: re.sub(r"^\s*<!--\s*", "", text)),
+                    validate.parse_json(),
+                    {"payload": list},
+                    validate.get(("payload", -1)),
+                    list,
+                    validate.get(-1),
+                    {"player": {"params": [dict]}},
+                    validate.get(("player", "params", 0)),
+                    {
+                        validate.optional("hls"): validate.url(path=validate.endswith(".m3u8")),
```
back-to, line 73: > Unable to validate response text: Unable to validate key 'hls': Unable to validate URL attribute 'path': '/video_hls.php' does not end with '.m3u8'

some urls don't use `.m3u8`

`https://vk.com/video_hls.php?dirs=&path=&extra=&c_extra=&video=&tag=&c_tag=&hash=&lang_id=`

src/streamlink/plugins/vk.py
bastimeyer: That should be resolved with the latest change.

PR #4271, opened 2021-12-28
plugins.dogus: update and cleanup

back-to:


src/streamlink/plugins/dogus.py
```diff
@@ -1,57 +1,43 @@
...
-                                                    headers={"Referer": self.url})
+        root = self.session.http.get(self.url, schema=validate.Schema(validate.parse_html()))
+
+        # https://www.ntv.com.tr/canli-yayin/ntv?youtube=true
+        yt_iframe = root.xpath("string(.//iframe[contains(@src,'youtube.com')]/@src)")
+        # https://www.startv.com.tr/canli-yayin
+        dm_iframe = root.xpath("string(.//iframe[contains(@src,'dailymotion.com')]/@src)")
+        # https://www.kralmuzik.com.tr/tv/kral-tv
+        # https://www.kralmuzik.com.tr/tv/kral-pop-tv
+        yt_script = root.xpath("string(.//script[contains(text(), 'youtube.init')]/text())")
```
bastimeyer, line 25: It's probably better to add a `[1]` XPath predicate to each node, because otherwise it could return `@src`/`text()` values of multiple nodes. So for example
```
string(.//iframe[contains(@src,'youtube.com')][1]/@src)
```

Same in the XPath query down below.

Issue #4300, opened 2022-01-19
ValueError: Padding is incorrect.


raoalinawaz-cys:
### Checklist

- [X] This is a plugin issue and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

`streamlink --http-header "User-Agent=Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.71 Safari/537.36" --http-header "Referer=https://www.log2base2.com/" "https://d2bzjah96pmxyc.cloudfront.net/C/Functions/Recursion/index.m3u8" best -l debug -o out1.ts`

`hls_audio_160k_v4.m3u8` has these `#EXT-X-BYTERANGE` lines.

#EXT-X-BYTERANGE:228432@0
#EXT-X-BYTERANGE:224288
#EXT-X-BYTERANGE:224864

...

And streamlink is setting these Range headers for HTTP GET requests.

Range: bytes=0-228431
Range: bytes=0-224287
Range: bytes=0-224863
...

Comment by:
@kikuyan

### Debug log

```text
streamlink --http-header "User-Agent=Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.71 Safari/537.36" --http-header "Referer=https://www.log2base2.com/" "https://d2bzjah96pmxyc.cloudfront.net/C/Functions/Recursion/index.m3u8" best -l debug -o out1.ts
[cli][debug] OS: Windows 10
[cli][debug] Python: 3.9.8
[cli][debug] Streamlink: 3.0.3
[cli][debug] Requests(2.26.0), Socks(1.7.1), Websocket(1.2.1)
[cli][debug] Arguments:
[cli][debug] url=https://d2bzjah96pmxyc.cloudfront.net/C/Functions/Recursion/index.m3u8
[cli][debug] stream=['best']
[cli][debug] --loglevel=debug
[cli][debug] --output=out1.ts
[...]

PR #4302, opened 2022-01-19
stream.hls: refactor segment byterange calculation

bastimeyer:
- Move "Range" header value calculation to separate class:
  Don't cache offsets by segment URL, and instead refer to the previous
  sequence. Initialization maps always require an existing offset value.
- Refactor HLSStreamWriter.fetch and related methods
- Add byterange tests

----

rel #4300, #4301

`EXT-X-BYTERANGE`
https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.2
`EXT-X-MAP`
https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.5

The spec does not say anything about mandatory offset values for initialization map byteranges, so I'm not sure if what I've implemented is correct. It currently raises a StreamError when the offset is missing on a map's byterange, and the download will fail. It could also default to zero in this case, but I'm not sure if this is implied by the specification, because maps can't refer to the previous segment like regular byterange tags do. Am I missing something?

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -1,10 +1,10 @@
 import logging
 import re
 import struct
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict
 from concurrent.futures import Future
 from threading import Event
-from typing import List, NamedTuple, Optional, Union
+from typing import List, NamedTuple, Optional, Tuple, Union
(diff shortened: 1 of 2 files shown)


gravyboat:
I don't think you're missing anything. It seems unclear in the spec to me regarding what that offset value should be.

PR #4303, opened 2022-01-20
Update http_server.py

zjzh:
I refactored code with set comprehension which is more concise and efficient，how do think this change which has practical value？

### DIFF
--- a/src/streamlink_cli/utils/http_server.py
+++ b/src/streamlink_cli/utils/http_server.py
@@ -33,9 +33,7 @@ class HTTPServer:
 
         addrs = set()
         try:
-            addrs = {info[4][0] 
-                     for info in socket.getaddrinfo(socket.gethostname(), self.port, 
-                                                    socket.AF_INET)}
+            addrs = {info[4][0] for info in socket.getaddrinfo(socket.gethostname(), self.port, socket.AF_INET)}
         except socket.gaierror:
             pass
 
--- a/src/streamlink_cli/utils/http_server.py
+++ b/src/streamlink_cli/utils/http_server.py
@@ -33,9 +33,9 @@ class HTTPServer:
 
         addrs = set()
         try:
-            for info in socket.getaddrinfo(socket.gethostname(), self.port,
-                                           socket.AF_INET):
-                addrs.add(info[4][0])
+            addrs = {info[4][0] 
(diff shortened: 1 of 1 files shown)


src/streamlink_cli/utils/http_server.py
```diff
@@ -33,9 +33,9 @@ def addresses(self):
 
         addrs = set()
         try:
-            for info in socket.getaddrinfo(socket.gethostname(), self.port,
-                                           socket.AF_INET):
-                addrs.add(info[4][0])
+            addrs = {info[4][0] 
+                     for info in socket.getaddrinfo(socket.gethostname(), self.port, 
+                                                    socket.AF_INET)}
```
bastimeyer, line 38: You're redefining the set from the line above and there's another set.add() call down below. Efficiency-wise, there won't be any difference here, and regarding conciseness, please don't wrap this over multiple lines, at least not the parameters of the `socket.getaddrinfo` call, which was already unnecessary.

zjzh: Thanks for your comments, I change the format, andyes there is another set.add(), but it should improve the performance of 36-38 line and it should not affect another set.add() call down below.

back-to:
This change seems unnecessary

bastimeyer:
Let's close this.

The efforts of trying to clean up some code are appreciated, especially from a first-time contributor, thank you, but this change in particular doesn't do anything relevant to the performance of Streamlink / Streamlink CLI, as claimed in the OP. This is just a change for the sake of a change, which doesn't improve anything. Apart from this, the redefinition of the set hasn't been addressed and it doesn't make sense having a set-comprehension when there's a set.add call afterwards anyway, so this unnecessarily mixes styles which wasn't an issue before.

Code cleanups like this should be done in larger chunks on the entire codebase and not just some random lines of code in random modules (which unfortunately are also untested / not covered by tests), and they need to make sense, readability-wise, code-style-wise and performance-wise.

zjzh:
Thanks for your comments. The redefinition would not affect the program because there is no statements is in the  `addrs = set()` and `addrs.add("127.0.0.1")`. We do not remove  the definition because if it occur exception, following code fragments could still run normally. And I also agree with you that it maybe better that the code cleanups is done in larger chunks, and our tool could complete the task in larger chunks and owing to some consideration, we still not open the tool.

PR #4309, opened 2022-01-25
release: 3.1.1

bastimeyer:
Just checked the installer from the nightly build from a couple of minutes ago and it's working fine, as expected.


gravyboat:
I can release this when I get done with work tonight.

Issue #4277, opened 2021-12-31
Python 3.6 has reached its EOL


bastimeyer:
Python 3.6 has had its official EOL on 2021-12-23
https://endoflife.date/python
https://www.python.org/dev/peps/pep-0494/#lifespan

Dropping Python 3.6 and setting the minimum version requirement to 3.7 enables the following main features:

- forward references for type hints (with `__future__` import)
- contextvars and asyncio improvements
- dataclasses as a mutable and subclassable alternative to named tuples
- `collections.OrderedDict` has become obsolete, as ordered `dict`s are now standardized (not an implementation detail anymore like in 3.6)

https://docs.python.org/3/whatsnew/3.7.html

Dropping 3.6 doesn't have to be done right now, but it's something we should keep in mind for one of the next releases.

PR #4330, opened 2022-02-03
installer: upgrade python from 3.9.8 to 3.9.10

bastimeyer:
We could probably also upgrade to 3.10.2, but I don't feel like testing/validating this now...

https://www.python.org/downloads/release/python-3910/


garoto:
Unrelated to this commit, but are there any plans on deprecating Python v3.6 as the minimum requirement for a `pip` install?

bastimeyer:
@garoto #4277
This won't be deprecated. Support for py36 will be removed with the next major version bump. A date doesn't exist yet for 4.0.0.

garoto:
Understood and thanks for the reply.

PR #4335, opened 2022-02-05
plugins.filmon: fix/update

### LINKED ISSUE #4332: plugins.filmon: all channels fail to start. Debug show invalid server response
### Checklist

### Streamlink version

Latest stable release

### Description

I have tried diferent versions of streamlink including the latest. Ubuntu and windows.

When running streamlink.exe https://www.filmon.com/tv/itv1 low --player-external-http --player-external-http-port 23555 --player-continuous-http 

it says channel sleep and invaild server responce.

### Debug log

```text
C:\Program Files (x86)\Streamlink\bin>streamlink.exe https://www.filmon.com/tv/itv1 low --player-external-http --player-external-http-port 23555 --player-continuous-http --loglevel debug
[cli][debug] Streamlink: 3.1.1
[cli][debug] Arguments:
[cli][debug]  url=https://www.filmon.com/tv/itv1
[cli][debug]  stream=['low']
[cli][debug]  --player-continuous-http=True
[cli][debug]  --player-external-http=True
[cli][debug]  --player-external-http-port=23555
[cli][debug]  --ffmpeg-ffmpeg=C:\Program Files (x86)\Streamlink\ffmpeg\ffmpeg.exe
[cli][info] Found matching plugin filmon for URL https://www.filmon.com/tv/itv1
[plugins.filmon][debug] Found channel ID: 11
[plugins.filmon][debug] invalid server response
[plugins.filmon][debug] channel sleep 1
[plugins.filmon][debug] invalid server response
[plugins.filmon][debug] channel sleep 2
[plugins.filmon][debug] invalid server response
[plugins.filmon][debug] channel sleep 3
[plugins.filmon][debug] invalid server response
[plugins.filmon][debug] channel sleep 4
[plugins.filmon][debug] invalid server response
[plugins.filmon][debug] Reset cached channel: itv1
error: Unable to find 'self.api.channel' for 11
```

### LINKED ISSUE #4334: Getting session_key
### Checklist

### Description

I've tried for months to figure out how to get the session_key, live headers, curl and more and all to no avail. Could you help me in how it's done??
Thanks,
Jason

mkbloke:
- use alternative/current sources for live and VOD stream info
- update schemas
- update URL regex
- update tests
- add support for mp4 VOD sources
- add note to docs regarding mp4 VODs
- use TLSSecLevel1Adapter()

closes #4332

### DIFF
--- a/src/streamlink/plugins/filmon.py
+++ b/src/streamlink/plugins/filmon.py
@@ -6,8 +6,10 @@ from urllib.parse import urlparse, urlunparse
 from streamlink.exceptions import PluginError, StreamError
 from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
+from streamlink.plugin.api.http_session import TLSSecLevel1Adapter
 from streamlink.stream.hls import HLSStream, HLSStreamReader, HLSStreamWorker, Sequence
 from streamlink.stream.hls_playlist import load as load_hls_playlist
+from streamlink.stream.http import HTTPStream
 
 log = logging.getLogger(__name__)
 
@@ -122,24 +124,32 @@ class FilmOnAPI:
     def __init__(self, session):
         self.session = session
 
-    channel_url = "http://www.filmon.com/api-v2/channel/{0}?protocol=hls"
-    vod_url = "http://www.filmon.com/vod/info/{0}"
+    channel_url = "https://www.filmon.com/ajax/getChannelInfo"
+    vod_url = "https://vms-admin.filmon.com/api/video/movie?id={0}"
 
     stream_schema = {
-        "quality": validate.text,
+        "quality": str,
         "url": validate.url(),
         "watch-timeout": int
     }
-    api_schema = validate.Schema(
+    channel_schema = validate.Schema(
         {
-            "data": {
+            "streams": validate.any(
+                {str: stream_schema},
+                [stream_schema]
+            )
+        }
+    )
+    vod_schema = validate.Schema(
+        {
+            "response": {
                 "streams": validate.any(
-                    {validate.text: stream_schema},
+                    {str: stream_schema},
                     [stream_schema]
                 )
             }
         },
-        validate.get("data")
+        validate.get("response")
     )
 
     def channel(self, channel):
@@ -150,11 +160,13 @@ class FilmOnAPI:
 
             # retry for 50X errors
             try:
-                res = self.session.http.get(self.channel_url.format(channel))
+                res = self.session.http.post(self.channel_url,
+                                             data={"channel_id": channel, "quality": "low"},
+                                             headers={"X-Requested-With": "XMLHttpRequest"})
                 if res:
                     # retry for invalid response data
                     try:
-                        return self.session.http.json(res, schema=self.api_schema)
+                        return self.session.http.json(res, schema=self.channel_schema)
                     except PluginError:
                         log.debug("invalid or non-JSON data received")
                         continue
@@ -165,7 +177,7 @@ class FilmOnAPI:
 
     def vod(self, vod_id):
         res = self.session.http.get(self.vod_url.format(vod_id))
-        return self.session.http.json(res, schema=self.api_schema)
+        return self.session.http.json(res, schema=self.vod_schema)
 
 
 @pluginmatcher(re.compile(r"""
@@ -183,7 +195,7 @@ class FilmOnAPI:
             (?P<is_group>group/)
         )(?:channel_id=)?(?P<channel>[-_\w]+)
         |
-        vod/view/(?P<vod_id>\d+)-
+        vod/view/(?P<vod_id>[^/?&]+)
     )
 """, re.VERBOSE))
 class Filmon(Plugin):
@@ -220,14 +232,28 @@ class Filmon(Plugin):
         vod_id = self.match.group("vod_id")
         is_group = self.match.group("is_group")
 
+        adapter = TLSSecLevel1Adapter()
+        self.session.http.mount("https://filmon.com", adapter)
+        self.session.http.mount("https://www.filmon.com", adapter)
+        self.session.http.mount("https://vms-admin.filmon.com/", adapter)
+
+        # get cookies
+        self.session.http.get(self.url)
+
         if vod_id:
             data = self.api.vod(vod_id)
             for _, stream in data["streams"].items():
-                streams = HLSStream.parse_variant_playlist(self.session, stream["url"])
-                if not streams:
-                    yield stream["quality"], HLSStream(self.session, stream["url"])
+                if stream["url"].endswith(".m3u8"):
+                    streams = HLSStream.parse_variant_playlist(self.session, stream["url"])
+                    if not streams:
+                        yield stream["quality"], HLSStream(self.session, stream["url"])
+                    else:
+                        yield from streams.items()
+                elif stream["url"].endswith(".mp4"):
+                    yield stream["quality"], HTTPStream(self.session, stream["url"])
                 else:
-                    yield from streams.items()
+                    log.error("Unsupported stream type")
+                    return
         else:
             if channel and not channel.isdigit():
                 _id = self.cache.get(channel)
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -80,7 +80,7 @@ egame                   egame.qq.com         Yes   No
 eltrecetv               eltrecetv.com.ar     Yes   Yes   Streams may be geo-restricted to Argentina.
 euronews                euronews.com         Yes   No
 facebook                facebook.com         Yes   Yes
-filmon                  filmon.com           Yes   Yes   Only SD quality streams.
+filmon                  filmon.com           Yes   Yes   Some VODs are mp4 which may not stream, use -o to download
 foxtr                   fox.com.tr           Yes   No
 funimationnow           - funimation.com     --    Yes   :ref:`Requires session cookies <cli:Authenticating with FunimationNow>`
                         - funimationnow.uk
--- a/tests/plugins/test_filmon.py
+++ b/tests/plugins/test_filmon.py
@@ -31,7 +31,10 @@ class TestPluginCanHandleUrlFilmon(PluginCanHandleUrl):
     ('https://www.filmon.com/tv/channel-4', [None, 'channel-4', None]),
     ('https://www.filmon.tv/tv/55', [None, '55', None]),
     ('http://www.filmon.tv/group/comedy', ['group/', 'comedy', None]),
-    ('http://www.filmon.tv/vod/view/10250-0-crime-boss', [None, None, '10250'])
+    ('http://www.filmon.tv/vod/view/10250-0-crime-boss', [None, None, '10250-0-crime-boss']),
+    ('http://www.filmon.tv/vod/view/10250-0-crime-boss/extra', [None, None, '10250-0-crime-boss']),
+    ('http://www.filmon.tv/vod/view/10250-0-crime-boss?extra', [None, None, '10250-0-crime-boss']),
+    ('http://www.filmon.tv/vod/view/10250-0-crime-boss&extra', [None, None, '10250-0-crime-boss'])
 ])
 def test_match_url(url, expected):
     Filmon.bind(Mock(), "tests.plugins.test_filmon")


src/streamlink/plugins/filmon.py
```diff
@@ -220,14 +231,28 @@ def _get_streams(self):
...
+                    streams = HLSStream.parse_variant_playlist(self.session, stream["url"])
+                    if not streams:
+                        yield stream["quality"], HLSStream(self.session, stream["url"])
+                    else:
+                        yield from streams.items()
+                elif stream["url"].endswith(".mp4"):
+                    if not mp4_warning_displayed:
+                        log.warning("Detected .mp4 video, this may not stream directly to your player. "
+                                    "In that case, try using -o/--output to save it to a file instead.")
+                        mp4_warning_displayed = True
```
bastimeyer, line 251: > Note: the warning for .mp4 video sources is due to the moov atom being located
> at the end of file. JW Player appears to get around this via range headers,
> sourcing data from the end of file first and then starting from the beginning
> to play the content as usual.

Are `HTTPStream`s actually needed? If the content is always available via HLS and the HTTPStream is just used by the site as a fallback, then it should be removed from the plugin. Otherwise the HTTPStream class should be subclassed and updated with the range header requests, if it's possible. Just adding a warning message isn't particularly great and should only be added as a last resort.

bastimeyer (commented):
Thanks for the PR. Apart from the HTTPStream concerns, the diff is looking good.

I can't verify the plugin though, because I'm getting SSL/TLS errors, even when setting `--http-no-ssl-verify`.
```
[cli][info] Found matching plugin filmon for URL https://www.filmon.com/tv/zdf
error: Unable to open URL: https://www.filmon.com/tv/zdf (HTTPSConnectionPool(host='www.filmon.com', port=443): Max retries exceeded with url: /tv/zdf (Caused by SSLError(SSLError(1, '[SSL: WRONG_SIGNATURE_TYPE] wrong signature type (_ssl.c:997)'))))
```

As you've already said in the gitter/matrix channel, the CI didn't start for some reason. You'll have to push to your branch again to re-trigger it.

mkbloke:
As I mentioned in the last paragraph of https://github.com/streamlink/streamlink/issues/4334#issuecomment-1030891141, the problem is regarding the minimum TLS version required in Open SSL's default configuration on newer distributions.
```
$ cat .openssl.cnf 
openssl_conf = default_conf

[ default_conf ]
ssl_conf = ssl_sect

[ssl_sect]
system_default = system_default_sect

[system_default_sect]
MinProtocol = TLSv1.2
CipherString = DEFAULT:@SECLEVEL=1

$ streamlink filmon.com/tv/e4 
[cli][info] Found matching plugin filmon for URL filmon.com/tv/e4
error: Unable to open URL: https://filmon.com/tv/e4 (HTTPSConnectionPool(host='filmon.com', port=443): Max retries exceeded with url: /tv/e4 (Caused by SSLError(SSLError(1, '[SSL: WRONG_SIGNATURE_TYPE] wrong signature type (_ssl.c:1131)'))))

$ OPENSSL_CONF=~/.openssl.cnf streamlink filmon.com/tv/e4 
[cli][info] Found matching plugin filmon for URL filmon.com/tv/e4
Available streams: low (worst), high (best)
```

src/streamlink/plugins/filmon.py
mkbloke: Unfortunately, it would seem not all content is available via HLS.  Take https://www.filmon.com/vod/view/277-judex for example, it only returns `.mp4` streams:
```
    "streams": {
      "low": {
        "is_free": true,
        "quality": "low",
        "url": "https://hls-dvr-gv.filmon.com/etv/Judex_1963_Channing_Pollock_Francine_Berge..mp4",
        "name": "SD",
        "watch-timeout": 86000
      },
      "high": {
        "is_free": true,
        "quality": "high",
        "url": "https://hls-dvr-gv.filmon.com/etv/Judex_1963_Channing_Pollock_Francine_Berge..mp4",
        "name": "HD",
        "watch-timeout": 86000
      }
    }
```
whereas others, such as https://www.filmon.com/vod/view/8762-john-lennon-the-ultimate-collection only return HLS streams:
```
    "streams": {
      "low": {
        "is_free": true,
        "quality": "low",
        "url": "https://hls-dvr-gv.filmon.com/uk01.filmon.com/vod/_definst_/content_filmon/source/141/7476/108477/108477/playlist.m3u8",
        "name": "SD",
        "watch-timeout": 86000
      },
      "high": {
        "is_free": true,
        "quality": "high",
        "url": "https://hls-dvr-gv.filmon.com/uk01.filmon.com/vod/_definst_/content_filmon/source/141/7476/108560/108560/playlist.m3u8",
        "name": "HD",
        "watch-timeout": 86000
      }
    }
```
I agree that the warning message is not great and I fully expected you to mention it!

mkbloke: After a bit of searching, I found the following project: https://github.com/danielgtaylor/qtfaststart

It appears to have a licence that would allow me to pull the relevant bits out of it and include them within a Streamlink plugin.  I'm very glad somebody else has already done this side of the work; I did not fancy trying to understand this via the specifications, even if I could get the official ones, but ISO/IEC 14496-12:2022 is 198 CHF (186.71 EUR)!!!

Here's the output of before and after using the tool on an mp4 file downloaded from FilmOn (I modified the output to show the position values as well):
```
$ qtfaststart/bin/qtfaststart -l test-orig.mp4 
ftyp at pos: 0 (32 bytes)
free at pos: 32 (8 bytes)
mdat at pos: 40 (276024271 bytes)
moov at pos: 276024311 (4438061 bytes)

$ qtfaststart/bin/qtfaststart -l test-mod.mp4 
ftyp at pos: 0 (32 bytes)
moov at pos: 32 (4438061 bytes)
mdat at pos: 4438093 (276024271 bytes)
```
The modified mp4 file does indeed seem to HTTP stream correctly with a test plugin I created.

I'll have a play with this and see if I can get it all working.

bastimeyer: Thanks for taking a look at this, but that's way too much of an overkill just for that single plugin. And writing this for a specific plugin is not a good idea and should ideally be done in a more generic way, eg. via a `HTTPStreamMPEG4` class which detects if the moov atom is at the end and adjusts the output automatically, so it can be used by other plugins as well, regardless whether the mp4 file is suitable for streaming via http or not and needs fixing.

Btw, there's a fork with some bugfixes here, since the repo you've linked is unmaintained since a long time ago:
https://github.com/danielyaa5/qtfaststart2

I'd say just leave it as it is. Maybe add a note to the docs instead of outputting a warning message, but I don't really mind having a warning message in the plugin.

mkbloke: OK.  Yes, a more generic way is probably the way to go for an implementation.

Thanks for the link to the maintained repo.

I don't have strong feelings either way about the warning message.  If you are happy to keep it and have no other issues with the PR, then you can merge it as is I guess.  Otherwise, I can put a note in the docs and remove the mp4 detection, so the code will just fall through to the "Unsupported stream type" message instead.

bastimeyer: As said, this can't be merged in the current state because no CI checks have run and passed as there was an outage at GH, and the master branch has certain protection rules when merging. You will have to push again.

bastimeyer:
A custom OpenSSL config with Python 3.10 on Arch Linux won't work. Apparently Python 3.10 hardcodes the list of allowed TLS ciphers with its default build arguments (which other distros / systems might be using as well):
https://wiki.archlinux.org/title/OpenSSL#Python_3.10_and_%22ca_md_too_weak%22_errors

A workaround is adding a different adapter for https requests for that specific plugin:
https://stackoverflow.com/a/59971965/15776887
https://stackoverflow.com/a/63349178/15776887
https://docs.python.org/3/library/ssl.html

"Transport Adapters" for specific hosts:
https://docs.python-requests.org/en/latest/user/advanced/#transport-adapters

```py
from requests import adapters
import ssl

class TLSAdapter(adapters.HTTPAdapter):
    def init_poolmanager(self, *args, **kwargs):
        ctx = ssl.create_default_context()
        ctx.set_ciphers("DEFAULT:@SECLEVEL=1")

        return super().init_poolmanager(
            *args,
            **kwargs,
            ssl_context=ctx
        )

adapter = TLSAdapter()
self.session.http.mount("https://filmon.com", adapter)
self.session.http.mount("https://www.filmon.com", adapter)
```

mkbloke:
That works nicely, thank you.  Note, I also needed to add `self.session.http.mount("https://vms-admin.filmon.com/", adapter)` to get VODs working too.

mkbloke:
Hmm, looks like it might need to be conditional for non-Windows systems?

src/streamlink/plugins/filmon.py
```diff
@@ -165,7 +180,19 @@ def channel(self, channel):
...
+class TLSAdapter(adapters.HTTPAdapter):
+    def init_poolmanager(self, *args, **kwargs):
+        ctx = ssl.create_default_context()
+        ctx.set_ciphers("DEFAULT:@SECLEVEL=1")
+
+        return super().init_poolmanager(
+            *args,
+            **kwargs,
+            ssl_context=ctx
+        )
```
bastimeyer, line 195: Let's call this `TLSSecLevel1Adapter` and let's move this to `plugin.api.http_session`, so that it can be imported by other plugins as well.

Is the keyword-arg-after-double-star-syntax supported on all python versions? I just mindlessly wrote this. Otherwise kwargs needs to be updated.

bastimeyer (commented):
> it might need to be conditional for non-Windows systems?

No. This is a Python 3.10 thing and the new default of the standard library's `ssl` module.

From your branch in my W10 VM:

```
$  streamlink -l debug 'https://filmon.com/tv/zdf'
[cli][debug] Python:     3.10.2
[cli][debug] Streamlink: 1.3.1+795.ga9e51c7
[cli][debug] Requests(2.27.1), Socks(1.7.1), Websocket(1.2.3)
[cli][debug] Arguments:
[cli][debug]  url=https://filmon.com/tv/zdf
[cli][debug]  --ffmpeg-ffmpeg=C:\Program Files (x86)\Streamlink\ffmpeg\ffmpeg.exe
[cli][info] Found matching plugin filmon for URL https://filmon.com/tv/zdf
error: Unable to open URL: https://filmon.com/tv/zdf (HTTPSConnectionPool(host='filmon.com', port=443): Max retries exceeded with url: /tv/zdf (Caused by SSLError(SSLError(1, '[SSL: WRONG_SIGNATURE_TYPE] wrong signature type (_ssl.c:997)'))))
```

src/streamlink/plugins/filmon.py
mkbloke: It seems like the test problem with SSL  is specific to the GH Windows 2019 environment and Python 3.6.8 only.

I don't know the answer to this and unless somebody can (be bothered to) figure it out, perhaps it's easier not to worry about implementing the custom TLSAdaptor until Streamlink drops Python 3.6?

The SSL issue with FilmOn has existed for quite some time now, so putting off this fix for a bit doesn't seem like too big a deal in my opinion.

I would propose we go back to https://github.com/streamlink/streamlink/commit/e6c352d3872f2077a5010c0c7cb93097444ec193 for the time being and merge that.  Additional changes can be made in the future.

bastimeyer: > perhaps it's easier not to worry about implementing the custom TLSAdaptor until Streamlink drops Python 3.6?
> putting off this fix for a bit doesn't seem like too big a deal in my opinion

That would mean that the plugin can't be used on py310 on any OS, as it hardcodes the ciphers, which is bad.

Just disable it on py36 for now, so that the tests pass. Or add the `HTTPSession.mount()` calls to `Plugin._get_streams()`, so that plugin initialization doesn't create the adapter instance where the error gets raised. Then the tests will pass, too. What then happens during runtime on systems using py36 where the SECLEVEL=1 cipher preset can't be used for some reason doesn't matter, as py36 is officially EOL and soon to be dropped from Streamlink.

mkbloke: OK.  Yes, I realised you will be dropping 3.6 soon, which is why I suggested waiting, but if you're happy with the runtime behaviour in Windows and 3.6 likely not working until that happens, then I'm OK with that.  It's does seem likely that almost all people using Streamlink under Windows will not be running it with 3.6 anyway, so all things considered it's probably not a much of an issue.

PR #4345, opened 2022-02-12
plugin.api.http_session: add class TLSSecLevel1Adapter()

mkbloke:
Creates an adapter with OpenSSL security level 1.  Example usage:

from streamlink.plugin.api.http_session import TLSSecLevel1Adapter

adapter = TLSSecLevel1Adapter()
self.session.http.mount("https://filmon.com", adapter)
self.session.http.mount("https://www.filmon.com", adapter)

---
~I have not added this to `__all__`, is that reasonable?~
There are no tests; I'm not really sure how to test it.

merge before #4335

### DIFF
--- a/src/streamlink/plugin/api/http_session.py
+++ b/src/streamlink/plugin/api/http_session.py
@@ -1,3 +1,4 @@
+import ssl
 import time
 from typing import Any, Callable, List, Pattern, Tuple
 
@@ -220,4 +221,12 @@ class HTTPSession(Session):
         return res
 
 
-__all__ = ["HTTPSession"]
+class TLSSecLevel1Adapter(requests.adapters.HTTPAdapter):
+    def init_poolmanager(self, *args, **kwargs):
+        ctx = ssl.create_default_context()
+        ctx.set_ciphers("DEFAULT:@SECLEVEL=1")
+        kwargs["ssl_context"] = ctx
+        return super().init_poolmanager(*args, **kwargs)
+
+
+__all__ = ["HTTPSession", "TLSSecLevel1Adapter"]


src/streamlink/plugin/api/http_session.py
```diff
@@ -220,4 +221,15 @@ def request(self, method, url, *args, **kwargs):
...
 
+class TLSSecLevel1Adapter(requests.adapters.HTTPAdapter):
+    def init_poolmanager(self, *args, **kwargs):
+        ctx = ssl.create_default_context()
+        ctx.set_ciphers("DEFAULT:@SECLEVEL=1")
+        return super().init_poolmanager(
+            *args,
+            **kwargs,
+            ssl_context=ctx
+        )
```
bastimeyer, line 232:
```suggestion
        kwargs["ssl_context"] = ctx
        return super().init_poolmanager(*args, **kwargs)
```

bastimeyer (changes requested):
> For websites using TLS version < 1.3

The commit message is wrong. This has nothing to do with TLS 1.3. This simply allows a negotiation of old/insecure ciphers which are disabled by default in Python 3.10, as it now hardcodes more secure ciphers (seclevel=2) instead of relying on the system configuration via openssl, etc. This is a compile flag and it can be changed to `--with-ssl-default-suites=openssl` to make it work with the system's openssl config for example.

In case of FilmOn, where this adapter with the ciphers override is needed, the site is running via TLS 1.2. The protocol version doesn't matter here though and the latest one supported by client and server is automatically chosen in the protocol negotiation. Python 3.10 still allows TLS 1.2.

----

https://docs.python.org/3/library/ssl.html#ssl.SSLContext

> Changed in version 3.10: The default cipher suites now include only secure AES and ChaCha20 ciphers with forward secrecy and security level 2. RSA and DH keys with less than 2048 bits and ECC keys with less than 224 bits are prohibited. [PROTOCOL_TLS](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS), [PROTOCOL_TLS_CLIENT](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_CLIENT), and [PROTOCOL_TLS_SERVER](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_SERVER) use TLS 1.2 as minimum TLS version.

https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_security_level.html

> Level 0
> Everything is permitted. This retains compatibility with previous versions of OpenSSL.
> 
> Level 1
> The security level corresponds to a minimum of 80 bits of security. Any parameters offering below 80 bits of security are excluded. As a result RSA, DSA and DH keys shorter than 1024 bits and ECC keys shorter than 160 bits are prohibited. All export cipher suites are prohibited since they all offer less than 80 bits of security. SSL version 2 is prohibited. Any cipher suite using MD5 for the MAC is also prohibited.
> 
> Level 2
> Security level set to 112 bits of security. As a result RSA, DSA and DH keys shorter than 2048 bits and ECC keys shorter than 224 bits are prohibited. In addition to the level 1 exclusions any cipher suite using RC4 is also prohibited. SSL version 3 is also not allowed. Compression is disabled.

----

> I have not added this to `__all__`, is that reasonable?

It should be added there, as it's part of the module's public interface.

PR #4346, opened 2022-02-13
typo fix "steams" to "streams" in opencv-face.py

dylan-roussin:
Fixed a typo


gravyboat:
Thanks.

bastimeyer:
This example file, which was written four years ago and is the last one that hasn't been removed yet, is pretty much irrelevant to the project and should be removed as well. It doesn't even use any Streamlink functionality apart from URL resolving, so it technically never got outdated and was thus never removed. Fixing this random typo while leaving other typos is just a pointless change.

PR #4349, opened 2022-02-14
plugins.openrectv: get URLs from subs_trial_media

### LINKED ISSUE #4348: Openrec.tv: cannot play livestream when it's in `member_trial`
### Checklist

### Streamlink version

Latest stable release

### Description

Cannot access to a subscription livestream with free-trial, with or without logging in. I can do so with my browser. 

### Debug log

```text
G:\>streamlink https://www.openrec.tv/live/n9ze3ylno84 best --loglevel=debug
[cli][debug] OS:         Windows 7
[cli][debug] Python:     3.8.7
[cli][debug] Streamlink: 3.1.1
[cli][debug] Arguments:
[cli][debug]  url=https://www.openrec.tv/live/n9ze3ylno84
[cli][debug]  stream=['best']
[cli][debug]  --player=D:\Program Files\SMPlayer\mpv\mpv.exe
[cli][debug]  --ffmpeg-ffmpeg=D:\Program Files\!cli\ffmpeg.exe
[cli][info] Found matching plugin openrectv for URL https://www.openrec.tv/live/n9ze3ylno84
[plugins.openrectv][debug] Got valid detail response
[plugins.openrectv][debug] Found video: きどまだSeason2　＃20 (n9ze3ylno84)
error: No playable streams found on this URL: https://www.openrec.tv/live/n9ze3ylno84
```

```text

G:\>streamlink https://www.openrec.tv/live/n9ze3ylno84 best --loglevel=debug --openrectv-email=**** --o
penrectv-password=****
[cli][debug] OS:         Windows 7
[cli][debug] Python:     3.8.7
[cli][debug] Streamlink: 3.1.1
[cli][debug] Arguments:
[cli][debug]  url=https://www.openrec.tv/live/n9ze3ylno84
[cli][debug]  stream=['best']
[cli][debug]  --player=D:\Program Files\SMPlayer\mpv\mpv.exe
[cli][debug]  --ffmpeg-ffmpeg=D:\Program Files\!cli\ffmpeg.exe
[cli][debug]  --openrectv-email=****
[cli][debug]  --openrectv-password=********
[cli][info] Found matching plugin openrectv for URL https://www.openrec.tv/live/n9ze3ylno84
[plugins.openrectv][debug] Logged in as ****
[plugins.openrectv][debug] Got valid detail response
[plugins.openrectv][debug] Found video: きどまだSeason2　＃20 (n9ze3ylno84)
error: No playable streams found on this URL: https://www.openrec.tv/live/n9ze3ylno84
```

fireattack:
Fix #4348

### DIFF
--- a/src/streamlink/plugins/openrectv.py
+++ b/src/streamlink/plugins/openrectv.py
@@ -33,7 +33,7 @@ class OPENRECtv(Plugin):
         validate.optional("subs_trial_media"): {
             "url": validate.any(None, validate.url()),
             "url_ull": validate.any(None, validate.url()),
-        }        
+        }
     })
 
     _subscription_schema = validate.Schema({
--- a/src/streamlink/plugins/openrectv.py
+++ b/src/streamlink/plugins/openrectv.py
@@ -29,7 +29,11 @@ class OPENRECtv(Plugin):
             "url": validate.any(None, validate.url()),
             "url_public": validate.any(None, validate.url()),
             "url_ull": validate.any(None, validate.url()),
-        }
+        },
+        validate.optional("subs_trial_media"): {
+            "url": validate.any(None, validate.url()),
+            "url_ull": validate.any(None, validate.url()),
+        }        
     })
 
     _subscription_schema = validate.Schema({
@@ -133,7 +137,7 @@ class OPENRECtv(Plugin):
                 m3u8_file = subs_data["data"]["items"][0]["media"]["url"]
             # streaming
             elif mdata["onair_status"] == 1:
-                m3u8_file = mdata["media"]["url_ull"]
+                m3u8_file = mdata["media"]["url_ull"] or mdata["subs_trial_media"]["url_ull"]
             # archive
             elif mdata["onair_status"] == 2 and mdata["media"]["url_public"] is not None:
                 m3u8_file = mdata["media"]["url_public"].replace("public.m3u8", "playlist.m3u8")


src/streamlink/plugins/openrectv.py
```diff
@@ -133,7 +137,7 @@ def _get_streams(self):
                 m3u8_file = subs_data["data"]["items"][0]["media"]["url"]
             # streaming
             elif mdata["onair_status"] == 1:
-                m3u8_file = mdata["media"]["url_ull"]
+                m3u8_file = mdata["media"]["url_ull"] or mdata["subs_trial_media"]["url_ull"]
```
bastimeyer, line 140: This will raise a `KeyError` if the `media` or `subs_trial_media` items do not exist on the `mdata` dict when trying to access their `url_ull` item. Both `media` and `subs_trial_media` are set to optional in the validation schema, so it's not guaranteed that `mdata["media"]` and `mdata["subs_trial_media"]` exist. `media` is the main problem here if you're using `subs_trial_media` as a fallback, unless `media` is always included in the API response. That makes flagging it as optional questionable though.

The whole plugin is in a rather bad shape and could need a proper cleanup/rewrite. For example, it queries the site's API every time for each metadata lookup (stream title and author) in addition to the regular stream URL lookup. The validation schemas are outdated and their return values are being read everywhere in an insecure way, as mentioned above. In addition to that, it uses outdated HTTPSession methods and the overall code style of how error messages are logged is not ideal either. This could all be done in a much simpler/better way.

That's all unrelated to this PR though, so please ignore what I just said.

What needs to be fixed however is how the `url_ull` items are read from the `mdata["media"]` and `mdata["subs_trial_media"]` dictionaries from the validation schema return value.

fireattack: @bastimeyer : both `media` and `subs_trial_media` are indeed always includes in the API call AFAIK.
If no media available (as in this case), it would just be `null` for all the properties (URLs).

Anything needs to be done in this case? Simply mark it as non-optional?

bastimeyer: > If no media available (as in this case), it would just be `null` for all the properties (URLs).

It should be fine keeping it as optional then. This can be fixed in a proper plugin cleanup at some point later if someone decides to do it.

bastimeyer: You'll have to fix the trailing whitespace in the validation schema though.

fireattack: Thanks, that's was a rookie mistake.

bastimeyer:
Thanks

PR #4356, opened 2022-02-19
Chocolatey package publishing from Github Actions

scowalt:
### Context

The `streamlink-chocolatey` package has gotten very out-of-date, since it required a manual update for each version. The most sustainable way to maintain a chocolatey package is to incorporate it into the deploy process of the app itself. This change aims to do just that.

### Implementation

I've made a few changes in `main.yml`

1. Created a `Generate version number` job to share some code between jobs
2. Modified the `New release` job to output the `sha256` of the windows executable, which chocolatey needs
3. Created the `Chocolatey release` job, which prepares the package, creates the package, and uploads it
4. Create the `version` job to help deduplicate lots of code that checks for the version.

I've also added the files that support the chocolatey build to `releases/chocolatey`. I can put these files _anywhere_ in the repo if you have a preference. 

### Validation

I've run a subset of the new job in my fork of the repo to iron out most potential issues. You can see the job run [here](https://github.com/scowalt/streamlink/actions/runs/1869277074). The job intentially failed because I didn't give it a proper API key to publish the package.

Unfortunately, I can't test the full end-to-end without modifying the chocolatey repository. If there are issues that stem from this change, I'm happy to ship a follow-up fix. I'm confident that this won't regress any of the existing Github Actions functionality.

### Follow-up

This change depends on the `CHOCOLATEY_API_KEY` environment variable. To make things simple, I'm happy to provide you with my API key to add to the Github secrets. (The most "correct" thing to do would be to create a special helper accout _just_ for shipping this package. However, I don't want to add another layer of complexity to this change)

Once this change has landed and is stable, I recommend deleting the `streamlink-chocolatey` repository.

If this change is well-received, I can almost certainly code up the same thing for `streamlink-twitch-gui`

cc @Conjuro @bastimeyer

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -92,9 +92,22 @@ jobs:
       - name: Build
         run: make --directory=docs html man
 
+  version:
+    name: Generate version number
+    runs-on: ubuntu-20.04
+    outputs:
+      version: ${{ steps.getVersion.outputs.version }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Generate verison number
+        id: getVersion
+        run: echo ::set-output name=version::$(python setup.py --version | sed 's/+/_/')
+
   windows-installer:
     name: Windows installer
     runs-on: ubuntu-20.04
+    needs:
+      - version
     steps:
       - uses: actions/checkout@v2
         with:
@@ -112,7 +125,7 @@ jobs:
           sudo apt install -y nsis imagemagick inkscape
       - name: Installer file name
         id: installer
-        run: echo ::set-output name=filename::streamlink-$(python setup.py --version | sed 's/+/_/')
+        run: echo ::set-output name=filename::streamlink-${{ needs.version.outputs.version }}
       - name: Build
         run: ./script/makeinstaller.sh "${{ steps.installer.outputs.filename }}"
       - name: Upload artifact
@@ -154,8 +167,11 @@ jobs:
     name: New release
     if: github.repository == 'streamlink/streamlink' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
     needs:
+      - version
       - deploy-documentation
     runs-on: ubuntu-20.04
+    outputs:
+      exeSha: ${{ steps.generateSha.outputs.exeSha }}
     steps:
       - uses: actions/checkout@v2
         with:
@@ -176,7 +192,7 @@ jobs:
         run: make --directory=docs man
       - name: Installer file name
         id: installer
-        run: echo ::set-output name=filename::streamlink-$(python setup.py --version | sed 's/+/_/')
+        run: echo ::set-output name=filename::streamlink-${{ needs.version.outputs.version }}
       - name: Download installer artifact
         uses: actions/download-artifact@v2-preview
         with:
@@ -186,6 +202,10 @@ jobs:
         env:
           RELEASE_KEY_PASSPHRASE: ${{ secrets.RELEASE_KEY_PASSPHRASE }}
         run: ./script/build-and-sign.sh
+      - name: Generate sha256 for chocolatey release
+        id: generateSha
+        run: |
+          echo "::set-output name=exeSha::$(sha256sum "${STREAMLINK_DIST_DIR}/${{ steps.installer.outputs.filename }}.exe" | awk '{ print $1}')"
       - name: Github release
         env:
           RELEASES_API_KEY: ${{ secrets.RELEASES_API_KEY }}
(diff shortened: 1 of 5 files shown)


bastimeyer:
> The `streamlink-chocolatey` package has gotten very out-of-date, since it required a manual update for each version. The most sustainable way to maintain a chocolatey package is to incorporate it into the deploy process of the app itself.

The package has become out of date because a new version hasn't been published in a long while, not because of lack of automation. Streamlink's Windows installer bundles Python 3.9 which isn't supported on EOL Windows versions, and the Chocolatey validation servers are running on those outdated versions of Windows for some reason. As mentioned in the chocolatey package repo thread, the chocolatey admins can work around that by disabling package verification. The choco package could also bundle Python 3.8 instead of relying on Streamlink's Windows installer, but that would mean building up-to-date 3.8 versions from source or bundling out-of-date 3.8 releases, which the Windows installer avoids explicitly.

----

Packaging is a manual review process and it depends on the target systems, so it shouldn't be automated here.

If we would package for every platform here, then it would become unmanagable very quickly, and any kind of package fixes, simple or not, would require patching here and publishing a new release of the application itself. Or a package fix would need to wait for the next regular release, which isn't ideal either.

In the past we had the Debian packaging files in this repo, but it was removed for that very reason.

I'm also not a fan of having the Windows installer build config here, especially now since we're finally having some kind of dependency lockfile for it. I'd much more prefer having a separate repo for that. Then building Python from source and having multiple installers with different Python versions could even be explored.

release/chocolatey/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,35 @@
+$ErrorActionPreference = 'Stop';
```
gep13, line 1: Have you tested _not_ having this file?  Since this is a standard installation of an exe, it is likely that Chocolatey' auto uninstaller can take care of removing it, and therefore this file may not be required at all, and can simply be removed.

release/chocolatey/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,27 @@
+
+$ErrorActionPreference = 'Stop';
+
+$packageName= 'Streamlink'
+$toolsDir   = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$url        = "https://github.com/streamlink/streamlink/releases/download/$env:chocolateyPackageVersion/streamlink-$env:chocolateyPackageVersion.exe"
```
gep13, line 6: Have you considered converting this into what we refer to as an embedded package?  That is where you would put the exe directly within the Chocolatey nupkg file, and install from there, rather than downloading the exe during the installation.  This would mean that the package is completely contained, and can be installed anywhere without the need to reach back out to the internet to grab the installer.

.github/workflows/main.yml
```diff
@@ -195,3 +215,28 @@ jobs:
...
+      - name: Package release
+        working-directory: release/chocolatey
+        run: |
+          choco pack
+        shell: pwsh
+      - name: Push release
+        working-directory: release/chocolatey
+        run: |
+          $version = "${{ needs.version.outputs.version }}"
+          choco push --api-key ${{ secrets.CHOCOLATEY_API_KEY }} streamlink.$version.nupkg
```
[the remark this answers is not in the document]
gep13: Also, I don't know if it makes a difference, but are you missing `shell: pwsh` from this step?

gep13 (changes requested):
Just a couple small suggestions here.  Happy to discuss further if you have any questions.

gep13:
> @bastimeyer said...
and the Chocolatey validation servers are running on those outdated versions of Windows for some reason

We want Chocolatey packages to work on as many machines as possible, having any dependences defined on the package as is required to make the underlying application work correctly. We have found that testing on the current package-verifier base image helps in this regard.  However, we are currently in the process of updating the base image to a newer version.

> @bastimeyer said...
As mentioned in the chocolatey package repo thread, the chocolatey admins can work around that by disabling package verification.

Was a request made to one of the moderators to provide an exemption from the package-verifier?  This is definitely something that can be done, but we need to know that this is required.  It can be requested from the package moderation page for the package version in question.

scowalt:
> Packaging is a manual review process and it depends on the target systems, so it shouldn't be automated here.

Totally understand that perspective. I'm concerned about the sustainability of having to manually deploy to chocolatey with every release.

What do you think about having this repository send an event to the `streamlink-chocolatey` repository when there's a new release? This should satisfy all of:
1. Having the publishing process be automated
2. Not having any packaging stuff in the `streamlink` repository
3. Allow package modifications and re-releases to happen out-of-band from the `streamlink` release process

gep13:
> @scowalt said...
One way we could do this is set up a job with CircleCI that [triggers when a CURL request is sent](https://circleci.com/docs/2.0/triggers/#trigger-a-job-using-curl-and-your-api-token)

Unless you are already using CircleCI for other things, then there is no _need_ to introduce this. You can trigger a workflow on a GitHub Repository from another repository using the repository_dispatch event.  There is even a GitHub Action to help with this:

https://github.com/peter-evans/repository-dispatch

So you could take the work that you have done in this PR and move it to a workflow in the streamlink-chocolatey repository, and then update the workflow in this repository to send a trigger when a release is completed.

scowalt:
Understood. In that case, closing this out.

PR #4363, opened 2022-02-21
tests.plugins: implement should_match_groups

bastimeyer:
This should simplify testing plugin URL capture groups and make it more consistent between tests.

Please see the commit messages of both commits.


gravyboat (approved):
Looks really good @bastimeyer, thanks!

PR #4364, opened 2022-02-24
plugins.lnk: add new plugin

justinas:
[LNK](https://en.wikipedia.org/wiki/LNK_(television_station)) is a Lithuanian broadcasting group with several TV channels under it. Those are freely streamed at https://lnk.lt/tiesiogiai .

### DIFF
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -98,6 +98,7 @@ invintus                player.invintus.com  Yes   Yes
 kugou                   fanxing.kugou.com    Yes   --
 linelive                live.line.me         Yes   Yes
 livestream              livestream.com       Yes   --
+lnk                     lnk.lt               Yes   No
 lrt                     lrt.lt               Yes   No
 ltv_lsm_lv              ltv.lsm.lv           Yes   No    Streams may be geo-restricted to Latvia.
 mediaklikk              - mediaklikk.hu      Yes   No    Streams may be geo-restricted to Hungary.
--- a/docs/plugin_matrix.rst
+++ b/docs/plugin_matrix.rst
@@ -98,7 +98,7 @@ invintus                player.invintus.com  Yes   Yes
 kugou                   fanxing.kugou.com    Yes   --
 linelive                live.line.me         Yes   Yes
 livestream              livestream.com       Yes   --
-lnk                     lnk.lt               Yes   No
+lnk                     lnk.lt               Yes   No    Streams may be geo-restricted to Lithuania.
 lrt                     lrt.lt               Yes   No
 ltv_lsm_lv              ltv.lsm.lv           Yes   No    Streams may be geo-restricted to Latvia.
 mediaklikk              - mediaklikk.hu      Yes   No    Streams may be geo-restricted to Hungary.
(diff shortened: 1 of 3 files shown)


bastimeyer (commented):
Thanks for the PR. Adding a plugin for this site should be fine since this includes a live news channel.

Are the channel IDs static? According to the `https://lnk.lt/api/main/content-page-by-program/tiesiogiai` API endpoint, the IDs are `episodeId`s with certain `endDate`s, so this suggests that they are dynamic.

The plugin code also needs some rework. I have fixed and improved it locally, so if you don't mind, I can add a commit to your PR branch.

mkbloke:
It looks like some of the content is geo-restricted.  Ideally, you should note that in the plugin matrix.  You could probably test for it, provide an info level message then just return.

`data["videoInfo"]["contentRestrict"]` appears to be the bool you should look for.  When true, `data["videoInfo"]["videoUrl"]` is an empty string.

You could also look at using schema validation for the returned JSON.

Also, as it's available, you could consider adding `self.id = data["videoInfo"]["id"]` and `self.title = data["videoInfo"]["title"]` to the plugin metadata.

bastimeyer:
Yes, I have done all of that already, but before pushing onto someone else's branch, I usually ask first.

justinas:
Thank you both for the quick review.

@bastimeyer 

> Are the channel IDs static? According to the `https://lnk.lt/api/main/content-page-by-program/tiesiogiai` API endpoint, the IDs are `episodeId`s with certain `endDate`s, so this suggests that they are dynamic.

The `endDate` seems to refer for the currently broadcasted programme, however the ID itself seems to remain static for the channel for a longer time. E.g. `https://lnk.lt/api/video/video-config/137535` currently includes this:

```json
    "airDate": "2020-12-28T12:58:16.674",
    "airDateText": "2020-12-28 12:58",
```

which makes me believe that the IDs do not change very often :slightly_smiling_face: It might be more correct to request `https://lnk.lt/api/main/content-page-by-program/tiesiogiai` first, but I'm not sure if it's worth the hassle and complexity if the channel IDs are stable.

> The plugin code also needs some rework. I have fixed and improved it locally, so if you don't mind, I can add a commit to your PR branch.

Absolutely, please go ahead! I am very new to the codebase, so I can't exactly say I know better.

@mkbloke Thank you for your notes, will try to look into the issues you mentioned as well.

mkbloke:
@bastimeyer, OK.  I was only making suggestions for the OP.  That wasn't intended for you, it just happened to come after your reply.

bastimeyer:
Let's wait and check if the IDs change. If they change, then this static channel map obviously can't stay.

justinas:
Hi again, reporting a couple days later, the IDs have not changed.

```console
$ date -u
Sun Feb 27 08:04:35 PM UTC 2022
$ curl -sS https://lnk.lt/api/main/content-page-by-program/tiesiogiai | jq -r '.components[1].component.channels[] | [.channel, .episodeId] | @csv'
"LNK",137535
"BTV",137534
"2TV",95343
"Info TV",137748
"TV1",106791
```

PR #4374, opened 2022-03-03
docs: replace plugin matrix

### LINKED ISSUE #4369: docs: replace plugin matrix with metadata comment block in each plugin file
## Motivation

The plugin matrix is and always has been a total mess.

- https://github.com/streamlink/streamlink/blame/3.1.1/docs/plugin_matrix.rst
- https://streamlink.github.io/plugin_matrix.html

The table that's defined via restructured-text is ugly, restrictive, limited in space and requires footnote workarounds for long URLs. Adding new entries is always cumbersome and prone to errors, and with the usage of the Furo theme for the rendered HTML docs, the navigation menu on the right is also completely empty, as it relies on sections and subsections in the document content. A table is not the right fit for this.

Lots of valuable info is also missing, like proper plugin descriptions/introductions, or other things like the type of returned streams and what not.

## Solution

A solution which I've had on the radar for a long time now would be having a mandatory block comment at the top of each plugin file which contains annotations for all this plugin metadata. The static table could then be replaced by a custom sphinx extension which parses the AST of each plugin file and yields ReST sections for each plugin with a short datalist generated from the available metadata in the header block comment, similar to the datalist here, except without line breaks:
https://streamlink.github.io/applications.html

This would enable a navigation via the sidebar on the right (due to the sections) and listing multiple/long URLs would be much easier without having to rely on workarounds (due to the datalists). And the most important thing is that the space limitation of the table would be gone (due to datalists and an arbitrary number of entries - no columns anymore).

A dynamically generated plugins page from parsed plugin ASTs could also link to each plugin's CLI arguments without having to specify them in the metadata.

## Plugin metadata format

How would this header block comment look like? The annotation format should not conflict with other already existing formats, so this is what I would suggest:

```py
"""
$key value
$key value
$key value
"""

import ...
```

The parser could then simply split the comment via this regex:
```regex
\n\$([a-z]+) ([^\n]+)
```

## Metadata fields

The plugin matrix currently has the "name", "urls", "live", "vod", and "notes" columns. Since we're not bound to a certain number of columns with this, we can organize this a bit differently.

- `description` (required, needs to be defined first)
  A short plugin description, eg. some kind of platform name or a name for a TV/news network, etc.
- `url` (required, can be set multiple times)
  One of the supported URLs. Similar to the current style of the plugin matrix, so no scheme or paths (if not required otherwise)
- `type` (required)
  Must be either `live`, `vod` or `live, vod`
- `region` (optional)
  A list of countries/regions (separated by comma), or "various"
- `account` (optional)
  A short description when an account or subscription is needed
- `notes` (optional, needs to be defined last)
  A short text with additional notes/hints

Alternatives:

- `info` instead of `description`, as it's shorter
- Have separate `live` and `vod` fields with value "Yes" or "No"
- `authentication` instead of `account`, as "account" doesn't always fit the description

----

I've already moved all the metadata from the table of the plugin matrix to each plugin file in my local repo, but before I continue with this and start writing the parser, I need to open this thread and ask for a format which we can agree on.

Another issue is that almost no plugin currently has the "description" text which I defined as mandatory above, so I'd suggest that this should be added later. I surely don't want to look up all this now.

The tests also need to get updated, as well as the "developing" page in the docs.

bastimeyer:
Resolves #4369 

This replaces the plugin matrix with a dynamically generated list of plugins via a custom sphinx extension that parses a mandatory header comment in each plugin. See #4369 for the motivation of this.

----

### First commit

This adds the metadata header comment to each plugin. The data was copied from the old plugin matrix and in some cases adjusted a bit. Some plugins received additional/missing data, like zattoo for example, where URLs were copied from the plugin matcher regex. In case of zatto, this was very obvious, because of the large number of URLs and the previous notes. It's very likely that there are similar plugins that did not get updated. This can be fixed in the future.

The metadata currently also does not include any `$description` tags. These need to be added later and then be made mandatory once every plugin has got one. Some of the plugin descriptions can be found in each plugin's git history, because lots of old header comments got removed in the past. Checking GitHub issues and pull requests can also help.

The metadata format is very strict and gets linted via the tests added in the third commit.

### Second commit

This renames the doc's `plugin_matrix` page to `plugins` and will thus make old links go 404 once merged. Since the plugin data is not represented as a table anymore, it doesn't really make sense calling the menu "plugin_matrix". The link in the main menu has always been called "Plugins" because it refers to the page's top header text.

### Third commit

This finally removes the plugin matrix and replaces it with the new sphinx extension, and it also adds tests which lint the plugin files and their strict metadata format.

The sphinx extension does not lint/validate the metadata and relies on the tests for validity. Since we're building docs and are running tests in parallel, this could cause confusion, but I don't think it's a problem. The tests are all atomic and parametrized, so any errors when something gets changed or added should be trivial to resolve, and it doesn't matter if the docs fail or produce broken output.

Plugin arguments which get read from each plugin's AST are also not 100% safe, as it's just a simple lexical analysis. Extending it with more validations where the plugin arguments are defined (and that the imports are actually correct) would be possible, but it would bloat up the extension's logic by quite a bit and it's not worth it. This issue shouldn't matter too much, as plugins get reviewed before merging anyway.

### Fourth commit

Updates the "developing" docs page.


bastimeyer:
Not sure why the preview doesn't include any plugin arguments.
This could be caused by netlify running on an older python version. I'll check this tomorrow.

bastimeyer:
Instead of bumping the python version on netlify, I made the AST parsing in the sphinx extension python 3.6 and 3.7 compatible. This now correctly builds with plugin arguments.

I've also updated the developing page in the docs.

gravyboat:
Finally had some time to review the various commits this evening. Looks good and the format makes sense to me!

PR #4379, opened 2022-03-04
docs: fix plugin metadata example

bastimeyer:
I accidentally used the wrong format in the doc's plugin metadata example.

----

With the plugin metadata changes being merged, should we publish a new release tomorrow (3.2.0)?


gravyboat:
> With the plugin metadata changes being merged, should we publish a new release tomorrow (3.2.0)?

I think so yeah.

PR #4384, opened 2022-03-07
FIx typo

KG7x:


src/streamlink/plugins/pluzz.py
```diff
@@ -45,10 +45,10 @@ def _get_streams(self):
         # Retrieve geolocation data
         country_code = self.session.http.get(self.GEO_URL, schema=validate.Schema(
             validate.parse_json(),
-            {"reponse": {"geo_info": {
+            {"response": {"geo_info": {
                 "country_code": str
             }}},
-            validate.get(("reponse", "geo_info", "country_code"))
+            validate.get(("response", "geo_info", "country_code"))
```
bastimeyer, line 51: This is a data response from the server. Your automated tool for checking spelling errors would break the plugin.

gravyboat:
I don't think this is worth merging in. Changelog commit messages should not be changed even if there are errors. The others are all pretty minor. @bastimeyer thoughts?

bastimeyer (changes requested):
Old changelogs and vendor files like versioneer should not be spellchecked.
The rest, apart from the `QUALITY_WEIGTHS_EXTRA` dict that has been there since the livestreamer fork and been kept in order to not break anything, are just irrelevant comments.

Spellchecking docs is fine, comments too (I guess), but this PR (currently) includes a lot of unnecessary stuff.

src/streamlink/plugins/pluzz.py
KG7x: You can close, I just posted where you can correct errors

Issue #4389, opened 2022-03-08
Re-add Latina.pe


malvinas2:
### Checklist

- [X] This is a plugin request and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+request%22)

### Description

Latina (formerly Frecuencia Latina) is a southamerican television network located in Peru.
The site switched back to footprint.net, opening the stream by just using the url isn't possible any longer. 
(probably geo-blocked to Peru). 

### Input URLs

latina.pe/tvenvivo

https://live-latina-mdstrm.secure.footprint.net/live-stream-dai/319fed2157fa4c48ae2f81928b21f8a4/manifest/f624fec49dadc88894ebeff1d974eab019af7757/5ce7109c7398b977dc0744cd/9ecdb01e-f8d0-42c0-8835-c8513c68574b/3.m3u8?aid=5ce70c587398b977dc0743b5&uid=2qdmtnyxqykgaq1bhsp6m1xljughmfmp&sid=aumr1eucaiuelbxyocsdah8axbysvga4&pid=rh5pcbnrqphhxs0t47orcnhmj6zwog8f&ref=https%3a%2f%2fwww.latina.pe%2f&without_cookies=false&dnt=true&es=live-latina-mdstrm.secure.footprint.net&proto=https&pz=us&access_token=nzvx5r0h5nfbfpmwdkld4p5cdulfgsf819yqu6donwkh68no9awt63irlslj0e2hwjobzuod2fr&manifestfilter=video_height:240-720&hash=0477df08cf37282c87750

bastimeyer:
The plugin was just removed in December because it broke when the site switched to a different streaming provider. See #4269. And now they've switched to a different provider again. Adding plugins for sites which are unstable doesn't make much sense.

malvinas2:
In general, I fully understand your concerns. 

But if I review the development history in the plugin mentioned, a first version was created in early 2020. 
A few minor adjustments followed, but they were more of a cosmetic nature.
All of this I would consider as one version.
The plugin was discontinued at the end of 2021 because the stream could now be extracted from the website without any tools. The effort to remove a plugin should be minimal.
So the current change would be the second in two years.

PR #4395, opened 2022-03-13
plugins.mdstrm: new plugin for mdstrm.com livestreams

back-to:
https://www.mediastre.am/en/about-us

closes https://github.com/streamlink/streamlink/issues/4389
ref https://github.com/streamlink/streamlink/pull/4269

```
$ streamlink https://mdstrm.com/live-stream/5a7b1e63a8da282c34d65445
[cli][info] Found matching plugin mdstrm for URL https://mdstrm.com/live-stream/5a7b1e63a8da282c34d65445
[plugins.mdstrm][debug] iframe=https://mdstrm.com/live-stream/5a7b1e63a8da282c34d65445
[cli][info] Available streams: 240p (worst), 360p, 480p, 720p (best)
```

```
$ streamlink https://www.latina.pe/tvenvivo
[cli][info] Found matching plugin mdstrm for URL https://www.latina.pe/tvenvivo
[plugins.mdstrm][debug] iframe=https://mdstrm.com/live-stream/5ce7109c7398b977dc0744cd?autoplay=false&player=
[plugins.mdstrm][debug] programmatic_url=https://mdstrm.com/programmatic/5ce7109c7398b977dc0744cd?autoplay=false
[cli][info] Available streams: 240p (worst), 360p, 480p, 720p (best)
```

### DIFF
--- a/src/streamlink/plugins/mdstrm.py
+++ b/src/streamlink/plugins/mdstrm.py
@@ -1,6 +1,8 @@
 """
 $description CDN hosting live content for various websites in the Americas.
 $url mdstrm.com
+$url latina.pe/tvenvivo
+$url saltillo.multimedios.com/video
 $type live
 """
 import logging
--- a/src/streamlink/plugins/mdstrm.py
+++ b/src/streamlink/plugins/mdstrm.py
@@ -1,0 +1,140 @@
+"""
+$description CDN hosting live content for various websites in the Americas.
+$url mdstrm.com
+$type live
+"""
+import logging
+import re
+from urllib.parse import urlparse
+
+from streamlink.plugin import Plugin, pluginmatcher
+from streamlink.plugin.api import validate
+from streamlink.stream.hls import HLSStream
+from streamlink.utils.url import update_scheme
+
+log = logging.getLogger(__name__)
+
+
+@pluginmatcher(re.compile(
+    r"https://(?:www\.)?latina\.pe/tvenvivo"
+))
+@pluginmatcher(re.compile(
+    r"https://saltillo\.multimedios\.com/video"
+))
+@pluginmatcher(re.compile(
+    r"https://mdstrm\.com/live-stream/\w+"
+))
+class MDStrm(Plugin):
+    _re_ad = re.compile(r"parent\._dai_session\s*=\s*'([^']+)';")
+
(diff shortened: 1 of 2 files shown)


malvinas2:
It works here with no problems. Thank you @back-to !

src/streamlink/plugins/mdstrm.py
```diff
@@ -0,0 +1,140 @@
...
+        url_iframe = update_scheme("https://", url_iframe, force=False)
+        log.debug(f"iframe={url_iframe}")
+        root = self.session.http.get(url_iframe,
+                                     schema=validate.Schema(validate.parse_html()))
+
+        schema = validate.Schema(validate.xml_xpath_string(".//div[@id='message']/text()"),
+                                 validate.any(None, str))
+        error_msg = validate.validate(schema, root)
+        if error_msg:
+            log.error(f"{error_msg}")
```
bastimeyer, line 73: Is this intentional that there's no `return` statement here? If the content is unavailable / geo-blocked, then it should stop here (I guess), because otherwise it'll log
```
[plugins.mdstrm][debug] Failed to find window.MDSTRM.OPTIONS
```

src/streamlink/plugins/mdstrm.py
```diff
@@ -0,0 +1,140 @@
+"""
+$description CDN hosting live content for various websites in the Americas.
+$url mdstrm.com
+$type live
+"""
```
bastimeyer, line 5: Please add all the supported URLs to the plugin metadata

```suggestion
"""
$description CDN hosting live content for various websites in the Americas.
$url mdstrm.com
$url latina.pe/tvenvivo
$url saltillo.multimedios.com/video
$type live
"""
```

bastimeyer (commented):
The plugin implementation is looking fine and should be good to be merged apart from the missing metadata.

I'm not a fan of the trailing closing parenthesis and indentation style though which makes it unnecessarily more difficult to read and edit later on for the sake of being more compact. I don't want to complain, but if we decide to add a more strict linting config in the future (like python-black for example), then this will most certainly generate unnecessary diffs. For now this is valid coding style though.

src/streamlink/plugins/mdstrm.py
back-to: it is intended,
i don't know if `".//div[@id='message']/text()"` is used for other messages

PR #4396, opened 2022-03-13
cli: add "overwrite" to --force help text

pzhlkj6612:
Hi. Today I'm going to download a stream. I tried multiple times and then encountered the following prompt:

```
File foo.ts already exists! Overwrite it? [y/N] 
```

Alright, I want to overwrite the existing file without confirmation, but what options should I specify? Searching for the "overwrite" word in [Command-Line Interface](https://streamlink.github.io/cli.html) turned up nothing, so I made this PR to improve the docs.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -555,7 +555,7 @@ def build_parser():
         "-f", "--force",
         action="store_true",
         help="""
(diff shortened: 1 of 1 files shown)


bastimeyer:
Thanks

pzhlkj6612:
Hi @bastimeyer , thank you for correcting the PR title and the commit message. Now they look more clear and informative.

PR #4405, opened 2022-03-23
installer: drop Windows installer build config

bastimeyer:
Drop installer build config in favor of the external build config at:
https://github.com/streamlink/windows-installer

- remove installer build scripts+configs
- remove Streamlink config file in win32 subdirectory
- remove windows-installer CI job
  - update steps of the release CI job
- remove removed-plugins file
  - remove src/streamlink/plugins/.removed
  - update package manifest
  - update tests
  - update developing docs
- rewrite install docs

----

Resolves #4390 

Merging this will break @scowalt's chocolatey package (due to the new download location) and @beardypig's portable builds (due to the removed config file).

Now the important question, do we already want to remove the build config, or should we wait and do it after the next release? As you can see, I've published the `3.2.0-1` installer release, but we could keep both versions, the one here and the installers at the other repo, for a short while:
https://github.com/streamlink/windows-installer/releases/tag/3.2.0-1

This `3.2.0-1` installer release doesn't mean though that I'm 100% confident with it. I've only tested it in my W10 VM and didn't touch anything W7 related. The custom Python embedded builds are all working, but they are different from the official ones, as mentioned [here](https://github.com/streamlink/streamlink/issues/4390#issuecomment-1073001230), so there's lots of potential for unknown issues. And on top of that, the FFmpeg builds are new, the removed plugins file got replaced with a `RMDir /r "$INSTDIR\pkgs"` command before unpacking/copying files in the installer, and the default install path got changed in the x86_64 installers.

The reviews I've got so far were quick "It's working for me" comments, without any code / build setup reviews. Since I don't want to cause any issues for Windows users in the next release with the switch of the build config, I would really appreciate if one could at least take a deeper look and preferably build it themselves once. Also please see if something's unclear in the installer repo's readme and the docs changes here.

Btw, even though we're not building the nightly installer anymore, I've kept the scheduled CI runs, so that dependency release issues can be caught.

PR #4410, opened 2022-03-28
chore: remove unneeded release script

bastimeyer:
The script is not needed and would otherwise require an update to fix the format.

I'm also thinking about changing the release format, at least the changelog file itself. Adding the entire shortlog bloats it up tremendously, and writing changelogs in PyCharm has become really unresponsive due to its size when rendering the markdown while writing. The commit history can already be read in git itself, so why even include a shortlog in the changelog file?

The shortlog could be generated for the docs and GH releases page only, where it actually makes sense for people to see. Maybe also splitting the changelog.md into multiple files for each major release version number could make sense, so that the docs could be split into multiple pages, because [that](https://streamlink.github.io/changelog.html) is already a bit ridiculous.

In addition to that, GH has added a new feature last year for the releases pages where contributors get highlighted if they are mentioned via the \@-notation in the release's body text. That could easily be generated by the deploy script.


gravyboat (approved):
Do you want to wait till https://github.com/streamlink/streamlink/pull/4405 is merged?

Issue #3195, opened 2020-09-22
Hello, ask for help with plugin dev


sadovsf:
Hello, i was trying to reach you (to be specific @back-to contributor) to ask how do you find out details for loading stream links from web page.I would like to be able to maintain some of the deprecated myself but when i open pretty much any plugin (obsolete or not) there are often details to get to stream i have no idea how i could find out. Is there some place you could send me to learn or could you maybe contact me with some hints? Contact details are in my github profile.

I'am really sorry for choosing this way to contact you but i failed to find any other :-(

mkbloke:
It mostly involves opening your browser debugger (in Chrome and Firefox with key F12).  You need to find something that links the given page with an eventual stream/playlist URL.  That could involve looking at API requests in the network tab or extracting the relevant JSON from the page source.  It's a matter of reverse engineering a sequence of events if API requests are made, for example, and then implementing those in the plugin with the necessary values extracted from the page as part of the API request.

Unfortunately, I don't have any links to hand which might help you to learn this.  Perhaps other people might have links to some tutorials about it though.

I can only suggest you start to look at your browser debugger while visiting a streaming site and try to learn from that, perhaps in unison with looking at a working streamlink plugin so that you can see how the plugin implements the necessary logic to mimic what happens in your browser when you visit a site.

sadovsf:
I see, that is what I was mostly trying but usually it seems like these links are just appearing out of nowhere 😁

I will try to reverse ingeneer some stream based on existing plugin that is I good point. Maybe you could add page into documentation a ou some usual steps you are going through when working with new stream? Along with some not so trivial example? I believe this would help a lot to community to make their own plugins and lift up some of the implementations or fixes from your back.

bastimeyer:
There is no point in documenting this, as it's different for every site implementing live streams. There are standards for different streaming protocols, but that's about it and Streamlink implements those standards. You'll have to debug it to the point where those standardized protocols are being used on the site and then plug them into the available interfaces in Streamlink while reverse engineering the required steps in a plugin, if it's necessary. If something deviates from those standards, sometimes this can be worked around with custom logic. Everything else requires knowledge in web development and modern web standards, so if you don't have that, you're out of luck.

And if you want to write a plugin without any prior knowledge of the Streamlink APIs, take a look at other plugins first which were recently updated and see how it's done.

sadovsf:
[...]

Issue #117, opened 2016-11-02
Augment Documentation / Website with a section "Primer / How To develop plugins"


WolfganP:
Just as proposed on https://github.com/chrippa/livestreamer/issues/1187 , I believe this will incentive contributions to the project.

After submitting my analysis on a streaming site to help create a plugin (https://github.com/chrippa/livestreamer/issues/1156) that didn't get any traction, I was looking at the code and searching around for documentation on how to develop a plugin myself, but with no luck other than a short answer at the forums (https://groups.google.com/d/msg/livestreamer/TmVKnud0KRc/hDvURFUjrM0J) and it felt a little overwhelming to reverse engineer the plugin and related code code.

I believe that with a little help / article on plugin development, it will be possible for many contributors to add additional community plugins (albeit simple ones). Other option may be take a plugin and post it as a sample with full code comments (ie which functions are mandatory to be developed for a new site to be supported, what those functions should return to core code, etc)
Thx, Wolf

beardypig:
I've written a fair number of plugins, and previously submitted a fairly large number of commits to livestreamer (under a different moniker).
I can give writing up an intro to writing plugins ago :-)

gravyboat:
@beardypig That would be awesome. I looked at the plugins recently and realized that we don't really test any of them. I know it's more to ask, but if you wanted to write up the docs for both writing a plugin and writing plugin tests that would be cool. If not I totally understand.

beardypig:
I can take a look :-)

WolfganP:
That would be awesome. Thx a lot!

dgoiko:
Hello.

We've to implement plugins for 30~ websites, all Spanish radios and televisions. I could find no documentation on plugin creation.

Did anyone posted the mentioned docs? I dont think I'll be allowed to create the plugins if I've to spend several hours reverse-enginering the thing, right now we're using m3u's directly, and will probably stick to that dirty and manual way in order to avoid studying SL code in depth. However, we've allways posted bugfixes or improvements made on well-documented programs.

Thanks in advance.

voltagex:
I was going to open a new issue for this, but it's not clear from https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md how to set up a development environment for streamlink, especially if you're relatively new to Python.

gravyboat:
[...]

PR #4415, opened 2022-03-31
docs: split CLI into submenus

bastimeyer:
- move any sections except for "Command-line usage" (argparse content)
  into separate, logical submenus
- rename certain section headers
- fix all references
- slightly rephrase certain things

----

The motivation for this is that the CLI docs page is way too large and that the actual CLI informations (argparse data) are buried below the tutorial, config file, plugin-sideloading, proxy, streaming-protocol, metadata, and other plugin related sections. This is especially bad, because the tutorial is meant to be read once for new users. For users who are not new, this is distracting and annoying.

Adding more plugin specific docs (eg. adding Twitch low latency docs) to the current monolithic page would make this even more problematic.

I've therefore split the CLI docs into multiple logical submenus, rearranged and rephrased some stuff.

New content and other improvements can be done afterwards.

One minor issue I didn't manage to overcome was the `toctree` and the `self` reference. It's currently impossible to list sections of the own document (`cli.rst` - which includes the "Command-line usage" section), so it's a little bit confusing, as it doesn't link to the actual subsection. See sphinx-doc/sphinx#2103.

**This will break any existing links on this issue tracker or anywhere else that are not argparse related.** I don't think is much of an issue though. The most important links, like anchors to argparse args will remain the same. And the man page also won't be affected by this change.

My aim is to split the changelog in a similar way next, by major release version number. The changelog is currently a symlink to the `CHANGELOG.md` file at the root dir, so I'm not sure yet what the best solution for this will be.


bastimeyer:
See https://deploy-preview-4415--streamlink.netlify.app/

bastimeyer:
"Command-line usage" gets linked twice on the index page due to the toctree issue mentioned above, but that can be ignored.

WolfganP:
Thanks for the documentation reorg, the cli docs are much clearer with the new subsections.

I'll take this opportunity to ask again (https://github.com/streamlink/streamlink/issues/117 wow... 5 years passed by already :-) ) for improving documentation for (casual?) developers of plugins. To be honest, I tried to make a simple one for some sites I use often, or improve **generic** when it fails, but I found that I spent a lot of time doing reverse engineering and fiddling to try to get a solution, and found myself discouraged and jump instead to find the stream manually with the browser console in the end.

So if it's possible to add a section for plugin development (ie under sideloading docs category?), and take an stable plugin and be very very verbose on commenting it (ie mandatory functions, input params, expected return information, error handling, high level interaction with the core modules, etc etc) I think there will be more and more participation of casual devs that may use streamlink as a easily customizable solution for those rebel sites :-)

I know a major improvement to the plugin loading mechanism is being worked on (https://github.com/streamlink/streamlink/issues/3814) so maybe it will be a good opportunity to help improve the plugin dev docs as well?

Please please please don't misinterpret me, I'm a happy user of streamlink and grateful to you all for your dedication the the app, just want to get my hands on it from time to time for the sake of the mental challenge, and probably some others are on the same page as me.

Thanks again all.

bastimeyer:
> improving documentation for (casual?) developers of plugins
> I spent a lot of time doing reverse engineering and fiddling

This is exactly what a plugin implementation is, reverse engineering and fiddling, trying to get the stream URL which can then be plugged into the constructor method of the various stream implementations supported by Streamlink, like HLS, DASH, etc. There is no generic solution which could be documented, because it varies every single time. One needs to at least have a basic understanding of web standards, as I've already explained in another issue.
https://github.com/streamlink/streamlink/issues/3195#issuecomment-697529984

Having a documented example plugin doesn't make much sense, because it would almost never translate into any real plugin implementations and it would become outdated eventually without someone spending additional maintenance time on it which I think is unnecessary.

The Streamlink APIs are all there for you to read and study, and written test cases are the best examples with real usage data. If someone is dedicated enough, learning from existing code instead of bad and potentially outdated documentation is the much better way.

If one wants to submit a new plugin but is unsure if it has been done right, then this is no problem at all and reviews from other devs will most certainly solve all the issues and questions. For private third party plugin implementations, I honestly don't care if Streamlink's lack of proper documentation causes problems for others. All the Streamlink utility functions are optional anyway. And the discussion forum is always there for questions.

The potential of having more plugin submissions if there were better documentation is also questionable IMO. But if someone feels like writing thorough and up2date documentation in order to help the project and attract new contributors, then please submit a pull request. So far, nobody has been willing to spend any time on it (for five years).

gravyboat (approved):
🤦‍♂️ Thought I marked this as approved when I looked at it a few days ago. I'm not concerned about the ToC issue nor am I concerned about breaking any existing links. It happens and the split is absolutely worth it for how much easier it makes doc navigation.

Issue #4358, opened 2022-02-21
plugin: session memory leak


justpaul:
### Checklist

- [X] This is a bug report and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

Reproduce:
```python
import streamlink
import gc

s = streamlink.Streamlink()
del s
gc.collect()

len([obj for obj in gc.get_objects() if isinstance(obj, streamlink.Streamlink)]) == 1
```

@lru_cache used on a class method causes a memory leak:
```python
@lru_cache(maxsize=128)
    def resolve_url(self, url: str, follow_redirect: bool = True) -> Tuple[Type[Plugin], str]:
```
created sessions are not released from the memory.

### Debug log

```text
-
```

bastimeyer:
Btw, I have no idea why the `Plugin.bind` classmethod was implemented (via 2e27f21a3eee8bfbc86135993c642853d6a37264 back in the Livestreamer days in 2013). It's not a particularly nice solution for storing the session on a plugin. It could very easily be done in each plugin's constructor method, especially now since we're returning plugin classes in the `resolve_url` method instead of plugin instances.

Changing that however would be a major breaking change. But since there are also plans for the plugin resolver to switch to a resolver that only operates on pre-built lists of URL regexes instead of loading all builtin plugins at once in order to access their URL regexes, it doesn't make much sense change the `Plugin.bind` method now.

PR #4424, opened 2022-04-03
docs: update API page, add type annotations

bastimeyer:
This updates the API page in the docs, which has been a total mess since years (or basically since forever).

### Changes

Docs / Sphinx:

- Set better autodoc defaults (see docs/conf.py)
  https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html
- Restructure API page and add sections for easier navigation

Streamlink:

- Update all class/method/attribute docstrings that get parsed by autodoc
  - Fix typos and spelling/grammar errors (I didn't read everything, but checked most)
  - This isn't a full rewrite of the existing docstrings, just a cleanup, so there's still lots of room for improvements.
- Add a couple of new docstrings for relevant methods and attributes
- **Where possible**, add type annotations instead of defining parameter and return types via docstrings.

### Issues

Autodoc:

- Autodoc renders certain parameters incorrectly, eg. `session` instead of `session_` (HTTPStream / HLSStream). These method signatures need to be updated in the future because the underscore doesn't make any sense. Fixing that would be a breaking change, so I suggest waiting for the py37 removal (EOL June 2023) when positional-only and keyword-only parameters become available, so this can be done in one go.
- The styling of the autodoc output is not particularly great and it gets really messy when methods have large signatures with type annotations. This can only be worked around with ugly CSS hacks, which isn't worth it.

Streamlink:

- Most types can't be annotated because of import cycle issues. Example of this are `streamlink.stream.Stream` which can't be imported in `streamlink.plugin.plugin`, or `streamlink.session.Streamlink` (in various places). That unfortunately requires a redesign. Annotating `Plugin.session` is also currently impossible because of the `Plugin.bind` classmethod which I've [commented on a few weeks ago](https://github.com/streamlink/streamlink/issues/4358#issuecomment-1046918030).

### Next

1. Improve the streaming protocol parameters documentation, because the parameters get passed to the various stream implementations and `requests.request()`, and my intention was to link to the classes/methods.

### Future improvements

1. Add intersphinx config for the standard library, requests and other dependencies, so that the docs can link to the documentations of the classes and methods which are mentioned in Streamlink's docstrings.
   https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html


bastimeyer:
https://deploy-preview-4424--streamlink.netlify.app/api.html
vs

gravyboat (approved):
The changes look good, big 👍 for fixing the grammar issues and lack of clarity on some of the entries. I'll mark as approved instead of merging right away in case anyone else wants to review just in case I missed something.

PR #4430, opened 2022-04-04
docs: split cli/plugin, add cli/plugin/twitch

bastimeyer:
This adds auth, ads and low latency documentation for the Twitch plugin. Some of it is based on what I have already posted on the Streamlink Twitch GUI wiki a while back.

I didn't include any player specific configurations here, because I don't want to publish information which is potentially incorrect or which is incomplete. This can be updated in the future if one wants to do more thorough research.

If you feel like this is too much text, then we can remove the short history lessions from the auth and ads sections. I included it because it explains the current state of the plugin a bit.


gravyboat (approved):
> If you feel like this is too much text, then we can remove the short history lessons from the auth and ads sections. I included it because it explains the current state of the plugin a bit.

I don't think it's too much text. Authentication, ads, and low latency streaming are asked about or mentioned time and time again on various sites like Reddit whenever people bring up Streamlink so having some extra documentation there makes sense.

PR #4435, opened 2022-04-07
docs: add intersphinx config

bastimeyer:
- add intersphinx mapping for `requests`, skip Python standard library
- fix `requests.Session.request` references in docs and docstrings

----

https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html

I didn't include references to the Python standard library, because it adds external links everywhere, which is annoying.

Since this downloads data from external sites, packages like Debian will probably patch out these changes, but that's nothing we care about here.

### DIFF
--- a/docs/cli/protocols.rst
+++ b/docs/cli/protocols.rst
@@ -81,11 +81,11 @@ Parameters are passed to the following methods of their respective stream implem
 Protocol prefix      Method references
 ==================== =======================
 ``httpstream://``    - :py:meth:`streamlink.stream.HTTPStream`
-                     - :py:meth:`requests.request`
+                     - :py:meth:`requests.Session.request`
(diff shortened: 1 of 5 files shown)


bastimeyer:
master:
- https://streamlink.github.io/latest/cli/protocols.html#available-parameters
- https://streamlink.github.io/latest/api.html#httpstream

with intersphinx:
- https://deploy-preview-4435--streamlink.netlify.app/cli/protocols.html#available-parameters
- https://deploy-preview-4435--streamlink.netlify.app/api.html#streamlink.stream.HTTPStream

----

I'll maybe take a look at the API guide tomorrow, but that should be it for me with docs changes for now.

gravyboat (approved):
> I didn't include references to the Python standard library, because it adds external links everywhere, which is annoying.

👍 This would be overkill any way.

PR #4440, opened 2022-04-08
build: add pyproject.toml, switch to versioningit

### LINKED ISSUE #4321: Build system definition via pyproject.toml (PEP 518)
### Checklist

### Description

As mentioned in #4113, Streamlink will have to switch to the declarative build system via `pyproject.toml` eventually (PEP 518). Implicit setuptools-based build systems via standalone `setup.py` files are deprecated.

Since this requires a major version bump, this should ideally be done in one go with other breaking changes like the removal of py36 support.

Metadata definitions (PEP 621) could be moved too, but don't have to (AFAIA).

- https://www.python.org/dev/peps/pep-0518/
- https://www.python.org/dev/peps/pep-0621/
- https://packaging.python.org/en/latest/tutorials/packaging-projects/
- https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/
- https://setuptools.pypa.io/en/latest/build_meta.html

----

The problem Streamlink currently has with this transition is that it uses [versioneer](https://github.com/python-versioneer/python-versioneer) for defining its version string, and this doesn't work well with pyproject.toml projects. The main issue here are editable installs via `pip install -e .`, and the version does not get set correctly. Versioneer also unfortunately doesn't support pyproject.toml build systems yet, so the version still has to be defined in setuptools' `setup()` call.

An alternative to versioneer is [pypa/setuptools_scm](https://github.com/pypa/setuptools_scm), but one big issue here is how it defines its default version format. While versioneer simply uses the format of `git describe --tags --dirty` (as defined in [`setup.cfg`](https://github.com/streamlink/streamlink/blob/3.1.1/setup.cfg#L57-L59) and in [`src/streamlink/_version.py`](https://github.com/streamlink/streamlink/blob/3.1.1/src/streamlink/_version.py#L315-L337)), setuptools_scm "normalizes" the git describe output and automatically applies pre/post/dev release tags according to [PEP 440](https://www.python.org/dev/peps/pep-0440/). I still haven't figured out how to disable this behavior and keep the current version format.

bastimeyer:
- Add basic pyproject.toml with build-system information
- Drop versioneer in favor of versioningit
- Delete versioneer install module and remove its setup.cfg config
- Replace `src/streamlink/_version.py` and load streamlink's version
  string via versioningit. The module's `__version__` export will get
  replaced by a static string on build-time when creating sdists/wheels,
  which results in the old versioneer behavior with support for
  editable installs.
- Make versioningit use the same version format as versioneer
- Update MANIFEST.in and .coveragerc

----

Resolves #4321 

This needs to be **thoroughly** reviewed before it can be merged. I'm opening this as a draft for now.

I've set the required setuptools version to >=45, which is the one which breaks compatibility with python 2. A higher version could be set here, but it's probably not necessary. [At least version 42 is necessary for PEP 517/518](https://setuptools.pypa.io/en/latest/history.html).

The version string format generated by [versioningit](https://github.com/jwodder/versioningit) remains the same, so it'll be one of these, depending on the current state
- `{version}`
- `{version}+{distance}.{vcs}{rev}`
- `{version}+{distance}.{vcs}{rev}.dirty`

Editable installs via `pip install -e .` should be working. Please see `src/streamlink/_version.py`, which loads the current version from versioningit unless the file's `__version__` export was replaced during build-time (sdist/bdist) with a static version string.

----

The next step would be moving the project metadata from `setup.py` to `pyproject.toml`, and dependencies from `{dev,docs}-requirements.txt` to the `project.optional-dependencies` table of `pyproject.toml`. Maybe also properly defining dev- and doc-dependency version ranges would be a wise idea, at least minimum versions. I'd also love to have a dependency lockfile here, but that can be done later.


bastimeyer:
This will also require an update of the install dependencies in the docs and dev-requirements.

pip 10.0.0 introduced support for PEP 517/518, but there were many changes afterwards, and I think it'd be best to require at least pip version 21.0.0, which was released in January 2021. This release also dropped support for Python 2.

I will update this after #4441 was merged.

PR #4441, opened 2022-04-09
build: move project metadata to pyproject

bastimeyer:
Follow up of #4440 with the pyproject.toml metadata changes, according to PEP 621.
https://peps.python.org/pep-0621/
https://setuptools.pypa.io/en/stable/history.html#v61-0-0

See #4321

----

1. Move the the project metadata from setup.cfg (soon to be deprecated by setuptools) to pyproject.toml according to PEP 621 and bump setuptools requirement to >= 61 (requires at least py37)
2. ~~Remove {dev,docs}-requirements.txt in favor of package "extra" dependencies that can be installed like this `streamlink[dev]`, `streamlink[docs]`~~
3. Move last remaining setuptools configs from setup.cfg to pyproject.toml

----

This leaves setup.cfg with only flake8 related content. Flake8 doesn't support pyproject yet AFAIK, so the file could be renamed to `.flake8`, similar to `.coveragerc` which includes configs for the coverage package, but that's rather pointless, because the stuff will be moved eventually anyway.

I've tried to keep the resulting package metadata as close to the currently included metadata. There are some changes though (ignoring the changes on master):

```sh
$ pip install .
$ diff \
  <( \
    curl -sSL 'https://files.pythonhosted.org/packages/67/97/22af99220a108d24eb635c9f1693ea52e1ffe5b0c93098dae7258c2a6a4d/streamlink-3.2.0.tar.gz' \
    | bsdtar -xOzf - 'streamlink-3.2.0/PKG-INFO' \
  ) \
  src/streamlink.egg-info/PKG-INFO
```
```diff
3c3
< Version: 3.2.0
---
> Version: 3.2.0+28.g678ff680
8c8,33
< License: Simplified BSD
---
> License: Copyright (c) 2011-2016, Christopher Rosell
>         Copyright (c) 2016-2022, Streamlink Team
>         All rights reserved.
>         
>         Redistribution and use in source and binary forms, with or without
>         modification, are permitted provided that the following conditions are met:
>         
>         1. Redistributions of source code must retain the above copyright notice, this
>            list of conditions and the following disclaimer.
>         2. Redistributions in binary form must reproduce the above copyright notice,
>            this list of conditions and the following disclaimer in the documentation
>            and/or other materials provided with the distribution.
>         
>         THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
>         ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
>         WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
>         DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
>         ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
>         (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
>         LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
>         ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
>         (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
>         SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
>         
>         
> Project-URL: Home-Page, https://github.com/streamlink/streamlink
23d47
< Classifier: Programming Language :: Python :: 3.6
120c144
<   [streamlink-plugins]: https://streamlink.github.io/plugin_matrix.html
---
>   [streamlink-plugins]: https://streamlink.github.io/plugins.html
```


bastimeyer:
Just rebased the branch...

I dropped the commit which moves the `{dev,docs}-requirements.txt` contents to `project.optional-dependencies`, because updating these extra deps while having an editable install of streamlink makes pip overwrite that install with a regular one, which is annoying. Let's keep the requirements files for now.

Other changes after rebasing:
- updated the install docs
- bumped min. pip version to 21

bastimeyer:
There are some more issues which I've just noticed. The sdist and bdist archives include the entire repo files, which means they are ignoring the MANIFEST.in. I have no idea why this happens. If this can't be fixed, then these changes will have to be dropped. Setuptools still considers pyproject metadata experimental, btw.

The `entry-points` field also had to be added to the `dynamic` attribute.

bastimeyer:
I've commented earlier (and then deleted my post again) that I noticed issues with the sdist tarballs and that it somehow includes every file in the git repo instead of only the package files and stuff defined in MANIFEST.in.

I still have no idea why this happened when building the sdist tarball and I first thought that this was caused by these changes here or #4440. All I know is that this started happening after the move to pyproject.toml, even after deleting the `build/` and `src/streamlink.egg-info/` dirs, running `setup.py clean --all`, and even after rolling back to `3.2.0`.

After checking `3.2.0`, `master` and this PR branch again in a clean git clone, clean venv, and updated pip+setuptools, everything seems to be working reliably and deterministically, but I can't figure out why this sdist issue is sometimes happening in my development venv. As long as this doesn't affect the release sdists though, this shouldn't be an issue, but it confuses me.

----

Regarding the current state of this PR, I had to add `entry-points` to the `dynamic` attribute in the recent push, because `entry_points` gets set in `setup.py`, as it checks if wheels are built for Windows which then include the additional `streamlinkw` entry. The `data-files` config does not need an entry in the `dynamic` attribute for some reason. This may be because it has been deprecated by setuptools. I don't think it will be possible anymore in the future to include man files and shell completions in wheels.

- https://setuptools.pypa.io/en/latest/deprecated/distutils/setupscript.html#installing-additional-files
- https://setuptools.pypa.io/en/latest/userguide/datafiles.html
- https://setuptools.pypa.io/en/latest/references/keywords.html
- https://setuptools.pypa.io/en/latest/history.html?highlight=data_files#v58-4-0

----

So the only thing that may be of concern here is that the most recent version of setuptools (`62.1.0`) still consideres metadata in pyproject.toml as experimental.
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#configuring-setuptools-using-pyproject-toml-files

> Support for project metadata in `pyproject.toml` is still experimental and may be removed (or change) in future releases.

bastimeyer:
Just a quick force-push in order to resolve the merge conflicts.

This is still considered too new to be merged, as setuptools 61 is required (2022-03-24) and setuptools still considers pyproject metadata experimental itself.
https://setuptools.pypa.io/en/latest/history.html

bastimeyer:
Rebased to master and fixed merge conflicts.

Also changed:
- bumped setuptools requirement from 61 to 64 (2022-08-11), which adds proper support for editable installs via PEP660
  https://setuptools.pypa.io/en/latest/history.html#v64-0-0
- added third commit which renames `setup.cfg` to `.flake8`, since only flake8 config stuff is left

I'll keep this PR as a draft, since setuptools still doesn't consider pyproject.toml project metadata stable

bastimeyer:
Rebased to master once again, with changes of #5141 applied to pyproject.toml

bastimeyer:
Rebased and fixed merge conflicts once more (#5143 and #5149)

Issue #4455, opened 2022-04-14
Remove `streamlink.plugin.api.utils.itertags`


bastimeyer:
[`streamlink.plugin.api.utils.itertags`](https://github.com/streamlink/streamlink/blob/3.2.0/src/streamlink/plugin/api/utils.py#L16-L28) has become obsolete ever since `lxml` was added as a dependency to Streamlink for parsing HTML.

`itertags` is a hacky implementation via regexes, which is not only slow, but it's also impossible to correctly parse HTML nodes with regular expressions, so it shouldn't be used when better and much faster solutions are available. It also always requires unescaping tag values, which is annoying.

We've already updated and replaced lots of plugins which were previously using it, but there are still some left:
```
$ GIT_PAGER=cat git grep -F 'from streamlink.plugin.api.utils import' a1ce471f
a1ce471f:src/streamlink/plugins/cdnbg.py:from streamlink.plugin.api.utils import itertags
a1ce471f:src/streamlink/plugins/facebook.py:from streamlink.plugin.api.utils import itertags
a1ce471f:src/streamlink/plugins/funimationnow.py:from streamlink.plugin.api.utils import itertags
a1ce471f:src/streamlink/plugins/senategov.py:from streamlink.plugin.api.utils import itertags
a1ce471f:src/streamlink/plugins/vrtbe.py:from streamlink.plugin.api.utils import itertags
a1ce471f:tests/test_plugin_utils.py:from streamlink.plugin.api.utils import itertags
```

- [x] cdnbg
- [x] facebook
- [x] funimationnow
- [x] senategov
- [x] vrtbe

Once every last plugin has been updated, the entire `streamlink.plugin.api.utils` module can be removed, as it only contains the `itertags` function and some other useless export aliases which are not even used anymore in Streamlink's codebase.

If we care about plugin-API stability (something which has never been discussed), removing this would be considered a breaking change. Since we've just dropped py36, that's something which could be included in the 4.0.0 release.

bastimeyer:
Regarding facebook, as I already posted on Gitter/Matrix, I have a dev branch with a plugin rewrite here:
https://github.com/streamlink/streamlink/compare/master...bastimeyer:plugins/facebook/rewrite
I lost interest though, because I don't have a FB account, and all new ones I've created got blocked pretty much immediately. That's why my rewrite is incomplete and basically just a translation and improvement of the old logic, which means that the changes are most likely not working.

----

[...]

PR #4457, opened 2022-04-14
plugins.funimationnow: replace itertags

bastimeyer:
ref #4455 

Completely untested changes. This is just the equivalent of the itertags-based logic. Returns the value of the `value` attribute of the first `input` element with the `name` attribute value set to `self.CSRF_NAME`, or `None`.


mkbloke:
I wouldn't worry about this being untested.  The plugin was in need of a complete rewrite before and will still require that.

That shouldn't stop this being committed, as it does at least rid the plugin of its dependency on `itertags`.

I have had for some time a  WIP on the funimation plugin that has written out the `itertags` dependency, but I have yet to get around to finishing it...

PR #4458, opened 2022-04-14
plugins.senategov: remove plugin

bastimeyer:
ref #4455 

I don't know how much sense it makes fixing this plugin. It's broken since the introduction of the pluginmatcher logic (#3821), and it was still referencing `self.url_re` the whole time, which didn't exist anymore, leading to the `AttributeError: 'SenateGov' object has no attribute 'url_re'` error. Nobody has reported a plugin issue since then and the plugin has been very very niche to begin with. The two URLs defined in the tests are working now though. I didn't check anything else.


back-to:
> It's broken since the introduction of the pluginmatcher logic

just remove it completely

gravyboat:
> just remove it completely

Agreed.

PR #4471, opened 2022-04-19
plugins.trovo: new plugin

### LINKED ISSUE #4436: trovo.live
### Checklist

- [X] This is a plugin request and not a different kind of issue
- [X] [I have checked the list of open and recently closed plugin requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+request%22)

### Description

trovo(https://trovo.live/) is a streaming media platform launched by Tencent. I have seen the Plugin Request record in issues before(#3069 , #3076 ), but it was not considered because trovo was in the beta stage at that time. At present, trovo has removed the beta logo and ended the beta stage. Hope to re-apply for plugin support for this platform.

### Input URLs

1. https://trovo.live/BloodyTV1
2. https://trovo.live/Sibuyas
3. https://trovo.live/SpaK

mkbloke:
closes #4436

### DIFF
--- a/src/streamlink/plugins/trovo.py
+++ b/src/streamlink/plugins/trovo.py
@@ -1,0 +1,181 @@
+"""
+$description Global video game live streaming and video hosting platform, owned by Tencent.
+$url trovo.live
+$type live, vod
+"""
+
+import logging
+import random
+import re
+
+from streamlink.plugin import Plugin, pluginmatcher
+from streamlink.plugin.api import validate
+from streamlink.stream.hls import HLSStream
+from streamlink.utils.url import update_scheme
+
+log = logging.getLogger(__name__)
+
+
+@pluginmatcher(re.compile(r"""
+        https?://(?:www\.)?trovo\.live/
+        (?:
+            (?:
+                (?:clip|video)/(?P<video_id>[^/?&]+)
+            )
+            |
+            (?P<user>[^/?&]+)
+        )
+""", re.VERBOSE))
+class Trovo(Plugin):
+    @staticmethod
+    def generate_qid():
+        return f"{random.getrandbits(40):010x}".upper()
+
+    def get_vod(self, video_id):
+        json = self.session.http.post(
+            f"https://gql.trovo.live/?qid={self.generate_qid()}",
+            json=[{
+                "operationName": "batchGetVodDetailInfo",
+                "variables": {
+                    "params": {
+                        "vids": [video_id],
+                    },
+                },
+                "extensions": {
+                    "persistedQuery": {
+                        "version": 1,
+                        "sha256Hash": "ceae0355d66476e21a1dd8e8af9f68de95b4019da2cda8b177c9a2255dad31d0",
+                    },
+                },
+            }],
+            schema=validate.Schema(
+                validate.parse_json(),
+                [{
+                    "data": {
+                        "batchGetVodDetailInfo": {
+                            "VodDetailInfos": validate.any(
+                                {
+                                    video_id: {
+                                        "streamerInfo": {
+                                            "userName": str,
+                                        },
+                                        "vodInfo": {
+                                            "playInfos": [{
+                                                "desc": validate.all(validate.transform(lambda s: s.lower()), str),
+                                                "playUrl": validate.url(),
+                                            }],
+                                            "vid": str,
+                                            "title": str,
+                                            "categoryName": str,
+                                            "playbackRights": {
+                                                "playbackRightsSetting": str,
+                                                "playbackRights": str,
+                                            },
+                                        },
+                                    },
+                                },
+                                {},
+                            ),
+                        },
+                    },
+                }],
+                validate.get((0, "data", "batchGetVodDetailInfo", "VodDetailInfos", video_id)),
+            ),
+        )
+
+        if not json:
+            log.error("Video not found")
+            return
+
+        log.debug(json["vodInfo"]["playbackRights"])
+        self.id = json["vodInfo"]["vid"]
+        self.author = json["streamerInfo"]["userName"]
+        self.title = json["vodInfo"]["title"]
+        self.category = json["vodInfo"]["categoryName"]
+
+        for s in json["vodInfo"]["playInfos"]:
+            q = s["desc"]
+            if "(source)" in q:
+                q = f"source_{q.replace('(source)', '')}"
+            yield q, HLSStream(self.session, update_scheme("https:", s["playUrl"]))
+
+    def get_live(self, user):
+        json = self.session.http.post(
+            f"https://api-web.trovo.live/graphql?qid={self.generate_qid()}",
+            json=[{
+                "operationName": "live_LiveReaderService_GetLiveInfo",
+                "variables": {
+                    "params": {
+                        "userName": user,
+                    },
+                },
+            }],
+            schema=validate.Schema(
+                validate.parse_json(),
+                validate.any(
+                    [{
+                        "data": {
+                            validate.optional("live_LiveReaderService_GetLiveInfo"): {
+                                "streamerInfo": {
+                                    "userName": str,
+                                },
+                                "programInfo": {
+                                    "id": str,
+                                    "title": str,
+                                    "streamInfo": [{
+                                        "desc": validate.all(validate.transform(lambda s: s.lower()), str),
+                                        "playUrl": validate.transform(lambda s: s.replace(".flv?", ".m3u8?")),
+                                    }],
+                                },
+                                "categoryInfo": {
+                                    "shortName": str,
+                                },
+                                "isLive": int,
+                            },
+                        },
+                    }],
+                    [{
+                        "errors": [{
+                            "message": validate.transform(lambda s: s.replace('\\"', '"')),
+                        }],
+                    }],
+                ),
+                validate.get(0),
+            ),
+        )
+
+        if "errors" in json:
+            errors = [e["message"] for e in json["errors"]]
+            log.error(f"API error(s): {', '.join(errors)}")
+            return
+
+        if json["data"]["live_LiveReaderService_GetLiveInfo"]:
+            json = json["data"]["live_LiveReaderService_GetLiveInfo"]
+        else:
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/trovo.py
```diff
@@ -0,0 +1,167 @@
...
+        self.id = json["vodInfo"]["vid"]
+        self.author = json["streamerInfo"]["userName"]
+        self.title = json["vodInfo"]["title"]
+        self.category = json["vodInfo"]["categoryName"]
+
+        for s in json["vodInfo"]["playInfos"]:
+            q = s["desc"]
+            if "(source)" in q:
+                q = f"source_{q.replace('(source)', '')}"
+            yield q, HLSStream(self.session, s["playUrl"])
```
bastimeyer, line 90: The `playUrl` values seem to be `http://` based. Accessing `https://` works, so `update_scheme("https:", s["playUrl"])` should be used here.

Same thing for live streams.

src/streamlink/plugins/trovo.py
```diff
@@ -0,0 +1,167 @@
...
+                },
+                "extensions": {
+                    "persistedQuery": {
+                        "version": 1,
+                        "sha256Hash": "ceae0355d66476e21a1dd8e8af9f68de95b4019da2cda8b177c9a2255dad31d0",
+                    },
+                },
+            }],
+            schema=validate.Schema(
+                validate.parse_json(), [{
```
bastimeyer, line 44: Please don't put multiple parameters on the same line like this. Same down below. This will guarenteed cause an unnecessary diff if we switch to a different linting config in the future.

src/streamlink/plugins/trovo.py
```diff
@@ -0,0 +1,167 @@
...
+                        validate.optional("live_LiveReaderService_GetLiveInfo"): {
+                            "streamerInfo": {
+                                "userName": str,
+                            },
+                            "programInfo": {
+                                "id": str,
+                                "title": str,
+                                "streamInfo": [{
+                                    "desc": validate.all(validate.transform(lambda s: s.lower()), str),
+                                    "playUrl": validate.transform(lambda s: s.replace(".flv?", ".m3u8?")),
```
bastimeyer, line 115: The site is using a progressive HTTP stream with a Flash video container. Is this only the case for non-authenticated users? I'm asking because the site asks for a login to be able to watch source quality streams. And VODs are available in HLS regardless.

src/streamlink/plugins/trovo.py
```diff
@@ -0,0 +1,167 @@
...
+
+from streamlink.plugin import Plugin, pluginmatcher
+from streamlink.plugin.api import validate
+from streamlink.stream.hls import HLSStream
+
+log = logging.getLogger(__name__)
+
+
+@pluginmatcher(re.compile(
+    r"https?://(?:www\.)?trovo\.live/([^/?&]+)(?:/([^/?&]+))?"
```
bastimeyer, line 19: The capture groups could be named and the distinction between live streams and vods/clips be improved, but it should be fine as it is now.

bastimeyer (commented):
Thanks for the PR. The plugin seems to be working fine.

One issue I observed while testing was a desync of video and audio in live streams, which is probably due to the HLS change.

src/streamlink/plugins/trovo.py
```diff
@@ -0,0 +1,167 @@
...
+                }],
+                validate.get((0, "data", "batchGetVodDetailInfo", "VodDetailInfos"))
+            ),
+        )
+
+        if video_id not in json:
+            log.error("Video not found")
+            return
+        else:
+            json = json[video_id]
```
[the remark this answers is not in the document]
mkbloke: You mean like this?
```
json = json.get(video_id)
if not json:
    log.error("Video not found")
    return
```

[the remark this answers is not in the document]
bastimeyer:
```
validate.get((0, "data", "batchGetVodDetailInfo", "VodDetailInfos", video_id))
...
if not json:
    log.error(...)
    return
```

src/streamlink/plugins/trovo.py
```diff
@@ -0,0 +1,167 @@
...
+                "operationName": "batchGetVodDetailInfo",
+                "variables": {
+                    "params": {
+                        "vids": [video_id],
+                    },
+                },
+                "extensions": {
+                    "persistedQuery": {
+                        "version": 1,
+                        "sha256Hash": "ceae0355d66476e21a1dd8e8af9f68de95b4019da2cda8b177c9a2255dad31d0",
```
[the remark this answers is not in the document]
mkbloke: Ah, I meant to mention that, but forgot.  These hashes seem to be computed from queries defined within the sourced JS.  I looked at trying to compute it and it looked rather difficult, not to mention it would be unmaintainable if it were done.  It does not appear to be an sha256 sum alone, the code loops through an array of integers in the sourced JS, amongst other things.  All rather dreadful.

The hash used here was obtained from the network traffic and has remained constant in the week or so that I have observed it.

src/streamlink/plugins/trovo.py
mkbloke: I hadn't noticed that, thanks.

src/streamlink/plugins/trovo.py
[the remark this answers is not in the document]
bastimeyer: GraphQL persistedQuery IDs are just cached GQL queries on the server side. I didn't find this exact ID in my network activity while browsing some VODs, so the query could've already been changed and a different ID is now used.
https://www.apollographql.com/docs/apollo-server/performance/apq/
This isn't much of a problem now, but it could be if they change their database models. Let's just keep it for now.

src/streamlink/plugins/trovo.py
mkbloke: I don't know about logged-in verses not, as I have not created an account.  I thought it would be better to use HLS rather than an HTTP stream, but given the A/V de-sync issue you have noticed (I had not, although I did not watch the output for very long), perhaps it would be better to just HTTP stream the FLV content?

src/streamlink/plugins/trovo.py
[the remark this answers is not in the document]
mkbloke: I looked at some of the docs on Apollo regarding the persisted queries last week.  In what I saw, they explain how it works, but not in great detail.  The implementation seems to be simple if you pull in their JS to take care of it all for you behind the scenes.  I did wonder if it would be possible to use the full base query, but I'm not familiar with GQL and wasn't sure about how to use what I found (if it was even the right thing).

src/streamlink/plugins/trovo.py
bastimeyer: A progressive HTTP stream would definitely be worse. I've checked two streams, and both had desync with their HLS streams. It could be possible though that it was also the case on the site, which I didn't check. I will take another look later. If there is indeed a desync issue, then ffmpeg could help.

src/streamlink/plugins/trovo.py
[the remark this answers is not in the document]
mkbloke: It's still there for me, by the way - in the payload data of the POST requests:

https://gql.trovo.live/?chunk=1&reqms=1650397043989&qid=9BF05FD913&cli=4&client_info={"device_info":"{\"tid\":\"16503970247684763\"}"}&from=/clip/lc-387702298160635191?ltab=videos&locale=en-US&resolution=760*1280
```
[
  {
    "operationName": "getVersionUpdateTips",
    "variables": {},
    "extensions": {
      "persistedQuery": {
        "version": 1,
        "sha256Hash": "669c9bb4d507500703acf252ee398ea17eb94f2d0b42b3d283c451e3aeb9bb8f"
      }
    }
  },
  {
    "operationName": "batchGetVodDetailInfo",
    "variables": {
      "params": {
        "vids": [
          "lc-387702298160635191"
        ]
      }
    },
    "extensions": {
      "persistedQuery": {
        "version": 1,
        "sha256Hash": "ceae0355d66476e21a1dd8e8af9f68de95b4019da2cda8b177c9a2255dad31d0"
      }
    }
  },
  ....
```
and:

https://gql.trovo.live/?chunk=1&reqms=1650397624783&qid=5C418F880F&cli=4&client_info={"device_info":"{\"tid\":\"16503976105681262\"}"}&from=/video/ltv-106639746_106639746_387702298489518804?ltab=videos&locale=en-US&resolution=760*1280
```
[
  {
    "operationName": "getVersionUpdateTips",
    "variables": {},
    "extensions": {
      "persistedQuery": {
        "version": 1,
        "sha256Hash": "669c9bb4d507500703acf252ee398ea17eb94f2d0b42b3d283c451e3aeb9bb8f"
      }
    }
  },
  {
    "operationName": "batchGetVodDetailInfo",
    "variables": {
      "params": {
        "vids": [
          "ltv-106639746_106639746_387702298489518804"
        ]
      }
    },
    "extensions": {
      "persistedQuery": {
        "version": 1,
        "sha256Hash": "ceae0355d66476e21a1dd8e8af9f68de95b4019da2cda8b177c9a2255dad31d0"
      }
    }
  },
  ....
```

src/streamlink/plugins/trovo.py
mkbloke: I think I've made it better...

src/streamlink/plugins/trovo.py
mkbloke: How much de-sync were you seeing?  I've watched a couple of streams for a bit and cannot seem to see an obvious de-sync.  I was watching streams with player cams shown as a picture in picture.  It's not always easy to tell, but generally the A/V looked to be in sync as far as I could see.

bastimeyer: It could've been the streams themselves which I checked yesterday. There definitely was a delay of 2 seconds or so. I can't see anything like this right now, so I really don't know what was going on there.

Apart from the massive delay caused by the ridiculous targetduration on the playlists, the HLS streams should be working fine, I guess.

PR #4479, opened 2022-04-21
Clean up coverage.py and flake8 configs

bastimeyer:
Tool configs should all be moved to `pyproject.toml` instead of using tool-specific files or `setup.cfg`. Having a central and standardized config file makes reading build and tool configs much more easy.

#4441 also tries to move project metadata from `setup.cfg` to `pyproject.toml`, but this is not yet considered stable according to the latest setuptools release, so we shouldn't merge this yet. This means that `setup.cfg` will remain for another while now.

Flake8 also unfortunately doesn't support `pyproject.toml` and the developers don't want to implement it, so unless we switch to a wrapper/fork, we're stuck with `setup.cfg` here too.

There are a couple of flake8 forks and wrappers, but only one seems to be reasonable.
- [pyproject-flake8](https://github.com/csachs/pyproject-flake8)
  Wrapper for flake8 which monkeypatches the config import logic, but doesn't guarantee API stability
- [flake9](https://gitlab.com/retnikt/flake9)
  Fork of flake8 with added pyproject.toml config logic, but seems to be unmaintained and is a couple of releases behind flake8
- [life4/flakehell](https://github.com/life4/flakehell)
  Wrapper with much more added features, but is discontinued.
- [flakehell/flakehell](https://github.com/flakehell/flakehell)
  Fork/continuation of life4/flakehell. Requires a different config. Incompatible with flake8 >=4.0.0.
- [flakeheaven/flakeheaven](https://github.com/flakeheaven/flakeheaven)
  Another fork/continuation of life4/flakehell. Compatible with most recent flake8 releases. Requires different configuration, but has a proper [documentation](https://flakeheaven.readthedocs.io/en/latest/index.html). The best solution, IMO.

Alternatively, we could also switch to a different linter, but that seems to be additional and unnecessary work. Black for example would require a ton of code changes, and I don't agree with some of its code style opinions.

----

1. The first commit moves the coverage.py config from `.coveragerc` to `pyproject.toml` and gets rid of one unnecessary file in the root directory.
2. The second commit cleans up the outdated flake8 config in `setup.cfg`.
   - removes versioneer stuff
   - removes old and unnecessary "per-file-ignores"
   - removes unneeded builtins overrides
   - adds `count = True`, so that it doesn't have to be added as CLI argument


bastimeyer:
Unless there are any concerns here, this should be good for merging. No functional changes, just configs being moved to the right place and configs being cleaned up.

Btw, I also had a look at flakeheaven, but I couldn't find a way to configure flake8 plugins like flake8-import-order, where we set a different style than the plugin's default. Other than that, the flakeheaven worked fine, albeit a tiny bit slower.

gravyboat:
I'm fine with flakeheaven.

> Alternatively, we could also switch to a different linter, but that seems to be additional and unnecessary work.

Let's not do that now. 👍

PR #4491, opened 2022-04-25
plugins.nicolive: fix timeshift-offset option

ekangmonyet:
Fix `AttributeError: 'NicoLive' object has no attribute 'hls_stream_url'` when `--niconico-timeshift-offset` option is used


bastimeyer (approved):
Thanks

PR #4497, opened 2022-04-29
changelog: remove git shortlogs

bastimeyer:
This rewrites the changelog and GitHub release body. The goal was to get rid of the git shortlog in the changelog.md file, as it bloats up the entire file by a lot and is mostly unnecessary. I've replaced each shortlog with a simple link to the git commit history between the previous tag and the current one on GitHub, where commit diffs are just one click away, which is much better in most cases.

The changelog is used in a couple of different places and removing the git shortlog has different implications:

1. **The changelog file itself**
   - **accessed via the git repo on GitHub**  
     A link to the commit history with access to diffs makes more sense than a shortlog
   - **accessed via a locally cloned git repo**
     The commit log and diffs can be read locally
   - **accessed via a downloaded tarball / zip archive from GitHub**
     A link to the commit history with access to diffs makes more sense than a shortlog
   - **accessed via the source-dist tarballs uploaded to PyPi / GitHub releases**
     Removing the git shortlog could be a problem here, but I don't think it's too big of an issue
2. **The docs**  
   A link to the commit history with access to diffs makes more sense than a shortlog
3. **As part of the GitHub release body message**  
   The new release script automatically generates a shortlog in a addition to a new contributors list with GitHub account names for automatic highlighting, so nothing will be changed here. Having a shortlog on the releases page makes sense, because the changelog often doesn't include everything and refers to the shortlog for minor plugin changes/fixes.

----

This doesn't split the changelog in the docs into multiple pages like I originally intended, as it would require writing a new Sphinx extension. Removing the shortlogs however already solves most of the navigation / legibility issues of the current changelog page in the docs.

----

I haven't verified these changes in my test repo with a test-release yet, but the dry-runs all looked fine. Opening this as a draft for now, because it won't be part of 4.0.0 anyway.


gravyboat:
This looks really good to me. It's very clear how everything works and the usage of the diff in the changleog file itself is a good idea.

bastimeyer:
This should now be ready for review. As mentioned, I did some test releases and it should be fine. I deleted these test releases, but if you want confirmation, I can push another test release again.

I unfortunately made a mistake in #4503 and forgot about the wheels in the dist directory which are now also uploaded to 4.0.1 on GitHub. This PR would fix it again and would only upload the tarball (plus signature).

There's also the option of replacing the streamlink-bot token with a simple GitHub actions token when creating a GitHub release. streamlink-bot has become obsolete and was only added in the past because some API OAuth token was needed to create releases or to push commits (in other repos). This is now fully automated via `GITHUB_TOKEN` and the according token permissions. I've modified the script so that both kinds of API tokens can be used. It's just a simple choice of which GH account will be used for GH releases. The streamlink-bot one however is non-temporal and thus has the potential to be stolen/abused. And I believe it's a deprecated OAuth token, too.

gravyboat:
> The streamlink-bot one however is non-temporal and thus has the potential to be stolen/abused. And I believe it's a deprecated OAuth token, too.

Yeah I'd prefer we no longer use this account for security purposes if we don't have to. I don't see any reason to support both.

bastimeyer:
> were there any changes made since I looked at this a few days ago

This is the diff compared to the very first push
https://github.com/streamlink/streamlink/compare/717fffa..d2becf1

So basically just a few clean-ups, bugfixes and minor changes in regards to the API key. Nothing functional.

Considering that there could be more problems with 4.0.x, I'd suggest not merging this yet. It's just an improvement for the readability of the changelog anyway (with the added contributors list).

bastimeyer:
One final update:

- ignore merge commits in contributors list
- don't include release commit in shortlog or contributors list
- use `GITHUB_TOKEN` for deploying to GH releases (instead of custom OAuth token from streamlink-bot)
- rename the file and make it consistent with other script file names (using a second commit to retain git history)

----

Test release for demonstration purposes:

- https://github.com/bastimeyer-test/streamlink-test/releases/tag/99.0.0
- https://github.com/bastimeyer-test/streamlink-test/compare/master...99.0.0
- https://github.com/bastimeyer-test/streamlink-test/runs/6252478448?check_suite_focus=true#step:9:12

.github/workflows/main.yml
```diff
@@ -148,8 +150,8 @@ jobs:
         run: ./script/build-and-sign.sh
       - name: Github release
         env:
-          RELEASES_API_KEY: ${{ secrets.RELEASES_API_KEY }}
-        run: ./script/github_releases.py "${STREAMLINK_DIST_DIR}"/*
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
back-to, line 153: there is no `secrets.GITHUB_TOKEN` in
https://github.com/streamlink/streamlink/settings/secrets/actions

bastimeyer: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

PR #4505, opened 2022-05-01
release: 4.0.1

bastimeyer:
Package updates won't be necessary.


bastimeyer:
I usually don't want to merge my own PRs in this repo, but in case nobody has merged this and tagged `4.0.1` within the next couple of hours, I'll be doing it myself because I want this to be fixed as soon as possible, so packagers can do their work.

PR #4509, opened 2022-05-02
docs: update dependencies section in install docs

bastimeyer:
`versioningit` needs to be documented here, because packagers who use setuptools directly will need it.

I've also reformatted the dependencies table in one go and added notes about PEP517/PEP518 and the 4.0.0 release.


bastimeyer:
https://deploy-preview-4509--streamlink.netlify.app/install.html#dependencies

PR #4517, opened 2022-05-07
stream.segmented: join worker+writer on close

### LINKED ISSUE #4516: SegmentedStreamWriter.close() does not reliably finish before CLI exits (race condition)
### Checklist

### Streamlink version

### Background
I was doing some works on latest `master` branch, adding some feature on Segmented/HLS streams, and it requires some cleanup, so I added them in the end of `SegmentedStreamWriter.close()`, like

```python
        self.closed = True
        self.reader.close()
        self.executor.shutdown(wait=True, cancel_futures=True)

        __my_extra_cleanup()
```

And I found my cleanup code does not always being executed, when HLS streams end normally.

### Problem
When an HLS stream ends normally, the whole shutdown process is triggered by the last line of `SegmentedStreamWriter.run()`, `self.close()`. Then `SegmentedStreamWriter.close()` calls `SegmentedStreamReader.close()`, then the iteration loop of `stream_cli.main:read_stream()` will reach its end, then `main()` exits.

The problem is, `SegmentedStreamWriter.run() -> SegmentedStreamWriter.close()` was run in a separated thread, which means `SegmentedStreamWriter.close()` cannot reliably finish its work before main thread exits.

### To reproduce
To reliably trigger the problem, adding a sleep to original `SegmentedStreamWriter.close()`, like

```python
        self.closed = True
        self.reader.close()
        self.executor.shutdown(wait=True, cancel_futures=True)

        time.sleep(3)
        log.debug("SegmentedStreamWriter.close() ends")
```

Then run the CLI with a short HLS stream, the `SegmentedStreamWriter.close() ends` message never appears.

### Debug log

```text
......

[cli][info] Stream ended
```

### Expected result
```text
......

[cli][info] Stream ended
[stream.segmented][debug] SegmentedStreamWriter.close() ends
```

yhhsin:
A simple, maybe ulgy, attempt to fix #4516 .
There may be more beautiful approaches.

### DIFF
--- a/src/streamlink/stream/segmented.py
+++ b/src/streamlink/stream/segmented.py
@@ -3,7 +3,7 @@ import queue
 from concurrent import futures
 from concurrent.futures import Future, ThreadPoolExecutor
 from sys import version_info
-from threading import Event, Thread
+from threading import Event, Thread, current_thread
 from typing import Any, Optional
 
 from streamlink.buffers import RingBuffer
@@ -228,9 +228,11 @@ class SegmentedStreamReader(StreamIO):
         self.writer.close()
         self.buffer.close()
 
-    def final_wait(self):
-        self.writer.join()
-        self.worker.join()
+        current = current_thread()
+        if current is not self.worker:  # pragma: no branch
+            self.worker.join(timeout=self.timeout)
+        if current is not self.writer:  # pragma: no branch
+            self.writer.join(timeout=self.timeout)
 
     def read(self, size):
         return self.buffer.read(
--- a/src/streamlink/stream/segmented.py
+++ b/src/streamlink/stream/segmented.py
@@ -228,6 +228,10 @@ class SegmentedStreamReader(StreamIO):
         self.writer.close()
         self.buffer.close()
 
+    def final_wait(self):
+        self.writer.join()
+        self.worker.join()
+
     def read(self, size):
         return self.buffer.read(
             size,
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -1073,6 +1073,10 @@ def main():
                 try:
                     log.info("Closing currently open stream...")
                     stream_fd.close()
+                    try:
+                        stream_fd.final_wait()
+                    except AttributeError:
+                        pass
                 except KeyboardInterrupt:
                     error_code = 130
     else:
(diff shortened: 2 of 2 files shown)


yhhsin:
Note: Simply putting `writer.join()` in `SegmentedStreamReader.close()` doesn't work, because `SegmentedStreamWriter.close()` also calls `SegmentedStreamReader.close()`, which creates a deadlock.

yhhsin:
Another approach:

In `SegmentedStreamReader.close()`, try to join `worker` and `writer` thread, like:
```diff
     def close(self):
         self.worker.close()
         self.writer.close()
         self.buffer.close()
+        try:
+            self.writer.join()
+        except RuntimeError:
+            pass
+        try:
+            self.worker.join()
+        except RuntimeError:
+            pass

```

Note about the try-catch: Since `SegmentedStreamReader.close()` and `SegmentedStreamWriter.close()` calls each other, putting `join()` like above will inevitably join current thread, which causes a RuntimeError.

bastimeyer (commented):
Thanks for the PR.

The following diff should be enough to fix the issue with the `SegmentedStreamWorker` and `SegmentedStreamWriter` deamon threads. No need to add a special method to `SegmentedStreamReader`, which is just one of the stream (base) classes, and catching an `AttributeError` when trying to call this method from the CLI on other stream implementations that don't have this method.

```diff
diff --git a/src/streamlink/stream/segmented.py b/src/streamlink/stream/segmented.py
index d097be24..26fbdac1 100644
--- a/src/streamlink/stream/segmented.py
+++ b/src/streamlink/stream/segmented.py
@@ -3,7 +3,7 @@ import queue
 from concurrent import futures
 from concurrent.futures import Future, ThreadPoolExecutor
 from sys import version_info
-from threading import Event, Thread
+from threading import Event, Thread, current_thread
 from typing import Any, Optional
 
 from streamlink.buffers import RingBuffer
@@ -228,6 +228,12 @@ class SegmentedStreamReader(StreamIO):
         self.writer.close()
         self.buffer.close()
 
+        current = current_thread()
+        if current is not self.worker:  # pragma: no branch
+            self.worker.join(timeout=self.timeout)
+        if current is not self.writer:  # pragma: no branch
+            self.writer.join(timeout=self.timeout)
+
     def read(self, size):
         return self.buffer.read(
             size,
```

PR #4523, opened 2022-05-11
session: fix deprecated imports

bastimeyer:
- Import urllib3 directly instead of importing from requests.packages
- Remove unneeded urllib3.contrib.pyopenssl.DEFAULT_SSL_CIPHER_LIST
  override, as it's not used on py3
- Update ipv4/ipv6 tests, add http-disable-dh tests

----

Motivation:
https://github.com/psf/requests/blob/v2.27.1/requests/packages.py#L11-L20

The `urllib3.contrib.pyopenssl` stuff in `--http-disable-dh` should've been removed with py2:
https://urllib3.readthedocs.io/en/latest/reference/contrib/pyopenssl.html

----

The `--http-disable-dh` logic could also be changed in the future, so that no global `urllib3.util.ssl_.DEFAULT_CIPHERS` override has to be done. Similar to #4345, TLS cipher updates could be applied in a new `SSLContext` (created via [`urllib3.util.ssl_.create_urllib3_context`](https://github.com/urllib3/urllib3/blob/1.26.9/src/urllib3/util/ssl_.py#L250-L285)) that gets added in a custom  `requests.HTTPAdapter` which gets mounted by Streamlink's `HTTPSession` when `--http-disable-dh` is set.

### DIFF
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -4,9 +4,10 @@ from functools import lru_cache
 from socket import AF_INET, AF_INET6
 from typing import Any, Dict, Optional, Tuple, Type
 
-import requests
-import requests.packages.urllib3.util.connection as urllib3_connection
-from requests.packages.urllib3.util.connection import allowed_gai_family
+# noinspection PyPackageRequirements
+import urllib3.util.connection as urllib3_util_connection
+# noinspection PyPackageRequirements
+import urllib3.util.ssl_ as urllib3_util_ssl
 
 from streamlink import __version__, plugins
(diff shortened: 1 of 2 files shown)


bastimeyer:
Also, why is this still here?
https://github.com/streamlink/streamlink/blame/810e0eb9b5eca988a8ffb088907b8bbdec27e000/src/streamlink/plugin/api/http_session.py#L18-L23

bastimeyer:
> Do you want to do that in this PR?

No, this is unrelated to these changes here.

PR #4527, opened 2022-05-14
docs: rewrite API guide

bastimeyer:
- Use unopinionated Apple HLS stream example URL
- Update Stream string representations
- Fix class/method/attribute references


bastimeyer:
Old:
https://streamlink.github.io/api_guide.html

New:
https://deploy-preview-4527--streamlink.netlify.app/api_guide.html

Issue #4528, opened 2022-05-14
ValueError: Padding is incorrect. With --hls-segment-key-uri option.


kamikazechaser:
### Checklist

- [X] This is a bug report and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest build from the master branch

### Description

Possibly related to #4300, but since that one has been specifically addressed, opening a new one here.

`localhost:8080/key`  returns the unmangled key based on the `_decrypt_key` implementation [here](https://github.com/streamlink/streamlink/issues/2253#issuecomment-455571775). I assume if that was the issue, it would throw an AES/decryption error instead.

> Another problem is that the HLSStreamWriter.fetch method (which calls create_request_params) is run by other threads from the writer's thread-pool, so a sequential order of requests is not guaranteed.

Sounds like this might be the issue. Not sure though.

### Debug log

```text
streamlink -l debug https://vcdn.spayee.in/spees/w/o/607d7f740cf265c8e0215b0d/v/60a0c1310cf20fa1ed1a0bf0/u/626cf77b0cf20486a589e9e8/t/39bca27e4d2ece835d78adaa651a019e/p/assets/videos/607d7f740cf265c8e0215b0d/2021/05/16/60a0c1310cf20fa1ed1a0bf0/index.m3u8 500k --hls-segment-key-uri "http://localhost:8080/key"
[cli][info] Loaded config from deprecated path, see CLI docs for how to migrate: /home/kamikaze/.streamlinkrc
[cli][debug] OS:         Linux-5.10.52-antix.1-amd64-smp-x86_64-with-glibc2.31
[cli][debug] Python:     3.9.2
[cli][debug] Streamlink: 4.0.1+29.g9e4a6577
[cli][debug] Requests(2.26.0), Socks(1.7.1), Websocket(1.3.2)
[cli][debug] Arguments:
[...]

PR #3768, opened 2021-06-02
plugin.api.http_session: enforce_content_length

bastimeyer:
Resolves #2365 

This *should* resolve #2365, but I can't validate it myself with data of real HLS streams and don't want to spend time writing sophisticated tests for the entire upstream stack where this bug occurs.

See https://github.com/psf/requests/issues/4956#issuecomment-573325001

PR #4532, opened 2022-05-16
stream.hls: catch ValueError on segment decryption

bastimeyer:
Catch ValueError when decrypting and unpadding segments, and stop
the stream on error after logging the error message.

----

`ValueError`s can get raised by both `unpad` and `decryptor.decrypt()`. This change therefore also includes the garbage data cut-off when the segment length is not a multiple of the AES block size (before decrypting).

We've had a discussion before where we were not sure why the garbage data cut-off was ever implemented. And to be honest, it doesn't make any sense having it, because if the encrypted segment is not a multiple of the AES block size, then it's an invalid segment and cutting off any remaining bytes surely won't help with fully decrypting it. The result will most likely be junk or incomplete data.

I believe that this was added in c38a543c726f9b207a80836edb09f9ae696f0871 to avoid `ValueError`s being raised when calling `decryptor.decrypt()` when the segment length is invalid. This seemed to be a better solution to the `Content-Length` issue which could occur that resulted in partial/incomplete segment downloads (fixed by #3768), and it meant that the stream could continue.

**Any opinions on whether the garbage truncation should be removed?** It doesn't hurt having it, but it's pretty much useless IMO. The `test_hls_encrypted_aes128_incorrect_block_length` test was added by me in b218259f08a16fe328f24ba901a8f207d62415e6 just to have the code path covered.

----

As you can see in the diff, it now logs the `Error while decrypting segment {sequence.num}` error message instead of showing an exception call stack.  This makes it much more clear that it's an issue with the stream, and not Streamlink.

However, if an invalid decryption key is used (eg. #4528), the error message is still not 100% clear. Invalid decryption keys will result in nonsense data after decrypting the segment, so the `unpad` function call will raise a `ValueError` in 99.9% of all cases. I don't think it's too much of a problem though because it can't be distinguished from corrupted/broken segment downloads anyway, so adding something like "Invalid key?" isn't always a good fit.

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -219,16 +219,23 @@ class HLSStreamWriter(SegmentedStreamWriter):
                 self.close()
                 return
 
-            data = result.content
-            # If the input data is not a multiple of 16, cut off any garbage
-            garbage_len = len(data) % AES.block_size
-            if garbage_len:
-                log.debug(f"Cutting off {garbage_len} bytes of garbage before decrypting")
-                decrypted_chunk = decryptor.decrypt(data[:-garbage_len])
-            else:
-                decrypted_chunk = decryptor.decrypt(data)
+            try:
+                data = result.content
+                # If the input data is not a multiple of AES.block_size (16 bytes), cut off any garbage
+                garbage_len = len(data) % AES.block_size
+                if garbage_len:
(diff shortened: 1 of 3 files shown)


bastimeyer:
> **codecov/project/tests** — 99.98% (target 100.00%)

Do I seriously have to open a new PR because codecov doesn't understand that I have force-pushed and fixed the test coverage issue? It is still showing the code of the first push...
https://app.codecov.io/gh/streamlink/streamlink/compare/4532/changes#D1L116

PR #4538, opened 2022-05-21
Fix typing issues

bastimeyer:
This will fix lots of typing issues that can be detected using the following mypy config. These commits don't just include simple type annotation changes, but also rearrange and fix a couple of minor things, which is why I didn't create just a single big commit for this.

https://mypy.readthedocs.io/en/stable/index.html

```toml
[tool.mypy]
python_version = 3.7
show_error_codes = true
show_error_context = true
show_column_numbers = true
ignore_missing_imports = true
warn_no_return = false
```

As you can see, there's no commit in this PR for adding mypy to the CI, because it's been a rather unstable experience. mypy has crashed a lot while fixing all these issues and its reports were often flaky and inconsistent, despite clearing and/or disabling its cache. In my test repo, it crashed during its CI run, showing that it's not suitable at the moment.

----

There's also a ton of stuff which doesn't make sense to me. For example, it complains that the custom `log.trace` method doesn't exist (as the base logging class is retrieved dynamically during runtime via `logging.setLoggerClass()` / `logging.getLoggerClass()`), which makes sense for a static type checker, but it only complains about that in the ustream plugin for some reason and nowhere else.

Then there's also the issue with some custom HLSStream and SegmentedStream implementations, which currently requires having a ton of ignore pragmas. Named tuples for example can't be subclassed, and dataclasses (or the [attrs](https://www.attrs.org/en/stable/overview.html) library) would be required as a replacement for that, which requires a bigger rework. I want to do this eventually though.

And py37 is unfortunately a bit limited in terms of the typing module, so certain things can't be used without a stub-module, which I don't want to add.

### DIFF
--- a/src/streamlink/logger.py
+++ b/src/streamlink/logger.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from logging import CRITICAL, DEBUG, ERROR, INFO, WARNING
 from pathlib import Path
 from threading import Lock
-from typing import IO, List, Optional, Union
+from typing import IO, List, Optional, TYPE_CHECKING, Union
 
 
 FORMAT_STYLE = "{"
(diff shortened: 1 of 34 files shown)


bastimeyer:
Rebased and fixed two minor issues that came up after #4540

PR #4543, opened 2022-05-23
docs: temporarily disable intersphinx

bastimeyer:
The TLS cert of https://python-requests.org/ is currently broken due to a domain suspension. The non-TLS version is broken too.

See
https://github.com/psf/requests/issues/6140

Unless intersphinx gets disabled, docs tests and docs builds will fail, as it can't pull the linked docs data from requests.
Once the issue has been resolved, intersphinx can be re-enabled.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,7 +19,7 @@ needs_sphinx = '3.0'
 extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.autosectionlabel',
-    'sphinx.ext.intersphinx',
+    # 'sphinx.ext.intersphinx',
     'ext_argparse',
     'ext_github',
     'ext_plugins',


bastimeyer:
```
loading intersphinx inventory from https://docs.python-requests.org/en/stable/objects.inv...

Warning, treated as error:
failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://docs.python-requests.org/en/stable/objects.inv' not fetchable due to <class 'requests.exceptions.SSLError'>: HTTPSConnectionPool(host='docs.python-requests.org', port=443): Max retries exceeded with url: /en/stable/objects.inv (Caused by SSLError(CertificateError("hostname 'docs.python-requests.org' doesn't match either of '*.ispapi.net', 'ispapi.net'")))
make: *** [Makefile:45: html] Error 2
```

bastimeyer:
Going to merge this now, as it prevents future PRs from failing.

PR #4544, opened 2022-05-23
Add mypy config

bastimeyer:
- Add mypy and available typing stubs to dev-requirements
- Add mypy config
- Add py.typed to streamlink + streamlink_cli packages (PEP 561):
  This is required, so that mypy can import types from streamlink when
  installed in editable mode. Don't publish this file for now.
  Once various streamlink APIs are properly typed, typing stub files
  can be added instead.

----

Follow-up of #4538.

This doesn't add mypy to the CI (yet), but it defines its config, so it can be run locally without yielding typing errors.

In #4538 I had the `ignore_missing_imports` option set to true, but this is now set to false and import errors are fixed by adding typing stubs of supported dependencies to `dev-requirements.txt`, and adding `py.typed` ([PEP 561](https://peps.python.org/pep-0561/)) to `streamlink` and `streamlink_cli`. Imports of dependencies without typing information get suppressed with the first commit of this PR.

See https://mypy.readthedocs.io/en/stable/running_mypy.html#ignore-missing-imports

The `py.typed` files do not get published in the sdists and wheels. They are only meant for enabling imports from the `streamlink` and `streamlink_cli` packages when type-checking in editable installs.

----

See my test-repo for a new CI linting config, as a separate workflow:
- https://github.com/bastimeyer-test/streamlink-test/commit/f9a0e36c9efb4e8a3c8af3d4950c38eaedba772c
- https://github.com/bastimeyer-test/streamlink-test/actions/runs/2372653558

Having a separate linting workflow allows tests and linters to fail independently. It adds 3 more checks to the PRs though, but I don't think it's too bad. It currently also runs on push in my test repo which I would remove if included here, so that it only runs in PRs. flake8 now only runs once on py37, and the other file permission/encoding checks also only run once.

As I mentioned in #4538, mypy crashed a couple of times while I was fixing the typing issues. Apparently, this is caused by bad caching data of incremental runs (there's a PR open for a couple of weeks that may fix this). For some reason, it also failed in my initial CI test, but that may be due to it running multiple time for each input and thus relying on cached data. Incremential checks can be disabled via [`--no-incremental`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-incremental), which I have added now and it hasn't crashed since (I've re-run the CI job a few times).

This means I would feel comfortable adding the new CI linting config to this PR now, since incremential checks are irrelevant for CI checks anyway, but I would like to get some feedback first, also because of the separate workflow.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,54 @@
+name: Lint
+
+on:
+  pull_request: {}
+
+jobs:
+  files:
+    name: Files
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v3
+      - name: File permissions
+        run: "! grep -Ev '^644' <(git ls-files src/ tests/ | xargs stat '--format=%a %n')"
+      - name: File encodings
+        run: "! grep -E -e 'UTF-[^8]' -e 'UTF-[^ ]+ \\(with BOM\\)' <(git ls-files src/ tests/ | xargs stat '--format=%a %n')"
+      - name: Line terminators
+        run: "! grep 'with CRLF line terminators' <(git ls-files | file -nNf-)"
+      - name: No unicode bidirectional control characters
+        run: "! git grep -EIn $'[\\u2066\\u2067\\u2068\\u2069\\u202A\\u202B\\u202C\\u202D\\u202E]'"
+
+  style:
+    name: Code style
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+        with:
+          python-version: "3.7"
+      - name: Install dependencies
+        run: ./script/install-dependencies.sh
+      - name: flake8
+        run: >
+          python -m flake8
+
+  typing:
+    name: Typing
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+        with:
+          python-version: "3.10"
+      - name: Install dependencies
+        run: ./script/install-dependencies.sh
+      - name: mypy
+        run: >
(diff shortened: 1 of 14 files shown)


gravyboat:
I'll try to make time to do a deeper dive once I'm done working.

gravyboat:
Went ahead and took a look at this tonight, your branches look fine to me.

>  It adds 3 more checks to the PRs though, but I don't think it's too bad.

I don't think this is bad either and actually prefer the linting being broken out from the testing, breaking it out in to three distinct sections is also a good idea. This makes it much more clear what failed and why, especially for new contributors or those less familiar with the code base.

> Incremential checks can be disabled via [--no-incremental](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-incremental), which I have added now and it hasn't crashed since (I've re-run the CI job a few times).

I don't see an issue with this unless throwing away the cache greatly increases the build times, it doesn't seem to matter much based on the testing branch you've built out and all three finish more than fast enough.

> This means I would feel comfortable adding the new CI linting config to this PR now, since incremental checks are irrelevant for CI checks anyway, 

👍 from me. It's a big improvement with minimal impact to build times and configuration.

bastimeyer:
Added the new CI workflow. Only runs on pull request, not on push.

gravyboat:
@bastimeyer Is this ready to go or is there more you want to add?

bastimeyer:
No, this should be good.

I could see adding a few more linting steps in the future though, but that's unrelated to mypy, which is the whole motivation of this PR.

The only thing that may be confusing is the addition of the `py.typed` files to `streamlink` and `streamlink_cli`, but as explained, this is necessary for mypy to import typing information from outside of these packages, like when `streamlink_cli` is importing from `streamlink` or the tests importing from both. Since we currently don't care about other projects trying to import streamlink type informations, we don't publish the files in the sdists/bdists. If we care about publishing type informations in the future, then this should be done via stub files (.pyi) without runtime code.

PR #4547, opened 2022-05-24
plugins.useetv: log if no link has been found

LaneSh4d0w:
**Why this PR ?**

This PR has been made to verify if no link has been found. Indeed, USeeTV doesn't provide all his channels worldwide. Some channels are blocked for Indonesian people only, and some others need a subscription to work (see beIN Asia as an example). Some channels like SeaToday would work, but channels like this one : 
image
will only show a Geo-restriction message above the player, telling the end-user he has no access to the stream. 

This also reflects inside the player, meaning no link can be scraped.

### DIFF
--- a/src/streamlink/plugins/useetv.py
+++ b/src/streamlink/plugins/useetv.py
@@ -12,6 +12,7 @@ from streamlink.plugin.api import validate
 from streamlink.stream.dash import DASHStream
 from streamlink.stream.hls import HLSStream
 
+log = logging.getLogger(__name__)
 
 @pluginmatcher(re.compile(r"https?://(?:www\.)?useetv\.com/"))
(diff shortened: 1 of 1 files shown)


mkbloke:
Thanks.  You'll also need to add the following if this is going to be merged:
```
```
Add it after the imports with one blank line between and another two blank lines after to pass the linting check.

bastimeyer (commented):
The elif block is unnecessary, as both if-blocks above already check for the URL being truthy.

Plugins which return/yield no streams already show an error message on the CLI, btw. If you want to show geo-restrictions or account-restrictions, it's better to check the HTTP response with certain XPath queries. Otherwise, it'll show this message even when the user has entered an invalid URL that still gets matched by the plugin.

Issue #4329, opened 2022-02-03
Streamlink corrupts URLs when given master playlist URL


AlyoshaVasilieva:
### Checklist

- [X] This is a bug report and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

(Region-locked to Canada)

1. Access the URL `https://www.cbc.ca/bistro/order?mediaId=1991453251854&limit=10&sort=dateAired` and take the value of `key` next to `PlatformLoader`
2. Access that URL, which at least for me is `https://link.theplatform.com/s/ExhSPC/media/guid/2655402169/1991453251854/meta.smil?feed=Player%20Selector%20-%20Prod&format=smil&mbr=true&manifest=m3u`
3. Take the URL listed as `<video src` (the proper form, not HTML-escaped): `https://cbcrcolympic-cbc.akamaized.net/hls/live/2025675/geo_allow_ca/1avymu0xew6/master-archive.m3u8?startTime=1643803194&endTime=1643809839&hdnea=ip=XXX.XXX.XXX.XXX~st=1643881385~exp=1643881775~acl=/*~id=705ce015-7a4c-4b2b-80dc-384ae25c52e7~hmac=11e7382f8de240f90b8b92e752e566dacabb09d6a58e053028a3dec5cdd6736c`
4. Pass that URL to streamlink

Expected result: it works. Browsers work fine.  
Actual result: Streamlink shows 403 Forbidden, but the URL it lists is not a valid URL and was not listed in the playlist.

I'm pretty sure there's a bug here, because manually building a playlist's URL and giving it to Streamlink *does* work:
1. Download the `master-archive.m3u8` playlist
[...]

PR #4552, opened 2022-05-25
stream.hls: parse M3U8 from Response obj directly

bastimeyer:
- Accept both `str` and `requests.Response` in `M3U8Parser.parse`
- Always set encoding of variant and media HLS playlists to utf-8
  https://datatracker.ietf.org/doc/html/rfc8216#section-9
- Remove old custom utf-8 overrides

----

Passing the HTTP Response object to the parser and iterating its content prevents having to keep a copy of the entire response content in memory. Support for `str` is kept for backwards compatibility and because some tests rely on it and would need to get rewritten.

The content is now also always read as UTF-8, as defined by RFC 8216. This was previously guessed by `chardet`/`charset_normalizer` if no HTTP response headers were set (see #4329) and it required custom overrides if there were issues while figuring out the unknown encodings. All HLS tests have been using implicit utf-8 encoding the entire time. We could add tests for invalid encodings in the future, but I don't think it's important.

----

I have a couple more changes planned for the `HLSStream`+ and `M3U8`+ implementations. But that's not ready yet and unrelated to this PR. Just want to quickly talk about that.

One of the changes is using `typing.Generic` + `typing.TypeVar` for making it easier to subclass HLS streams, the parser and other stuff, without having to suppress invalid type informations or method signatures (see the Twitch plugin for example). Using dataclasses instead of named tuples is also one of the things which will make subclassing/extending HLS logic easier.

Another change is passing the parsed master playlist object to the HLSStreams created by `parse_variant_playlist`, so additional data can be read by the media playlists and their parsers. It currently only adds the master playlist URL to the `{Muxed,}HLSStream` for the `to_manifest_url` method (I also want to change this interface eventually, but that may be a breaking change). I noticed the master playlist issue while rewriting Twitch's low latency stuff two days ago, because there's metadata in the master playlist that should be read by the media playlist. It's also relevant for `EXT-X-SESSION-DATA` and `EXT-X-SESSION-KEY`, which are currently not implemented.

### DIFF
--- a/src/streamlink/plugins/raiplay.py
+++ b/src/streamlink/plugins/raiplay.py
@@ -16,13 +16,6 @@ from streamlink.stream.hls import HLSStream
 log = logging.getLogger(__name__)
 
 
-class RaiPlayHLSStream(HLSStream):
-    @classmethod
-    def _get_variant_playlist(cls, res):
-        res.encoding = "UTF-8"
-        return super()._get_variant_playlist(res)
-
-
 @pluginmatcher(re.compile(
     r"https?://(?:www\.)?raiplay\.it/dirette/(\w+)/?"
 ))
@@ -56,7 +49,7 @@ class RaiPlay(Plugin):
             log.error("Geo-restricted content")
             return
 
-        yield from RaiPlayHLSStream.parse_variant_playlist(self.session, stream_url).items()
+        yield from HLSStream.parse_variant_playlist(self.session, stream_url).items()
 
 
 __plugin__ = RaiPlay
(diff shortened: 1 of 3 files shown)


gravyboat:
> One of the changes is using typing.Generic + typing.TypeVar for making it easier to subclass HLS streams...

This sounds like a valid change and I don't see any downside to it.

> Another change is passing the parsed master playlist object to the HLSStreams created by...

I'm not too concerned with breaking changes. I'm not really sure what other feedback to provide here, what would be the downside of implementing this change?

----

Is there more you want to do in this PR? If not I'll merge it, the changes look good.

bastimeyer:
No, there's nothing to be added here apart from new tests which test invalid encodings, which is not that useful.

`requests.Response.iter_lines` reference here, btw:
https://github.com/psf/requests/blob/v2.27.1/requests/models.py#L794

Compared to `requests.Response.text`:
https://github.com/psf/requests/blob/v2.27.1/requests/models.py#L846

PR #4556, opened 2022-05-30
plugin.api: update useragents

mkbloke:
Maybe worth popping this into 4.1.0.

All these were taken from the various sub-pages at: https://www.whatismybrowser.com/guides/the-latest-user-agent/


bastimeyer:
Thanks.

One thing to note is that this changes the version numbers of both Chromium and Firefox to three-digit numbers, which both browser vendors were recently very cautious about due to lots of sites parsing the user agent strings (which is bad) and relying on certain patterns (which is even worse). But since these browsers have already been released as a stable version, that shouldn't be an issue we need to worry about.

I'll see if the UA string updates can be automated when I get the time. There's an API available on whatismybrowser.com that can be used and a GH actions workflow could be added that opens a PR once a month:
https://developers.whatismybrowser.com/api/docs/v2/integration-guide/#software-version-numbers-request-and-response

bastimeyer:
Yes, I noticed that after writing the message. Access to the API endpoint is unfortunately not available via their free plan. I've sent them an email and asked if they support FOSS projects like this with free access. Let's see if and what they reply.

PR #4564, opened 2022-06-03
Added support for Skyline website

LightKitty:
Added support for Skyline website.
The Skyline website address is https://www.skylinewebcams.com/

### DIFF
--- a/src/streamlink/plugins/skyline.py
+++ b/src/streamlink/plugins/skyline.py
@@ -1,0 +1,101 @@
+import logging
+import re
+from urllib.parse import urljoin, urlparse
+
+from streamlink.exceptions import (
+    NoPluginError,
+)
+from streamlink.plugin import Plugin, pluginmatcher
+from streamlink.stream import HLSStream
+
+SKYLINE_VERSION = '2022-06-02'
+
+log = logging.getLogger(__name__)
+
+
+@pluginmatcher(re.compile(r'https?://www\.skylinewebcams\.com/(?P<url>.+)'))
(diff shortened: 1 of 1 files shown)


bastimeyer:
Thanks for submitting a pull request.

I'm sorry, but we're not going to add new plugins for static webcam locations, as you can read in the contribution guidelines:

> 11. Sites which are static cameras of a physical location

Please check the guidelines before opening a PR, or better, ask on the issue tracker for feedback before writing a new plugin, so you don't have to waste your time.

hellishvictor:
> Thanks for submitting a pull request.
> 
> I'm sorry, but we're not going to add new plugins for static webcam locations, as you can read in the contribution guidelines: https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#plugin-requests
> 
> > 11. Sites which are static cameras of a physical location
> 
> Please check the guidelines before opening a PR, or better, ask on the issue tracker for feedback before writing a new plugin, so you don't have to waste your time.

There is a plugin for EarthCam, which is almost the same than Skyline.

bastimeyer:
Yes, and that plugin was added during the Livestreamer days before Streamlink was forked and before our plugin wubmission rules were decided. Those old plugins which would violate those rules if being submitted now have been kept for as long as they've been working. Plugins which started to break and which don't require an entire rework have consistently been removed, and so will Earthcam should it break.

You can always sideload plugins which were not pulled into the master branch.
https://streamlink.github.io/cli/plugin-sideloading.html

PR #4568, opened 2022-06-04
stream.hls: cache parsed multivariant playlist

bastimeyer:
- Move `M3U8Parser.base_uri` to `M3U8.uri`
- Add `multivariant` to `{,Muxed}HLSStream` and deprecate `url_master`:
  - Set `multivariant` in `HLSStream.parse_variant_playlist` instead of
    `url_master`, so that stream instances can keep a reference of
    their parsed multivariant playlist.
    This is required for implementing the `EXT-X-SESSION-{DATA,KEY}`
    HLS tags, or for accessing custom tags in HLS subclasses.
  - Update `to_manifest_url`
    TODO: rename method / refactor interface
- Fix subclasses in various plugins

----

The `url_master` parameter was not removed from `{,Muxed}HLSStream` for backwards compatibility reasons. This can be done later when other HLS method signatures get changed in the future.

Instead of "master", I chose the name "multivariant", because that's how master playlists will be called in future versions of the HLS spec. This avoids one unnecessary parameter name change later on and keeps it consistent with future spec implementations.

### DIFF
--- a/src/streamlink/plugins/ltv_lsm_lv.py
+++ b/src/streamlink/plugins/ltv_lsm_lv.py
@@ -49,7 +49,7 @@ class LTVHLSStream(HLSStream):
         streams = super().parse_variant_playlist(*args, **kwargs)
 
         for stream in streams.values():
-            stream.args["url"] = copy_query_url(stream.args["url"], stream.url_master)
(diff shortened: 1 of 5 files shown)


gravyboat:
Nice job getting ahead of this!

Issue #4562, opened 2022-05-31
segmentation error with LANG=fr_FR.iso885915@euro


novazur972:
### Checklist

- [X] This is a bug report and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

Segmentation fault with my locale for several plugins (pluz, tf1)

$ locale
LANG=fr_FR.iso885915@euro
LC_CTYPE="fr_FR.iso885915@euro"
LC_NUMERIC="fr_FR.iso885915@euro"
LC_TIME="fr_FR.iso885915@euro"
LC_COLLATE=C
LC_MONETARY="fr_FR.iso885915@euro"
LC_MESSAGES="fr_FR.iso885915@euro"
LC_PAPER="fr_FR.iso885915@euro"
LC_NAME="fr_FR.iso885915@euro"
LC_ADDRESS="fr_FR.iso885915@euro"
LC_TELEPHONE="fr_FR.iso885915@euro"
LC_MEASUREMENT="fr_FR.iso885915@euro"
LC_IDENTIFICATION="fr_FR.iso885915@euro"
LC_ALL=

BUT works fine with LANG=C

$ LANG=C streamlink --loglevel debug http://www.france.tv/france-2/direct.html
[cli][debug] OS:         Linux-5.15.41-gentoo-x86_64-Intel-R-_Core-TM-_i5-9600K_CPU_@_3.70GHz-with-glibc2.34
[cli][debug] Python:     3.9.12
[cli][debug] Streamlink: 4.1.0
[cli][debug] Requests(2.27.1), Socks(1.7.1), Websocket(1.3.2)
[cli][debug] Arguments:
[cli][debug]  url=http://www.france.tv/france-2/direct.html
[cli][debug]  --loglevel=debug
[cli][debug]  --player=mpv
[cli][info] Found matching plugin pluzz for URL http://www.france.tv/france-2/direct.html
[plugins.pluzz][debug] Country: MQ
[plugins.pluzz][debug] Video ID: 006194ea-117d-4bcf-94a9-153d999c59ae
[...]

PR #4575, opened 2022-06-05
cli: list all dependencies in debug output

bastimeyer:
- Require importlib-metadata as fallback on Python < 3.8
- Add importlib_metadata to streamlink_cli.compat
- List all dependencies in `log_current_versions`
- Update tests

----

All of Streamlink's dependencies should be listed in the debug output. This is especially useful for figuring out issues like #4562.

```
$ streamlink -l debug
[cli][debug] OS:         Linux-5.18.1-1-git-x86_64-with-glibc2.35
[cli][debug] Python:     3.10.4
[cli][debug] Streamlink: 4.1.0+7.g0b539fd1
[cli][debug] Dependencies:
[cli][debug]  isodate: 0.6.1
[cli][debug]  lxml: 4.9.0
[cli][debug]  pycountry: 22.3.5
[cli][debug]  pycryptodome: 3.14.1
[cli][debug]  PySocks: 1.7.1
[cli][debug]  requests: 2.27.1
[cli][debug]  websocket-client: 1.3.2
[cli][debug] Arguments:
[cli][debug]  --player=mpv

Use -h/--help to see the available options or read the manual at https://streamlink.github.io
```

----

Since `importlib.metadata` is only available on py38 and above, the `importlib-metadata` fallback requirement needs to be added for py37.
https://github.com/python/importlib_metadata#compatibility
https://pypi.org/project/importlib-metadata/

The version restriction of `<4.3` is intentional because of `flake8`, which sets the same requirement:
https://github.com/PyCQA/flake8/blob/4.0.1/setup.cfg#L45
This unfortunately causes a conflict with sphinx, which requires `>=4.4` for py37-py39:
https://github.com/sphinx-doc/sphinx/blob/v4.5.0/setup.py#L32

Since we don't build docs and thus don't install docs requirements on py37, we don't need to care about sphinx's dependency version conflict. But since we obviously run tests and lint code on py37, the version requirement of flake8 needs to be set to avoid introducing version conflicts when installing streamlink first and then streamlink's dev-requirements, because pip won't downgrade in this case.

This means that the version conflict does only affect packagers of streamlink who publish the latest versions for py37. This is already unlikely, and then it can simply be patched out if needed.

It probably doesn't matter anyway and the version conflict can probably be ignored... Flake8 is running fine with the latest importlib-metadata version. **So should I remove the version restriction again?**

```
$ pip install -U importlib-metadata
...
flake8 4.0.1 requires importlib-metadata<4.3; python_version < "3.8", but you have importlib-metadata 4.11.4 which is incompatible.
Successfully installed importlib-metadata-4.11.4
$ flake8
0
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,6 +38,7 @@ package_dir =
   =src
 packages = find:
 install_requires =
+  importlib-metadata ; python_version<"3.8"
   isodate
   lxml >=4.6.4,<5.0
   pycountry
--- a/src/streamlink_cli/compat.py
+++ b/src/streamlink_cli/compat.py
@@ -3,6 +3,11 @@ import sys
 from pathlib import Path
 from typing import BinaryIO, TYPE_CHECKING
 
+try:
+    import importlib.metadata as importlib_metadata  # type: ignore[import]  # noqa: F401
+except ImportError:
+    import importlib_metadata  # type: ignore[import]  # noqa: F401
+
 
 is_darwin = sys.platform == "darwin"
 is_win32 = os.name == "nt"
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -3,6 +3,7 @@ import errno
 import logging
 import os
 import platform
+import re
 import signal
 import sys
 from contextlib import closing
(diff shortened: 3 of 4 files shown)


bastimeyer:
Decided to remove the version restriction.

PR #4594, opened 2022-06-14
ci: update deploy-docs script

bastimeyer:
- Replace SSH key with new streamlinkbot PAT
- Update and refactor deploy-docs.sh
- Update CI workflow

----

Since there's still no reponse from beardypig and nobody has access to the streamlink-bot, let's update the docs-deploy script and replace the SSH key with a PAT from the new streamlinkbot account.

While the scope of PATs still can't be configured and it enables write permissions everywhere where the PAT's account has write permissions, security-wise, it's more than enough and nothing to worry about compared to the encrypted SSH key.

The PAT has been added to the repo secrets (`DOCS_DEPLOY_TOKEN`) and been copied to PM, and write permissions for streamlinkbot have been added to `streamlink/streamlink.github.io`.

I've tested the changes via a temp PAT on my test account and a test deploy-repo by running the script locally. If something shouldn't work after merging this, we can easily revert it if we're unable to fix it.


script/deploy-docs.sh
```diff
@@ -93,6 +78,8 @@ https://github.com/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}
 EOM
 
 echo Pushing changes
+git config --unset 'http.https://github.com/.extraheader' || true
```
bastimeyer, line 81: This is needed because of the GitHub CI runner configuration when pushing via https and PATs.
The `|| true` is set here because of the `set -e` BASH options from above.

bastimeyer: Example: `streamlink-twitch-gui-wiki` -> `streamlink-twitch-gui.wiki`:
https://github.com/streamlink/streamlink-twitch-gui-wiki/commit/ab68051a597d1ebf2f60e120b3bff2dfcc0ea92d
https://github.com/streamlink/streamlink-twitch-gui-wiki/runs/6873416348?check_suite_focus=true#step:3:10

PR #4610, opened 2022-06-19
docs: update dependencies

bastimeyer:
This updates the furo theme to the latest version and adds an edit button and scroll-to-top button. Except for some minor details, everything should stay the same, but please check thoroughly before merging.

Noticeable changes

- CLI arguments and API reference are displayed much clearer
- Code highlighting colors are different
- Some tables don't define a colgroup anymore and have thus different sizes

----

This also replaces recommonmark (abandoned project) with myst-parser for parsing markdown (changelog). No differences here.

And sphinx gets set to `>=4.0.0`, which was already a requirement of the previously used version of furo. Sphinx 5.0.0 does not have any interesting changes, and both furo and myst-parser require at least 4.0.0. We've been using the latest version the whole time, so this is only relevant for packagers who package offline documentation.

----

- https://www.sphinx-doc.org/en/master/changes.html
- https://github.com/pradyunsg/furo/blob/2022.06.04.1/docs/changelog.md
- https://github.com/readthedocs/recommonmark#recommonmark
- https://myst-parser.readthedocs.io/en/latest/intro.html

### DIFF
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,4 +1,4 @@
-sphinx>=3.0
-furo==2021.09.08
+sphinx >=4.0.0
+furo ==2022.06.04.1
 recommonmark>=0.5.0
 versioningit >=2.0.0, <3
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,4 +1,4 @@
 sphinx >=4.0.0
 furo ==2022.06.04.1
-recommonmark>=0.5.0
+myst-parser >=0.18.0
 versioningit >=2.0.0, <3
--- a/docs/_static/styles/custom.css
+++ b/docs/_static/styles/custom.css
@@ -40,6 +40,35 @@ html {
   font-size: 100% !important;
 }
 
+h1, h2 {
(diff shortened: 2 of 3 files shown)


bastimeyer:
Old:
https://streamlink.github.io/latest/index.html

New:
https://deploy-preview-4610--streamlink.netlify.app/index.html

gravyboat:
Text color choices are clearer on the new one which is nice as is the spacing. Only thing I noticed while browsing around is that the `github` and `issues` buttons no longer look like buttons, but are just links:
broken_icons

I don't think it's problematic enough to avoid blocking this merge however.

mkbloke:
> Only thing I noticed while browsing around is that the github and issues buttons no longer look like buttons, but are just links

That's different to what I see.  In Firefox 101.0.1 they are both showing like this for me:

menu

bastimeyer:
The GH buttons have not been changed. Their data, including stylesheets, gets loaded from a JS snippet, as you can see here:
https://github.com/bastimeyer/streamlink/blame/docs/update-deps/docs/_templates/sidebar/github-buttons.html

PR #4614, opened 2022-06-20
Update thirdparty.rst

asabino2:

### DIFF
--- a/docs/thirdparty.rst
+++ b/docs/thirdparty.rst
@@ -85,3 +85,12 @@ OBS-Streamlink
 
 
 Help us extend this list by sending us a pull request on Github. Thanks!
+
+Streamproxy
+-----------
+:Description: A proxy for livestreams
+:Type: Server
+:OS: |Windows/docker/linux comming soon|
+:Author: `Alexander Sabino <https://github.com/asabino2>`_
+:Website: https://github.com/asabino2/streamproxy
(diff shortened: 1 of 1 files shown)


bastimeyer:
Thanks for the PR, but apart from the docs issues you'd first have to fix here, I see several issues with your project which make me not want to include it in Streamlink's docs, at least for now:

1. It's only 7 days old and doesn't have any users other than you
2. There is no project license and the one [set in the package.json](https://github.com/asabino2/streamproxy/blob/e1fd9e7fb07ab60bb88176d14c1b5352f949d428/package.json#L17) is invalid
3. You've [published a 118MiB setup.exe binary](https://github.com/asabino2/streamproxy/releases/tag/streamproxy-1.0.5) for a very simple NodeJS application [without any information on how it's built](https://github.com/asabino2/streamproxy/blob/e1fd9e7fb07ab60bb88176d14c1b5352f949d428/package.json#L11). The docker container [is not much better](https://hub.docker.com/layers/asabino2/streamproxy/latest/images/sha256-7009792f27e7e9eb436f864af375008c3bf3b1ae1c5e2e6ac16ed502bbb250ba?context=explore).
4. There are several dangerous remote execution security issues in your project because you don't sanitize user inputs:
   - https://github.com/asabino2/streamproxy/blob/e1fd9e7fb07ab60bb88176d14c1b5352f949d428/src/main.js#L243
   - https://github.com/asabino2/streamproxy/blob/e1fd9e7fb07ab60bb88176d14c1b5352f949d428/src/main.js#L405
   - https://github.com/asabino2/streamproxy/blob/e1fd9e7fb07ab60bb88176d14c1b5352f949d428/src/main.js#L412

PR #4639, opened 2022-07-06
release: 4.2.0

bastimeyer:
Preliminary changelog with the release date set for tomorrow, but as usual, nothing is set in stone here.

Please check for any mistakes or if something important is missing.

There are still a few PRs/issues open, so it might make sense waiting a bit more.

----

Regarding the changes of #4568, the `url_master` parameter of `HLSStream` and `MuxedHLSStream` was deprecated in favor of `multivariant`, but no proper deprecation was added to the docs, and only the parameter docstrings were updated. Is that something you think should be added to the doc's deprecations page?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,33 @@
 # Changelog
 
+## streamlink 4.2.0 (2022-07-09)
+
+Release highlights:
+
+- Added: new Windows portable builds ([#4581](https://github.com/streamlink/streamlink/pull/4581))
+- Added: more dependency versions to debug log header ([#4575](https://github.com/streamlink/streamlink/pull/4575))
+- Added: parsed multivariant playlist reference to `HLSStream` and `MuxedHLSStream` ([#4568](https://github.com/streamlink/streamlink/pull/4568))
+- Fixed: unnecessary delay when closing `DASHStream`s ([#4630](https://github.com/streamlink/streamlink/pull/4630))
+- Fixed: `FFmpegMuxer` not closing sub-streams concurrently ([#4634](https://github.com/streamlink/streamlink/pull/4634))
+- Fixed: threading issue when closing `WebsocketClient` connections ([#4608](https://github.com/streamlink/streamlink/pull/4608))
+- Fixed: handling of `PluginError`s when outputting JSON data via `--json` ([#4590](https://github.com/streamlink/streamlink/pull/4590))
+- Fixed: broken YouTube plugin when setting custom authentication headers ([#4576](https://github.com/streamlink/streamlink/pull/4576))
+- Fixed: "source" Twitch VODs not being considered "best" ([#4625](https://github.com/streamlink/streamlink/pull/4625))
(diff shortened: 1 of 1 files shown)


gravyboat:
Changelog looks good. Let's wait on those last few PRs before we cut this though.

> the url_master parameter of HLSStream and MuxedHLSStream was deprecated in favor of multivariant, ... 

I don't think a deprecation notice is really needed is it? We're really just matching newer standards.

gravyboat:
@bastimeyer I think everything that needs to get in should be in, the vimeo update would be nice to have but I don't think that's ready. Anything else you want to get in before I merge this and cut the release?

mkbloke:
Given that the Vimeo issue has been open for a year and half, it probably doesn't matter if it doesn't go into this release, although as you say, it would be nice.

The other thing about that PR is not only the Vimeo plugin itself, but whether or not the changes in `stream.hls` that are required for the Vimeo plugin fix are fully decided upon also.

I will try and get back to it shortly in any case.

gravyboat:
@mkbloke Definitely not a big deal so don't inconvenience yourself, it's just the only PR currently open that will be merged at some point.

bastimeyer:
#4643 should be merged before 4.2.0 gets released.

gravyboat:
@bastimeyer #4643 is in.

bastimeyer:
Why was the "New release" step skipped?
https://github.com/streamlink/streamlink/runs/7260805016

bastimeyer:
That's the run condition:
https://github.com/streamlink/streamlink/blob/3ce369477d0b08472638375e6fbfb4b78e8f5a83/.github/workflows/main.yml#L113

The only recent change was this, but that shouldn't be relevant, as the workflow was already triggered:
https://github.com/streamlink/streamlink/blob/3ce369477d0b08472638375e6fbfb4b78e8f5a83/.github/workflows/main.yml#L5-L6

bastimeyer:
Ok, none of the action workflow runs was triggered by the tag-push.
https://github.com/streamlink/streamlink/actions

Tag `4.2.0` exists though:
https://github.com/streamlink/streamlink/releases/tag/4.2.0

Why is it saying that I tagged it? I didn't...

bastimeyer:
Since I don't know what you did and what happened exactly, @gravyboat, I will delete the tag now and push a new one.

bastimeyer:
So apparently, the `branches-ignore` list is responsible for this:
https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-filters-to-target-specific-branches-or-tags-for-push-events

> If you define only tags/tags-ignore or only branches/branches-ignore, the workflow won't run for events affecting the undefined Git ref. If you define neither tags/tags-ignore or branches/branches-ignore, the workflow will run for events affecting either branches or tags.

gravyboat:
@bastimeyer For future reference when I tag all I do is pull the latest down, then `git tag x.x.xx` then `git push origin x.x.xx`. Nothing fancy.

PR #4656, opened 2022-07-16
cli: rewrite progress output

### LINKED ISSUE #4642: cli: threaded read_stream() for undelayed process termination
[`streamlink_cli.main.read_stream()`](https://github.com/streamlink/streamlink/blame/4.1.0/src/streamlink_cli/main.py#L372-L431) is in need of an update. It's the heart of the `streamlink_cli` package and it's responsible for reading stream data and writing it to the output.

The issue is that reading stream data is single-threaded by reading from a simple `stream.read()` iterator, and the polling of the player process for termination or rather the detection of the end of the write-pipe thus depends on the stream-buffer's thread locks while reading the stream data, which is pretty bad when read calls stall for several seconds due to an empty stream buffer while waiting for new data.

What this means is that closing the player often doesn't get detected immediately, and ending the streamlink process takes an unnecessarily long time, as it's waiting for the next stream data first before it can continue exiting gracefully. This is especially noticable when reading HLS streams with long segment durations and long playlist refresh times.

The download-progress output is also tightly coupled with the current stream iterator and its updates are therefore timely inconsistent, which is bad.

----

What should be done in order to fix this is

1. Wrap the stream reader in a separate thread. This can already be done via `streamlink.stream.wrappers.StreamIOThreadWrapper`, but it'd add a secondary stream buffer, which is probably not ideal. I haven't had a closer look at this yet.
2. Poll the player process for termination in a separate thread as long as the stream reader thread is alive. Abort the stream reader when the player was closed.
3. Move the download progress output into a separate thread, so it can update in consistent time intervals instead of relying on the pace of the stream reader.

I have already experimented with that, and apart from the progress output which would need some code refactoring, streams get closed immediately which previously caused a delay of several seconds. The challenge however is implementing this so that current functionality doesn't break, because there are no tests written for that (the sparse test coverage comes as a side effect from mocked cmdline tests).

bastimeyer:
- Split `streamlink_cli.utils.progress` into `progress` + `terminal`
  - `terminal`:
    Responsible for calculating character output widths and for writing
    data to the output text stream
  - `progress`:
    Responsible for calculating the download speed and for formatting
    the progress output
- Use a separate thread for the progress output, with the output being
  written in constant time intervals, independent of the data that gets
  fed by the stream iterator of the main thread
- Output format changes:
  - Remove "prefix" from output format (full path already gets logged)
  - Avoid showing download speeds until a time threshold is reached
  - Use binary values for file size and download speed values
  - Add leading zeros to time units (only ever grow output text size)
- Don't import from alias module in `streamlink_cli.main`
- Split, move and rewrite tests

----

The download progress on the master branch is currently bound to the main thread and data of the stream iterator in `read_stream()`. As explained in #4642, this has several issues. Rewriting the progress output is one of the requirements for being able to resolve the linked issue.

The interface of the `Progress` class is written with the current `stream_iterator` implementation in `read_stream()` in mind, which will change according to my plans described in #4642, but that shouldn't be a big issue with how the `put()` method is implemented.


bastimeyer:
Added the missing tests.

I decided to not test the threading logic of the `Progress` thread, as it requires a ridiculous amount of boilerplate code and it also requires adding a workaround for freezegun (in regards to the `threading` module) which then unfortunately interferes with pytest's test runner outputs. Testing the `update()` logic is good enough.

src/streamlink_cli/utils/progress.py
```diff
@@ -1,153 +1,161 @@
...
+
+
+class Progress(Thread):
+    def __init__(
+        self,
+        output: Optional[TerminalOutput] = None,
+        formatter: Optional[ProgressFormatter] = None,
+        interval: float = 0.25,
+        history: int = 20,
+        threshold: int = 2,
```
bastimeyer, line 80: These values could be tweaked.

Four updates per seconds seems reasonable, and so is keeping a download speed history (the result is the average in this time frame) of twenty seconds which ensures that segmented streams with large segment durations (but still less than 20 seconds) don't make the download speed go up and down in between segment downloads and segment queue wait times.

gravyboat: Let's see if anyone has feedback after the next release. If so they can be modified but these seem reasonable to me.

gravyboat:
Looks great @bastimeyer!

PR #4660, opened 2022-07-17
stream.ffmpegmux: cache executable lookups

bastimeyer:
- Remove redundant commands with `.exe` suffixes.
  `shutil.which` will pick those up via `PATHEXT` on Windows.
- Move executable lookup to `FFMPEGMuxer.resolve_command`, so its inputs
  and return values can be cached using `functools.lru_cache` without
  relying on a Streamlink session instance as input
- Keep interface of `FFMPEGMuxer.is_usable` and `FFMPEGMuxer.command`
- Add proper tests

----

`FFMPEGMuxer.is_usable` currently doesn't get cached, which means that when generating a list of `MuxedHLSStream`s for example, `ffmpeg` will get looked up every single time, and additionally when a the stream gets opened.
- https://github.com/streamlink/streamlink/blob/4.2.0/src/streamlink/stream/hls.py#L748
- https://github.com/streamlink/streamlink/blob/4.2.0/src/streamlink/stream/ffmpegmux.py#L119

I didn't remove `avconv` yet, even if it's totally useless nowadays (the libav fork of ffmpeg was abandoned in 2018). There might still be someone though using an old Debian or CentOS release which defaults to libav.

### DIFF
--- a/src/streamlink/stream/ffmpegmux.py
+++ b/src/streamlink/stream/ffmpegmux.py
@@ -3,7 +3,9 @@ import logging
 import subprocess
 import sys
 import threading
+from functools import lru_cache
 from shutil import which
+from typing import Optional
 
 from streamlink import StreamError
 from streamlink.compat import devnull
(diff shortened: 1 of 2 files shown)


bastimeyer:
I'll submit another PR which will update the other ffmpegmuxer tests once this one got merged, as it's unrelated to the changes here.

bastimeyer:
Any reviews? The changes are fairly simple.

gravyboat:
Sorry @bastimeyer. I actually looked at this 3 days ago but guess I closed the tab instead of merging.

PR #4664, opened 2022-07-18
plugins.nbcnews: Plugin fix

Grabien:
The stream.nbcnews.com API host used in the plugin recently stopped working. It fixes of plugin for the new API endpoint.

### DIFF
--- a/src/streamlink/plugins/nbcnews.py
+++ b/src/streamlink/plugins/nbcnews.py
@@ -26,14 +26,14 @@ class NBCNews(Plugin):
     token_url = 'https://tokens.playmakerservices.com/'
 
     api_schema = validate.Schema(
-        validate.transform(parse_json), 
+        validate.transform(parse_json),
         {
             'videoSources': [{
                 'cdnSources': {
                     'primary': [{
                         'sourceUrl': validate.url(),
                     }]
-                }                
+                }
             }],
         },
         validate.get('videoSources'),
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/nbcnews.py
```diff
@@ -9,7 +9,8 @@
 
 from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
-from streamlink.stream.hls import HLSStream
+from streamlink.stream import HLSStream
```
bastimeyer, line 12: Don't change the HLSStream import

src/streamlink/plugins/nbcnews.py
```diff
@@ -9,7 +9,8 @@
...
+from streamlink.utils import parse_json
```
bastimeyer, line 13: Don't import this. The `validate.parse_json()` method should be used.

src/streamlink/plugins/nbcnews.py
```diff
@@ -21,23 +22,30 @@ class NBCNews(Plugin):
...
+                    }]
+                }
             }],
         },
         validate.get('videoSources'),
         validate.get(0),
+        validate.get('cdnSources'),
+        validate.get('primary'),
+        validate.get(0),
+        validate.get('sourceUrl'),
```
bastimeyer, line 44: `validate.get` can receive a tuple of strings which will call `__getitem__` on each successive object, which is much better than calling `validate.get` lots of times.

bastimeyer (changes requested):
Thanks for the PR. The rest is looking fine.

The plugin could use a rewrite/cleanup though, but that's not important right now... (parsing HTML and querying XPath instead of matching a regex, moving schemas away from class attribute definitions, replacing single-quote strings, using `self.id` instead of `video_id`, etc)

PR #4669, opened 2022-07-20
stream.ffmpegmux: rewrite tests

bastimeyer:
Follow-up of #4660 

This replaces the very repetitive tests of the ffmpegmuxer with parametrized tests, as the setup is exactly the same everywhere. Also some minor adjustments of the tests added in #4660, as the test fixtures could be improved and shared between the newly added tests.

Second commit adds tests for `--ffmpeg-verbose` and `--ffmpeg-verbose-path`. That's mainly just a coverage improvement though.


gravyboat:
This is a big improvement, nice job!

PR #4680, opened 2022-07-27
refactor: is_darwin/is_win32 compat exports

bastimeyer:
- Move `is_darwin` from `streamlink_cli.compat` to `streamlink.compat`
- Remove `is_win32` from `streamlink_cli.compat`
- Replace last remaining `unittest.skipIf` test decorators with the
  `posix_only`/`windows_only` pytest decorators from the `tests` module
- Remove `is_win32` check when importing from `ctypes` in `named_pipe`

### DIFF
--- a/src/streamlink/compat.py
+++ b/src/streamlink/compat.py
@@ -1,5 +1,8 @@
 import os
+import sys
 
+
+is_darwin = sys.platform == "darwin"
 is_win32 = os.name == "nt"
 
 # win/nix compatible devnull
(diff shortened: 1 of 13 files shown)


bastimeyer:
The main intention for this change are the test decorators, but I also have an eye on rewriting & improving the `streamlink.cache` module, and part of it is the incorrect cache-file path on macOS, so having `is_darwin` defined in `streamlink_cli` doesn't make sense:
https://github.com/streamlink/streamlink/blob/d92658f633f69eb4e85bb8bd2831751dc5fdce7b/src/streamlink/cache.py#L9-L14

PR #4681, opened 2022-07-28
docs: remove dead packages and add package availability notes

### LINKED ISSUE #4670: Remove dead packages from install docs
The install docs currently include a couple of packages which haven't been updated in a very long time.
https://streamlink.github.io/install.html

These packages are

- **Ubuntu nilarimogard/webupd8 PPA**
  https://launchpad.net/~nilarimogard/+archive/ubuntu/webupd8/+packages?field.name_filter=streamlink&field.status_filter=published&field.series_filter=
  Latest updates:
  - 2.3.0 (2021-08-12) - Ubuntu 21.04 Hirsuite Hippo (non-LTS) only
  - 2.1.0 (2021-06-01) - Ubuntu 20.10 Groovy Gorilla (non-LTS) only
  - 1.7.0 (2020-10-20) - Ubuntu 20.04 Focal Fossa (LTS) and Ubuntu 18.04 Bionic Beaver (LTS)
- **OpenBSD**
  https://openports.se/multimedia/streamlink
  Latest Updates:
  - 2.0.0 (2021-02-14)
  - 1.7.0 (2020-10-29)

I think it's wrong listing those packages, as it's confusing for new users, especially Ubuntu users. Not to mention that suggesting to set up a third party PPA is a terrible thing to do.

The default Ubuntu repos inherit the `streamlink` + `python3-streamlink` packages from Debian, but those are always outdated due to the nature of stable point-releases, which makes me not want to include those instead (hence the reason why the PPA was added in the past).
- 18.04: https://packages.ubuntu.com/bionic/streamlink (0.10.0)
- 20.04: https://packages.ubuntu.com/focal/streamlink (1.3.1)
- 21.10: https://packages.ubuntu.com/impish/streamlink (2.0.0)
- 22.04: https://packages.ubuntu.com/jammy/streamlink (3.1.1)
- 22.10: https://packages.ubuntu.com/kinetic/streamlink (4.2.0)

**So, is there any reason to keep those packages listed in the install docs?**

bastimeyer:
Closes #4670 

Since nobody has had anything to comment about my suggested removal of dead packages from the install docs, I'm just going to submit a PR.

Please see the rendered docs fist. I've tried to be as precise and short as possible in the added notes. I don't mind rewriting/improving it though.

The removals are done in separate commits, in case we want to revert this later on, should the package maintainers start maintaining the packages again.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -180,6 +180,7 @@ openSUSE                             Simon Puchert <simonpuchert at alice.de>
 Solus                                Joey Riches <josephriches at gmail.com>
 Void                                 Michal Vasilek <michal at vasilek.cz>
 Windows binaries                     Sebastian Meyer <mail at bastimeyer.de>
+Linux AppImages                      Sebastian Meyer <mail at bastimeyer.de>
 ==================================== ===========================================
 
 
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -140,9 +140,6 @@ Python pip                           See the `PyPI package and source code`_ sec
                                         sudo xbps-install streamlink
 ==================================== ===========================================
 
-Please see the `PyPI package and source code`_ or `AppImages`_ sections down below
-if a package is not available for your distro or platform, or if it's out of date.
-
 .. _Arch Linux: https://www.archlinux.org/packages/community/any/streamlink/
 .. _Arch Linux (aur, git): https://aur.archlinux.org/packages/streamlink-git/
 .. _Debian (sid, testing): https://packages.debian.org/unstable/streamlink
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -188,15 +173,28 @@ Fedora                               Mohamed El Morabity <melmorabity at fedorap
...
+
+Packaging is not done by the Streamlink maintainers themselves except for
+the `PyPI package <PyPI package and source code_>`_,
+the `Windows installers + portable builds <Windows binaries_>`_,
+and the `Linux AppImages <AppImages_>`_.
+
+If a package is not available for your operating system / distro or your system's architecture, or if it's out of date
+or broken, then please contact the respective package maintainers or package-repository maintainers.
+
+Please open an issue or pull request on GitHub if an available and maintained package is missing from the install docs.
```
gravyboat, line 195: Do we even want to say this? I feel like people will interpret this incorrectly and open issues because packages are missing any way. Can we make the assumption that package maintainers know what to do?

gravyboat (commented):
Sorry for not commenting on the previous issue @bastimeyer, I don't even remember reading it so I must have missed it. I think the removal plan is good. Some small feedback about the wording in the package availability section.

docs/install.rst
bastimeyer: Package maintainers usually don't know or don't care about external documentation. What they care about is the availability of software in their package repos.

I've reworded the section a bit and put more emphasis on when to open issues / PRs for unlisted packages.

gravyboat:
Looks great to me!

PR #4691, opened 2022-07-31
plugin.api.validate: ListSchema, re.Pattern and NoneOrAllSchema

bastimeyer:
This adds new schema validation methods that are useful for commonly used patterns.

### ListSchema

The list instance validation schema (eg. `[int, str]`) only validates a subset of the input, which means it defines any items that can match. This is often not what plugin authors have in mind when trying to validate a list where each indexed item needs to be validated.

The `ListSchema` (via `validate.list(...)`) on the other hand therefore checks for equal length of the input list and then validates each item according to the schema.

### re.Pattern

Instead of having to define the `validate.transform(re.compile(...).search)` validation pattern, a simple `re.compile(...)` is now supported, which will call the `search()` method internally and return either `None` or an instance of `re.Match`. Other methods like `findall()` for example can still be called via `validate.transform()`.

### NoneOrAllSchema

This avoids having to repeatedly define the `validate.any(None, validate.all(...))` validation pattern via the newly added `validate.none_or_all(...)` schema. This is especially useful for `re.Pattern` or `validate.xml_xpath_string(...)` validations.

I am not really happy with the naming of this though. Any better suggestions than `none_or_all`?

### DIFF
--- a/src/streamlink/plugin/api/validate/__init__.py
+++ b/src/streamlink/plugin/api/validate/__init__.py
@@ -4,6 +4,7 @@ from streamlink.plugin.api.validate._schemas import (  # noqa: I101, F401
     SchemaContainer,
     AllSchema as all,
     AnySchema as any,
+    ListSchema as list,
     TransformSchema as transform,
     OptionalSchema as optional,
     GetItemSchema as get,
--- a/src/streamlink/plugin/api/validate/__init__.py
+++ b/src/streamlink/plugin/api/validate/__init__.py
@@ -4,6 +4,7 @@ from streamlink.plugin.api.validate._schemas import (  # noqa: I101, F401
     SchemaContainer,
     AllSchema as all,
     AnySchema as any,
+    NoneOrAllSchema as none_or_all,
     ListSchema as list,
     TransformSchema as transform,
     OptionalSchema as optional,
--- a/src/streamlink/plugin/api/validate/_schemas.py
+++ b/src/streamlink/plugin/api/validate/_schemas.py
@@ -29,6 +29,13 @@ class AnySchema(_CollectionSchemaContainer):
     """
 
 
+class NoneOrAllSchema(_CollectionSchemaContainer):
+    """
+    Collection of schemas where every schema must be valid. If the initial input is None, all validations will be skipped.
+    The last validation result gets returned.
+    """
+
+
 class ListSchema(_CollectionSchemaContainer):
     """
     Collection of schemas where every indexed schema must be valid, as well as the input type and length.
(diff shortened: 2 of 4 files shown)


bastimeyer:
The lintint failure is caused by a new major version of flake8 with unrelated linting errors. Will take a look at this later and then rebase here once fixed.

gravyboat:
These changes look really good @bastimeyer, lowering the complexity is always appreciated. Thank you!

> I am not really happy with the naming of this though. Any better suggestions than none_or_all?

Based on what the code does and the description you've added doc wise, what about changing `AllSchema` to `AllValidSchema`, and then `NoneOrAllSchema` to `NoneOrAllValidSchem`. I don't know, these don't seem like major improvements over what you have, but they do clarify a bit more without having to look at the code/docs what is actually happening. The only other thing I can think of is for `NoneOrAllSchema` it could potentially be `NoneIsSkippedAllValidSchema` but that's a mouthful.

bastimeyer:
I was talking about renaming the `none_or_all` export as part of the public interface, not any internal class names. Class names are only relevant for the ValidationError messages, and the existing ones are all fine and shouldn't be changed. Changing `none_or_all` obviously requires changing the `NoneOrAllSchema` class name as well.

I was thinking about maybe renaming it to `optional_all`, but that's not as specific/clear, as it could suggest that validations could optionally fail, which they can not.

`all_or_nothing` sounds stupid.

gravyboat:
> I was thinking about maybe renaming it to optional_all, but that's not as specific/clear, as it could suggest that validations could optionally fail, which they can not.

Yeah I don't think this is is very helpful. I don't think it's worth stressing about too much, I can't think of much better than `none_or_all` and you're right that `all_or_nothing` sounds like we're in some sort of anime showdown.

bastimeyer:
Let's keep the `validate.none_or_all(...)` name. If nobody has any further comments here, then this should be good.

There's one minor concern I have with the `ListSchema`, but I don't think it's important. Originally, I had implemented this as a `SequenceSchema`, which would also take tuples, sets and frozensets as inputs when validating, but then I figured that it didn't make much sense considering that usually JSON data is being validated, so these other data structures are useless. And there was the `tuple() is tuple()` issue that came up when writing tests, as the schema is supposed to create new instances of the sequences, so I changed my mind.

Btw, I've already modified, fixed and rewritten 52 plugins based on these new validation changes the last couple of days. Not sure though how I will submit these changes, as there are a couple of complete rewrites that shouldn't be included in commits of basic/simple validation pattern changes.

gravyboat:
> Btw, I've already modified, fixed and rewritten 52 plugins based on these new validation changes the last couple of days. Not sure though how I will submit these changes, as there are a couple of complete rewrites that shouldn't be included in commits of basic/simple validation pattern changes.

What about only breaking out the full rewrites in to separate PRs, and all the small fixes in to one PR? Seems like a hassle to have a PR for every small fix just because it's a different plugin.

bastimeyer:
That's what I was planning to do.

bastimeyer:
Any further comments? If not, then please merge, so I can submit (some of) the plugin changes I was talking about.

PR #4702, opened 2022-08-04
plugins: refactor validation schemas

bastimeyer:
Follow-up of #4691 

These changes only affect plugins where no major changes were needed for fixing and cleaning them up.

Split into two commits. Please see individual diffs and commit messages:

1. Simple validation schema code refactorings
2. Simple validation schema code refactorings where schemas were defined as class attributes

I haven't checked any of these plugins (yet). This is supposed to be a code refactoring, not a plugin fix. However, since I changed some regexes, it might be worth checking some plugins first though before merging, but should there be any issues, it's most likely due to the plugin code being outdated and not because of my changes. In that case, broken plugins need to be fixed in a new PR.

And as mentioned in #4691, plugins which required more changes to the validation schemas or the remaining implementation will be submitted in separate PRs later. Those for example will include changes from regexes to XPath queries, or other rewrites.

----

Btw, after reviewing the diff of this PR before submitting it, I realized that `validate.all(re.Pattern, re.Match)` is probably not ideal in terms of the error messages that get shown when the regex doesn't match and returns `None` (so that the `re.Match` type doesn't validate). A better solution would be adding a `RegexSchema` where the regex has to match and where better error messages can be shown. However, that can be implemented and changed in the plugins later on.

### DIFF
--- a/src/streamlink/plugins/albavision.py
+++ b/src/streamlink/plugins/albavision.py
@@ -90,46 +90,40 @@ class Albavision(Plugin):
         return is_token_based_site
 
     def _get_live_url(self):
-        live_url_re = re.compile(r"""LIVE_URL\s*=\s*['"]([^'"]+)['"]""")
         schema = validate.Schema(
             validate.xml_xpath_string(".//script[contains(text(), 'LIVE_URL')]/text()"),
-            validate.any(None, validate.all(
-                validate.transform(live_url_re.search),
-                validate.any(None, validate.all(
+            validate.none_or_all(
+                re.compile(r"""LIVE_URL\s*=\s*(?P<q>['"])(.+?)(?P=q)"""),
+                validate.none_or_all(
                     validate.get(1),
                     validate.url(),
-                )),
-            )),
+                ),
+            ),
         )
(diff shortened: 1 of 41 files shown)


gravyboat:
Both commits look good to me. I don't think it's worth checking anything but the most highly used plugins. If anyone complains about them not working they can be fixed at that point.

> ...However, that can be implemented and changed in the plugins later on.
👍

bastimeyer:
Slight update: reverting the changes of the zdf_mediathek plugin.
I will submit a separate PR for this which replaces the schema's  regex with an XPath query.

bastimeyer:
Don't merge this yet. There's another issues with the regexes. It's expecting instances of `str` or `bytes`, but `validate.xml_xpath_string` returns `_ElementUnicodeResult`. FIxing this in `xml_xpath_string` probably makes the most sense.

bastimeyer:
Once #4708 and #4709 have been merged, I will rebase this branch.

I will also change the couple of `re.Pattern, re.Match` validations to `validate.regex(re.Pattern)`, so that better error messages can be shown, as described in the OP, so that no additional PR is needed.

PR #4707, opened 2022-08-05
plugins.huajiao: rewrite and fix plugin

bastimeyer:
Not sure how relevant this plugin actually is. It's been using an old and long-time deprecated base64 decoding API which was removed from the standard library in 3.9, so everyone trying to open streams using those python versions would see an error, and there haven't been any plugin issue reports.

https://docs.python.org/3/whatsnew/3.9.html
> base64.encodestring() and base64.decodestring(), aliases deprecated since Python 3.1, have been removed

----

Apparently, there are also HLS URLs in the "feed" data, but I haven't seen any of those. The website does use HTTP video streams, and so does the plugin when trying to access a stream. I kept it though, just in case. And I also upgraded the plugin to HTTPS and removed the verify=false option on the session.

### DIFF
--- a/src/streamlink/plugins/huajiao.py
+++ b/src/streamlink/plugins/huajiao.py
@@ -1,11 +1,10 @@
 """
-$description Chinese live streaming platform for live video game broadcasts and individual live streams.
+$description Chinese live-streaming platform for live video game broadcasts and individual live streams.
 $url huajiao.com
 $type live
 """
 
 import base64
-import json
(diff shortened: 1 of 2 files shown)


gravyboat:
What do you think about merging this then removing the plugin since it's broken and no one has complained? Could always revert later to add it back in.

bastimeyer:
It's working now. No need to remove it.

PR #4712, opened 2022-08-06
plugins.ltv_lsm_lv: fix plugin

bastimeyer:
The plugin was broken due to site changes.

- moved and updated validation schemas
- simplified channel ID extraction
- reformatted code


gravyboat:
@bastimeyer When you have a few minutes can you rebase and resolve the merge conflict please?

bastimeyer:
I accidentally included some old changes in #4702 which were not supposed to be part of that PR, so the merge conflicts occurred here. Those are fixed now, and I also added a check for offline streams. The plugin on the master branch not only is outdated, but the changes which were not supposed to be in #4702 also break the plugin in another way (forgot to rename the API URL class attribute). Also fixed now.

PR #4734, opened 2022-08-14
release: 4.3.0

bastimeyer:
Anything missing? Any unsubmitted changes?

Release date is set for today, but I will update each day at midnight UTC if it's not ready yet, as usual.

Unmerged PRs:
- #4733

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 # Changelog
 
+## streamlink 4.3.0 (2022-08-15)
+
+Release highlights:
+
+- Improved: CLI download progress output ([#4656](https://github.com/streamlink/streamlink/pull/4656))
+- Fixed: consecutive FFmpeg executable lookups not being cached ([#4660](https://github.com/streamlink/streamlink/pull/4660))
+- Fixed: `--ffmpeg-verbose-path` not expanding `~` to the user's home directory ([#4688](https://github.com/streamlink/streamlink/pull/4688))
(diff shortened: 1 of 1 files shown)


gravyboat:
@bastimeyer The only thing that's worth discussing prior to merge and release is the issue @scowalt brought up here: https://github.com/streamlink/streamlink-chocolatey/pull/72#pullrequestreview-1072016449 about the way the windows releases are tagged that creates an inconsistency between the installer and the chocolatey package.

bastimeyer:
Changelog updated and release date changed to tomorrow.

> the way the windows releases are tagged

The Windows releases are tagged correctly, like any other packaging version scheme, namely `major.minor.patch-pkgversion`. If chocolatey doesn't support this, then it's not our fault.

gravyboat:
@bastimeyer Yes I agree and so does Scowalt that the versioning used is appropriate and that the other chocolatey packaging utilities are in the wrong, just wanted to bring it up 👍

gravyboat:
@bastimeyer If there isn't anything blocking let's do the release today so maybe swap back the date? I was planning on cutting the new tag as soon as we resolved the above concern.

bastimeyer:
> let's do the release today

I would prefer having it tagged tomorrow, because I will need to update the Windows builds and AppImages afterwards, and I don't want to do this now because I'm not at my desktop computer right now and don't have everything set up here. There also haven't been any comments from anyone else, in case something's missing.

gravyboat:
Okay that's fine by me. Do you want me to try and cut the release in the morning tomorrow (PST)? I can do so before work if that helps.

bastimeyer:
Sure

PR #4739, opened 2022-08-16
cli: implement --player-external-http-continuous

### LINKED ISSUE #4738: cli: option for running --player-external-http in "non-continuous" mode
### Checklist

### Description

I frequently run streamlink with the `player-external-http` flags within a tmux session to have the ability to swap between devices. This works quite well, but I'll occasionally encounter errors where streamlink is still running idle off in a session somewhere long after the target stream has wrapped up. It would be nice if we could be able to customize behavior of the app when a target stream ends (or is down for X amount of time to allow for unintended disconnects on the source stream's end). 

The current behavior is largely fine for desktop use, but the ability to get more granular control would be appreciated for headless use cases.

bastimeyer:
Resolves #4738 

No tests, because [no tests have ever been written for `output_stream_http()`](https://app.codecov.io/gh/streamlink/streamlink/blob/master/src/streamlink_cli/main.py) or the entire [`streamlink_cli.utils.http_server` module](https://app.codecov.io/gh/streamlink/streamlink/blob/master/src/streamlink_cli/utils/http_server.py).
- https://github.com/streamlink/streamlink/blame/4.3.0/src/streamlink_cli/main.py#L164-L274
- https://github.com/streamlink/streamlink/blame/4.3.0/src/streamlink_cli/utils/http_server.py

----

There are two choices that can be made for implementing the new CLI argument:
1. The current implementation, with `--player-external-http-continuous=bool` requiring a boolean value and affecting the logic of the main argument `--player-external-http`.
   The issue with this is that it's not consistent with the logic of the `--player-http` and `--player-continuous-http` arguments, which are individual arguments. It's also not consistent with the naming scheme (which I just realized), so it probably has to be changed to `--player-external-continuous-http`.
2. Having two individual CLI arguments, `--player-external-http` and `--player-external-http-non-continuous`, similar to `--player-http` and `--player-continuous-http`. As you can see, the default logic of `--player-external-http` already is "continuous", so the second argument name has to be negated, which is not great.

Thoughts?

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -482,12 +482,31 @@ def build_parser():
         useful to allow external devices like smartphones or streaming boxes to
         watch streams they wouldn't be able to otherwise.
 
-        Behavior will be similar to the continuous HTTP option, but no player
-        program will be started, and the server will listen on all available
+        The default behavior is similar to the --player-continuous-http option,
+        but no player program will be started, and the server will listen on all available
         connections instead of just in the local (loopback) interface.
+
+        Optionally, the --player-external-http-continuous option allows for disabling
+        the continuous run-mode, so that Streamlink will stop when the stream ends.
 
         The URLs that can be used to access the stream will be printed to the
         console, and the server can be interrupted using CTRL-C.
+        """
+    )
+    player.add_argument(
+        "--player-external-http-continuous",
+        type=boolean,
+        metavar="{yes,true,1,on,no,false,0,off}",
+        default=True,
+        help="""
+        Set the run-mode of --player-external-http to continuous or non-continuous.
+
+        In the continuous run-mode, Streamlink will keep running after the stream has ended
+        and will wait for the next HTTP request being made unless it gets shut down via CTRL-C.
+
+        If set to non-continuous, Streamlink will stop once the stream has ended.
+
+        Default is true.
         """
     )
     player.add_argument(
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -197,6 +197,7 @@ def output_stream_http(
     initial_streams: Dict[str, Stream],
     formatter: Formatter,
     external: bool = False,
+    continuous: bool = True,
     port: int = 0,
 ):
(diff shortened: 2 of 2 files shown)


nbgibson:
Of the two paths listed here I'd have to lean towards the former (renamed to be in spec with the rest of the command schema of course), though I'm not too much a fan of either for the reasons listed above. As is I think there is potential for a larger rework of of how the server option tree is handled, but that's beyond the scope of this particular PR.

My reasoning is:

- As is we already have to pass a separate flag for configuring ports when the "better" flow would be to just optionally tack on a desired target port to [```--player-external-http```](https://streamlink.github.io/cli.html?highlight=player+external+http+continuous#cmdoption-player-external-http), so having another flag or two isn't exactly new. It's not ideal, but it's also not exactly a deviation at this time.
- The flow here isn't really a deviation from ```--player-external-http``` imo as that process flow (from my understanding) is still on the "desktop" side of things albiet with a web based serving method. ```--player-external-continuous-http``` and so forth are all off in headless land.

bastimeyer:
> 1. It's also not consistent with the naming scheme (which I just realized), so it probably has to be changed to `--player-external-continuous-http`

Hm, that's not consistent with `--player-external-http-port` then... I think I will just leave it as it is. The parameter naming scheme isn't particularly well thought out anyway (eg. `--verbose-player`), and using natural language for parameter names instead of logical/layered suffixes makes everything an inconsistent mess anyway.

> I think there is potential for a larger rework

Yes. After Livestreamer got forked into Streamlink, the CLI was never cleaned up and it's a total mess of historically grown features built on top of each other over several years by lots of different contributors. That's why nobody has tried refactoring or rebuilding the CLI package, and it's still a mess today. And starting from scratch is even more difficult if feature parity needs to be achieved without introducing (many) breaking changes. One feature in regards to the external http server for example was being able to serve multiple clients at once, but implementing this on top of the current code just doesn't make sense from a maintenance perspective.

> just optionally tack on a desired target port to --player-external-http

I think this has been avoided intentionally when it was implemented, because optional argument values are a bit weird. The argument could be changed to `nargs="?", action="store"`, which would make the value optional.

PR #4747, opened 2022-08-18
plugin: pluginargument decorator

bastimeyer:
This PR changes the plugin argument definitions from

```py
from streamlink.plugin import Plugin, PluginArgument, PluginArguments

@pluginmatcher(...)
class Foo(Plugin):
    arguments = PluginArguments(
        PluginArgument(...),
        PluginArgument(...),
        PluginArgument(...),
    )
```

to

```py
from streamlink.plugin import Plugin, pluginargument, pluginmatcher

@pluginmatcher(...)
@pluginargument(...)
@pluginargument(...)
@pluginargument(...)
class Foo(Plugin):
    ...
```

As mentioned in https://github.com/streamlink/streamlink/issues/4741#issue-1340972848, having plugin arguments definied in a similar declarative way as the URL matchers makes it much more consistent and easier to read. The previous definitions were all rather messy.

This also simplifies the parsing of each plugin module's AST via the plugins extension of the docs and the planned URL-cache pre-build step (#4741).

These changes are fully backwards compatible and custom plugins don't need to change anything. I didn't add tests for the old definition style though, but that's not important.

The docs and man page should stay the same, except for the changes of the help texts and some argument rearrangements.

I also fixed some of the help texts, but didn't do a very careful job here. There's still lots of room for improvement. Since this requires testing plugins in some cases, I couldn't be bothered, so I only fixed the obvious stuff.

As usual, review the commits separately.

### DIFF
--- a/docs/ext_plugins.py
+++ b/docs/ext_plugins.py
@@ -77,7 +77,6 @@ class PluginArguments(ast.NodeVisitor, IDatalistItem):
         super().__init__()
         self.pluginname = pluginname
         self.arguments = []
-        self.found = False
         self.visit(pluginast)
 
     def generate(self) -> Iterator[str]:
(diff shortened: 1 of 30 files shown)


bastimeyer:
- https://deploy-preview-4747--streamlink.netlify.app/cli.html#plugin-options
- https://deploy-preview-4747--streamlink.netlify.app/plugins.html

Issue #4744, opened 2022-08-17
plugin: remove `Plugin.bind` classmethod


bastimeyer:
As mentioned in https://github.com/streamlink/streamlink/issues/4358#issuecomment-1046918030, the current implementation of the `Plugin` and `Session` classes are flawed and cause references of the last initialized `Session` instance to be kept in memory, meaning that Python's garbage collector can never free this part of memory, including every built-in plugin that's held by the `Session`'s `plugins` dict (related to #4741).

The `Plugin.bind` classmethod was implemented in 2e27f21a3eee8bfbc86135993c642853d6a37264 during the Livestreamer days in order to store the `session` as a class attribute on each `Plugin` class and in order to set the plugin's `module` name when it gets loaded. Later on, the `logger` class attribute was added, which was then deprecated after the new logging style was implemented in Streamlink. And there's also the `_user_input_requester` class attribute that was added after the fork.

All of these class attributes are unnecessary except for the `session`:
- The `module` name already exists as a reference in the `Plugin` base class's constructor via `self.__class__.__module__`
  **edit:** It might be necessary to change the output type of `Session.resolve_url` again, and make it return the module name, so it can be used by streamlink_cli for reading plugin config files and resolving plugin arguments. It currently relies of the `module` class attribute, and this needs to be removed.
- The `logger` is unused except for some methods on the `Plugin` base class that could easily be updated
- The `_user_input_requester` should live on the `session` instance, not on the `Plugin` classes

So in order to fix the current implementation, the `session` instance should be passed to each `Plugin`'s constructor as the first argument instead of just the `url` string. The other attributes could then be set by the base class's constructor on the class instances if we care about that. As said, they are all unnecessary and could be changed otherwise.

Changing the `Plugin` class constructor would be a breaking change however. It could temporarily be worked around though by checking the signature of custom plugins and providing a fallback path with a deprecation message that's kept for a bit, similar to the plugin matcher API.

gravyboat:
Sounds like a good plan to me. I don't care much if it's a breaking change. We don't owe anyone non-breaking changes.

PR #4754, opened 2022-08-20
tests: rewrite plugin cookies tests

bastimeyer:
In preparation for #4744

The old tests were in need of a rewrite (similar to the ffmpegmux tests), and updating them for the removal of `Plugin.bind()` doesn't make much sense, so here's a rewrite based on pytest with a much easier setup that can be replaced very easily when `Plugin.bind()` gets removed. I also applied some minor style clean-ups and added typing information...

### DIFF
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -4,6 +4,7 @@ import operator
 import re
 import time
 from functools import partial
+from http.cookiejar import Cookie
 from typing import Any, Callable, ClassVar, Dict, List, Match, NamedTuple, Optional, Pattern, Sequence, Type, Union
 
 import requests.cookies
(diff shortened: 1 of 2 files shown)


bastimeyer:
> **codecov/project/streamlink_cli** Successful in 1s — 69.85% (+1.96%) compared to 2c09253

Where is this coming from? Looks like flaky coverage reports or rather the merging of Linux and Windows reports on the side of codecov which appear to be fixed again:
https://app.codecov.io/gh/streamlink/streamlink/compare/4754/changes

gravyboat:
> Where is this coming from? Looks like flaky coverage reports or rather the merging of Linux and Windows reports...

Agreed, I don't see anything in here that would impact the coverage like this.

The test changes are really good. It's much easier to understand what is going on, thanks Bastimeyer.

PR #4764, opened 2022-08-22
Add file path to progress output

### LINKED ISSUE #4752: No filename displayed in HLS livestream downloading since 4.3.0
### Checklist

### Streamlink version

Latest stable release

### Description

In 4.2.0 and earlier versions, when downloading hls livestream, the destination filename is always being displayed in the latest prompted progress line.

But from 4.3.0, the filename is no longer displayed, only size/duration/speed. Like this:

image

For users who need to download multiple streams in the same time, this change might make it much harder to monitor the progress of different files.

### Debug log

```text
N/A
```

bastimeyer:
Resolves #4752

This refactors the `progress` module, removes the `terminal` module that was added in #4656, adds support for substitutions in format strings with dynamic/variable lengths, and finally adds a new format string that includes the file path. Paths now get truncated in a proper way, and the whole available space is used. Once not enough space is available for the path to be shown in a proper/usable way (at least 15 characters), then it'll fall back to the next format.

Didn't test on Windows. There's a special case when adding space characters to the output, where it removes one character. Since the non-space characters don't get covered by this special case, it's possible that the line will overflow. Need to check this later.

Also didn't test with paths that include CJK characters with widths larger than 1. After the changes of #4656, the whole character width calculation was unused, as there was no user-generated content in the progress output. Now that's obviously changed with the addition of the path, but the `cut()` method should still do its job according to the tests, so I don't think there will be an issue here.


bastimeyer:
This needs more work. As said, I didn't test on Windows before submitting this, so I didn't notice the test failure (the space character I was talking about), but there are other issues as well.

To no surprise, there are reflowing issues on Windows when changing the width of the terminal application(s), because lines don't get cleared properly (not an issue with the code - there's always the carriage return character), but there also appears to be an issue with how the path gets truncated. I will have to check later.

bastimeyer:
Everything should be working as intended now.

If there are any text duplication issues when resizing the terminal application, then it's because of the terminal application not respecting carriage return characters on the line that's getting updated by the progress output. All standard terminals on Windows as well as the one on macOS have this issue, which is quite ridiculous. Konsole for example works beautifully. Not an issue with this implementation or the ones before though.

https://user-images.githubusercontent.com/467294/186149552-01ff1c4a-7fbe-4ac8-925a-cb521cbbcfe6.mp4

gravyboat:
Thanks for the helpful comments, it made reviewing this easier.

Issue #4414, opened 2022-03-31
plugins.okru: Could not find metadata and cannot record stream


akali752:
### Checklist

- [X] This is a plugin issue and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

Recently when I tried to record ok.ru live stream, it always shows "[plugins.okru][error] Could not find metadata" (it was prefectly fine and no issue like 2 months ago). The stream is live and there is playable streams in the URL, and I also can get m3u8 URL by using youtube-dl.

```
streamlink "https://ok.ru/video/3178148273793"
[cli][info] Found matching plugin okru for URL https://ok.ru/video/3178148273793
[plugins.okru][error] Could not find metadata
error: No playable streams found on this URL: https://ok.ru/video/3178148273793
```

### Debug log

```text
streamlink "https://ok.ru/video/3178148273793"  --loglevel debug
[cli][debug] OS:         macOS 10.13.6
[cli][debug] Python:     3.10.2
[cli][debug] Streamlink: 3.2.0
[cli][debug] Requests(2.27.1), Socks(1.7.1), Websocket(1.3.1)
[cli][debug] Arguments:
[cli][debug]  url=https://ok.ru/video/3178148273793
[cli][debug]  --loglevel=debug
[cli][info] Found matching plugin okru for URL https://ok.ru/video/3178148273793
[plugins.okru][error] Could not find metadata
error: No playable streams found on this URL: https://ok.ru/video/3178148273793
```

bastimeyer:
> [plugins.okru][error] Could not find metadata

I can't reproduce this.

The site doesn't appear to have changed anything and Streamlink's okru plugin does successfully find the matadata required for getting the HLS stream URL, as you can see here:

```
$ streamlink -l debug 'https://ok.ru/video/3178148273793'
[cli][debug] OS:         Linux-5.17.1-1-git-x86_64-with-glibc2.35
[cli][debug] Python:     3.10.4
[cli][debug] Streamlink: 3.2.0+9.gfeec5f4
[...]

PR #4765, opened 2022-08-22
plugins.huya: Trim whitespace at the beginning of the HTML response

ohaiibuzzle:
This patch fixes an issue where, at random, some stream (eg. https://huya.com/25886853) would fail with

```
[cli][info] Found matching plugin huya for URL https://huya.com/25886853
error: Unable to validate response text: ValidationError(Callable):
  iselement(None) is not true
```

It appears that the issue is caused by `lxml.etree.HTMLParser`, which in some cases would parse the empty line and whitespaces before `<!DOCTYPE html>` and incorrectly assumes that the page is empty, then returns an empty result, causing subsequent validators to fail

By simply running `lstrip` on the HTML output to get rid of them, lxml parses the content correctly.

```sh
>>> data[0:70]
'\n        <!DOCTYPE html>\n        <html lang="zh-CN">\n            <head'
>>> parser = HTMLParser()
>>> tree = HTML(data, parser)
>>> tree
>>> parser.error_log
<string>:2:1:ERROR:HTML:ERR_DOCUMENT_EMPTY: Document is empty
>>> data = data.lstrip()
>>> data[0:70]
'<!DOCTYPE html>\n        <html lang="zh-CN">\n            <head>\n       '
>>> tree = HTML(data, parser)
>>> tree
<Element html at 0x10467c1c0>
```

### DIFF
--- a/src/streamlink/plugins/huya.py
+++ b/src/streamlink/plugins/huya.py
@@ -33,6 +33,7 @@ class Huya(Plugin):
 
     def _get_streams(self):
         data = self.session.http.get(self.url, schema=validate.Schema(
+            validate.transform(str.lstrip),
             validate.parse_html(),
             validate.xml_xpath_string(".//script[contains(text(),'var hyPlayerConfig = {')][1]/text()"),
             validate.none_or_all(


bastimeyer:
> It appears that the issue is caused by `lxml.etree.HTMLParser`, which in some cases would parse the empty line and whitespaces before `<!DOCTYPE html>` and incorrectly assumes that the page is empty, then returns an empty result, causing subsequent validators to fail

That is not how `lxml`'s `HTMLParser` works. The parser ignores leading whitespace and properly parses the HTML document if it's fed all the data. Streamlink also doesn't import `HTMLParser` directly, but `HTML`, which is a wrapper around it.

If there was an issue with the parser, then this doesn't belong into the huya plugin, let alone Streamlink, and it would need to be reported on the lxml issue tracker.

```py
>>> from lxml.etree import HTML, HTMLParser
>>> HTML("\n     \n" * 1000 + "<!doctype html><html><head><title>foo").find(".//title").text == "foo"
True
>>> p = HTMLParser(); p.feed("\n     \n" * 1000 + "<!doctype html><html><head><title>foo"); p.close().find(".//title").text == "foo"
True

>>> from streamlink.utils.parse import parse_html
>>> parse_html("\n     \n" * 1000 + "<!doctype html><html><head><title>foo").find(".//title").text == "foo"
True
```

> at random

This looks like a different issue, maybe an incomplete HTTP response, leading to partial data.

ohaiibuzzle:
> This looks like a different issue, maybe an incomplete HTTP response, leading to partial data.

At least to me this does not seem like so, I did dump the HTML response and it appears correct

> at random

What I mean is that some random stream pages (one of them being the example above) would cause this kind of behavior. On those page, this is consistently the case like this

```sh
>>> from lxml.etree import HTML, HTMLParser
>>> import requests
>>> data = requests.get("https://huya.com/25886853").text
>>> data[0:70]
'\n        <!DOCTYPE html>\n        <html lang="zh-CN">\n            <head'
>>> parser = HTMLParser()
>>> HTML(data, parser)
>>> parser.error_log
<string>:2:1:ERROR:HTML:ERR_DOCUMENT_EMPTY: Document is empty
>>> HTML(data.lstrip(), parser)
<Element html at 0x103559e00>
```

Interestingly after further testing, I found out that if you leave the data as a bytestream (which streamlink isn't doing here, if I'm not being silly, by the point we're parsing the HTML it's already `str`) it parses correctly

```sh
>>> from lxml.etree import HTML, HTMLParser
>>> import requests
>>> data = requests.get("https://huya.com/25886853").content
>>> data[0:70]
b'\n        <!DOCTYPE html>\n        <html lang="zh-CN">\n            <head'
>>> parser = HTMLParser()
>>> HTML(data.lstrip(), parser)
<Element html at 0x103591580>
>>> parser.error_log
                                   # blank
>>> 
```

ohaiibuzzle:
> ```python
> >>> from lxml.etree import HTML, HTMLParser, LXML_VERSION
> >>> import requests
> >>> LXML_VERSION
> (4, 9, 0, 0)
> >>> data = requests.get("https://huya.com/25886853").text
> >>> type(data)
> <class 'str'>
> >>> data[0:70]
> '\n        <!DOCTYPE html>\n        <html lang="zh-CN">\n            <head'
> >>> parser = HTMLParser()
> >>> HTML(data, parser)
> <Element html at 0x7ff48e06dec0>
> ```

Here is the exact same thing on my machine
```py
>>> from lxml.etree import HTML, HTMLParser, LXML_VERSION
>>> import requests
>>> LXML_VERSION
(4, 9, 1, 0)
>>> data = requests.get("https://huya.com/25886853").text
>>> type(data)
<class 'str'>
>>> data[0:70]
'\n        <!DOCTYPE html>\n        <html lang="zh-CN">\n            <head'
>>> parser = HTMLParser()
>>> HTML(data, parser)
>>> parser.error_log
<string>:2:1:ERROR:HTML:ERR_DOCUMENT_EMPTY: Document is empty
>>> 
```

I will see if I can get some other machines to reproduce this issue to rule out any configuration being variables.

ohaiibuzzle:
Some (limited) testing shows that this behavior seems specific to macOS 12.5.1, 12.5 doesn't exert the same behavior, so does Windows. Probably something with the C compilers which is used to set `lxml` up, so I guess this should go upstream

image

bastimeyer:
> macOS

Take a look at https://github.com/streamlink/streamlink/issues/4414#issuecomment-1086620667

You can't use your system's `libxml2` version when building `lxml` from source. You need to install a more recent version of `libxml2` via homebrew (or any other method / package manager) and build `lxml` using that.

----

lxml built using the system's libxml2 version 2.9.4:
```py
>>> import lxml.etree
>>> lxml.etree.LIBXML_VERSION
(2, 9, 4)
>>> import requests
>>> data = requests.get("https://huya.com/25886853").text
>>> lxml.etree.HTML(data)
>>>
```

building lxml using homebrew's libxml2 2.10.0:
```
$ uname -sr
Darwin 21.4.0
$ python -V
Python 3.9.13
$ brew info libxml2 | head -n1
==> libxml2: stable 2.10.0 (bottled), HEAD [keg-only]
$ LDFLAGS="-L/usr/local/opt/libxml2/lib" \
  CPPFLAGS="-I/usr/local/opt/libxml2/include" \
  PKG_CONFIG_PATH="/usr/local/opt/libxml2/lib/pkgconfig" \
  pip install -U --no-binary=:all: lxml==4.9.1

```

lxml built using homebrew's libxml2 2.10.0:
```py
>>> import lxml.etree
>>> lxml.etree.LIBXML_VERSION
(2, 10, 0)
>>> import requests
>>> data = requests.get("https://huya.com/25886853").text
>>> lxml.etree.HTML(data)
<Element html at 0x1084e2f00>
>>>
```

PR #4768, opened 2022-08-25
plugin: remove `Plugin.bind()`, change `Plugin.__init__()` and `Session.resolve_url()`

### LINKED ISSUE #4358: plugin: session memory leak

### LINKED ISSUE #4744: plugin: remove `Plugin.bind` classmethod

bastimeyer:
Resolves #4744 
Fixes #4358

This introduces breaking changes to the `Plugin.__init__()` method, as well as `Session.resolve_url()` and `Session.resolve_url_no_redirect()`.

**See the commit messages of each individual commit for proper descriptions and explanations.**

Current `master`:
```py
>>> import streamlink
>>> import gc
>>> streamlink.__version__
'4.3.0+18.g60e87243'
>>> s = streamlink.Streamlink()
>>> del s
>>> gc.collect()
0
>>> [obj for obj in gc.get_objects() if isinstance(obj, streamlink.Streamlink)]
[<streamlink.session.Streamlink object at 0x7f59380a36d0>]
```

Removal of `Plugin.bind()`:
```py
>>> import streamlink
>>> import gc
>>> streamlink.__version__
'4.3.0+21.gf34a33cc'
>>> s = streamlink.Streamlink()
>>> del s
>>> gc.collect()
3835
>>> [obj for obj in gc.get_objects() if isinstance(obj, streamlink.Streamlink)]
[]
```

----

I've decided to add compatibility wrappers for old-style `Plugin` constructors, so that breaking changes for custom plugin implementations can be avoided for a short while and deprecation messages be shown.

~~I still need to update the docs (which I just remembered while writing this). But even then,~~ I want to keep this PR open for a bit, so I can think about all this and check if something's wrong or missing.

### DIFF
--- a/docs/deprecations.rst
+++ b/docs/deprecations.rst
@@ -1,6 +1,33 @@
 Deprecations
 ============
 
+streamlink 5.0.0
+----------------
+
+Deprecation of Plugin.__init__(self, url)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+With the removal of the ``Plugin.bind()`` class method which was used for setting up the
+:py:class:`Streamlink <streamlink.session.Streamlink>` session instance and module name in each plugin class,
+the :py:class:`Plugin <streamlink.plugin.Plugin>` constructor's signature was changed and it now requires
+the ``session`` and ``url`` arguments. Implementors of custom plugins should define variable positional arguments and keyword
+arguments when subclassing and adding a custom constructor (``*args, **kwargs``), and the ``url`` should be accessed via
+``self.url`` after calling the constructor of the super class.
+
+Compatibility wrappers were added for old custom plugin implementations, and a deprecation message will be shown until
+the compatibility wrappers will get removed in a future release.
+
+
+Session.resolve_url() return type changes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Due to the changes of the :py:class:`Plugin <streamlink.plugin.Plugin>` class mentioned above, the return value of
+:py:meth:`Streamlink.resolve_url <streamlink.Streamlink.resolve_url>` and
+:py:meth:`Streamlink.resolve_url_no_redirect <streamlink.Streamlink.resolve_url_no_redirect>` had to be changed
+from ``tuple[type[Plugin], str]`` to ``tuple[str, type[Plugin], str]``, and both methods now return the resolved plugin name
+as the first item, in addition to the plugin class and resolved URL.
+
+
 streamlink 4.2.0
 ----------------
 
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -1,4 +1,5 @@
 import ast
+import inspect
 import logging
 import operator
 import re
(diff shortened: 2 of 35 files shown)


bastimeyer:
Btw, as a next step (ignore this for now in this PR though), since the session is not a class attribute on the `Plugin` class anymore and is thus not defined as `None` initially, typing information can finally be added, so that calls to `self.session.http.get()` and the likes can be resolved by the IDE/editor and static code analyzers like mypy.

I have stashed some local changes, but these changes also required adding wrappers for direct HTTP-verb-methods of Streamlink's `HttpSession` subclass, like `get`, `post`, `head`, `patch`, `put`, `options` and `delete`, as the `schema` keyword and others are not recognized due to the typing stubs of `requests`. Otherwise, it shows typing warnings when calling `self.session.http.get(url, schema=...)`.

bastimeyer:
Rebased to master and fixed merge conflicts introduced by #4771

bastimeyer:
I've reverted the branch back to e2e1eb8249478f03a92b6dc2cfc8c74eccadb398, as the merge conflict caused by #4779 will be reverted once #4780 gets merged.

I have no idea though why reverting back to an old commit that was already run causes GitHub to create a temporary merge with master while running additional CI jobs (WTF?):
https://github.com/streamlink/streamlink/runs/8056558175?check_suite_focus=true#step:2:488

bastimeyer:
Rebased to master now after #4780 was merged. No need to unnecessarily wait for this.

This is ready for review now.

gravyboat:
Looks good, thanks for the thorough commit message on [e88750](https://github.com/streamlink/streamlink/commit/e8875035676de7c1dc98eef8018e80b5077f5f37), made it a lot more clear.

PR #4791, opened 2022-08-31
Fix misinfo for LiveProxy

MagnusStrom:
LiveProxy does not support windows

### DIFF
--- a/docs/thirdparty.rst
+++ b/docs/thirdparty.rst
@@ -44,7 +44,7 @@ LiveProxy
 ---------
 
 :Description: Local proxy server between Streamlink and an URL
-:OS: |Windows| |MacOS| |Linux|
+:OS: |MacOS| |Linux|
 :Author: `back-to <https://github.com/back-to>`_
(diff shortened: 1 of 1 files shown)


bastimeyer:
> LiveProxy does not support windows

What makes you think that?

It's an OS-agnostic python project, with dependencies on streamlink and yt-dl / yt-dlp.
@back-to 

- https://github.com/back-to/liveproxy
- https://github.com/back-to/liveproxy/blob/35cad27645a0f0a078e6b802218c954bffb69642/setup.py#L47-L54

It even has Windows-specific logic:
https://github.com/back-to/liveproxy/blob/35cad27645a0f0a078e6b802218c954bffb69642/liveproxy/main.py#L32-L40

back-to:
@MagnusStrom 

> LiveProxy does not support windows

it should work again with the latest version or after this commit

https://github.com/back-to/liveproxy/commit/cf5b345c5f66f093e12920e24d48482163e1baba

PR #4795, opened 2022-09-01
cli: refactor setup of session options

bastimeyer:
- Merge `setup_options()` and `setup_http_session()` and always
  initialize all session options:
  It doesn't make sense having this split up into two functions, just
  so that some CLI functions can have a partially configured session
  with just the HTTP options set
- Move merged session setup logic into the `argparser` module
  and replace the dozens of if-statements and `session.set_option()`
  calls with a list of "CLI argument -> session option" mappings
- Add tests and update `streamlink_cli.main` tests respectively

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -3,17 +3,20 @@ import numbers
 import re
 from string import printable
 from textwrap import dedent
-from typing import Dict, List, Optional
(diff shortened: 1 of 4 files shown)


gravyboat:
@bastimeyer Sorry for taking so long to get to this, been a busy few days. Could you please rebase since the other PR was merged to address this conflict?

PR #4797, opened 2022-09-02
cli: remove custom `ignored` contextmanager

bastimeyer:
and replace it with `contextlib.suppress`

----

Available since Python 3.4
https://docs.python.org/3/library/contextlib.html#contextlib.suppress

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -6,7 +6,7 @@ import platform
 import re
 import signal
 import sys
-from contextlib import closing
+from contextlib import closing, suppress
 from functools import partial
 from gettext import gettext
 from itertools import chain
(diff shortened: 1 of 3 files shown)


bastimeyer:
Btw, this PR doesn't replace the

```py
try:
    ...
except ...:
    pass
```

pattern that is used throughout the entire code. Some linting tools / code analyzers suggest replacing this with the `suppress` contextmanager

```py
with contextlib.suppress(...):
    ...
```

because it makes the intention of the code author "more clear" (according to a discussion of some Python devs), but IMO this makes writing tests more difficult and "unthorough", because with the explicit try/except block, code coverage immediately reveals if the suppressed exception actually gets tested.

I will submit some automated (and adjusted) code cleanups later (via [sourcery.ai](https://sourcery.ai/), a proprietary "AI" code analyzer - only good for one-time clean-ups), and this suggestion won't be included.

PR #4802, opened 2022-09-04
plugin: add session typing information

bastimeyer:
- Add typing information to `Plugin.session`
- Implement `http_session` stub file due to the `HTTPSession` subclass
  of `requests.Session` which adds additional keywords to the
  `request()` method, including all other HTTP-verb methods
- Add `flake8-pyi` and `typing_extensions` to dev-requirements.txt
  (`typing_extensions` is not a runtime dependency)

----

With the removal of the `Plugin.bind()` classmethod in #4768, now that the `Streamlink` session instance gets set on each `Plugin` instance, typing information can finally be added, as the `session` attribute doesn't have to be defined as `None` anymore initially.

Adding typing informations finally allows plugin implementors to see all the stuff that's defined on the `Session`, including the `HTTPSession` via `self.session.http`. However, since Streamlink subclasses requests's `Session` and adds additional keywords to the `request()` method and thus also to all HTTP-verb methods like `get()`, `post()`, etc., this introduces typing issues. The most common problem by far is of course the addition of the `schema` keyword, which also changes the return type of the various methods from `Response` to `Any`.

I originally tried to avoid adding a stub file and directly adding the missing HTTP-verb methods to the `HTTPSession` subclass and adding typing information there, but that got very messy. The repetitive typing information therefore needs to be defined in a stub file. See [PEP561](https://peps.python.org/pep-0561/).

Unfortunately though, it doesn't seem to be possible to import already defined type aliases from the `types-requests` package, so I had to copy all required definitions for the other args+keywords provided by requests.

----

The [`typing_extensions`](https://github.com/python/typing_extensions) dependency had to be added because mypy is [configured for python 3.7](https://github.com/streamlink/streamlink/blob/8a7717f549b3f67fe3d596dbe1e562f17d48a8a8/pyproject.toml#L52-L53), and [`TypeAlias` wasn't part of the `typing` module yet in 3.7](https://docs.python.org/3/library/typing.html#typing.TypeAlias). Otherwise a `error: Module "typing" has no attribute "TypeAlias"` would be raised.

The new `Y037` linting error also has to be ignored because of an apparent mypy bug (`error: Type application has too many types (1 expected)`) when union types like `a | b | c` are defined instead of `Union[a,b,c]`. This is all a bit weird, because the copied type alias defintions are using both styles. Maybe it's because of the re-defintions... The union types syntax in stub files is valid on Python 3.7, btw:
https://typing.readthedocs.io/en/latest/source/stubs.html#syntax

----

Opening this as a draft for now.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -6,6 +6,7 @@ requests-mock
 freezegun>=1.0.0
 flake8
 flake8-import-order
+flake8-pyi
 shtab
 versioningit >=2.0.0, <3
 
@@ -14,3 +15,4 @@ lxml-stubs
 types-freezegun
 types-requests
 types-urllib3
+typing_extensions
--- a/setup.cfg
+++ b/setup.cfg
@@ -55,6 +55,8 @@ ignore =
   # W503 - line break before binary operator
   # https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator
   W503,
+  # Y037 - Use PEP 604 union types instead of typing.Union
+  Y037,
 extend-exclude =
   build/,
   dist/,
--- a/src/streamlink/plugin/api/http_session.pyi
+++ b/src/streamlink/plugin/api/http_session.pyi
@@ -1,0 +1,345 @@
+from collections.abc import Callable, Iterable, Mapping, MutableMapping, Sequence
+from typing import Any, IO, Union
+
+from _typeshed import SupportsItems, SupportsRead
+from requests import PreparedRequest, Response, Session
+from requests.adapters import HTTPAdapter
+from requests.auth import AuthBase
+from requests.cookies import RequestsCookieJar
+from typing_extensions import TypeAlias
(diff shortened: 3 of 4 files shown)


bastimeyer:
I've changed the return type of the `request()` and HTTP-verb methods to `Any`, because `Response | Any` doesn't work well. My idea was to have the **default return type** included in addition to `Any`, so that code that actually calls methods or reads attributes on `Response` objects can have typing information, but apparently this results in the `Response | Response` typing information in PyCharm, which causes warnings for all kinds of stuff. No idea why, but it's annoying and can be fixed/improved later.

Once this PR is merged, I will open another one with typing information added to the various stream implementations. Adding this here unnecessarily bloats up the diff.

gravyboat:
Thanks @bastimeyer. Bummer about the `Response | Any` types since it would be really nice, seems weird that it would infer the `Any` as `Response` but as you mentioned something along the line must be causing weird inference.

PR #4804, opened 2022-09-06
docs: handle suppressed plugin arguments

bastimeyer:
and remove AST parser workaround for py37, as the currently used version
of the Furo theme doesn't build on py37 anyway


bastimeyer:
Let me clean this up a bit more...

Issue #4825, opened 2022-09-13
stream.ffmpegmux: stream with independent segments (audio+video) cannot be remuxed


vidanio:
### Checklist

- [X] This is a plugin issue and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

I am using a Debian 11 stable with kernel 5.10 as you can see in the debug log below.
Tried to send this URL: https://videodelivery.net/9fb3596948ddf463fde0ec4b85625b24/manifest/video.m3u8 that has independent segments for audio and video. Streams were closed, does not start re-muxing.
Cmdline used:
streamlink --stdout hls://https://videodelivery.net/9fb3596948ddf463fde0ec4b85625b24/manifest/video.m3u8 --default-stream best --loglevel debug | ffmpeg -re -i - -c copy 

### Debug log

```text
[cli][debug] OS:         Linux-5.10.0-16-amd64-x86_64-with-glibc2.31
[cli][debug] Python:     3.9.2
[cli][debug] Streamlink: 4.3.0
[cli][debug] Dependencies:
[cli][debug]  isodate: 0.6.0
[cli][debug]  lxml: 4.9.1
[cli][debug]  pycountry: 20.7.3
[cli][debug]  pycryptodome: 3.15.0
[cli][debug]  PySocks: 1.7.1
[cli][debug]  requests: 2.28.1
[cli][debug]  websocket-client: 1.4.1
[cli][debug] Arguments:
[cli][debug]  url=hls://https://videodelivery.net/9fb3596948ddf463fde0ec4b85625b24/manifest/video.m3u8
[cli][debug]  --loglevel=debug
[cli][debug]  --stdout=True
[cli][debug]  --default-stream=['best']
[cli][info] Found matching plugin hls for URL hls://https://videodelivery.net/9fb3596948ddf463fde0ec4b85625b24/manifest/video.m3u8
[plugins.hls][debug] URL=https://videodelivery.net/9fb3596948ddf463fde0ec4b85625b24/manifest/video.m3u8; params={}
[...]

PR #4781, opened 2022-08-28
stream.ffmpegmux: show warning if unavailable

bastimeyer:
There is way too much confusion about FFmpeg and muxed streams.

The problem is that it's not transparent at all which streams are muxed and which are not. This is due to how streams get returned, as a mapping of name->stream dicts or tuples with "alt" deduplications already done by the `Stream` implementations themselves or the `plugin.streams()` method, and only names get shown by the CLI.

The HLS implementation with external audio streams is also not ideal, because it silently falls back to streams without external audio. That needs to be changed, ideally with the implementation of alternative video streams (#3579).

So in order to at least give the user some feedback, let's add warning messages when FFmpeg is not available on the system.
The warning messages get only shown once if multiple checks are done, thanks to the caching added recently.

I decided to add two warning messages and tried to make them as short and precise as possible. Referring to the CLI argument name is not 100% correct, but since the session option does have the same name, it's fine. The empty set is also always a subset, so that should be fine, too.

----

UStreamTV plugin which requires FFmpeg:
```
$ streamlink 'https://www.ustream.tv/nasahdtv'
[cli][info] Found matching plugin ustreamtv for URL https://www.ustream.tv/nasahdtv
Available streams: 252p+a84k (worst), 252p+a95k, 360p+a84k, 360p+a95k, 486p+a84k, 486p+a95k, 720p+a84k, 720p+a95k (best)

$ streamlink --ffmpeg-ffmpeg=/dev/null 'https://www.ustream.tv/nasahdtv'
[cli][info] Found matching plugin ustreamtv for URL https://www.ustream.tv/nasahdtv
[stream.ffmpegmux][warning] FFmpeg was not found. See the --ffmpeg-ffmpeg option.
[stream.ffmpegmux][warning] Muxing streams is unsupported! Only a subset of the available streams can be returned!
error: No playable streams found on this URL: https://www.ustream.tv/nasahdtv
```

YouTube plugin with adaptive streams - #4777:
```
$ streamlink 'https://www.youtube.com/watch?v=0Jt8Wu1osgI'
[cli][info] Found matching plugin youtube for URL https://www.youtube.com/watch?v=0Jt8Wu1osgI
Available streams: audio_mp4a, audio_opus, 144p (worst), 240p, 360p, 480p, 720p, 720p60, 1080p60, 1440p60, 2160p60 (best)

$ streamlink --ffmpeg-ffmpeg=/dev/null 'https://www.youtube.com/watch?v=0Jt8Wu1osgI'
[...]

Issue #4520, opened 2022-05-09
hls-multi from arte.tv has seconds of hanging video


msdemlei:
### Checklist

- [X] This is a bug report and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

When pulling video from arte, e.g.

```
streamlink --output o.mp4 \
  https://www.arte.tv/de/videos/108210-039-A/mit-offenen-karten-im-fokus/ \ 
  worst
```

the resulting video stream runs fine for a few seconds.  Then the video freezes, and mpv emits a few lines like

```
Invalid video timestamp: 5.960000 -> 5.920022
AV: 00:00:05 / 00:02:29 (3%) A-V:  0.041 Dropped: 48
```

I suppose the 48 refers to dropped frames; that would roughly coincide with the duration of the video freezes.  Video then resumes for a few seconds, then freezes again, etc (persumably once per HLS segment).  Audio is fine all the time in mpv.  Vlc stutters in both audio and video.

For me, that already has been the behaviour on streamlink 3.2.0 and 3.0.3.  2.0.0 as packagaged by Debian was fine (but its artetv plugin is now broken).

Since I suspect it matters here, ffmpeg version is 4.3.4-0+deb11u1.

### Debug log

```text
[cli][debug] OS:         Linux-5.16.16-x86_64-with-glibc2.31
[cli][debug] Python:     3.9.2
[cli][debug] Streamlink: 4.0.1
[cli][debug] Requests(2.25.1), Socks(1.7.1), Websocket(0.57.0)
[cli][debug] Arguments:
[cli][debug]  url=https://www.arte.tv/de/videos/108210-039-A/mit-offenen-karten-im-fokus/
[cli][debug]  stream=['216p']
[cli][debug]  --loglevel=debug
[cli][debug]  --output=o.mp4
[cli][info] Found matching plugin artetv for URL https://www.arte.tv/de/videos/108210-039-A/mit-offenen-karten-im-fokus/
[utils.l10n][debug] Language code: en_US
[...]

PR #4843, opened 2022-09-21
FFmpeg validation

bastimeyer:
We've recently added [warning log messages to the `FFMPEGMuxer`'s resolver when FFmpeg could not be found](https://github.com/streamlink/streamlink/pull/4781), so that the user can be informed that muxing is not supported. However, there are still a couple of issues left which I think should be addressed:

1. The chosen/resolved FFmpeg executable does not get validated/verified:
   This is a problem when the user has set an invalid FFmpeg executable, or when it is broken and doesn't work. In case of invalid executables, the output and stream muxing are pretty much undefined, and broken executables lead to empty stream outputs, and unless verbose logging is enabled, no error messages get printed, making the user believe there's an issue with the input stream.
2. The FFmpeg version does not get logged:
   Logging the FFmpeg version makes debugging issues easier. The recent issue which comes to mind is:
   #4825 
3. Certain streams might require a certain FFmpeg version in order to correctly mux streams:
   This PR doesn't add any version checks, but this could be added in the future if it's needed, based on this work.
   https://github.com/streamlink/streamlink/issues/4520#issuecomment-1129428877

Opening the PR as a draft, as it's unfinished work.

----

The first commit implements a generic `ProcessOutput` utility class for **asynchronously** spawning a subprocess and reading its stdout/stderr streams.

This is necessary, because using APIs like [`subprocess.run(capture_output=True)`](https://docs.python.org/3/library/subprocess.html#subprocess.run) or [`asyncio.create_subprocess_exec().communicate()`](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.subprocess.Process.communicate) read the entire stdout/stderr streams into memory until the process terminates on its own. This is bad when executing unknown binaries. And reading streams of a spawned subprocess in separate threads as well as dealing with timeouts is unnecessarily difficult to implement in regards to making everything thread-safe (especially when writing tests, just see the messy HLS stream tests for example).

Asynchronously iterating output streams and using `asyncio.sleep()` makes this fairly easy and straightforward to implement, but testing stuff unfortunately requires adding two more dev-requirements:

1. [`pytest-asyncio`](https://pypi.org/project/pytest-asyncio/) - already installed as a transitive dependency, so no changes here
2. [`mock`](https://pypi.org/project/mock/) - compatibility backport of the `unittest.mock` stdlib implementation for py37 (`AsyncMock` is missing)
   `mock` was already removed from the dev-requirements in d04767fa5b812a4e971e454226cecb25b4d975ae, and the only gotchas were the [`args`/`kwargs` properties in `Mock.call_args`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args) which are missing on py37, but that doesn't justify adding compatibility imports everywhere

Another issue I ran into while writing tests was `freezegun`, which currently affects the internals of `asyncio` in its most recent version, and there's an open PR on their repo which will change that. I managed to get the tests working (after lots of issues with `asyncio.sleep()`), but I'm not sure how the next `freezegun` release(s) will affect the tests. We'll see...

----

The second commit adds FFmpeg executable validation and version logging. The validation can be turned off by setting `--ffmpeg-no-validation`.

Still need to update and fix tests.

I've included some links in the code to the FFmpeg version string definitions, which have been stable for 12 years now.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,7 +1,9 @@
 pip>=9
 pytest
+pytest-asyncio
 pytest-cov
 coverage[toml]
+mock ; python_version<"3.8"
 requests-mock
 freezegun>=1.0.0
 flake8
--- a/src/streamlink/session.py
+++ b/src/streamlink/session.py
@@ -81,6 +81,7 @@ class Streamlink:
             "stream-segment-timeout": 10.0,
             "stream-timeout": 60.0,
             "ffmpeg-ffmpeg": None,
+            "ffmpeg-no-validation": False,
             "ffmpeg-fout": None,
             "ffmpeg-video-transcode": None,
             "ffmpeg-audio-transcode": None,
@@ -167,6 +168,9 @@ class Streamlink:
         ffmpeg-ffmpeg            (str) Specify the location of the
                                  ffmpeg executable use by Muxing streams
                                  e.g. ``/usr/local/bin/ffmpeg``
+
+        ffmpeg-no-validation     (bool) Disable FFmpeg validation and version logging.
+                                 default: ``False``
 
         ffmpeg-verbose           (bool) Log stderr from ffmpeg to the
                                  console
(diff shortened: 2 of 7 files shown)


bastimeyer:
I'll open a new PR in a couple of days once I've finished updating the tests.

PR #4889, opened 2022-10-16
plugins.tf1 : add 'Stream' channels

BZHDeveloper:
Allow 'Stream' channels. No account needed

### DIFF
--- a/src/streamlink/plugins/tf1.py
+++ b/src/streamlink/plugins/tf1.py
@@ -18,20 +18,20 @@ log = logging.getLogger(__name__)
 
 
 @pluginmatcher(re.compile(
-    r"https?://(?:www\.)?(?:tf1\.fr/([\w-]+)/direct|(lci)\.fr/direct)/?"
+    r"https?://(?:www\.)?(?:tf1\.fr/(?P<live>[\w-]+)/direct/?|(?P<lci>tf1info|lci)\.fr/direct/?|tf1.fr/stream/(?P<stream>[\w-]+))"
 ))
 class TF1(Plugin):
     api_url = "https://mediainfo.tf1.fr/mediainfocombo/{}?context=MYTF1&pver=4001000"
 
-    def api_call(self, channel, useragent=useragents.CHROME):
-        url = self.api_url.format(f"L_{channel.upper()}")
+    def api_call(self, channel_id, useragent=useragents.CHROME):
+        url = self.api_url.format(channel_id)
         req = self.session.http.get(url,
                                     headers={"User-Agent": useragent})
         return self.session.http.json(req)
 
-    def get_stream_urls(self, channel):
+    def get_stream_urls(self, channel_id):
         for useragent in [useragents.CHROME, useragents.IPHONE_6]:
-            data = self.api_call(channel, useragent)
+            data = self.api_call(channel_id, useragent)
 
             if 'delivery' not in data or 'url' not in data['delivery']:
                 continue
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/tf1.py
```diff
@@ -17,21 +17,23 @@
 log = logging.getLogger(__name__)
 
 
-@pluginmatcher(re.compile(
-    r"https?://(?:www\.)?(?:tf1\.fr/([\w-]+)/direct|(lci)\.fr/direct)/?"
-))
+@pluginmatcher(re.compile(r"""
+    https?://(?:www\.)?(?:tf1\.fr/(?P<live>[\w-]+)/direct/?|
+    (?P<lci>tf1info|lci)\.fr/direct/?|
+    tf1.fr/stream/(?P<stream>[\w-]+))
```
bastimeyer, line 23: The regex should be reformatted. This is barely legible. Literal dots also need to be escaped.

src/streamlink/plugins/tf1.py
```diff
@@ -42,16 +44,30 @@ def get_stream_urls(self, channel):
...
+            channel_id = ""
+            if m.group("live"):
+                channel = m.group("live")
+                channel_id = "L_" + channel.upper()
+            elif m.group("lci"):
+                channel = "LCI"
+                channel_id = "L_LCI"
+            elif m.group("stream"):
+                channel = m.group("stream")
+                channel_id = "L_FAST_v2l-" + m.group("stream")
```
bastimeyer, line 57: Please remove the unnecessary `if m:` block.

Also please don't use the `+` operator for string concatenation, use f-strings.

src/streamlink/plugins/tf1.py
```diff
@@ -42,16 +44,30 @@ def get_stream_urls(self, channel):
...
             log.debug("Found channel {0}".format(channel))
-            for sformat, url in self.get_stream_urls(channel):
+            for sformat, url in self.get_stream_urls(channel_id):
                 try:
```
bastimeyer, line 60: The `try: ... except PluginError: ...` block should also be removed. It's unnecessary.

DASH streams and HLS multivariant streams should also not be yielded in a loop and instead the resulting dict of the `HLSStream.parse_variant_playlist` and `DASHStream.parse_manifest` should be returned by `Plugin._get_streams` and the loop be interrupted, so potential unnecessary/duplicate streams don't get included in the streams result.

src/streamlink/plugins/tf1.py
```diff
@@ -17,21 +17,23 @@
...
 class TF1(Plugin):
     api_url = "https://mediainfo.tf1.fr/mediainfocombo/{}?context=MYTF1&pver=4001000"
 
-    def api_call(self, channel, useragent=useragents.CHROME):
-        url = self.api_url.format(f"L_{channel.upper()}")
+    def api_call(self, channel_id, useragent=useragents.CHROME):
+        url = self.api_url.format(channel_id)
         req = self.session.http.get(url,
                                     headers={"User-Agent": useragent})
         return self.session.http.json(req)
```
bastimeyer, line 32: The API calls should ideally use a validation schema instead of simply returning the parsed JSON response and later checking for `if 'delivery' not in data or 'url' not in data['delivery']`. I'm saying this, because `data['delivery']['format']` is not checked for validity either and this will cause a `KeyError`/`LookupError` to be raised if it's missing instead of a proper validation error.

src/streamlink/plugins/tf1.py
```diff
@@ -17,21 +17,23 @@
...
 
-    def get_stream_urls(self, channel):
+    def get_stream_urls(self, channel_id):
         for useragent in [useragents.CHROME, useragents.IPHONE_6]:
```
bastimeyer, line 35: Is the useragent distinction for the API calls still relevant?

bastimeyer (commented):
Thanks for the PR.

Just a few comments regarding the old plugin implementation, as it should receive some cleanup.

I haven't actually tested the changes yet and will try to look at this tomorrow.

Issue #4918, opened 2022-11-02
plugins.bloomberg: Always get Could not Find JSON data error from today


chaoyic:
### Checklist

- [X] This is a plugin issue and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

Getting [plugins.bloomberg][error] Could not find JSON data. Invalid URL or bot protection... consistently. 
Used to be able to solve this buy trying VPN or a different network. This does not help now.

My guess is that Bloomberg updated something on their side and the plugin is now broken?
Or Bloomberg just update their "are you a robot" check?

Tried changing User-Agent header manually and it did not help. Can someone point out what is the issue here?

### Debug log

```text
PS C:\Users\***> streamlink -l debug https://www.bloomberg.com/live/us
[cli][debug] OS:         Windows 10
[cli][debug] Python:     3.10.8
[cli][debug] Streamlink: 5.0.1
[cli][debug] Dependencies:
[cli][debug]  isodate: 0.6.1
[cli][debug]  lxml: 4.9.1
[cli][debug]  pycountry: 22.3.5
[cli][debug]  pycryptodome: 3.15.0
[cli][debug]  PySocks: 1.7.1
[cli][debug]  requests: 2.28.1
[cli][debug]  websocket-client: 1.4.1
[cli][debug] Arguments:
[cli][debug]  url=https://www.bloomberg.com/live/us
[cli][debug]  --loglevel=debug
[cli][info] Found matching plugin bloomberg for URL https://www.bloomberg.com/live/us
[plugins.bloomberg][error] Could not find JSON data. Invalid URL or bot protection...
error: No playable streams found on this URL: https://www.bloomberg.com/live/us
```

bastimeyer:
They've updated their bot detection. I had a quick look at this, but I haven't managed to figure this out yet.

[...]

PR #4919, opened 2022-11-02
plugins.bloomberg: update/fix headers to avoid bot detection

mkbloke:
As far as my testing goes, all the previously added headers seem to be unnecessary.  Adding the `Accept-Encoding` header back seems to trigger bot detection, possibly because with requests it only includes `gzip` and `deflate`, but not `br`, which Firefox also includes.  Removing the header entirely seems to work.
```
$ streamlink https://www.bloomberg.com/live
[cli][info] Found matching plugin bloomberg for URL https://www.bloomberg.com/live
Available streams: 272p (worst), 360p, 720p_alt, 720p, 1080p (best)

$ streamlink https://www.bloomberg.com/news/videos/2022-10-26/rishi-sunak-and-the-uk-economy-showdown
[cli][info] Found matching plugin bloomberg for URL https://www.bloomberg.com/news/videos/2022-10-26/rishi-sunak-and-the-uk-economy-showdown
Available streams: 100k (worst), 176p, 272p, 360p, 540p, 720p_alt, 720p, 1080p (best)
```

### DIFF
--- a/src/streamlink/plugins/bloomberg.py
+++ b/src/streamlink/plugins/bloomberg.py
@@ -102,17 +102,7 @@ class Bloomberg(Plugin):
         return secureStreams or streams
 
     def _get_streams(self):
-        self.session.http.headers.update({
-            "authority": "www.bloomberg.com",
-            "upgrade-insecure-requests": "1",
-            "dnt": "1",
-            "accept": ";".join([
-                "text/html,application/xhtml+xml,application/xml",
-                "q=0.9,image/webp,image/apng,*/*",
-                "q=0.8,application/signed-exchange",
-                "v=b3"
-            ])
-        })
+        del self.session.http.headers["Accept-Encoding"]
 
         try:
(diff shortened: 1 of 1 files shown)


bastimeyer:
> possibly because with requests it only includes `gzip` and `deflate`, but not `br`

Brotli does only get added by urllib3 if the `brotlicffi` or `brotli` package is installed, hence why we're installing `brotli` for the CI test runners:
- 57023b1430bf5d3aa6f8d10cca70cbdd231c912a
- https://github.com/psf/requests/blob/v2.28.1/requests/utils.py#L62-L65
- https://github.com/urllib3/urllib3/blob/1.26.12/src/urllib3/util/request.py#L15-L24

The bot detection triggers regardless whether brotli is available or not, just checked it. No idea why removing the header works... :confused:

mkbloke:
Ah, I just read your comment here after writing: https://github.com/streamlink/streamlink/issues/4918#issuecomment-1301237360

Well, that's very strange indeed.

PR #4923, opened 2022-11-06
plugins.niconicochannelplus: add support for "nicochannel.jp"

pzhlkj6612:
For VoD and live on "[Niconico Channel Plus](https://portal.nicochannel.jp/)" (Japanese: ニコニコチャンネルプラス).

Thanks to the following people who provided useful information:

- @lokilin, raised a request in yt-dlp/yt-dlp#2537;
- 雷兰濑 (@railannad), raised a request in #4366;
- @Trung0246, analyzed HTTP data in #4366;
- @Lesmiscore, made [cross-reference](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls) between the above two issues.

closes  #4366

PR #4929, opened 2022-11-08
docs: add mypy for code linting

pzhlkj6612:
Hi. 

In #4923 (464eac5c3daddf2934ff47cc77ce91b30f468dff), I encountered the following error in CI:

```console
$ python -m mypy --no-incremental src/streamlink/ src/streamlink_cli/ tests/ docs/ setup.py
src/streamlink/plugins/niconicochannelplus.py: note: In member "get_master_playlist_url" of class "NicoNicoChannelPlus":
src/streamlink/plugins/niconicochannelplus.py:47:13: error: Need type annotation for "payload" (hint: "payload: Dict[<type>, <type>] = ...")  [var-annotated]
```

I didn't forget to check the code by flake8, but really didn't know to call mypy. To help future developers, I added it to the developing docs.
 
<br>

At ca4fff1453fbc10dcd583c8a88c922fb7ad16fd1 , I tested mypy 0.990 with Python 3.8.10 and 3.10.4 for Windows, and it only passed when using 3.10:

```console
> python --version; mypy --version
Python 3.8.10
mypy 0.990 (compiled: yes)
> mypy --no-incremental src/streamlink/ src/streamlink_cli/ tests/ docs/ setup.py
src\streamlink_cli\compat.py:8:5: error: Name "importlib_metadata" already defined (possibly by an import)  [no-redef]
src\streamlink_cli\compat.py:8:5: note: Error code "no-redef" not covered by "type: ignore" comment
```

I guess this is the reason to setup Python with the version of 3.10:

https://github.com/streamlink/streamlink/blob/ca4fff1453fbc10dcd583c8a88c922fb7ad16fd1/.github/workflows/lint.yml#L43-L45

### DIFF
--- a/docs/developing.rst
+++ b/docs/developing.rst
@@ -110,6 +110,8 @@ performing these checks locally avoids unnecessary build failures.
 
     # check code for linting errors
     flake8
+    # check code for typing errors
+    mypy
 
     # build the documentation
     make --directory=docs clean html


bastimeyer:
Thanks.

> I encountered the following error in CI

Typing information in plugins is mostly irrelevant. If you remove typing information from the method signatures, then the method bodies won't be analyzed by mypy (``), so you don't have to define strict typing informations when setting variables, etc.

> it only passed when using 3.10

That's because `importlib_metadata` is a dependency which only gets installed on old Python versions where it's not available in the standard library, so mypy will use a different code path when statically analyzing the `streamlink_cli.compat` module and its import statements. There's a missing `no-redef` ignore comment which hasn't been caught because the CI runner is running on a more recent version and thus doesn't take this code path.

> I guess this is the reason to setup Python with the version of 3.10

The Python version which mypy is testing against is defined in `pyproject.toml`, namely 3.7. The runner is on 3.10, because it was simply the latest one when the config file was created. Python 3.11 has just been released and CI runners haven't been upgraded yet.

> To help future developers, I added it to the developing docs.
> ```
> mypy --no-incremental src/streamlink/ src/streamlink_cli/ tests/ docs/ setup.py
> ```

Ideally, the files and modules which need to be tested should be defined in `pyproject.toml`, to avoid any specific CLI arguments. Having the files and modules defined in the docs as part of the mypy command is unnecessary duplication and will 100% be forgotten about on the next change. Let me fix/update this in a separate PR real quick, then you can rebase once merged and update the docs.

Please add the `# check code for typing errors` comment then, as linting is more about code style, not typing information, thanks.

docs/developing.rst
```diff
@@ -110,6 +110,7 @@ performing these checks locally avoids unnecessary build failures.
 
     # check code for linting errors
     flake8
+    mypy --no-incremental src/streamlink/ src/streamlink_cli/ tests/ docs/ setup.py
```
bastimeyer, line 113:
```suggestion

    # check code for typing errors
    mypy
```

bastimeyer (changes requested):

pzhlkj6612:
Very, informative, response! Thank you, @bastimeyer .

> > I guess this is the reason to setup Python with the version of 3.10
> 
> The Python version which mypy is testing against is defined in `pyproject.toml`, namely 3.7.

OK, I've found [`python_version`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-python_version) under `[tool.mypy]` .

> The runner is on 3.10, because it was simply the latest one when the config file was created. Python 3.11 has just been released and CI runners haven't been upgraded yet.

OK.

> Ideally, the files and modules which need to be tested should be defined in `pyproject.toml`, to avoid any specific CLI arguments. Having the files and modules defined in the docs as part of the mypy command is unnecessary duplication and will 100% be forgotten about on the next change.

I agree. Parameters are a bit annoyed when manually typing them in CLI.

> Let me fix/update this in a separate PR real quick, then you can rebase once merged and update the docs.

> Please add the `# check code for typing errors` comment then, as linting is more about code style, not typing information, thanks. The `--no-incremental` CLI argument should also not be used as a documentation advice. This is for preventing caching processed typing informations.

Done.

docs/developing.rst
pzhlkj6612: Thanks. I committed the change just now, sorry for the late.

bastimeyer (approved):
Thanks, appreciated.

PR #4930, opened 2022-11-08
tools: move list of files to mypy config

bastimeyer:
- Move list of files from CI config to mypy config in pyproject.toml, so that mypy can be run locally without having to specificy the same inputs as CLI arguments. Keep --no-incremental exclusive to CI runner.
- Update Python version of mypy CI runner.
- Fix typing ignore comment when importing importlib_metadata fallback in the streamlink_cli.compat module.

----

See #4929


bastimeyer:
https://github.com/streamlink/streamlink/actions/runs/3419334576/jobs/5692726802#step:5:12

PR #4936, opened 2022-11-10
stream.hls: add more debug messages to write calls

bastimeyer:
Adding more debug messages to `HLSStreamWriter.write()` will make debugging segment filtering much easier.

This should be done independently of the current Twitch ad filtering issues #4934.


bastimeyer:
Going to merge this now, as I'm going to open a PR with the updated Twitch ads filtering later before I'm going to bed, and this should be part of it. There's nothing to review here anyway.

PR #4941, opened 2022-11-10
logger: add new log level "all"

bastimeyer:
Logging extremely verbose data like HLS playlist contents for example doesn't really fit on the `trace` log level. Adding an ultimate verbosity log level makes more sense here, so we can add log calls to the HLS parser and similar things.

Is the name "all" fine? I've just renamed it from "paranoid" before submitting this PR, because I wasn't really happy with that name. "all" is the opposite of "none", which is at the other end of the log level list, and it's easier to remember, too.

Since there are two custom log levels now, I had to refactor a bit of code in order to prevent duplicate stuff and duplicate tests.

### DIFF
--- a/src/streamlink/logger.py
+++ b/src/streamlink/logger.py
@@ -8,6 +8,16 @@ from threading import Lock
 from typing import IO, List, Optional, TYPE_CHECKING, Union
 
 
+if TYPE_CHECKING:  # pragma: no cover
+    _BaseLoggerClass = logging.Logger
+else:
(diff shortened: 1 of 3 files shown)


gravyboat (approved):
Typically I would expect this to be at the `trace` level but as you noted it's even more verbose than what might be seen there so I'm fine with calling it `all` even though it's non-standard (who really cares as long as we get the info). I'm going to mark this as approved @bastimeyer and you can merge it whenever.

PR #4951, opened 2022-11-11
plugins.beyaztv: new plugin

kaypoush:
Hello, I'm adding support for Beyaz TV, a Turkish live TV channel privately owned by Kanal Beyaz. I tested the plugin, it should work and I also fixed the issues pointed out by flake8. I'm new to all this (including programming) so excuse me if I did anything wrong. If you go to https://m.beyaztv.com.tr/canli-yayin it will give you a http error 404 but if you add / at the end it will work, that's why I included the should not match line in the test file but I'm not sure if I did it the right way. It's only with the mobile version of the site that I had this issue. With the desktop version, adding / or not at the end of the url doesn't change anything.

### DIFF
--- a/src/streamlink/plugins/beyaztv.py
+++ b/src/streamlink/plugins/beyaztv.py
@@ -1,0 +1,29 @@
+"""
+$description Turkish live TV channel owned by Kanal Beyaz.
+$url beyaztv.com.tr
+$type live
+"""
+
+import re
+
+from streamlink.plugin import Plugin, pluginmatcher
+from streamlink.plugin.api import validate
+
+
+@pluginmatcher(re.compile(
+    r"https?://(?:www\.|m\.)?beyaztv\.com\.tr/canli-yayin",
+))
+class BeyazTV(Plugin):
+    def _get_streams(self):
+        video_id = self.session.http.get(self.url, schema=validate.Schema(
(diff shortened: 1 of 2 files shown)


back-to:
you don't need a plugin for this,
use the dailymotion URL

```
$ streamlink.exe https://www.dailymotion.com/BeyazTV best -l info
[cli][info] Found matching plugin dailymotion for URL https://www.dailymotion.com/BeyazTV
[cli][info] Available streams: 180p_alt (worst), 180p, 288p_alt, 288p, 477p_alt, 477p, 720p_alt, 720p, 1080p_alt, 1080p (best)
```

kaypoush:
Why is [cnews.py](https://github.com/streamlink/streamlink/blob/f7cbfae2621d6dd2f9f1189cf98a59ba64ff9721/src/streamlink/plugins/cnews.py) allowed then?

```streamlink https://www.dailymotion.com/CNEWS best```

back-to:
> Why does [cnews.py](https://github.com/streamlink/streamlink/blob/f7cbfae2621d6dd2f9f1189cf98a59ba64ff9721/src/streamlink/plugins/cnews.py) exist then?
streamlink https://www.dailymotion.com/CNEWS best
and all the other youtube based streams

old code,
a lot of plugins got already removed in the past.

PR #4952, opened 2022-11-11
plugins.twitch: add --twitch-access-token-param

bastimeyer:
Since authentication data for example can be set via `--twitch-api-header="Authorization=OAuth some-token"` and it is thus a useful generic plugin parameter, adding `--twitch-access-token-param` could be useful as well.

Having `--twitch-access-token-param` allows setting different `playerType` API request parameter values for example. The default value of this is `embed`, which makes Streamlink retrieve an access token for Twitch's embeddable web player. `site` is another valid value.

In the past we've had to change this multiple times and reverted everything back again (see #3210). Adding this parameter allows for customizations made by users without having to modify the plugin itself. Should there be any new developments in terms of special Twitch players with access tokens without embedded ads, then this parameter is a way to quickly test and adapt to that, without having to publish new releases and having to potentially revert stuff.

### DIFF
--- a/docs/cli/plugins/twitch.rst
+++ b/docs/cli/plugins/twitch.rst
@@ -80,9 +80,9 @@ segments from Twitch's HLS streams and pauses the stream output until regular co
 logic has seen several iterations since then, with the latest big overhaul in
 :ref:`Streamlink 1.7.0 in 2020 <changelog:streamlink 1.7.0 (2020-10-18)>`.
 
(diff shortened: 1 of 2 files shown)


bobvargas:
Hi with the new --twitch-accesstoken-param option where I can find a access token for the twitch frontpage option becuase I'm cofused with the chrome inspect option thanks.

PR #4966, opened 2022-11-13
plugins.tv8: filter out ad segments

fenopa:
I tried to add ad segment filtering thanks to bastimeyer's guidance here: https://github.com/streamlink/streamlink/discussions/4944


bastimeyer (commented):
Please don't add \@-mentions in commit messages. This will trigger notifications on every single fork.

bastimeyer:
Filtering should also only be applied when there's a stream discontinuity? Is this the case here?

fenopa:
> Filtering should also only be applied when there's a stream discontinuity. Is this the case here?

Yes, these segments comes under "#EXT-X-DISCONTINUITY" if that's what you mean.

PR #4967, opened 2022-11-13
plugins.tv8: filter out ad segments

fenopa:


bastimeyer:
Just had a look at the stream and the segment URLs. PR looks fine to me.

PR #4970, opened 2022-11-14
docs: add pipx install info

mkbloke:
- Adds info for using `pipx` to install into venv
- Clarifies installing master into a venv with `pip` and `pipx`

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -252,10 +252,13 @@ Version                              Installing
 .. _Specific tag/branch or commit: https://pip.pypa.io/en/stable/reference/pip_install/#git
 
 Virtual environment
-^^^^^^^^^^^^^^^^^^^
+-------------------
 
 Another method of installing Streamlink in a non-system-wide way is
 using `virtualenv`_, which creates a user owned Python environment instead.
+
+Install with ``virtualenv`` and ``pip`` commands
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. code-block:: bash
 
@@ -265,8 +268,11 @@ using `virtualenv`_, which creates a user owned Python environment instead.
     # Activate the environment
     source ~/myenv/bin/activate
 
-    # Install Streamlink in the environment
+    # *Either* install the latest Streamlink release from PyPI in the virtual environment
     pip install --upgrade streamlink
+
+    # *Or*, install the most up-to-date development version from master on GitHub
+    pip install --upgrade git+https://github.com/streamlink/streamlink.git
 
     # Use Streamlink in the environment
     streamlink ...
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -257,32 +257,60 @@ Virtual environment
...
+  The `pipx`_ command combines the functionality of the ``virtualenv`` and ``pip`` commands. It may be necessary to
+  install it first, either with a system package manager, or using ``pip``, as detailed in the `documentation <pipx_>`_.
 
-    # Activate the environment
-    source ~/myenv/bin/activate
+  By default, ``pipx`` will install packages into ``~/.local/pipx/venvs/PACKAGE_NAME`` and create symbolic links in
+  ``~/.local/bin``. If these directories do not exist, they will be created. If ``~/.local/bin`` is newly created,
+  logging out and logging back in again may update the ``$PATH`` environment variable. Sourcing the relevant shell
+  configuration file in the current session may also update ``$PATH``. Otherwise, follow the directions in the note
+  output by ``pipx`` to update ``$PATH`` on your system.
```
bastimeyer, line 294: This is way too specific for Streamlink's install page. Linking to the pipx docs is more than enough:
https://pypa.github.io/pipx/docs/

I'm also not sure if this is what it's doing on Windows or macOS, too, as I've never used pipx in my Windows or macOS VMs.

docs/install.rst
```diff
@@ -257,32 +257,60 @@ Virtual environment
...
 
-    # Use Streamlink in the environment
-    streamlink ...
+      # *Either* install the latest Streamlink release from PyPI in a virtual environment
+      pipx install streamlink
 
-    # Deactivate the environment
-    deactivate
+      # *Or*, install the most up-to-date development version from master on GitHub
+      pipx install --spec git+https://github.com/streamlink/streamlink.git streamlink
```
bastimeyer, line 302: The `--spec` arg is only used for the `run` command:
https://pypa.github.io/pipx/docs/

docs/install.rst
```diff
@@ -257,32 +257,60 @@ Virtual environment
...
+      # Use Streamlink in the environment
+      streamlink ...
+
+      # Deactivate the environment
+      deactivate
+
+      # Use Streamlink without activating the environment
+      ~/myenv/bin/streamlink ...
+
+- Install with ``pipx`` command
```
bastimeyer, line 285: If you want to add a `pipx` section, then it'd probably make more sense to add another subsection instead of splitting this up into a list with two items. `pipx` is just a wrapper for `virtualenv` and `pip`, so having a subsection makes more sense..

docs/install.rst
mkbloke: Ah, it looks like that changed a while back.  I should have checked against a more recent version of `pipx`.

docs/install.rst
mkbloke: I wasn't quite sure how to organise it.  Ideally, each would have been a sub section beneath the "Virtual environment" sub section, but I don't think that's possible?  So, maybe "Virtual environment with `pip`" and "Virtual environment with `pipx`" sub sections of "PyPI package and source code" sub section?

bastimeyer: Don't know what's the best. Let's have a look.

mkbloke:
I think the latest push fixes the stuff mentioned.  Hope the sub-section and sub-sub-section stuff is OK as arranged...

bastimeyer (approved):
Looks good, thanks.

The macOS note box at the bottom is probably unnecessary though and could be removed. With the new subsections, this makes it more confusing than actually useful.

PR #4974, opened 2022-11-16
cli: threaded read_stream()

### LINKED ISSUE #4642: cli: threaded read_stream() for undelayed process termination

bastimeyer:
Rewrite `streamlink_cli.main.read_stream()`, refactor progress thread data feeding, and move player polling into a separate thread which closes the reader once the player process terminates or gets killed.

This fixes the player polling issue, or rather the detection of its broken pipe while reading the stream, as stream read calls can stall for various reasons, e.g. due to segmented streams or the filtering of stream data which pauses the stream output and disables any timeouts.

----

Resolves #4642 

As mentioned in #4642, there are still no tests written for `read_stream()`. I will be thinking about writing ones for a bit first before I'm going to mark this PR as ready. I also haven't fully tested this in production yet, only with a simple player stdin pipe as output, and only on my Linux system.

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -7,16 +7,14 @@ import re
 import signal
 import sys
 from contextlib import closing, suppress
-from functools import partial
 from gettext import gettext
-from itertools import chain
 from pathlib import Path
+from threading import Event, Lock, Thread
 from time import sleep
-from typing import Any, Dict, Iterator, List, Optional, Type, Union
+from typing import Any, Dict, List, Optional, Type, Union
 
 import streamlink.logger as logger
 from streamlink import NoPluginError, PluginError, StreamError, Streamlink, __version__ as streamlink_version
-from streamlink.compat import is_win32
 from streamlink.exceptions import FatalPluginError
 from streamlink.plugin import Plugin, PluginOptions
 from streamlink.stream.stream import Stream, StreamIO
@@ -262,7 +260,7 @@ def output_stream_http(
 
         if stream_fd and prebuffer:
             log.debug("Writing stream to player")
-            read_stream(stream_fd, server, prebuffer, formatter)
+            read_stream(stream_fd, server, prebuffer)
 
         if not continuous:
             break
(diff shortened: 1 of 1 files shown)


bastimeyer:
I've updated and rewritten the PR changes. Please check.

The player polling now stops immediately via a `threading.Event` and `subprocess.Popen().poll()` instead of `subprocess.Popen().wait()`, and "Player closed" gets logged once in both cases, when the `write()` call fails (due to a broken pipe, etc) or when the player process was terminated/killed. I've also fixed the incorrect `stream.closed` check in the read-loop, which led to stream data not being fully read from the buffer when the stream ended regularly. The polling thread also did not get closed gracefully when the stream ended this way.

Still no tests though. I made some efforts, but it's a lot of boilerplate code dealing with three threads (main test thread, CLI "main" read thread and player polling thread) and testing all the timings between all possible scenarios.

bastimeyer:
Going to close this... I will open a new PR with a complete code refactor of the `read_stream()` method that moves the logic into a dedicated module with classes that can actually be tested instead of the current mess. Not finished yet though, but I'm 100% sure that I don't want to continue this work here, and I don't want to rebase again.

PR #4989, opened 2022-11-22
ci.github: temporarily downgrade Sphinx to 5.2.3

### LINKED ISSUE #4988: docs: `:option:` directive incorrectly links to man-page since 5.1.0
### Checklist

### Streamlink version

### Description

Not sure yet what exactly has changed (haven't had a look at the docs config and depedencies yet), but all links of CLI args now point towards the man page, which is supposed to be an orphan page.

https://github.com/streamlink/streamlink.github.io/commit/8c43e01ffba911423748fd3903b8daf83b91d39e

The CLI args are built via the custom `ext_argparse.py` sphinx extension which defines all the
```
.. option:: ...

   ...

```
directives which then get used as cross-reference targets by various
```
:option:`...`
```
directives.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-option

Since the man page also executes the custom `ext_argparse.py` extension for listing the available CLI arguments, there seems to be a priority conflict, or orphan pages don't get ignored anymore.

A fix could be moving the man page into a separate docs build, so that it's not an orphan page anymore and instead completely isolated (which it's supposed to be anyway).

### Debug log

```text
https://github.com/streamlink/streamlink/actions/runs/3461942676/jobs/5780273301
```

bastimeyer:
Resolves #4988 

I've created a bug report upstream, so if this gets fixed, I will be notified, so we can remove the downgraded version again. Only downgrading on the CI side, as I don't want to set the restriction in the docs-requirements.txt.

We should probably take a look at releasing 5.1.1 due to this issue.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,7 @@ jobs:
       - name: Install dependencies
         run: |
           ./script/install-dependencies.sh
-          python -m pip install -r docs-requirements.txt
+          python -m pip install -r docs-requirements.txt sphinx==5.2.3
       - name: Build
         run: make --directory=docs html man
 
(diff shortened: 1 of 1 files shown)


bastimeyer:
I got confused for a second, because the docs preview build on netlify was linking to the man page. Totally expected though, because I only changed the github CI config :man_facepalming:

PR #4990, opened 2022-11-22
release: 5.1.1

bastimeyer:
I'll merge this tomorrow.


gravyboat (approved):

bastimeyer:
It's so frustrating how slow the Windows CI runners are. I always have to give the merge a headstart of at least a couple of minutes, because if the merge and tag-push both run at the same time, it's possible that this conflicts with the docs build when publishing.

PR #5011, opened 2022-12-03
plugins.twitch: better access token error handling

### LINKED ISSUE #5009: plugins.twitch: Show error message when the provided OAuth token is not valid
### Checklist

### Description

I'd imagine the playlist server returning some kind of 403 error or similar when the authorization fails, and if this is the case, could there be an error message related to tokens instead of "No playable streams found"?
Would help with scripting a lot. Thanks.

bastimeyer:
- Refactor `TwitchAPI` class:
  - Set `CLIENT_ID` as a class attribute (for test references)
  - Allow `call()` to pass custom headers (currently unused)
  - Remove unneeded `json.dumps()`
- Handle `PlaybackAccessToken` error response and log error messages
- Add `Options.clear()` for resetting plugin options state after tests
- Add access token API tests

----

Resolves #5009 

This adds error logging to the access token processing and also adds tests for the access token responses and how the `--twitch-api-header` and `--twitch-access-token-param` influence the access token request.

----

Small side-note in regards to the added `Options.clear()` method:

I've already mentioned somewhere in the `Plugin.bind()` removal in 5.0.0 (#4744 / #4768) that `Plugin.options` will also need to get fixed at some point. Similar to the previously bound class attributes, `Plugin.options` is a class attribute that gets shared accross all plugin classes unless it gets overridden by [`streamlink_cli.main.setup_plugin_args()`](https://github.com/streamlink/streamlink/blob/5.1.2/src/streamlink_cli/main.py#L783). This means that when streamlink_cli is not involved in plugin construction, e.g. in tests or when using the Streamlink session API directly, setting any kind of plugin option always requires clearing the options, which is bad. I'm pretty sure that there are a couple of tests which are polluting the shared plugin options dict. I'll check once I get the time.

The reason why I didn't touch `Plugin.options` at the time is that removing `Plugin.bind()` was already a lot of work. In hindsight, this should've been done in one go, because changing it will probably be a breaking API change, because of how the plugins will need to get initialized. Currently, the options are set by the session instance before plugins are initialized, and it expects the options dict to be available on the plugin class. Options should however only be set on plugin instances, so there's no shared options dict, neither between plugin classes, nor between different instances of the same plugin.

I will need to have a closer look at this and figure out a better solution before making any further comments. This PR should be fine though.

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -8,7 +8,6 @@ $notes :ref:`Low latency streaming <cli/plugins/twitch:Low latency streaming>` i
 """
 
 import argparse
-import json
 import logging
 import re
 import sys
@@ -251,20 +250,23 @@ class UsherService:
 
 
 class TwitchAPI:
+    CLIENT_ID = "kimne78kx3ncx6brgo4mv6wki5h1ko"
+
     def __init__(self, session):
         self.session = session
         self.headers = {
-            "Client-ID": "kimne78kx3ncx6brgo4mv6wki5h1ko",
+            "Client-ID": self.CLIENT_ID,
         }
         self.headers.update(**dict(session.get_plugin_option("twitch", "api-header") or []))
         self.access_token_params = dict(session.get_plugin_option("twitch", "access-token-param") or [])
         self.access_token_params.setdefault("playerType", "embed")
 
-    def call(self, data, schema=None):
+    def call(self, data, schema=None, **kwargs):
         res = self.session.http.post(
             "https://gql.twitch.tv/gql",
-            data=json.dumps(data),
-            headers=self.headers
+            json=data,
+            headers={**self.headers, **kwargs.pop("headers", {})},
+            **kwargs,
         )
 
         return self.session.http.json(res, schema=schema)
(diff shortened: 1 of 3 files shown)


src/streamlink/plugins/twitch.py
```diff
@@ -421,18 +423,30 @@ def access_token(self, is_live, channel_or_vod):
             validate.union_get("signature", "value"),
         )
 
-        return self.call(query, schema=validate.Schema(
-            {"data": validate.any(
+        return self.call(query, acceptable_status=(200, 401), schema=validate.Schema(
```
gravyboat, line 426: Are these the only status codes that are ever returned from a "successful" request?

gravyboat (approved):
Looks good to me.

src/streamlink/plugins/twitch.py
bastimeyer: 400 and 403 could maybe be added, but apart from that, I don't think anything else will be useful when trying to parse the `{"error": str, ...}` response schema. The remaining 4xx errors or any 5xx errors will be handled by the `HTTPSession` where it'll raise a `PluginError` instead.

gravyboat: Sounds good to me. Figured it was worth double checking just in case so there's no issue when this goes live that will require a second release 👍

PR #5017, opened 2022-12-08
tests: refactor thread handshakes

bastimeyer:
This cleans up and simplifies the HLS test mixins by implementing a generic thread-handshake class for synchronizing code execution between two threads: application logic (producer thread) and tests (consumer/main thread), with optional error capturing. The previous implementation was using a similar setup in multiple different classes. This is now defined with a proper interface and proper method names, for much better readability of the tests.

I've also added support for async coroutines in case this is required at some point. And I added tests for the new `tests.testutils.handshake.Handshake` class, so that it doesn't get tested implicitly via the other tests.

This work is required for my code refactoring of the CLI's `read_stream()` stuff (see #4974 - not yet finished), because I didn't want to redefine the same handshake stuff over and over again. Including this work in a single PR would've been a bit too much, so that's why I'm opening this one here first.

### DIFF
--- a/tests/mixins/stream_hls.py
+++ b/tests/mixins/stream_hls.py
@@ -8,6 +8,7 @@ import requests_mock
 
 from streamlink import Streamlink
 from streamlink.stream.hls import HLSStream, HLSStreamWriter as _HLSStreamWriter
+from tests.testutils.handshake import Handshake
 
 
 TIMEOUT_AWAIT_READ = 5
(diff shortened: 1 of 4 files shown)


tests/mixins/stream_hls.py
```diff
@@ -154,10 +144,7 @@ def run(self):
             # at least one read was attempted
             self.read_once.set()
             # only read once per step
-            self.read_wait.wait()
-            self.read_wait.clear()
-
-            try:
+            with self.handshake(OSError) as cm:
```
gravyboat, line 147: What is `cm` standing for here?

gravyboat (approved):
@bastimeyer My review is primarily based on the prior commit you pushed but I don't see any significant differences between them unless I'm missing something. Looks good, minor question around a name but otherwise 👍

tests/mixins/stream_hls.py
bastimeyer: - https://github.com/streamlink/streamlink/pull/5017/files#diff-e39ed681d4d3630ba69d1d7b2252e96a1ea51add69bcfd7678d0a7c002984ea9R26-R38
- https://docs.python.org/3/reference/compound_stmts.html#with

PR #5021, opened 2022-12-11
plugins.turkuvaz: fix url match

fenopa:
I adjusted it to the changed live stream url.


bastimeyer (commented):
Please also update the plugin tests

fenopa:
Oh sorry, is it OK now?

PR #5033, opened 2022-12-16
Move plugin options to plugin instances and remove global plugin arguments

bastimeyer:
These are breaking changes of the `Streamlink` session and `Plugin` APIs.

Opening this as a draft for now, even though it's pretty much final, because I don't think that this has to be merged now. It simply fixes an old design flaw and is not too important and should be included with other breaking changes in the future in one go.

----

Two separate changes:

### 1. Moving `Plugin.options` to plugin instances

As mentioned in #5011, turning `Plugin.options` from a plugin class attribute into a plugin instance attribute is related to the removal of the `Plugin.bind()` class method in #4744 / #4768, and ideally should've been part of the `5.0.0` release. While `5.0.0` removed `Plugin.bind()`, the breaking changes were actually in the `Streamlink` session class where `Plugin.bind()` was called internally, and compatibility wrappers were added to the `Plugin` constructor, with deprecation messages for old-style parameters. This here is another set of breaking changes on the `Streamlink` session class and `Plugin` class, but without any deprecations.

Just to recap what I've said in #5011 and what I've included in the commit messages of this PR, the `Plugin` class previously shared a common `Options` instance as the `Plugin.options` class attribute. This class attribute was then overridden by `streamlink_cli.main.setup_plugin_args()` where plugin arguments were processed and added to the argparser while reading default values of the plugin arguments that were added to a new `Options` instance.

This design is bad for several reasons. First, this makes `streamlink_cli` a "soft" dependency of the `Plugin` and `Streamlink` classes, as all plugin options would otherwise be shared between all plugins on the same `Options` instance. Second, multiple instances of the same plugin always have the same set of options set (not really important, but still bad). And third, tests and third party projects not using streamlink_cli need to manually override/clear the `Plugin.options` class attribute when initializing multiple plugins, which is bad.

Now, after these changes:

- The `Plugin` class can receive an optional `Options` instance as a parameter on its constructor, which streamlink_cli uses for setting up individual plugin options. Having it in the `Plugin` constructor is important, because plugin subclasses need to be able to read plugin options immediately (see Twitch and TwitchAPI), and not after initialization.
- Since plugin options are not stored on plugin classes anymore, `Streamlink.get_plugin_option()` and `Streamlink.set_plugin_option()` have been removed. These methods were useful for sharing state between different parts of Streamlink, e.g. plugins and streams (see Twitch and TwitchHLSStream), but this can be done in different ways. The PR therefore also adds the `keywords` parameter to `HLSStream.parse_variant_playlist()` to be able to pass keywords to the `{,Muxed}HLSStream` classes. In case of the Twitch plugin, these keywords are `disable_ads` and `low_latency` which are read from the plugin options instead of `session.get_plugin_option()`.

### 2. Removal of `Argument.is_global`

I've already had this on my radar for a while. Since the changed CLI methods already had to deal with global plugin arguments, I decided to include the removal of global plugin arguments in this PR.

The reason for the removal of global plugin arguments is simple: it adds unnecessary complexity for absolutely no gain. The value which a global plugin argument stores is already stored in the session options and should be read from there.

`is_global` was added when we removed individual plugin arguments for muxing subtitles, but no other global plugin arguments were ever added. The only benefit it had was being able to include a list of plugins in the docs which were making use of `--mux-subtitles`, as the global plugin arguments allowed for parsing this data. This additional data in the docs is now gone as well, but only affects four plugins. If we want to have a list of certain features a plugin uses, then this can be done by extending the regular plugin metadata.

What plugin implementors have to do to fix this is very simple, remove any global plugin arguments and change the option lookup, e.g. `self.get_option("mux_subtitles")` to `self.session.get_option("mux-subtitles")`.

### DIFF
--- a/src/streamlink/options.py
+++ b/src/streamlink/options.py
@@ -1,7 +1,4 @@
-import warnings
 from typing import Any, Callable, ClassVar, Dict, Iterator, Mapping, Optional, Sequence, Union
-
-from streamlink.exceptions import StreamlinkDeprecationWarning
 
 
 class Options:
@@ -106,7 +103,6 @@ class Argument:
         sensitive: bool = False,
         argument_name: Optional[str] = None,
         dest: Optional[str] = None,
-        is_global: bool = False,
         **options,
     ):
         """
@@ -117,7 +113,6 @@ class Argument:
         :param sensitive: Whether the argument is sensitive (passwords, etc.) and should be masked
         :param argument_name: Custom CLI argument name without plugin name prefix
         :param dest: Custom plugin option name
-        :param is_global: Whether this plugin argument refers to a global CLI argument (deprecated)
         :param options: Arguments passed to :meth:`ArgumentParser.add_argument()`, excluding ``requires`` and ``dest``
         """
 
@@ -131,15 +126,6 @@ class Argument:
         self.prompt = prompt
         self.sensitive = sensitive
         self._default = options.get("default")
-        self.is_global = is_global
-        if is_global:
-            warnings.warn(
-                "Defining global plugin arguments is deprecated. Use the session options instead.",
-                StreamlinkDeprecationWarning,
-                # set stacklevel to 3 because of the @pluginargument decorator
-                # which is the public interface for defining plugin arguments
-                stacklevel=3,
-            )
 
     @staticmethod
     def _normalize_name(name: str) -> str:
@@ -153,7 +139,7 @@ class Argument:
         return self._argument_name or self._normalize_name(f"{plugin}-{self.name}")
 
     def argument_name(self, plugin):
-        return f"--{self.name if self.is_global else self._name(plugin)}"
+        return f"--{self._name(plugin)}"
 
     def namespace_dest(self, plugin):
         return self._normalize_dest(self._name(plugin))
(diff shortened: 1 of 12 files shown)


bastimeyer:
Rebased to master and fixed merge conflict

bastimeyer:
Rebased to master and fixed merge conflict again. No other changes.

I'm not sure about the `keywords` keyword arg in `HLSStream.parse_variant_playlist` though. This is just an ugly hack on an already bad design, just so that `HLSStream` and `MuxedHLSStream` are able to receive custom data from plugins, because those can't be retrieved from the session anymore. Ideally, the entire `parse_variant_playlist` classmethod should be reworked and how the HLS stream instances are returned, but that's very much out of scope of this PR.

bastimeyer:
Rebased to master and resolved new merge conflicts:
- remove the now deprecated global plugin arguments (#5140)
- funimation plugin was removed (#5128)

bastimeyer:
Rebased to master again and fixed merge conflicts. No other changes.

vstavrinov:
Please remove set_plugin_option from docs: 

and put there explanation and example of how to set plugin options  for now

bastimeyer:
> Please remove set_plugin_option from docs:

Have a look at the **latest** docs.
https://streamlink.github.io/latest/api/session.html#streamlink.session.Streamlink


```
Plugin.options
self.options.get()
self.options.set()
```

https://github.com/streamlink/streamlink/blob/628e8ab910664b3548a51b6c7b83dcf53bbc7965/src/streamlink/plugin/plugin.py#L273

PR #5044, opened 2022-12-24
cli: refactor console.exit() calls

bastimeyer:
Just a small cleanup in regards to `console.exit()` in the main CLI module. This replaces these direct `console.exit()` calls with a `CLIExit` exception that gets raised instead and then gets caught down in the `main()` function. This fixes static code analysis and makes writing tests a bit easier. And it allows for additional cleanup code in case of an error, but nothing of that kind is being implemented here yet. Also no tests for the handling of the `CLIExit` exceptions themselves, but that's not too important.

Also decided to rewrite the stream-url and JSON tests, because they were pretty bad. Still no improvements of the code being tested, though, and also no error codes, but this can all be done later.

Opening this as a draft for now, because I didn't have the time to check everything in detail yet...

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -41,6 +41,10 @@ streamlink: Streamlink = None  # type: ignore[assignment]
 log = logging.getLogger("streamlink.cli")
 
 
+class CLIExit(Exception):
+    pass
+
+
 def get_formatter(plugin: Plugin):
     return Formatter(
         {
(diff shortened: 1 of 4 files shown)


bastimeyer:
Rebased to master in order to fix missing CI status checks with the new branch protection rules.

In regards to the changes, I will still have to take a closer look. So far, this is just a simple replacement of the `console.exit(...)` calls with some upgraded tests, but ideally we want to include exit codes.

bastimeyer:
Rebased and fixed merge conflict. Still need to take a closer look at the changes.

bastimeyer:
Will open a new PR at some point in the future... This one is old, requires a rebase and lots of more changes.

Issue #3416, opened 2020-12-14
kanald broken


fenopa:
## Plugin Issue

- [x] This is a plugin issue and I have read the contribution guidelines.
- [x] I am using the latest development version from the master branch.

### Reproduction steps / Explicit stream URLs to test

Example link 1 (VOD): https://www.kanald.com.tr/sadakatsiz/fragmanlar/sadakatsiz-10-bolum-fragmani
Example link 2 (Live): https://www.kanald.com.tr/canli-yayin

### Log output

```
streamlink 'https://www.kanald.com.tr/sadakatsiz/fragmanlar/sadakatsiz-10-bolum-fragmani'
[cli][info] Found matching plugin dogan for URL https://www.kanald.com.tr/sadakatsiz/fragmanlar/sadakatsiz-10-bolum-fragmani
[plugin.dogan][error] Could not find the contentId for this stream
error: No playable streams found on this URL: https://www.kanald.com.tr/sadakatsiz/fragmanlar/sadakatsiz-10-bolum-fragmani
```

### Additional comments, etc.
Thanks in advance

[Love Streamlink? Please consider supporting our collective. Thanks!](https://opencollective.com/streamlink/donate)

mkbloke:
I have work in progress on this.  I'll just need to validate the changes I've made across all the supported sites for the plugin.

fenopa:
> 
> 
> I have work in progress on this. I'll just need to validate the changes I've made across all the supported sites for the plugin.

Thank you, I can help with test if you want.

mkbloke:
@fenopa, many thanks for offering to help test.  I'll just do some preliminary tests and then post something up here for you to try.

mkbloke:
@fenopa, many thanks for helping to test.  Here's the fixed/updated plugin:

https://raw.githubusercontent.com/mkbloke/streamlink/dogan/src/streamlink/plugins/dogan.py

I've done some validation to make sure my changes work on the below sites/URL paths before pushing the branch to GitHub (which turned out to be very necessary), so hopefully you will not have any issues.  The plugin supports the following sites/URL paths now:

- cnnturk.com/canli-yayin
- cnnturk.com/video/*
- dreamturk.com.tr/canli
- dreamturk.com.tr/canli-yayin-izle
- kanald.com.tr/*
- teve2.com.tr/canli-yayin
- teve2.com.tr/filmler/*
- teve2.com.tr/programlar/*

Thanks.

fenopa:
[...]

PR #5052, opened 2022-12-29
plugins.dogan: fix broken streams

fenopa:
Live streams other than kanald were broken. I tired to fix them. I checked all the example links from https://github.com/streamlink/streamlink/issues/3416#issuecomment-748613571 and from test file. All links are working again.


bastimeyer:
Please see #5053 instead.

PR #5054, opened 2022-12-29
plugins.vkplay: new plugin

Inokenty90:
Plugin to support live streaming on vkplay.live

```
streamlink https://vkplay.live/sguhoman
[cli][info] Found matching plugin vkplay for URL https://vkplay.live/sguhoman
[cli][info] Available streams: 240p (worst), 360p, 480p, 720p, 1080p (best)
```

### DIFF
--- a/src/streamlink/plugins/vkplay.py
+++ b/src/streamlink/plugins/vkplay.py
@@ -1,5 +1,5 @@
 """
-$description Russian live-streaming platform.
+$description Russian live-streaming platform for gaming and esports, owned by VKontakte.
 $url vkplay.live
 $type live
 """
@@ -7,8 +7,7 @@ $type live
 import logging
 import re
 
-from streamlink.exceptions import NoStreamsError
-from streamlink.plugin import Plugin, PluginError, pluginmatcher
+from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
 from streamlink.stream.hls import HLSStream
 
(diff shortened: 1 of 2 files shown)


bastimeyer:
Thanks for the PR. Adding a plugin for this site should be fine.

However, some code clean up needs to be done here first before this can be merged. Allow me to add some changes to your PR branch, so this doesn't get unnecessarily complicated. I will also add support for API error messages and for stream metadata.

bastimeyer:
Metadata:
```
$ streamlink 'https://vkplay.live/d3l_by' best -j | jq .metadata
{
  "id": "4462052257156",
  "author": "d3l_by",
  "category": "Говорим и смотрим",
  "title": "ВЛОГ2 уже на ютубе!"
}
```

Offline:
```
$ streamlink 'https://vkplay.live/sguhoman'
[cli][info] Found matching plugin vkplay for URL https://vkplay.live/sguhoman
error: No playable streams found on this URL: https://vkplay.live/sguhoman
```

Non-existent:
```
$ streamlink 'https://vkplay.live/foo'
[cli][info] Found matching plugin vkplay for URL https://vkplay.live/foo
[plugins.vkplay][error] Blog not found
error: No playable streams found on this URL: https://vkplay.live/foo
```

src/streamlink/plugins/vkplay.py
```diff
@@ -0,0 +1,89 @@
...
+
+        self.category, self.title, streamdata = data
+        if not streamdata:
+            return
+
+        self.id, streams = streamdata
+
+        for streamtype, streamurl in streams:
+            if streamtype == "live_hls":
+                return HLSStream.parse_variant_playlist(self.session, streamurl)
```
mkbloke, line 86: Not that it seems to cause much of an issue in practice, but `streamurl` could be an empty string here.  I guess in practice that might not actually happen when `streamtype == "live_hls"` anyway, so I don't know if you'd consider it worth checking for.

mkbloke (approved):
Looks good to me.

src/streamlink/plugins/vkplay.py
bastimeyer: I had a check for this before I commited my changes but then removed it. Let's add it back, though.

PR #5061, opened 2023-01-01
ci.github: ubuntu-20.04 -> ubuntu-latest

bastimeyer:
Just wanted to update the master branch protection rules and noticed that the Ubuntu runners are still on 20.04.

Both Linux and Windows runners should use the latest OS release versions respectively. Upgrading from Ubuntu 20.04 to 22.04 shouldn't cause any issues.

After this has been merged, I will make Python 3.11 a requirement for PRs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,7 +24,7 @@ jobs:
       # when changing the build matrix and changing the number of test runners
       matrix:
         os:
-          - ubuntu-20.04
+          - ubuntu-latest
           - windows-latest
         python:
           - "3.7"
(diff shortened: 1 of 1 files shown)


bastimeyer:
> After this has been merged, I will make Python 3.11 a requirement for PRs.

The branch protection has to be temporarily disabled before this can be merged, as "ubuntu-20.04, 3.10" is still a requirement.

bastimeyer:
I've updated the branch protection rules now.

Changing the required status reports means that all unmerged PRs will need to be rebased onto the latest commit on the master branch, and all old/merged PRs will be shown with a yellow dot, as the required status didn't exist at the time.

PR #5063, opened 2023-01-02
plugins.ceskatelevize: update for ct24, sport and sportplus; add decko

### LINKED ISSUE #5055: plugins.ceskatelevize: URL updates needed for CT24 and CT Sport
### Checklist

### Streamlink version

Latest stable release

### Description

Ceska Televize offers five channels. For three of them, ceskatelevize.py still works fine (CT1, CT 2, CT :D/Art).
However, the remaining two channels have recently changed their URLs., i.e.

CT 24: formerly  https://www.ceskatelevize.cz/zive/ct24/ , now redirects to
https://ct24.ceskatelevize.cz/#live

CT Sport: formerly  https://www.ceskatelevize.cz/zive/sport/ , now redirects to
https://sport.ceskatelevize.cz/#live

These two new URLs will not be recognised by ceskatelevize.py anymore as it basically looks for www.ceskatelevize.cz/zive/...
It yould be great if you could update ceskatelevize.py to enable it again for CT 24 and CT Sport.
Please note that CT Sport is geoblocked for the Czech Republic. This is not the case for CT 24.

### Debug log

```text
Debug für CT24 with its now outdated URL:
[streamlinksrv][debug] Arguments:
[streamlinksrv][debug]  url=https://www.ceskatelevize.cz/zive/ct24/
[streamlinksrv][debug]  stream=['best']
[streamlinksrv][info] Found matching plugin ceskatelevize for URL https://www.ceskatelevize.cz/zive/ct24/
[streamlinksrv][error] No playable streams found on this URL: https://www.ceskatelevize.cz/zive/ct24/
[streamlinksrv][debug] Send Offline clip
```

mkbloke:
@duunsupen, could you check this is working for all channels by [sideloading](https://streamlink.github.io/cli/plugin-sideloading.html) the updated plugin, please?  I think it's probably working OK, but they seem to detect VPS based proxies, so I just get the short clip telling me the channel isn't available for all except CT24.

---

For some reason, the subdomains are both different to the main channels in terms of HTML and JSON data, and different to each other as well.  The API response is slightly different for `sport` too.  All a bit strange.

I didn't bother going through the process of finding the id for CT24, as it seems unlikely it'll change.

I notice that sometimes the streaming URL gets an empty response, but it seems intermittent.  I've seen that happen in the web browser too via the network tab on the dev console.

closes #5055

### DIFF
--- a/src/streamlink/plugins/ceskatelevize.py
+++ b/src/streamlink/plugins/ceskatelevize.py
@@ -5,25 +5,138 @@ $type live
 $region Czechia
 """
 
+import json
 import logging
 import re
+from urllib.parse import urlparse
 
-from streamlink.plugin import Plugin, PluginError, pluginmatcher
+from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
 from streamlink.stream.dash import DASHStream
 
 log = logging.getLogger(__name__)
 
 
-@pluginmatcher(re.compile(
-    r"https://(?:www\.)?ceskatelevize\.cz/zive/\w+"
-))
+@pluginmatcher(re.compile(r"https?://ct24\.ceskatelevize\.cz/"))
+@pluginmatcher(re.compile(r"https?://decko\.ceskatelevize\.cz/"))
+@pluginmatcher(re.compile(r"https?://sport\.ceskatelevize\.cz/"))
+@pluginmatcher(re.compile(r"https?://(?:www\.)?ceskatelevize\.cz/zive/\w+"))
 class Ceskatelevize(Plugin):
-    _re_playlist_info = re.compile(r"{\"type\":\"([a-z]+)\",\"id\":\"([0-9]+)\"")
+    schema_playlist = {
+        "playlist": [{
+            "streamUrls": {
+                "main": validate.url(),
+            },
+        }],
+    }
 
-    def _get_streams(self):
-        self.session.http.headers.update({"Referer": self.url})
-        schema_data = validate.Schema(
+    def get_stream_url(self, video_id):
+        url = self.session.http.post(
+            "https://www.ceskatelevize.cz/ivysilani/ajax/get-client-playlist/",
+            data={
+                "playlist[0][type]": "channel",
+                "playlist[0][id]": video_id,
+                "requestUrl": "/ivysilani/embed/iFramePlayer.php",
+                "requestSource": "iVysilani",
+                "type": "html",
+                "canPlayDRM": "false",
+            },
+            schema=validate.Schema(
+                validate.parse_json(),
+                {
+                    "url": validate.url(),
+                },
+                validate.get("url"),
+            ),
+        )
+
+        return self.session.http.get(
+            url,
+            schema=validate.Schema(
+                validate.parse_json(),
+                self.schema_playlist,
+                validate.get(("playlist", 0, "streamUrls", "main")),
+            ),
+        )
+
+    def get_ct24(self):
+        self.id = 24
+        self.title = "ČT24"
+        return self.get_stream_url(self.id)
+
+    def get_decko(self):
+        self.id = 5
+        self.title = "Déčko"
+        return self.get_stream_url(self.id)
+
+    def get_sport(self, content):
+        video_id, key, date = validate.Schema(
+            validate.parse_html(),
+            validate.xml_xpath_string(".//section[@id='live']/@data-ctcomp-data"),
+            str,
+            validate.parse_json(),
+            {
+                "items": [{
+                    "items": [{
+                        validate.optional("video"): {
+                            "data": {
+                                "source": {
+                                    "playlist": [{
+                                        "id": int,
+                                        "key": str,
+                                        "date": str,
+                                        "noDrmData": {
+                                            "id": int,
+                                            "key": str,
+                                            "drm": int,
+                                            "quality": str,
+                                            "assetId": str,
+                                        },
+                                    }],
+                                },
+                            },
+                        },
+                    }],
+                }],
+            },
+            validate.get(("items", 0, "items", 0, "video", "data", "source", "playlist", 0)),
+            validate.union_get(("noDrmData", "id"), ("noDrmData", "key"), "date"),
+        ).validate(content)
+
+        self.id = video_id
+        self.title = "ČT sport"
+
+        return self.session.http.post(
+            "https://playlist.ceskatelevize.cz/",
+            data={
+                "data": json.dumps(
+                    {
+                        "contentType": "live",
+                        "items": [{
+                            "id": video_id,
+                            "key": f"{key}",
+                            "assetId": "CT4DRM",
+                            "playerType": "dash",
+                            "date": f"{date}",
+                            "requestSource": "front-sport",
+                            "drm": 0,
+                            "quality": "web",
+                        }],
+                    },
+                    separators=(',', ':'),
+                ),
+            },
+            schema=validate.Schema(
+                validate.parse_json(),
+                {
+                    "RESULT": self.schema_playlist,
+                },
+                validate.get(("RESULT", "playlist", 0, "streamUrls", "main")),
+            ),
+        )
+
+    def get_channel(self, content):
+        data = validate.Schema(
             validate.parse_html(),
             validate.xml_xpath_string(".//script[@id='__NEXT_DATA__'][text()]/text()"),
             str,
@@ -31,105 +144,67 @@ class Ceskatelevize(Plugin):
             {
                 "props": {
                     "pageProps": {
-                        "data": {
+                        validate.optional("data"): {
                             "liveBroadcast": {
-                                # "id": str,
-                                "current": validate.any(None, {
-                                    "channel": str,
-                                    "channelName": str,
-                                    "legacyEncoder": str,
-                                }),
-                                "next": validate.any(None, {
-                                    "channel": str,
-                                    "channelName": str,
-                                    "legacyEncoder": str,
-                                })
-                            }
-                        }
-                    }
-                }
-            },
-            validate.get(("props", "pageProps", "data", "liveBroadcast")),
-            validate.union_get("current", "next"),
-        )
-
-        try:
-            data_current, data_next = self.session.http.get(
-                self.url, schema=schema_data)
-        except PluginError:
-            return
-
-        log.debug(f"current={data_current!r}")
-        log.debug(f"next={data_next!r}")
-
-        data = data_current or data_next
-        video_id = data["legacyEncoder"]
-        self.title = data["channelName"]
-
-        _hash = self.session.http.get(
-            "https://www.ceskatelevize.cz/v-api/iframe-hash/",
-            schema=validate.Schema(str))
-        res = self.session.http.get(
-            "https://www.ceskatelevize.cz/ivysilani/embed/iFramePlayer.php",
-            params={
-                "hash": _hash,
-                "origin": "iVysilani",
-                "autoStart": "true",
-                "videoID": video_id,
-            },
-        )
-
-        m = self._re_playlist_info.search(res.text)
-        if not m:
-            return
-        _type, _id = m.groups()
-
-        data = self.session.http.post(
-            "https://www.ceskatelevize.cz/ivysilani/ajax/get-client-playlist/",
-            data={
-                "playlist[0][type]": _type,
-                "playlist[0][id]": _id,
-                "requestUrl": "/ivysilani/embed/iFramePlayer.php",
-                "requestSource": "iVysilani",
-                "type": "html",
-                "canPlayDRM": "false",
-            },
-            headers={
-                "x-addr": "127.0.0.1",
-            },
-            schema=validate.Schema(
-                validate.parse_json(),
-                {
-                    validate.optional("streamingProtocol"): str,
-                    "url": validate.any(
-                        validate.url(),
-                        "Error",
-                        "error_region"
-                    )
-                }
-            ),
-        )
-
-        if data["url"] in ["Error", "error_region"]:
-            log.error("This stream is not available")
-            return
-
-        url = self.session.http.get(
-            data["url"],
-            schema=validate.Schema(
-                validate.parse_json(),
-                {
-                    "playlist": [{
-                        validate.optional("type"): str,
-                        "streamUrls": {
-                            "main": validate.url(),
-                        }
-                    }]
+                                "id": str,
+                                "current": validate.none_or_all(
+                                    {
+                                        "id": str,
+                                        "channelName": str,
+                                        "isPlayable": bool,
+                                    },
+                                ),
+                                "next": validate.none_or_all(
+                                    {
+                                        "id": str,
+                                        "channelName": str,
+                                        "isPlayable": bool,
+                                    },
+                                ),
+                            },
+                        },
+                    },
                 },
-                validate.get(("playlist", 0, "streamUrls", "main"))
-            )
-        )
-        return DASHStream.parse_manifest(self.session, url)
+            },
+            validate.get(("props", "pageProps")),
+        ).validate(content)
+
+        if not data:
+            return
+
+        log.debug(f"data={data}")
+        data = data.get("data").get("liveBroadcast")
+        self.id = data.get("id")
+
+        data = data.get("current") or data.get("next")
+        if not data:
+            return
+
+        self.title = data.get("channelName")
+
+        return self.get_stream_url(self.id)
+
+    def _get_streams(self):
+        res = self.session.http.get(self.url)
+
+        if "://ct24" in res.url:
+            url = self.get_ct24()
+        elif "://decko" in res.url:
+            url = self.get_decko()
+        elif "://sport" in res.url:
+            url = self.get_sport(res.content)
+        else:
+            url = self.get_channel(res.content)
+
+        if url:
+            res = self.session.http.head(url, allow_redirects=True)
+            log.debug(f"res.url={res.url}")
+            p = urlparse(res.url).path
+            if not p.split("/")[-1].startswith(str(self.id)):
+                log.error("This stream is not available")
+                return
+            else:
+                return DASHStream.parse_manifest(self.session, res.url)
 
 
 __plugin__ = Ceskatelevize
--- a/tests/plugins/test_ceskatelevize.py
+++ b/tests/plugins/test_ceskatelevize.py
@@ -6,23 +6,17 @@ class TestPluginCanHandleUrlCeskatelevize(PluginCanHandleUrl):
     __plugin__ = Ceskatelevize
 
     should_match = [
-        "https://www.ceskatelevize.cz/zive/ct1/",
-        "https://www.ceskatelevize.cz/zive/ct2/",
-        "https://www.ceskatelevize.cz/zive/ct24/",
-        "https://www.ceskatelevize.cz/zive/ct26/",
-        "https://www.ceskatelevize.cz/zive/ct27/",
-        "https://www.ceskatelevize.cz/zive/ct28/",
-        "https://www.ceskatelevize.cz/zive/ct31/",
-        "https://www.ceskatelevize.cz/zive/ct32/",
-        "https://www.ceskatelevize.cz/zive/decko/",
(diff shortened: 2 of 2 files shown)


src/streamlink/plugins/ceskatelevize.py
```diff
@@ -5,131 +5,168 @@
...
+        "playlist": [{
+            "streamUrls": {
+                "main": validate.url(),
+            },
+        }],
+    }
 
-    def _get_streams(self):
-        self.session.http.headers.update({"Referer": self.url})
+    def get_stream_url(self, id):
```
bastimeyer, line 28: `id` shadows the built-in function. I usually avoid this, as it's showing up as a warning in pycharm via its default linting rules. Same below. Optional though, and just wanted to mention it.

src/streamlink/plugins/ceskatelevize.py
```diff
@@ -5,131 +5,168 @@
...
+            },
+            validate.get(("props", "pageProps", "data", "liveBroadcast")),
+            validate.union_get("id", ("current", "channelName")),
         )
-        return DASHStream.parse_manifest(self.session, url)
+
+        try:
+            id, channel_name = self.session.http.get(self.url, schema=schema_data)
+        except PluginError:
+            return
```
bastimeyer, line 153: I know you're just continuing the pattern from before, but it's probably not a good idea to suppress validation errors (PluginError) unless there is some fallback mechanic implemented, which there is none of that here.

bastimeyer (commented):
The pluginmatcher should be changed.

First, `https://www.ceskatelevize.cz/#live` is an invalid link, and second, `https://ct24.ceskatelevize.cz/` should work without the URL hash, as it doesn't change anything in regards to the initial HTTP request of the plugin.

Also, those do not match (which are links from the channel selection dropdown):
- `https://sport.ceskatelevize.cz/zive-vysilani#live`
- `https://decko.ceskatelevize.cz/zive` (ignore this if you don't want to support this channel)

Just split the pluginmatcher into multiple ones if you already need to make a distinction between regular channels and special case channels. Then check which matcher matched.

duunsupen:
Hi Guys, unfortunately I wasn't able to get it to work.
I use an Enigma2 STB with OpenHDF 7.0 (Python 3.97) took the ceskatelvize.py plugin from here:
https://github.com/mkbloke/streamlink/blob/ceskatelevize/src/streamlink/plugins/ceskatelevize.py and put it in 
/home/root/.config/, then restarted the box. When opening https://ct24.ceskatelevize.cz/#live I got the typical Streamlink message screen in Enigma2 when something doesn't work ("Offline"). See debug log below.

I am probably not much of a help here as I am either doing something wrong or used the wrong plugin or so...

root@gbquad4k:~# /usr/sbin/streamlinksrv manualstart debug
[streamlinksrv][info]  Mon Jan  2 21:45:08 2023 Server (1.8.3) started
[streamlinksrv][debug] Host:           gbquad4k
[streamlinksrv][debug] Port:           8088
[streamlinksrv][debug] OS:             Linux-4.1.20-1.9-armv7l-with-glibc2.34
[streamlinksrv][debug] Python:         3.9.7
[streamlinksrv][info]  Streamlink:     0.0.0+unknown
[streamlinksrv][debug] youtube-dl:     2021.12.17
[streamlinksrv][debug] optparser:      0.3.0
[streamlinksrv][debug] Requests(2.26.0), Socks(1.7.1), Websocket(1.2.1)
[session][debug] Plugin ceskatelevize is being overridden by /home/root/.config/streamlink/plugins/ceskatelevize.py
[streamlinksrv][error] Got exception: local variable 'resolved_url' referenced before assignment
['Traceback (most recent call last):', '  File "/usr/sbin/streamlinksrv", line 794, in Stream', '    config_files, plugin = setup_config_files(streamlink, url)', '  File "/usr/sbin/streamlinksrv", line 758, in setup_config_files', '    return (config_files, pluginclass(streamlink, resolved_url))', "UnboundLocalError: local variable 'resolved_url' referenced before assignment"]
[streamlinksrv][debug] Send Offline clip

mkbloke:
Thanks for trying.  I'd forgotten you were using streamlinksrv.  I'll get the plugin working under Streamlink, but what you are running and the Streamlink version in use is an unknown for us, that we cannot support.

When the plugin is working and merged into Streamlink master, you might have to contact the maintainers of streamlinksrv to get it working for you.

bastimeyer:
> [session][debug] Plugin ceskatelevize is being overridden by /home/root/.config/streamlink/plugins/ceskatelevize.py

As you can see, the plugin was sideloaded by the Streamlink session instance.

> [streamlinksrv][error] Got exception: local variable 'resolved_url' referenced before assignment

This, however, is not a Streamlink issue. You are using a third party project which embeds/implements Streamlink, and we don't support that here. One hint, `local variable 'resolved_url' referenced before assignment"` suggests that this "streamlinksrv" project has copied stuff from an old version of the `streamlink`/`streamlink_cli` package and didn't update stuff after the latest major version upgrades of Streamlink, so your software is incompatible. You'll have to figure this out on your own or contact the maintainer of that project.

bastimeyer:
> /home/root/.config/streamlink/plugins/

This is a deprecated path, btw. The actual plugin path is `${XDG_DATA_HOME:-${HOME}/.local/share}/streamlink/plugins/`.
https://streamlink.github.io/cli/plugin-sideloading.html

duunsupen:
"When the plugin is working and merged into Streamlink master, you might have to contact the maintainers of streamlinksrv to get it working for you."

That's what I am going to to in case it is needed.
Thanks a lot to all of you.

src/streamlink/plugins/ceskatelevize.py
bastimeyer: @mkbloke Could you please quickly fix this once you have the time, so we can merge this?

mkbloke:
Sorry for the delay.  I have fixed it up now, so I *think* everything should be resolved.  I've added support for Decko and the sometimes on-air SportPlus (ct29).  When SportPlus isn't on-air `https://www.ceskatelevize.cz/zive/ct29/` redirects to `/zive/`.  This redirect also occurs for `/zive/anything-unknown`.

src/streamlink/plugins/ceskatelevize.py
```diff
@@ -5,131 +5,177 @@
...
 
-log = logging.getLogger(__name__)
 
-
-@pluginmatcher(re.compile(
-    r"https://(?:www\.)?ceskatelevize\.cz/zive/\w+"
-))
+@pluginmatcher(re.compile(r"https?://ct24\.ceskatelevize\.cz/"))
+@pluginmatcher(re.compile(r"https?://decko\.ceskatelevize\.cz/"))
+@pluginmatcher(re.compile(r"https?://sport\.ceskatelevize\.cz/(?:zive-vysilani)?#live"))
```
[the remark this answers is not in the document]
mkbloke: I see what you're saying - the page data is still the same with or without the hash.  I think it will still need to be included to match if it's used and to avoid other unintended matches?  For that reason, I've settled on:
```
@pluginmatcher(re.compile(r"https?://sport\.ceskatelevize\.cz/(?:zive-vysilani)?(?:#live$)?$"))
```
Does that seem reasonable?

src/streamlink/plugins/ceskatelevize.py
```diff
@@ -5,131 +5,177 @@
...
+        hostname = urlparse(self.url).hostname
+
+        if hostname.startswith("ct24"):
+            url = self.get_ct24()
+        elif hostname.startswith("decko"):
+            url = self.get_decko()
+        elif hostname.startswith("sport"):
+            url = self.get_sport()
+        else:
+            url = self.get_channel()
```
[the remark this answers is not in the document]
mkbloke: Good point, although I've changed this now anyway - for reasons I'll explain when I've pushed my new changes...

src/streamlink/plugins/ceskatelevize.py
```diff
@@ -5,131 +5,178 @@
...
+        data = data.get("current")
+        if not data:
+            return
+
+        self.title = data.get("channelName")
+
+        return self.get_stream_url(self.id)
+
+    def _get_streams(self):
+        self.res = self.session.http.get(self.url)
```
bastimeyer, line 167: Please don't store HTTP responses on the plugin instance. The garbage collector can't let go of this, and it'll be kept in memory, which is unnecessary. It's also not a good style. Just pass the response to the methods as an argument.

src/streamlink/plugins/ceskatelevize.py
```diff
@@ -5,131 +5,178 @@
...
+
+        if "://ct24" in self.res.url:
+            url = self.get_ct24()
+        elif "://decko" in self.res.url:
+            url = self.get_decko()
+        elif "://sport" in self.res.url:
+            url = self.get_sport()
+        else:
+            url = self.get_channel()
```
bastimeyer, line 176: As said, it's better to check via the `self.matches` list, because it's independent of the actual URLs. The only downside (right now) is the numeric indexing, but that's not too bad.

Or is there a specific reason to check the response URL due to redirections?

mkbloke:
One thing about matching on the subdomain is that you need to follow the redirect first, as Sport and Decko are both linked from `/zive` as `/zive/sport` and `/zive/decko' respectively.  The links at the bottom of `https://www.ceskatelevize.cz/` link directly to the subdomains though.  Talking of which, I see Art is linked via a subdomain there too, so I might need to add that if the live channel is also available on the subdomain.  I can't check that until Art is back on-air though.

I've also had to modify the schema for off-air channels to accommodate different responses.

I see right now there are even more sports channels: ct25; ct28; ct29; ct31 and ct32 all linked off `/zive/`.  None of them are on-air until later this evening though, so I can't check them right now.

Given all of the above, I've just tagged this as WIP for now.

src/streamlink/plugins/ceskatelevize.py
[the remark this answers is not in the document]
bastimeyer: > I think it will still need to be included to match if it's used and to avoid other unintended matches?

No, because of the TLD which contains "sport."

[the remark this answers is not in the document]
mkbloke: I think I'm missing something here.  Unlike `get_ct24()`, `get_sport()` parses the page content of `self.url`, so don't I want to limit the path matches?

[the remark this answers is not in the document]
bastimeyer: I don't understand what you're saying. What does this have to do with the hash? Hashes are for client-side DOM stuff and not for HTTP requests, so the `#live` hash should not be included, because it won't affect the response content which you're parsing. And other matchers can't be matched, because the TLD of every matcher is distinctive.

It's just a detail whether the `#live` hash included or not, but including hashes in general is just unnecessary.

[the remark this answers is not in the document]
mkbloke: What regex would you be using here?  Perhaps that would help to clear things up.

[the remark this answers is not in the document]
bastimeyer: One without the `#live` hash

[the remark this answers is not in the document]
mkbloke: OK, so I think you want:
```
@pluginmatcher(re.compile(r"https?://sport\.ceskatelevize\.cz/(?:zive-vysilani)?$"))
```
While I recognise what you're saying about `#live` being redundant for the purposes of the page data returned:
```
$ streamlink 'https://sport.ceskatelevize.cz/#live'
error: No plugin can handle URL: https://sport.ceskatelevize.cz/#live
```
My argument here would be that users expect to be able to copy/paste the URL from the browser address bar as is, regardless of whether or not bits of it are technically unnecessary.

I don't want to drag this out any further, so I can put the above regex in if you feel that's what it should be.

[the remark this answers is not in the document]
bastimeyer: > users expect to be able to copy/paste the URL from the browser

Yes, absolutely, but why the `$` anchor?

As said, it's just a detail for being able to input the URL without the hash. If this is causing confusion or any issues, just include the hash in the regex.

mkbloke:
> Art is linked via a subdomain there too, so I might need to add that if the live channel is also available on the subdomain.

Does not look to be so.

> SportPlus channels: ct25; ct28; ct29; ct31 and ct32 all linked off /zive/

All tested and working as far as I can tell (the manifest is returned, but the geo-restriction means I couldn't test full streaming).

bastimeyer:
I'll have another look later today. Thanks for the updates.

bastimeyer:
Geo-restricted streams sometimes seem to fail instead of showing the geo-restriction video:
```
$ streamlink --player-no-close 'https://www.ceskatelevize.cz/zive/ct1/' best
[cli][info] Found matching plugin ceskatelevize for URL https://www.ceskatelevize.cz/zive/ct1/
[cli][info] Available streams: 288p (worst), 404p, 576p (best)
[cli][info] Opening stream: 576p (dash)
[utils.named_pipe][info] Creating pipe streamlinkpipe-206770-1-2460
[utils.named_pipe][info] Creating pipe streamlinkpipe-206770-2-5180
[stream.ffmpegmux][error] Pipe copy aborted: /tmp/streamlinkpipe-206770-1-2460
[stream.dash][warning] Download of segment: 000001.m4s aborted
[cli][error] Try 1/1: Could not open stream <DASHStream ['dash', 'https://ivys-vst09-5.o2tv.cz/aa/05a2b1f97d6c1e2f879e917d8a584fc1/1673674842326/a60bed1601d42fb4f25e275dc0dbfb9c3/dna-61924494876945083-pc/mpd']> (No data returned from stream)
error: Could not open stream <DASHStream ['dash', 'https://ivys-vst09-5.o2tv.cz/aa/05a2b1f97d6c1e2f879e917d8a584fc1/1673674842326/a60bed1601d42fb4f25e275dc0dbfb9c3/dna-61924494876945083-pc/mpd']>, tried 1 times, exiting
```

I can't see anything in the API responses or any HTTP response headers which would indicate geo-restriction, so there's probably not much that can be done plugin-wise. This looks more like an issue with the DASH implementation, which kills the ffmpeg process to early.

Other than that, the plugin should be fine. `self.id` gets set to an integer in `get_ct24` and `get_decko` which is not the correct type, but since there's no type checking in those methods, it doesn't matter.

bastimeyer:
While testing with a Czech VPN, the sports stream didn't seem to work:
```
$ streamlink --player-no-close 'https://sport.ceskatelevize.cz' best
[cli][info] Found matching plugin ceskatelevize for URL https://sport.ceskatelevize.cz
error: Unable to open URL: https://ivys-cdn.o2tv.cz/cdn/uri/get/?contentType=live&drmOnly=false&expiry=1673679694&id=422&playerType=dash&quality=web&region=4&skipIpAddressCheck=false&userId=7fa9c1d1512da9e6&token=759f17e50f37147a337cb873ef5edc4dc671a520&encryption=wv (404 Client Error: Not found for url: https://ivys-vst01-1.o2tv.cz/atip/8a51e3901a19162c015395fa69c1fb11/1673676098496/422-tv-pc.mpd?userId=7fa9c1d1512da9e6)
```

The result is a 404 response if the `drm` attribute of the JSON `POST` payload is set to `0`. If set to `1`, the correct DASH manifest gets returned which the website uses (the website sets `drm` to `1`), but it's of course not supported by Streamlink. :/

src/streamlink/plugins/ceskatelevize.py
```diff
@@ -5,131 +5,178 @@
...
-import logging
+import json
 import re
 
-from streamlink.plugin import Plugin, PluginError, pluginmatcher
+from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
 from streamlink.stream.dash import DASHStream
 
-log = logging.getLogger(__name__)
```
back-to, line 15: don't remove all logging, add debug logs for `data`

src/streamlink/plugins/ceskatelevize.py
```diff
@@ -5,131 +5,178 @@
...
+            validate.xml_xpath_string(".//script[@id='__NEXT_DATA__'][text()]/text()"),
+            str,
+            validate.parse_json(),
+            {
+                "props": {
+                    "pageProps": {
+                        validate.optional("data"): {
+                            "liveBroadcast": {
+                                "id": str,
+                                "current": validate.none_or_all(
```
back-to, line 139: https://www.ceskatelevize.cz/zive/ct28/

might be None
the stream is in `next`

```python
{
  'id': '28', 
  'current': None,
  'next': {'id': '28-1673778600', 'channelName': 'ČT sport Plus', 'isPlayable': True}
}
```

```python
                                "next": validate.none_or_all({
                                    "id": str,
                                    "channelName": str,
                                    "isPlayable": bool,
                                }),
```

```python
data = data.get("current") or data.get("next")
```

src/streamlink/plugins/ceskatelevize.py
```diff
@@ -5,131 +5,178 @@
...
+            validate.xml_xpath_string(".//section[@id='live']/@data-ctcomp-data"),
+            str,
+            validate.parse_json(),
+            {
+                "items": [{
+                    "items": [{
+                        validate.optional("video"): {
+                            "data": {
+                                "source": {
+                                    "playlist": [{
```
back-to, line 80: `404 Client Error: Not found for url`

use **noDrmData**

```python
    def get_sport(self, content):
        video_id, key, date = validate.Schema(
            validate.parse_html(),
            validate.xml_xpath_string(".//section[@id='live']/@data-ctcomp-data"),
            str,
            validate.parse_json(),
            {
                "items": [{
                    "items": [{
                        validate.optional("video"): {
                            "data": {
                                "source": {
                                    "playlist": [{
                                        "id": int,
                                        "key": str,
                                        "date": str,
                                        "noDrmData": {
                                            "id": int,
                                            "key": str,
                                            "drm": int,
                                            "quality": str,
                                            "assetId": str,
                                        }
                                    }],
                                },
                            },
                        },
                    }],
                }],
            },
            validate.get(("items", 0, "items", 0, "video", "data", "source", "playlist", 0)),
            validate.union_get(("noDrmData", "id"), ("noDrmData", "key"), "date"),
        ).validate(content)

        self.id = video_id
        self.title = "ČT sport"
```

src/streamlink/plugins/ceskatelevize.py
```diff
@@ -5,131 +5,178 @@
...
+            url = self.get_ct24()
+        elif "://decko" in res.url:
+            url = self.get_decko()
+        elif "://sport" in res.url:
+            url = self.get_sport(res.content)
+        else:
+            url = self.get_channel(res.content)
+
+        if url:
+            return DASHStream.parse_manifest(self.session, url)
```
back-to, line 179: you can detect geo blocking

```python
        if url:
            res = self.session.http.head(url, allow_redirects=True)
            log.debug(f"{res.url}")
            p = urlparse(res.url).path
            if not p.split("/")[-1].startswith(str(self.id)):
                log.error("This stream is not available")
                return
            else:
                return DASHStream.parse_manifest(self.session, res.url)
```

back-to (changes requested):

src/streamlink/plugins/ceskatelevize.py
mkbloke: OK, thanks.  I did seem to find that CT28 and others worked when available (as noted in https://github.com/streamlink/streamlink/pull/5063#issuecomment-1382009964), but I'll put this in.

src/streamlink/plugins/ceskatelevize.py
mkbloke: Thanks.

src/streamlink/plugins/ceskatelevize.py
mkbloke: Thanks.

src/streamlink/plugins/ceskatelevize.py
mkbloke: I guess this isn't in fact just for CT28, but any channel where `current` may be empty.  It wasn't something I'd seen, so I don't know when this might happen, but clearly it does.  Cheers.

luxorko:
ct1, ct2 streams nor working

bastimeyer:
You are commenting on a merge pull request. Open a new plugin issue and fill in all the required details.

PR #5072, opened 2023-01-06
logger: capture warnings

bastimeyer:
See #4785

This makes Streamlink's root logger capture warnings (disabled by default, enabled via streamlink_cli) and it replaces various warning log messages for deprecated (user-facing) stuff with `FutureWarning`s.
https://docs.python.org/3/library/exceptions.html#warnings
https://docs.python.org/3/library/warnings.html#warning-categories

The reason for this is that this allows for finer control over the kind of deprecation messages (API stuff for devs, or other stuff for end-users) and it uses the standard interfaces for warnings, which is preferred. Warnings from dependencies can also be caught this way, but I'm not sure how useful this is, because I modified the default warning message format, so it doesn't include the file+line of the warning, because it's not useful for Streamlink's deprecation messages (`FutureWarning`). This can be changed later on if it causes problems. I only realized this while writing this PR text.

----

Warnings can always be filtered via Python's `-W` parameter (in addition to the regular log level of Streamlink's root logger):
https://docs.python.org/3/library/warnings.html#describing-warning-filters

```
$ python -m streamlink --https-proxy ...
[warnings][futurewarning] The https-proxy option has been deprecated in favor of a single http-proxy option

$ python -W ignore -m streamlink --https-proxy socks5h://localhost:1920

$ python -W ignore::FutureWarning -m streamlink --https-proxy socks5h://localhost:1920
```

----

The changes of the logger's warning messages are this, when printed from a caught warning, not from regular `warning()` calls:

```
[streamlink][warning] The https-proxy option has been deprecated in favor of a single http-proxy option
-->
[warnings][futurewarning] The https-proxy option has been deprecated in favor of a single http-proxy option
```

So instead of the logger name and the "warning" log level, it logs "warnings" as logger name and uses the name of the warning as log level name.

----

As a next step, dedicated warning subclasses could be added for reusable code and better deprecation messages with links to the docs, and with better warning names, etc.

----

Opening as a draft for now...

### DIFF
--- a/src/streamlink/exceptions.py
+++ b/src/streamlink/exceptions.py
@@ -42,6 +42,21 @@ class StreamError(StreamlinkError):
     """
 
 
+# https://stackoverflow.com/a/49797717
+class _StreamlinkWarningMeta(type):
+    def __new__(mcs, name, bases, namespace, **kw):
+        name = namespace.get("__name__", name)
+        return super().__new__(mcs, name, bases, namespace, **kw)
+
+
+class StreamlinkWarning(UserWarning, metaclass=_StreamlinkWarningMeta):
+    pass
+
+
+class StreamlinkDeprecationWarning(StreamlinkWarning):
+    __name__ = "StreamlinkDeprecation"
+
+
 __all__ = [
     "StreamlinkError",
     "PluginError",
@@ -49,4 +64,6 @@ __all__ = [
     "NoPluginError",
     "NoStreamsError",
     "StreamError",
+    "StreamlinkWarning",
+    "StreamlinkDeprecationWarning",
 ]
--- a/src/streamlink/logger.py
+++ b/src/streamlink/logger.py
@@ -1,11 +1,14 @@
 import logging
 import sys
+import warnings
 from datetime import datetime
 from logging import CRITICAL, DEBUG, ERROR, INFO, WARNING
 from pathlib import Path
 from sys import version_info
 from threading import Lock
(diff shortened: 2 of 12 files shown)


bastimeyer:
Decided to change the warning message format, because of warnings from dependencies which should contain all the info necessary to find the cause of those warnings.

The paths of Streamlink deprecation warnings could be suppressed in a future change. Warnings should be subclassed anyway, as mentioned, so this could be done in this step.

New format:
```
[warnings][name] message
  path:line
```

Examples:

Streamlink deprecation warning
```
$ streamlink --https-proxy foo
[warnings][futurewarning] The `https-proxy` option has been deprecated in favor of a single `http-proxy` option
  /home/basti/repos/streamlink/src/streamlink/session.py:104
```

InsecureRequestWarning (urllib3)
```
$ streamlink --http-no-ssl-verify '...'
[cli][info] Found matching plugin hls for URL ...
[warnings][insecurerequestwarning] Unverified HTTPS request is being made to host '...'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
  /home/basti/venv/streamlink-311/lib/python3.11/site-packages/urllib3/connectionpool.py:1045
```

bastimeyer:
Rebased to master and added `StreamlinkWarning`, which suppresses the origin of the warning, as it's not needed for Streamlink-related warnings/deprecations which get logged.

This should cover everything now.

PR #5076, opened 2023-01-07
session: refactor option getter+setter methods

bastimeyer:
This replaces the logic of the `Streamlink.{get,set}_option()` methods with a new `StreamlinkOptions` subclass which uses mapped option getters+setters that apply the custom option logic instead.

The motivation for this is that this improves the runtime of said methods, as option mappers are looked up in Python dictionaries in constant time, so option names don't have to be checked one after another. The old implementation is also really really bad code-wise and not extensible.

In order to do this, the `Options` class had to be refactored a bit and extended:
1. Previously, the internally stored option keys were normalized with argument names of the `argparse.Namespace` in mind, so dashes were replaced with underscores. Since Streamlink's session options are defined using dashes, I changed the normalization the other way around, to make it consistent. For various plugin options, this doesn't have any effect (e.g. see various "purge_credentials" options). This is all properly tested and no further changes are required.
2. The second commit adds support for mapped getter and setter functions, so that keys can be looked up in O(1). This is why the change of key-name normalization was necessary, to be able to define getters and setters in a normalized way that's consistent with the default options of the Streamlink session.

The third commit then finally cleans up the Streamlink session and moves the special option logic into the newly added `StreamlinkOptions` subclass, and it also adds new deprecation messages (see commit message). The individual getters and setters were not modified except for
- `http-disable-dh`, which now has a proper getter
- `https-proxy`, which now shows a deprecation message on its getter (still returns the "https" proxy, even though both "http" and "https" get set by the setter, because the proxies can also be defined by requests via env vars)

I only added new tests for the mapped options implementation and `http{,s}-proxy` getter. Some of the Streamlink session options are not actually tested, e.g. some HTTP session attributes, but all the getters and setters should be covered, and adding tests for simple setattr/getattr calls isn't too important. Current code coverage (can't permalink with commit-id, so this will change):
https://app.codecov.io/gh/streamlink/streamlink/blob/master/src/streamlink/session.py

----

If this PR gets merged, then this will cause merge conflicts with two of my other open (draft) PRs and those will need a quick update:
- #5072 (deprecation messages)
- #5033 (`Argument.is_global` removal and `Streamlink.{get,set}_plugin_option()` removal)

### DIFF
--- a/src/streamlink/options.py
+++ b/src/streamlink/options.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Iterator, Mapping, Optional, Sequence, Union
+from typing import Any, Callable, ClassVar, Dict, Iterator, Mapping, Optional, Sequence, Union
 
 
 class Options:
@@ -8,6 +8,9 @@ class Options:
     Note: Option names are normalized by replacing "_" with "-".
           This means that the keys ``example_one`` and ``example-one`` are equivalent.
     """
+
+    _MAP_GETTERS: ClassVar[Mapping[str, Callable[[Any, str], Any]]] = {}
+    _MAP_SETTERS: ClassVar[Mapping[str, Callable[[Any, str, Any], None]]] = {}
 
     def __init__(self, defaults: Optional[Mapping[str, Any]] = None):
         if not defaults:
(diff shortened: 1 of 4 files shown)


gravyboat (approved):
Looks great @bastimeyer, the code is much easier to parse as you go through it.

PR #5077, opened 2023-01-08
build: update build script

bastimeyer:
Use `build` to build sdist and wheels instead of calling setuptools directly, which is deprecated.

See
https://github.com/pypa/build

----

I've checked the sdist tarball and the built wheels, and everything looked okay to me. The win32 and win-amd64 wheels also have the correct GUI entry scripts (streamlink**w**).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -132,7 +132,7 @@ jobs:
         run: |
           ./script/install-dependencies.sh
           python -m pip install -r docs-requirements.txt
-          python -m pip install --upgrade Jinja2 wheel twine
+          python -m pip install --upgrade Jinja2 build wheel twine
       - name: Build shell completions
         run: ./script/build-shell-completions.sh
       - name: Build man page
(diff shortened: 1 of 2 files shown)


gravyboat:
Looks good, nice improvement!

bastimeyer:
There's also
```sh
gpg ... --passphrase="${RELEASE_KEY_PASSPHRASE}" ...
```
which should be changed to
```sh
gpg ... --passphrase-fd 0 ... <<< "${RELEASE_KEY_PASSPHRASE}"
```
but I didn't want to touch this. The PyPI user+pass are also passed as parameters to `twine` in the deploy script, and this doesn't look like it can be changed other than env vars, which is better, but not ideal either. Doesn't matter too much though on the CI side.

----

Btw, I don't have the password for the encrypted signing key (462165daefd5b94422824de59349cd55ba3e9629) and it's also not on the protonmail account. I believe only @beardypig has it. If nobody else has it, then I'll send him an email (I don't think he'll see this message)

gravyboat:
I double checked and I don't have the password for the signing key.

PR #5088, opened 2023-01-12
exceptions: update docs, fix NoPluginError and remove url arg from NoStreamsError

bastimeyer:
- Make `NoPluginError` inherit from `StreamlinkError` instead of `PluginError`
- Update and fix docstrings of all Streamlink exception classes
- Update list of exceptions in API docs
- Fix exports

----

`NoPluginError` is only raised [here](https://github.com/streamlink/streamlink/blob/ae7583473626a09f870fbdb71e4a988a83664ea6/src/streamlink/session.py#L439) and caught in the main CLI module where URLs get resolved. It shouldn't inherit from `PluginError`, because it's something entirely different.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -87,9 +87,11 @@ DASHStream
 Exceptions
 ----------
 
-Streamlink has three types of exceptions:
+Streamlink has multiple types of exceptions:
 
-.. autoexception:: streamlink.StreamlinkError
-.. autoexception:: streamlink.PluginError
-.. autoexception:: streamlink.NoPluginError
-.. autoexception:: streamlink.StreamError
+.. autoexception:: streamlink.exceptions.StreamlinkError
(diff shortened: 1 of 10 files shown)


bastimeyer:
https://deploy-preview-5088--streamlink.netlify.app/api.html#exceptions

bastimeyer:
Btw, no idea why `NoStreamsError` has the `url` argument. It's not used at all, and also unneeded considering that the `Plugin` instance gets initialized with the `url` argument.
https://github.com/streamlink/streamlink/blob/ae7583473626a09f870fbdb71e4a988a83664ea6/src/streamlink/plugin/plugin.py#L383-L384

And the empty streams error message is coming from the main CLI module:
https://github.com/streamlink/streamlink/blob/ae7583473626a09f870fbdb71e4a988a83664ea6/src/streamlink_cli/main.py#L554-L555

PR #5092, opened 2023-01-14
chore: ignore coverage in TYPE_CHECKING blocks

bastimeyer:
Accidentally forgot to suppress coverage of the `TYPE_CHECKING` block added in #5090.
https://app.codecov.io/gh/streamlink/streamlink/pull/5090

This PR adds `TYPE_CHECKING` to the list of excluded lines via `pyproject.toml`. But since pycharm seems to ignore this when building coverage reports via the `coverage` module, I also added the regular `# pragma: no cover` directive.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ exclude_lines = [
   "def __repr__",
   "raise NotImplementedError",
   "if __name__ == \"__main__\":",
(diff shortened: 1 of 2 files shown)


bastimeyer:
Btw, codecov is once again showing completely nonsense data, both for PRs and for the master branch (which is outdated). :man_facepalming:

PR #5097, opened 2023-01-17
plugins.tvp: add tvp.info

### LINKED ISSUE #5094: tvp.info
### Checklist

- [X] This is a plugin request and not a different kind of issue
- [X] [I have checked the list of open and recently closed plugin requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+request%22)

### Description

existing tvp.py plugin supports vod.tvp.pl and some of tvp.info materials except one.

for example there are some daily materials which are not supported by this plugin.

### Input URLs

streamlink https://www.tvp.info/33822794/program-minela-8
streamlink https://www.tvp.info/65275202/13012023-0823

mkbloke:
```
$ streamlink --http-proxy ... 'https://www.tvp.info/65275202/13012023-0823' 360p
[cli][info] Found matching plugin tvp for URL https://www.tvp.info/65275202/13012023-0823
[cli][info] Available streams: 224p (worst), 270p, 360p, 450p, 540p, 720p, 1080p (best)
[cli][info] Opening stream: 360p (hls)
[cli][info] Player closed
[cli][info] Stream ended

$ streamlink 'https://www.tvp.info/65275202/13012023-0823'
[cli][info] Found matching plugin tvp for URL https://www.tvp.info/65275202/13012023-0823
[plugins.tvp][error] The content is not available in your region
error: No playable streams found on this URL: https://www.tvp.info/65275202/13012023-0823

$ streamlink 'https://www.tvp.info/'
[cli][info] Found matching plugin tvp for URL https://www.tvp.info/
error: No playable streams found on this URL: https://www.tvp.info/
```
closes #5094

### DIFF
--- a/src/streamlink/plugins/tvp.py
+++ b/src/streamlink/plugins/tvp.py
@@ -1,6 +1,7 @@
 """
 $description Live TV channels and VODs from TVP, a Polish public, state-owned broadcaster.
-$url stream.tvp.pl
+$url tvp.info
+$url tvp.pl
 $type live, vod
 $notes Some VODs may be geo-restricted. Authentication is not supported.
 """
@@ -19,6 +20,7 @@ from streamlink.stream.http import HTTPStream
 log = logging.getLogger(__name__)
 
 
+@pluginmatcher(re.compile(r"""https?://(?:www\.)?tvp\.info/"""), name="tvp_info")
 @pluginmatcher(re.compile(
     r"""
         https?://
@@ -42,7 +44,7 @@ class TVP(Plugin):
                 "Connection": "close",
             },
             schema=validate.Schema(
-                re.compile(r"window\.__channels\s*=\s*(?P<json>\[.+?])\s*;", re.DOTALL),
+                validate.regex(re.compile(r"window\.__channels\s*=\s*(?P<json>\[.+?])\s*;", re.DOTALL)),
                 validate.none_or_all(
                     validate.get("json"),
                     validate.parse_json(),
@@ -91,7 +93,7 @@ class TVP(Plugin):
                 "Referer": self.url,
             },
             schema=validate.Schema(
-                re.compile(r"window\.__api__\s*=\s*(?P<json>\{.+?})\s*;", re.DOTALL),
+                validate.regex(re.compile(r"window\.__api__\s*=\s*(?P<json>\{.+?})\s*;", re.DOTALL)),
                 validate.get("json"),
                 validate.parse_json(),
                 {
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/tvp.py
```diff
@@ -169,8 +172,58 @@ def _get_vod(self, vod_id):
...
+    def _get_tvp_info_vod(self):
+        data = self.session.http.get(
+            self.url,
+            schema=validate.Schema(
+                validate.parse_html(),
+                validate.xml_xpath_string(".//script[contains(text(), 'window.__videoData')]"),
+                validate.none_or_all(
+                    str,
+                    re.compile(r"window\.__videoData\s*=\s*(?P<json>{.*?});", re.DOTALL),
+                    validate.get("json"),
```
bastimeyer, line 184: The validation of `re.Pattern` can return `None`, so the `get()` call afterwards will fail with a bad validation error message from the raised `AttributeError`.
https://github.com/streamlink/streamlink/blob/5.1.2/src/streamlink/plugin/api/validate/_validate.py#L135

Use `validate.regex(re.Pattern)` if you want to ensure that a compiled pattern matches. If it doesn't, then a proper validation error message will be raised.
https://github.com/streamlink/streamlink/blob/5.1.2/src/streamlink/plugin/api/validate/_validate.py#L216

The regex itself could probably be improved by adding `\s*` in between the end of the JSON payload and the semicolon character.

src/streamlink/plugins/tvp.py
```diff
@@ -169,8 +172,58 @@ def _get_vod(self, vod_id):
         if data.get("HLS"):
             return HLSStream.parse_variant_playlist(self.session, data["HLS"][0]["src"])
 
+    def _get_tvp_info_vod(self):
+        data = self.session.http.get(
+            self.url,
+            schema=validate.Schema(
+                validate.parse_html(),
+                validate.xml_xpath_string(".//script[contains(text(), 'window.__videoData')]"),
```
bastimeyer, line 180: This implicitly gets the node texts of all matches. May be better to add `[1]/text()`.

bastimeyer (commented):
Didn't check running the plugin with a VPN. Only checked the HTTP requests on their site with a VPN and the plugin changes should be fine according to that.

src/streamlink/plugins/tvp.py
bastimeyer: Please also update the other schemas. I made a mistake here when I rewrote the plugin a couple of months ago.

- https://github.com/streamlink/streamlink/blame/5.1.2/src/streamlink/plugins/tvp.py#L45
- https://github.com/streamlink/streamlink/blame/5.1.2/src/streamlink/plugins/tvp.py#L94

PR #5111, opened 2023-01-22
release: 5.2.0

bastimeyer:
Release date is set to tomorrow. I will have a look later today or tomorrow if there's some stuff missing, e.g. in regards to the new warnings logging and deprecation messages.

There's also the potential of a release failure due to the recent changes in #5077 and #5081, as mentioned...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Changelog
 
+## streamlink 5.2.0 (2023-01-23)
+
+Release highlights:
+
+- Added: new stream read/output loop, to be able to detect player process termination while stream output is paused (ad filtering, etc.) ([#5024](https://github.com/streamlink/streamlink/pull/5024))
+- Added: support for named plugin matchers ([#5103](https://github.com/streamlink/streamlink/pull/5103), [#5107](https://github.com/streamlink/streamlink/pull/5107))
+- Added: Python warnings capturing to streamlink logger and added `StreamlinkWarning` ([#5072](https://github.com/streamlink/streamlink/pull/5072))
+- Changed: deprecation log messages to warnings, and added missing warnings for [previously deprecated things](https://streamlink.github.io/deprecations.html) ([#5072](https://github.com/streamlink/streamlink/pull/5072))
+- Deprecated: usage of `validate.text` in favor of `str` ([#5090](https://github.com/streamlink/streamlink/pull/5090))
+- Improved: `Streamlink` session option getters/setters ([#5076](https://github.com/streamlink/streamlink/pull/5076))
+- Fixed: incorrect inheritance of `NoPluginError` and removed unneeded `url` parameter from `NoStreamsError` ([#5088](https://github.com/streamlink/streamlink/pull/5088))
+- Fixed: error handling in Twitch access token acquirement ([#5011](https://github.com/streamlink/streamlink/pull/5011))
+- Fixed: dogan plugin ([#5053](https://github.com/streamlink/streamlink/pull/5053))
+- Fixed: ceskatelevize plugin, added sport/sportplus/decko ([#5063](https://github.com/streamlink/streamlink/pull/5063))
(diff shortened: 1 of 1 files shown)


bastimeyer:
I'll have a look at the warning log spam of certain warnings tomorrow before publishing 5.2.0. Not sure yet which warning filters are required.

gravyboat:
@bastimeyer Do you want to merge this yourself then or do you want me to merge now that it's the 23rd UTC?

bastimeyer:
These things are completely unrelated. I was talking about `Warning`s being logged with the new logging implementation, not other random log warning messages like in #5113, which is just a last minute addition for people who don't understand ad filtering.

I will merge this myself anyway because of the Windows installer/portable and Linux AppImages.

bastimeyer:
> I'll have a look at the warning log spam

```py
>>> pprint(warnings.filters)
[('always', None, <class 'Warning'>, None, 0),
 ('always', None, <class 'PendingDeprecationWarning'>, None, 0),
 ('always', None, <class 'DeprecationWarning'>, None, 0),
 ('ignore', re.compile('Trying to detect', re.IGNORECASE), <class 'Warning'>, re.compile('charset_normalizer'), 0),
 ('ignore', None, <class 'urllib3.exceptions.DependencyWarning'>, None, 0),
 ('ignore', None, <class 'getpass.GetPassWarning'>, None, 0),
 ('default', None, <class 'DeprecationWarning'>, '__main__', 0),
 ('ignore', None, <class 'DeprecationWarning'>, None, 0),
 ('ignore', None, <class 'PendingDeprecationWarning'>, None, 0),
 ('ignore', None, <class 'ImportWarning'>, None, 0),
 ('ignore', None, <class 'ResourceWarning'>, None, 0),
 ('ignore', None, <class 'pkg_resources.PEP440Warning'>, None, 0),
 ('always', None, <class 'urllib3.exceptions.SecurityWarning'>, None, 0),
 ('default', None, <class 'urllib3.exceptions.SubjectAltNameWarning'>, None, 0),
 ('default', None, <class 'urllib3.exceptions.InsecurePlatformWarning'>, None, 0),
 ('default', None, <class 'urllib3.exceptions.SNIMissingWarning'>, None, 0),
 ('default', None, <class 'requests.exceptions.FileModeWarning'>, None, 0)]
```

Probably not worth changing any default filtering rules.
https://docs.python.org/3/library/warnings.html#the-warnings-filter

The new `StreamlinkWarning` (subclassed from `UserWarning`) gets raised at most once anyway, so no custom rule is necessary until something new gets added which could raise unwanted duplicates in certain cases. For users of the Streamlink API, they can add filters themselves in case multiple sessions with deprecated stuff get initialized for example, so this all is only relevant for Streamlink CLI.

`urllib3.exceptions.SecurityWarning` is set to `always` [and always has been](https://github.com/urllib3/urllib3/blame/1.26.14/src/urllib3/__init__.py#L89), so when `--http-no-ssl-verify` gets set, every insecure HTTPS request will trigger a warning. A bit annoying (and the reason why I had a look), but it doesn't matter. It's not worth changing it to `default` or `once` just to prevent spam for [any kind of TLS problem](https://github.com/urllib3/urllib3/blame/1.26.14/src/urllib3/exceptions.py#L185-L212).

gravyboat:
@bastimeyer I'm not concerned at all about spamming for TLS issues. If a site isn't using TLS the plugin should be removed any way. Whoever runs the site is clearly incapable.

bastimeyer:
PyPI upload failed
https://github.com/streamlink/streamlink/actions/runs/3990883680/jobs/6845136470#step:10:1
:(
#5081

I'll have a look at uploading the files manually

bastimeyer:
Release 5.2.1 was successful with the reverted PyPI deploy changes.
Switching to a token-based auth is however something that should be done in the future. I will do a test release on the PyPI test index when I get the time. Not sure when though.

PR #5113, opened 2023-01-23
stream.hls: warn when writing after discontinuity

bastimeyer:
A warning message should've been added a while ago...

This adds the `written_once` attribute to the base `Buffer` class instead of the `HLSStreamWriter` class, because `HLSStreamWrite` would be the wrong place should a multi-buffer `HLSStream` ever get implemented for being able to write multiple files when a discontinuity occurs.

### DIFF
--- a/src/streamlink/buffers.py
+++ b/src/streamlink/buffers.py
@@ -27,6 +27,7 @@ class Buffer:
         self.current_chunk = None
         self.closed = False
         self.length = 0
+        self.written_once = False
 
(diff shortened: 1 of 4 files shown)


gravyboat (approved):
I'm marking this as approved as I'm not sure if there's more you want to do based on your message in https://github.com/streamlink/streamlink/pull/5111#issuecomment-1399688726, Looks good though and the messages are clear.

Issue #5084, opened 2023-01-10
globoplay.com


guiworldtv:
### Checklist

- [X] This is a plugin request and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+request%22)

### Description

hey friends, how are you? I would like you to include access to Globoplay (Brazil's main streaming service) in the streamlink
this streaming works outside brazil, as most owners here must be in the united states
some contents needs a vpn
globoplay need a subscription, make a option do put a login and password to acess the content
thanks

### Input URLs

https://globoplay.globo.com/tv-globo/ao-vivo/9909627/
https://globoplay.globo.com/
https://globoplay.globo.com/globonews/ao-vivo/6461772/

back-to:
> some contents needs a vpn
globoplay need a subscription

ref https://github.com/streamlink/streamlink/issues/1229

PR #5118, opened 2023-01-24
New Plugin - GloboPlay

guiworldtv:
Whats up
im trying to do a version of the plugin of globoplay for streamlink,that doesn't exist yet, I hope this can help in some way, I want globoplay to be able to enter here soon


bastimeyer (changes requested):
Well, thanks for the PR, but I'm going to close this.

There's so much wrong with this plugin implementation that I won't bother giving it a proper review. To be honest, what you've posted here raises at least a thousand questions. There are also syntax errors, which means you haven't actually run and tested this at all.

You also already received your answer here
https://github.com/streamlink/streamlink/issues/5084#issuecomment-1383121403
And apparently, the site is using DRM for live streams, so this won't work with Streamlink.

PR #5124, opened 2023-01-27
chore: add templates for GitHub discussions

bastimeyer:
This adds templates to the discussions forum, similar to the issue templates.

The markdown header is mainly copied from the stickied thread which I will remove once merged:
https://github.com/streamlink/streamlink/discussions/4578

The checklist is a pseudo-checklist, so that the list doesn't get included in the discussion thread. The "description" header unfortunately gets included, as it's the title of the textarea, which is the only user input (required when using a template).

Both templates are idential, because there's not much difference between the two categories. The function of the Q&A category is already explained in its title description.

I've tested this on my fork repo, and everything worked just fine. The YML schema is still considered public beta, but I don't expect anything to break here, as it's basically the same as the issue template schema.
https://docs.github.com/en/discussions/managing-discussions-for-your-community/syntax-for-discussion-category-forms

Anything missing or wrong, or is there something that needs to be added or improved?

### DIFF
--- a/.github/DISCUSSION_TEMPLATE/general.yml
+++ b/.github/DISCUSSION_TEMPLATE/general.yml
@@ -1,0 +1,52 @@
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ## Thanks for using the discussions forum!
+
+        **This place is only meant for help and support questions when using Streamlink, which can be answered by the community.**
+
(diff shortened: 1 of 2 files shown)


gravyboat:
Very nice. I'm not sure it's "mean" enough for when people ask us about their third party apps that use Streamlink (I would prefer to instantly close these) but let's see how it does.

PR #5125, opened 2023-01-28
stream.hls: refactor and fix attribute list parser

bastimeyer:
This fixes and cleans up the M3U8Parser's attribute list regex+parser and some individual attribute parser methods, and it adds missing unit tests.

The old attribute list regex was incorrect and was also missing some stuff. This is now fixed and it's a bit more strict when parsing the attribute list, but it's not doing that 100% according to the spec either, because the previous one also wasn't (and shouldn't, to be able to support broken/invalid playlists).
https://www.rfc-editor.org/rfc/rfc8216#section-4.2

One key difference here is that the entire attribute list gets discarded if there's a parsing issue with one of the attributes in the attributes list. Previously, this was not the case and the already parsed attributes of a specific tag were returned and it kept parsing via `re.finditer()`, which could lead to invalid results. That's also the reason, why the comma separator has been added to the regex.

It also adds a couple of warning log messages, which can be useful, even though the input string is not included in the log message. I didn't add warnings for everything, though.

But before I'm opening this for merging, I will check and see if this breaks the parsing of actual production playlists.

### DIFF
--- a/src/streamlink/stream/hls_playlist.py
+++ b/src/streamlink/stream/hls_playlist.py
@@ -4,7 +4,6 @@ import math
 import re
 from binascii import unhexlify
 from datetime import datetime, timedelta
-from itertools import starmap
 from typing import Any, Callable, ClassVar, Dict, Iterator, List, Mapping, NamedTuple, Optional, Tuple, Type, Union
 from urllib.parse import urljoin, urlparse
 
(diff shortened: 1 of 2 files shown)


gravyboat:
@bastimeyer Is there any further testing you want to do or is this ready to go? Code looks good.

gravyboat (approved):

Issue #4182, opened 2021-11-17
plugins.schoolism: new schoolism site (still considered beta) not working


Knauj:
### Checklist

- [X] This is a plugin issue and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

new user are at schoolism are redirected to https://beta.schoolism.com/login
all schoolism video urls are changed accordingly.

"error: No plugin can handle URL: https://beta.schoolism.com/s/course/library/lessons/50"

### Debug log

```text
error: No plugin can handle URL: https://beta.schoolism.com/s/course/library/lessons/50
```

gravyboat:
We don't do plugin updates for website betas since the beta can be constantly changing, once it's no longer in a beta state someone can take a look if they feel like it.

mkbloke:
I was going to say more or less the same as above, but also that due to Schoolism being a subscriber site, it's even more unlikely that this will get fixed, unfortunately.

CasualRascal:
I would like to add that while it carries "beta" in the name, it is not an actual beta but the production website. There is no way to access the older website and the URLs do not contain the `beta.` subdomain. 

This makes the plugin completely broken, not just partly broken.

gravyboat:
> I would like to add that while it carries "beta" in the name, it is not an actual beta but the production website.

As long as it carries the beta name whether it's on the primary domain or not it's still subject to changes. If someone feels like updating the plugin they're welcome to do so and it can be side loaded.

AntoineC4Ci:
The schoolism URL does not incudes "beta" anymore. (but the ébeta" tag is still present on the website header).

Even if I am not sure about what i was trying to do, I tried to change the following command : 

_C:\Users[OMITTED]> streamlink https://schoolism.com/s/course/library/lessons/25 best -o "location and file name.mp4" --schoolism-email EMAIL --schoolism-password PASSWORD
error: No plugin can handle URL: https://schoolism.com/s/course/library/lessons/25_


Then i used a webbrowser extention that detects the following HLS URLs from the schoolism webpage : 
[...]

PR #5127, opened 2023-01-29
plugins.schoolism: deleted

back-to:
impossible to update without an account

closes https://github.com/streamlink/streamlink/issues/4182


bastimeyer:
Agreed, let's remove it.

Issue #5145, opened 2023-02-08
tools: replace flake8 with ruff


bastimeyer:
`ruff` is an extremely fast linter for Python written in Rust, and a drop-in replacement for `flake8`.

- https://github.com/charliermarsh/ruff
- https://beta.ruff.rs/docs/
- https://beta.ruff.rs/docs/faq/

Before I'm going to open a PR, I'd like to discuss this first.

----

Ruff is still in its early development, but I don't see a problem switching to it already, especially because some big/popular Python projects have already shifted towards it (see the links above).

**The main benefits of switching are**

1. it's very fast, with support for caching (not too relevant for Streamlink though)
2. its migration is very easy compared to other linters like Black
3. it's compatible with the default Black config (except for the line-length)
4. it has near identical rule sets compared to flake8 (= pyflakes + pycodestyle)
5. it has lots of rules from various other linters and flake8 plugins built in, e.g.
   - isort (its rules are slightly different from the currently used flake8-import-order plugin with the pycharm profile)
   - flake8-commas
   - flake8-quotes
   - flake8-bugbear
   - flake8-comprehensions
   - flake8-simplify
6. there's no mandatory ini-config nonsense and its config is defined in pyproject.toml, where it belongs
7. rules with auto-fix upgrades can be added gradually, which is very much preferred when transitioning and cleaning up old code
8. various editor integrations are available

**Potential concerns**

1. installing binary wheels (or compiling its Rust code from source) is required, but wheels are all available on PyPI, so installing on any platform is not an issue at all (none of that is relevant for packagers though)
2. it doesn't have a plugin system (yet), but it's not relevant for Streamlink anyway and also not necessary
3. some upstream configs from built-in plugins are not available yet, resulting in a slightly different code-style that can't be adjusted

----

So if I was going to open a PR with a replacement for flake8, I would very much like to also fix some still outstanding style issues with Streamlink's code base, namely:

- enforce double quotes everywhere
- enforce trailing commas everywhere
- set the number of empty lines after import statements to 2 (the default is 1, but it's stupid, considering that functions and classes on the global module scope must be separated by 2 lines)
- maybe also pep8-naming conventions, but I'm not sure about that yet

[...]

PR #5149, opened 2023-02-09
tools: replace flake8 with ruff

bastimeyer:
Step 1 of 3+x, as detailed in #5145 

This replaces `flake8` and plugins with `ruff`, sets a similar config, and updates the docs.

The import linting config is missing and will be added in a subsequent PR via ruff's `isort` "I" rules, which are slightly different and require lots of module code changes. This PR only changes the bare minimum for having a short diff without any code changes.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,12 +27,12 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.7"
+          python-version: "3.11"
       - name: Install dependencies
         run: ./script/install-dependencies.sh
-      - name: flake8
+      - name: ruff
         run: >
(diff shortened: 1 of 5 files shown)


bastimeyer:
Apparently there was some kind of hiccup on the side of netlify which let the docs preview build fail and now after the manual re-build there's no status check. Link to the rendered docs here:
https://deploy-preview-5149--streamlink.netlify.app/developing.html#code-style

gravyboat:
This looks good, I don't have time to give the second PR the look it deserves right now, will have to do it later tonight.

bastimeyer:
#5150 only includes trivial changes, so apart from scrolling through the entire diff, there's not much going on. The diff view on GitHub unfortunately is (still) terrible, and looking at the diff locally is just way more efficient.

PR #5157, opened 2023-02-12
docs: rewrite, fix and improve CLI tutorial

bastimeyer:
The main motivation of these changes is having a short linkable example of how to escape URLs in various command-line shells, because this question comes up often and gets incorrectly (and annoyingly) reported as a bug.

Since there'd be a lot of repetition when showing examples for different shells, I decided to add the `sphinx-design` extension, to be able to have support for tabs (and other things that might be useful in the future). It's recommended by the furo-theme author and it's maintained by the developer of the `myst-parser` extension which Streamlink already uses for markdown support in the docs.
https://sphinx-design.readthedocs.io/en/furo-theme/index.html

This PR also rewrites and clarifies some sections and fixes broken links. I also replaced the tutorial's example URL with something more neutral, because the previous URL was set by chrippa, who apparently was a starcraft fan in the early days of Twitch, hence the day9tv example.

The examples are not super strict, because on BASH for example, the `nullglob` option could be set, but that's mostly irrelevant for interactive use. I tried to cover the most common pitfalls though when pasting URLs.

Please see the docs preview.

----

Both `sphinx-design` and `myst-parser` extensions are currently holding back the upgrade to Sphinx 6 if the furo-theme would get bumped to the latest version. A new release of the extensions should be coming up shortly though (highly requested by lots of people). I will have a look at bumping furo and thus Sphinx to 6.0.0 which will drop support for py37 and which will finally drop unneeded bloat like jquery and underscore. :vomiting_face:

The py37 drop also made me have a look at the netlify docs-preview build config, but that one is unfortunately still stuck on py38, and they've even made their [build-image](https://github.com/netlify/build-image) repo closed source for development pace reasons, with eventual Python upgrades promised [on their forums](https://answers.netlify.com/t/python-3-11-or-newer/84566) :disappointed:.

### DIFF
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,6 +1,7 @@
 sphinx >=4.0.0
 furo ==2022.09.15
 myst-parser >=0.18.0
+sphinx-design >=0.3.0
 versioningit >=2.0.0, <3
 
 docutils-stubs
--- a/docs/cli/tutorial.rst
+++ b/docs/cli/tutorial.rst
@@ -1,40 +1,53 @@
 Tutorial
---------
+========
+
+Introduction
+------------
 
 Streamlink is a command-line application, which means that the commands described
-here should be typed into a terminal. On Windows, you have to open either the
-`Command Prompt`_, `PowerShell`_ or `Windows Terminal`_, on macOS open the `Terminal <macOS-Terminal>`_ app,
-and if you're on Linux or BSD you probably already know the drill.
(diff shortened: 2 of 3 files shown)


bastimeyer:
https://deploy-preview-5157--streamlink.netlify.app/cli/tutorial.html

gravyboat:
Looks really good and hopefully this will reduce questions, nice job.

PR #5184, opened 2023-02-21
stream.dash: fix and refactor

bastimeyer:
1. Use the [`datetime.timezone.utc`](https://docs.python.org/3/library/datetime.html#datetime.timezone.utc) stdlib class instance instead of the custom `isodate.UTC` class instance (differs in `dst()`, but that's irrelevant) (`datetime.UTC` is only available on 3.11+)
   - https://github.com/python/cpython/blame/v3.11.0/Lib/datetime.py#L2409
   - https://github.com/python/cpython/blame/v3.10.0/Lib/datetime.py#L2301
   - https://github.com/gweis/isodate/blob/0.6.1/src/isodate/tzinfo.py#L13-L46
2. Fix default param in `stream.dash_manifest.count_dt()` (should be the current time when initializing)
3. Rename constants for consistency reasons
4. Use dataclass for DASH `Segment`s, add typing information, rename `range` to `byterange` (similar to HLS's `Segment` (currently still a NamedTuple)) and update old `range` references

### DIFF
--- a/src/streamlink/plugins/vinhlongtv.py
+++ b/src/streamlink/plugins/vinhlongtv.py
@@ -7,10 +7,8 @@ $region Vietnam
 
 import logging
 import re
-from datetime import datetime
+from datetime import datetime, timezone
 from hashlib import md5
-
-from isodate import UTC  # type: ignore[import]
 
 from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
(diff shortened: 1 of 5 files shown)


bastimeyer:
The `isodate.UTC` implementation was added to `dash_manifest` in
https://github.com/streamlink/streamlink/commit/32fff155161e6fedfe7dbc0fdae92160a44f2034#diff-12241c9242a7153fd343167744c89f6a343ff0a2738cf531bbab991a9ae95b50
because it was previously using a custom UTC implementation due to py2

PR #5185, opened 2023-02-21
stream.dash: refactor and reformat code

bastimeyer:
- Fix incorrect typing
- Fix error messages, as well as docstrings and comments
- Fix indentation
- Replace str.format with f-strings
- Use consistent order of `MPDNode.attr()` call keywords, but keep the order of the method's arguments
- Use keywords when initializing the `Segment` dataclass
- Flatten nested if-blocks in `SegmentTemplate.format_media()`
- Remove unnecessary if-block in `Representation.segments()`

### DIFF
--- a/src/streamlink/stream/dash_manifest.py
+++ b/src/streamlink/stream/dash_manifest.py
@@ -7,7 +7,7 @@ import re
 from collections import defaultdict, namedtuple
 from contextlib import contextmanager
 from itertools import count, repeat
-from typing import Optional, Tuple
+from typing import ClassVar, Optional, Tuple
 from urllib.parse import urljoin, urlparse, urlsplit, urlunparse, urlunsplit
 
 from isodate import Duration, parse_datetime, parse_duration  # type: ignore[import]
@@ -48,6 +48,9 @@ def freeze_timeline(mpd):
     mpd.timelines = timelines
 
 
+_re_segment_template = re.compile(r"(.*?)\$(\w+)(?:%([\w.]+))?\$")
+
+
 class MPDParsers:
     @staticmethod
     def bool_str(v):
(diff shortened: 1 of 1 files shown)


src/streamlink/stream/dash_manifest.py
```diff
@@ -461,97 +592,127 @@ def segment_numbers(self):
...
-        else:
-            for number, available_at in self.segment_numbers():
-                yield (self.make_url(self.media(Number=number, **kwargs)),
-                       available_at)
+        log.debug(f"Generating segment timeline for {self.root.type} playlist (id={self.parent.id}))")
+
+        if self.root.type == "static":
+            for segment, n in zip(self.segmentTimeline.segments, count(self.startNumber)):
+                url = self.make_url(self.media(Time=segment.t, Number=n, **kwargs))
+                available_at = datetime.datetime.now(tz=UTC)  # TODO: replace with EPOCH_START ?!
```
bastimeyer, line 630: I wasn't 100% sure here, and since I didn't want to include any logic changes, I added the TODO comment.

Manifests with static timelines should have segments always available, so I don't know why the current timestamp (even iterated in the loop) is being yielded here instead of just `EPOCH_START`.

src/streamlink/stream/dash_manifest.py
```diff
@@ -579,18 +740,21 @@ def segments(self, **kwargs):
         segmentTemplate = self.segmentTemplate or self.walk_back_get_attr("segmentTemplate")
 
         if segmentTemplate:
-            for segment in segmentTemplate.segments(RepresentationID=self.id,
-                                                    Bandwidth=int(self.bandwidth * 1000),
-                                                    **kwargs):
-                if segment.init:
-                    yield segment
-                else:
-                    yield segment
```
bastimeyer, line 588: This looks like this was done intentionally with future changes in mind in regards to the initialization segment. I still decided to replace it with `yield form` though...

bastimeyer:
CI failure is caused by a new ruff release, btw. I'll have a look and will then rebase...

bastimeyer:
It's a bug in ruff 0.0.250. I made a bug report, so no need to add an ignore comment for this. Just wait until this gets fixed.

PR #5208, opened 2023-02-28
stream.dash: add --dash-manifest-reload-attempts

bastimeyer:
This adds `--dash-manifest-reload-attempts` and the respective session option with a default value of 3, similar to `--hls-playlist-reload-attempts`.

HTTP requests of dynamic DASH stream manifests currently don't have a retry value set, so if a request fails due to a temporary connection/server issue, then no new segments can be queued. Failed requests still raise a `StreamError` on the worker thread, same as the HLS implementation, which needs to be fixed separately.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -808,6 +808,7 @@ def build_parser():
 
     transport = parser.add_argument_group("Stream transport options")
     transport_hls = parser.add_argument_group("HLS options", parent=transport)
+    transport_dash = parser.add_argument_group("DASH options", parent=transport)
     transport_ffmpeg = parser.add_argument_group("FFmpeg options", parent=transport)
 
     transport.add_argument(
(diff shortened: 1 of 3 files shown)


src/streamlink_cli/argparser.py
```diff
@@ -1029,11 +1030,22 @@ def build_parser():
         Skip to the beginning of a live stream, or as far back as possible.
         """,
     )
+
+    transport_dash.add_argument(
+        "--dash-manifest-reload-attempts",
+        type=num(int, min=0),
+        metavar="ATTEMPTS",
+        help="""
+        How many attempts should be done to reload the DASH manifest before giving up.
```
gravyboat, line 1039: Wording on this seems a bit odd, "how many attempts should be done" doesn't read very well. Pretty minor nit but something like "How many attempts to reload the DASH manifest should occur before giving up." or something else.

bastimeyer: I've just mindlessly copied that from the HLS param help text, which was added in 2017 and hasn't been changed since.

https://github.com/streamlink/streamlink/commit/f5d862e281d430baade64976e1a6840593cbf8a9#diff-cf813af4bcbb52b3cde4c96d819cf4834df10b6f2aefbe8f9efd4afa5f08649bR608-R609

gravyboat: Gotcha.

PR #5216, opened 2023-03-04
plugins.kanal7: new plugin

omergorur:
Add plugin for kanal7.com

This is my first plugin. If I made any mistakes, I would appreciate your help.


src/streamlink/plugins/kanal7.py
```diff
@@ -0,0 +1,33 @@
+"""
+$description Turkish live TV channel.
```
bastimeyer, line 2: The description must include more information about the sites which that plugin supports, like who's the owner, etc. Take a look at other plugins.
https://en.wikipedia.org/wiki/Kanal_7

src/streamlink/plugins/kanal7.py
```diff
@@ -0,0 +1,33 @@
...
+
+log = logging.getLogger(__name__)
+
+
+@pluginmatcher(re.compile(r"""
+    https?://(?:www\.)?
+    (?:
+        kanal7\.com/canli-izle
+    )
+""", re.VERBOSE))
```
bastimeyer, line 22: A verbose regex is unnecessary here, and so are the (non-capturing) parentheses.

```suggestion
@pluginmatcher(re.compile(
    r"https?://(?:www\.)?kanal7\.com/canli-izle",
))
```

src/streamlink/plugins/kanal7.py
```diff
@@ -0,0 +1,33 @@
...
+class KANAL7(Plugin):
+
+    def _get_streams(self):
+        hls_url = self.session.http.get(self.url, schema=validate.Schema(
+            re.compile(r"""hls\s*:\s*(?P<q>["'])(?P<hls_url>https?://.*?\.m3u8.*?)(?P=q)"""),
+            validate.any(None, validate.get("hls_url")),
+        ))
+        return HLSStream.parse_variant_playlist(self.session, hls_url)
```
bastimeyer, line 30: `hls_url` can be `None`.
```suggestion
        if hls_url:
            return HLSStream.parse_variant_playlist(self.session, hls_url)
```

src/streamlink/plugins/kanal7.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@pluginmatcher(re.compile(r"""
+    https?://(?:www\.)?
+    (?:
+        kanal7\.com/canli-izle
+    )
+""", re.VERBOSE))
+class KANAL7(Plugin):
+
```
bastimeyer, line 24: Why the uppercase class name? The blank line below is also unnecessary and might result in a linting error in the future.

bastimeyer (changes requested):
The kanal7 plugin was removed in #2839 because of static stream URLs, but it looks like the URLs linked in the commit message of the plugin removal aren't working anymore except for the first one.

When looking at the site's HTML/JS response, the HLS URL doesn't change, even though it does have an UUID4 token for its `app` query string parameter. Is it really a non-static URL?

src/streamlink/plugins/kanal7.py
omergorur: I based my plugin on the tv8.plugin. The class name was written in capital letters there.

omergorur:
> The kanal7 plugin was removed in #2839 because of static stream URLs, but it looks like the URLs linked in the commit message of the plugin removal aren't working anymore except for the first one.
> 
> When looking at the site's HTML/JS response, the HLS URL doesn't change, even though it does have an UUID4 token for its `app` query string parameter. Is it really a non-static URL?

Yes, it is non-static now. The static URL belongs to Kanal 7 Europe channel. Its streaming is quite different from the channel in Turkey.

src/streamlink/plugins/kanal7.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+import logging
+import re
+
+from streamlink.plugin import Plugin, pluginmatcher
+from streamlink.plugin.api import validate
+from streamlink.stream.hls import HLSStream
+
+log = logging.getLogger(__name__)
```
bastimeyer, line 14: Missing empty line after imports:
https://github.com/streamlink/streamlink/blob/350cb58a46bad34ce0aa5918572c19061b5f3b8e/pyproject.toml#L122

src/streamlink/plugins/kanal7.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@pluginmatcher(re.compile(
+    r"https?://(?:www\.)?kanal7\.com/canli-izle",
+))
+class Kanal7(Plugin):
+
```
bastimeyer, line 21: Unnecessary empty line

bastimeyer (commented):
You'll have to fix the linting issues.

> probably non-static

And before this can get merged, I want to see the UUID4 change.
Current one is `f99587ad-1637-494d-8255-da35b09d17a1`

omergorur:
> You'll have to fix the linting issues.
> 
> > probably non-static
> 
> And before this can get merged, I want to see the UUID4 change. Current one is `f99587ad-1637-494d-8255-da35b09d17a1`

Alright, let's wait and see. Thanks.

bastimeyer:
Still the same UUID

omergorur:
> Still the same UUID

You can close the pull request. If anything changes, I will let you know.

bastimeyer:
Still the same UUID after 3 1/2 days. No plugin required. Stream URL:
`https://kanal7.daioncdn.net/kanal7/kanal7.m3u8?app=f99587ad-1637-494d-8255-da35b09d17a1&ce=3`

PR #5227, opened 2023-03-09
plugins.vimeo: fix playerConfig regex

OnlyFart:
Parameters to reproduce

```
https://player.vimeo.com/video/803849296?h=1f71c5e810 best
```

image


bastimeyer:
Thanks.
The plugin requires a couple more fixes, but the regex fix for player.vimeo.com should suffice for now.

PR #5217, opened 2023-03-05
stream.dash: fix dynamic timeline-less streams

bastimeyer:
- Fix segment number iterator and segment availability-time iterator for dynamic timeline-less DASH streams by calculating the right segment number offset and basing the segment availability-time on the segment number, and not just on the current time alone
- Set a default value of 0 seconds for presentationTimeOffset
- Ensure that segment durations are known
- Never calculate negative segment number offsets
- Reduce unnecessary delay by starting with the upcoming segment in the generated timeline to be as close to the live-edge as possible (with minBufferTime and presentation offset+delay in mind)
- Add debug log output
- Update MPD manifest test fixture with better time data

Next:

- Consider removing the default value of suggestedPresentationDelay, as it seems to be a relict of Streamlink's initial DASH implementation
- Also take a look at properly synchronizing the generated timeline between multiple substreams, so that the correct segment numbers always get calculated in all threads
- Avoid interweaving debug log and make it flush per-thread

----

Closes #4607

Not 100% sure whether this will fix #4607, but the data in that thread is not enough to confirm that, so let's close it. Since this PR adds more debug log messages with actual timestamps, new issue threads can be opened if there's still a problem with DASH streams of that particular kind.

At least this PR fixes all issues with DASH streams from the steam plugin, and the correct segment numbers at the correct times should be queued and requested by the worker and writer threads respectively.

PR #5233, opened 2023-03-10
stream.dash: fix segment number iterator

bastimeyer:
Fixes #5217
Ref #5231


bastimeyer:
Segment requests need better debug info with proper timing data. I'll have a look at this tomorrow.

PR #5274, opened 2023-04-02
plugins.vk: support .ru domain

Schn4pper:
VK platform uses both vk.ru and vk.com domain names, but Streamlink only supports vk.com.

### DIFF
--- a/src/streamlink/plugins/vk.py
+++ b/src/streamlink/plugins/vk.py
@@ -1,13 +1,14 @@
 """
 $description Russian live-streaming and video hosting social platform.
 $url vk.com
+$url vk.ru
 $type live, vod
 """
 
 import logging
 import re
 from hashlib import md5
-from urllib.parse import parse_qsl, unquote, urlparse
+from urllib.parse import parse_qsl, unquote, urlparse, urlunparse
 
 from streamlink.exceptions import NoStreamsError
 from streamlink.plugin import Plugin, PluginError, pluginmatcher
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/vk.py
```diff
@@ -21,10 +21,10 @@
 
 
 @pluginmatcher(re.compile(
-    r"https?://(?:\w+\.)?vk\.com/videos?(?:\?z=video)?(?P<video_id>-?\d+_\d+)",
+    r"https?://(?:\w+\.)?vk(?:\.com|.ru)/videos?(?:\?z=video)?(?P<video_id>-?\d+_\d+)",
```
bastimeyer, line 24:
```suggestion
    r"https?://(?:\w+\.)?vk\.(?:com|ru)/videos?(?:\?z=video)?(?P<video_id>-?\d+_\d+)",
```

The dot needs to be escaped and can be moved out of the non-capture group

src/streamlink/plugins/vk.py
```diff
@@ -21,10 +21,10 @@
...
 ))
 @pluginmatcher(re.compile(
-    r"https?://(\w+\.)?vk\.com/.+",
+    r"https?://(\w+\.)?vk(?:\.com|.ru)/.+",
```
bastimeyer, line 27:
```suggestion
    r"https?://(\w+\.)?vk\.(?:com|ru)/.+",
```

Same here

bastimeyer (changes requested):
Thanks for the PR.

----

Please also update the plugin metadata at the top by adding another
```
$url vk.ru
```
entry under the
```
$url vk.com
```
line

bastimeyer:
There might also be problems with the plugin's `_get_cookies()` method which only gets cookies from the .com site.

PR #5281, opened 2023-04-09
stream.hls: check adapter for decryption key URI

bastimeyer:
This improves the error message for encrypted HLS streams where the key `URI` doesn't have a matching adapter registered. The optional `KEYFORMAT` attribute actually describes the key `URI`, but Streamlink doesn't use that attribute, so it's better to check for an available adapter, in case plugins implement it.
https://www.rfc-editor.org/rfc/rfc8216#section-4.3.2.4


bastimeyer:
hm... flaky tests...
Looks like the thread closing issue finally needs to get fixed in these tests. Will have a look later...

bastimeyer:
I've found the issue, but I'm going to fix that in another PR. Let me explain it here though, for reference...

The issue is a combination of a bug in the `HLSStreamWorker.iter_segments()` loop and a timing issue in the `requests_mock` package when the `HTTPSession.get_adapter()` method gets monkeypatched **during the call of the monkeypatched `HTTPSession.send()` method**. The newly added logic and its test rely on the original `HTTPSession.get_adapter()` method.
https://github.com/jamielennox/requests-mock/blob/1.10.0/requests_mock/mocker.py#L140-L205

The failing HLS tests only define a single segment and set the playlist's media sequence to `0`. This is the tag's default value and is perfectly valid:
https://www.rfc-editor.org/rfc/rfc8216#section-4.3.3.2

The `HLSStreamWorker` initializes its `playlist_end` attribute with `None`. This value then gets set to the number of the last segment if there's an `EXT-X-ENDLIST` tag in the HLS playlist, which is the case in these tests. This means that the value gets changed from `None` to `0` when the playlist gets parsed.

However, in `HLSStreamWorker.iter_segments()`, it only checks whether the stream ended by evaluating `self.playlist_end and sequence.num >= self.playlist_end`, which always returns `False` due to the segment number being `0`:
https://github.com/streamlink/streamlink/blob/d372059127c1b8624a0956cf8b92589f86fb189c/src/streamlink/stream/hls.py#L445-L448

This means that each test's worker thread keeps running until the end of the test where all worker+writer+reader threads get terminated by the test setup.

When `requests_mock` monkeypatches `HTTPSession.send()`, it makes other threads block, so that there's no issue during the patching (override and restore). But depending on the timing of the **worker thread** which keeps running and requesting the playlist (as the end of the stream was not detected correctly), **one of the writer threads** which is trying to find the adapter for the encryption key URI can incorrectly call the monkeypatched `HTTPSession.get_adapter()` method, which is not what we want. The call is supposed to raise an `InvalidSchema` exception, but the patched method never does that.

This means that the encryption key request continues regularly with the mocked response and the stream keeps running. It then fails at the very first test assertion which expects the stream to have ended due to the exception being raised by one of the writer threads. There's also a missing `self.await_close()` synchonization call here, but that's mostly irrelevant.

Fixing `HLSStreamWorker.iter_segments()` fixes the failing tests here. The tests also should get updated and properly await the end of the stream before making assertions.

This took me a bit to figure out, due to the timing issue in the `requests_mock` dependency. I'll submit a PR later today.

bastimeyer:
Calling `HTTPSession.get_adapter()` when there are other threads making more requests while the session instance is mocked unfortunately doesn't work reliably, so a slightly different implementation was needed.

This now lets the `self.session.http.get()` call raise the `InvalidSchema` exception when an adapter for the decryption key request could not be found. Since Streamlink overrides the `HTTPSession.request()` method with custom retry and error handling logic (which desperately need to be rewritten), some ugly error handling needed to be added. The tests also needed to be updated with a different mocking setup for raising the exception, and mocked request failures needed to be patched in order to avoid `time.sleep()` calls in the overridden `request()` method.

Second commit is optional, but it improves some of the tests.

gravyboat:
@bastimeyer The changes look good. This PR still has the WIP label, is it ready to go or is there more you want to add?

PR #5291, opened 2023-04-16
plugins.openrectv: fix 401 error

vinyl-umbrella:
OPENREC.tv recently made a small change.
When fetching m3u8, it now returns 401 if "https://www.openrec.tv" is not specified in the referer.
So I avoid that error by adding a referer.

### DIFF
--- a/src/streamlink/plugins/openrectv.py
+++ b/src/streamlink/plugins/openrectv.py
@@ -149,10 +149,10 @@ class OPENRECtv(Plugin):
                 log.error("There is no video file.")
 
             if m3u8_file is not None:
-                self.session.http.headers.update({"Referer": "https://www.openrec.tv/"})
                 yield from HLSStream.parse_variant_playlist(
                     self.session,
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/openrectv.py
```diff
@@ -149,6 +149,7 @@ def _get_streams(self):
                 log.error("There is no video file.")
 
             if m3u8_file is not None:
+                self.session.http.headers.update({"Referer": "https://www.openrec.tv/"})
```
bastimeyer, line 152: Probably better to use `self.url` instead, so the plugin doesn't rely on a hardcoded string literal.

And as an alternative to setting the header on the entire http session, `headers={...}` could be set as a keyword on the `parse_variant_playlist()` call.

vinyl-umbrella:
Thank you for the review and suggestions.

bastimeyer:
Thanks

PR #5301, opened 2023-04-19
docs: bump furo to 2023.03.27

bastimeyer:
This updates the furo docs theme to the latest version.

The main intention however is bumping Sphinx to 6.x, which drops the unnecessary jQuery+underscore.js bloat:
https://www.sphinx-doc.org/en/master/changes.html#release-6-0-0-released-dec-29-2022

Sphinx's minimum version requirement however is 5.x (defined by furo and the other build dependencies). I decided against bumping the min version to 6.x, because Sphinx is also used for building the man page. Some Linux distros are still on Sphinx 5 and I don't want to unnecessarily introduce incompatibilities for no reason.

Sphinx 6 also drops support for py37, but we've already dropped support for building docs on py37 a while ago in 46496db3698e4753154061aeb504e348270d3f9f

In order for being able to bump Sphinx to 5.x/6.x, myst-parser and sphinx-design needed an upgrade. Those have finally been published a couple of weeks and days ago respectively:
- https://github.com/executablebooks/MyST-Parser/releases
- https://github.com/executablebooks/sphinx-design/releases

----

Bumping furo also required some fixing due to changes of the theme's `logo` variable.

I couldn't spot any issues after the upgrade, but please check and compare the netlify preview build first.
https://deploy-preview-5301--streamlink.netlify.app/

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -50,4 +50,4 @@ jobs:
       - name: mypy
         run: |
           python -m mypy --no-incremental
-          python -m mypy --no-incremental docs
+          python -m mypy --no-incremental --python-version 3.8 docs
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,7 +1,7 @@
-sphinx >=4.0.0
-furo ==2022.09.15
-myst-parser >=0.18.0
-sphinx-design >=0.3.0
+sphinx >=5.0.0, <7
+furo ==2023.03.27
+myst-parser >=1.0.0, <2
+sphinx-design >=0.4.1, <1
 versioningit >=2.0.0, <3
 
 docutils-stubs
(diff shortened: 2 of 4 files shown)


bastimeyer:
> `/opt/hostedtoolcache/Python/3.11.3/x64/lib/python3.11/site-packages/sphinx/util/typing.py:214:2: error: Positional-only parameters are only supported in Python 3.8 and greater  [syntax]`

mypy's min Python version is set to 3.7:
https://github.com/streamlink/streamlink/blob/5.4.0/pyproject.toml#L140
The previous push failed due to the updated version of Sphinx which already dropped py37.

Unfortunately, changing the python version for specific modules can't be done in the mypy config, so the mypy command in the lint CI config needs to be changed. This isn't ideal, because it'll also need to get set when linting locally.

This workaround can be removed once py37 sees its EOL in two months:
https://endoflife.date/python

PR #5310, opened 2023-04-25
cli.output: path-only --player argument

### LINKED ISSUE #5305: proposal: remove optional player args from `--player` and only allow player paths (breaking change)
## Introduction

The `--player` argument value has always been used for building the **entire player command line**, consisting of the **player path** and **optional player arguments**. `--player-args` only was added in Livestreamer as an addition to `--player` in 1ff89b8c81ddca6231eec85d46d5b0f309391585 for being able to **append more player arguments** (half a year after the initial `PlayerOutput` implementation in 331350f297fa62a902f967a032f4555d05659cfc).

With that addition, `--player` should've immediately been turned into a path-argument for only resolving the player executable path and nothing else, so no optional player arguments, but this was never done, because it's a breaking change which requires users to update their config files (especially on Windows).

I was just in the process of refactoring `PlayerOutput` (as a follow-up of the other recent CLI output code refactorings), which is the reason why I'm bringing this up now and why I'm proposing making this breaking change now.

## The issue

The big issue here is how the resulting player "command-line" is built when executing the player process and also how it gets parsed when trying to detect the player executable for the `--title` CLI argument.

The current `PlayerOutput` implementation uses two different approaches for building the player command-line:

1. On POSIX systems, it's an argument vector (list of arguments) that gets parsed via `shlex.split()` and that then gets passed to `subprocess.Popen()` or `subprocess.call()`.
2. On Windows, it's a simple string, because that's how Windows works (which is terrible). That string later on gets parsed by the [win32 APIs](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) and it therefore needs to get formatted correctly, either by the application logic or by the Python stdlib.

While the `PlayerOutput`'s POSIX implementation is mostly fine, the Windows implementation is flawed.

### Player detection

First, the issue with the player detection. Detecting the player is necessary for implementing the `--title` argument (automated player-specific title args).

The goal here is to get the player-path's `basename` (file name) from the `--player` argument value, so that it can be compared with a list of known player executable names.

On POSIX, we're simply parsing the argument value via `shlex.split()` which handles quoted player paths, and then we're using the first token, which is the player path. Done.

On Windows however, this is not possible, because even though `shlex.split(posix=False)` exists for supporting Windows-style file paths, it won't work in all cases because `shlex` does only support Unix-style shell syntax, and worst of all, **quoted executable paths on Windows are optional** (more about that later).

That means that a workaround is required for detecting the player, which is the following one which was added by the initial `--title` implementation:
https://github.com/streamlink/streamlink/blob/78c319508c80a5c4cf72992e7b976d0ff741c81b/src/streamlink_cli/output/player.py#L77-L81
As you can see, it calls `startswith()`, because `os.path.basename()` apparently ignores the leading quote character if there is one.

```py
>>> os.path.basename('C:\\Program Files\\VideoLAN\\VLC\\vlc.exe')
'vlc.exe'
>>> os.path.basename('"C:\\Program Files\\VideoLAN\\VLC\\vlc.exe"')
'vlc.exe"'
>>> os.path.basename('C:\\Program Files\\VideoLAN\\VLC\\vlc.exe --play-and-exit')
'vlc.exe --play-and-exit'
>>> os.path.basename('"C:\\Program Files\\VideoLAN\\VLC\\vlc.exe" --play-and-exit')
'vlc.exe" --play-and-exit'
```

Not a particularly nice solution, but fixing that wouldn't justify introducing a breaking change yet.

### Player execution

Now to the problematic player execution part.

As explained, on POSIX systems, `subprocess.Popen()` and `subprocess.call()` receive a list of arguments. Those are coming from the tokenized `--player` and `--player-args` argument values, as well as the built player input value and optional player title, so there's not much interesting going on here.

But on Windows, the current implementation operates on strings, which was most likely done because of the parsing issues of the `--player` argument mentioned above. None of the user inputs are sanitized and only the generated player arguments get shell-escaped (input path and optional title) via `subprocess.list2cmdline()`. The rest is simple string concatentation with a `" "` separator.

`subprocess.Popen()` and `subprocess.call()` also support passing an argv on Windows, which is always the preferred way, so the resulting command-line string on Windows can be built internally by the stdlib. That however can't be done right now, because we can't reliably parse and tokenize the `--player` argument value on Windows, hence the string nonsense.

The big issue is about player paths with spaces and without quotation marks, as those are optional on Windows. When executing such a command, Windows will do the following: first, it tokenizes the command as usual, then it merges the tokens from left to right until a matching executable path gets found on the filesystem. For example, `C:\Program Files\sub dir\program name` gets resolved and executed in this order:

1. `C:\Program.exe`
6. `C:\Program Files\sub.exe`
7. `C:\Program Files\sub dir\program.exe`
8. `C:\Program Files\sub dir\program name.exe`

This essentially means that when the `--player` argument **value** doesn't include additional quotation marks, for any player in `C:\Program Files\` (which is the case on 99% of all Windows systems), Windows will try to find matching *rogue executables* up the root path first. Streamlink doesn't do any path validation, because it can't parse the input reliably:

```py
>>> shlex.split('"C:\\Program Files\\VideoLAN\\VLC\\vlc.exe"')
['C:\\Program Files\\VideoLAN\\VLC\\vlc.exe']
>>> shlex.split('"C:\\Program Files\\VideoLAN\\VLC\\vlc.exe" --play-and-exit')
['C:\\Program Files\\VideoLAN\\VLC\\vlc.exe', '--play-and-exit']
>>> shlex.split('C:\\Program Files\\VideoLAN\\VLC\\vlc.exe')
['C:Program', 'FilesVideoLANVLCvlc.exe']
>>> shlex.split('C:\\Program Files\\VideoLAN\\VLC\\vlc.exe --play-and-exit')
['C:Program', 'FilesVideoLANVLCvlc.exe', '--play-and-exit']

>>> shlex.split('"C:\\Program Files\\VideoLAN\\VLC\\vlc.exe"', posix=False)
['"C:\\Program Files\\VideoLAN\\VLC\\vlc.exe"']
>>> shlex.split('"C:\\Program Files\\VideoLAN\\VLC\\vlc.exe" --play-and-exit', posix=False)
['"C:\\Program Files\\VideoLAN\\VLC\\vlc.exe"', '--play-and-exit']
>>> shlex.split('C:\\Program Files\\VideoLAN\\VLC\\vlc.exe', posix=False)
['C:\\Program', 'Files\\VideoLAN\\VLC\\vlc.exe']
>>> shlex.split('C:\\Program Files\\VideoLAN\\VLC\\vlc.exe --play-and-exit', posix=False)
['C:\\Program', 'Files\\VideoLAN\\VLC\\vlc.exe', '--play-and-exit']
```

So player paths with spaces and without quotation marks pose a security issue on Windows, because no validation is being done. Parsing the user input is not reliable. And error messages when a path is invalid are also not great.

On Windows shells like Batch and PowerShell, player paths actually need to be defined as follows in order to be precise and safe:
```pwsh
streamlink -p "\`"C:\Program Files\...\`" optional player args" -a "more optional player args" ...
```
but the vast majority of users won't do that correctly.

The config file is a bit easier, because shell syntax doesn't need to be escaped:
```
player="C:\Program Files\..." optional player args
player-args=more optional player args
```

## How to fix

In order to fix this mess, `--player` needs to be turned into a path-only argument. Then the argument value doesn't have to be parsed, the player path can be reliably validated, and an argv can be built on Windows by parsing the `--player-args` value instead (which works just fine via the `shlex` module) and combining that with the player path and generated arguments.

This would fix any security issues and would help cleaning up and simplifying the `PlayerOutput` implementation (in a follow-up code refactor).

Another benefit is removing any kind of confusion around Streamlink's config file, where unlike on command-line shells, argument values don't require quotation marks for arbitrary argument values with spaces. For the `player` argument however, it currently does, because it's interpreted as a command-line.
https://github.com/streamlink/windows-builds/blob/5.4.0-1/files/config#L13-L31

## Follow-up plans

As mentioned, I was in the process of refactoring the player arguments logic in `PlayerOutput`. After this issue has been fixed and `PlayerOutput` been improved, more player-specific stuff could be added, like for example for IINA on macOS, which requires the `--stdin` argument instead of `-`, where currently the ugly `{playerinput}-stdin` workaround is needed to be able to use it.

## Feedback

Opinions on introducing this **significant breaking change** of the `--player` argument?

This will affect almost every single Windows user who's set a player path via the config file, and it will affect everyone on any OS who's set player arguments via `--player` instead of `--player-args`. The [docs](https://streamlink.github.io/cli.html#player-options) haven't done a great job over the years to make people use `--player-args` over `--player` for any custom player argument.

## Migration / Deprecation

I haven't made any plans yet on how to implement any automated migrations, if it's even possible. The Windows installer's NSIS code isn't particularly nice, and touching user configs isn't a great idea either.

Deprecations can't be added, unless we're implementing some kind of cheap detections for certain things like quoted player paths, but that doesn't catch every single case like unquoted player paths with spaces and player args on Windows.

## What would change

The changes are pretty easy to explain though

- Before (POSIX shell)
  ```sh
  streamlink -p "\"PLAYERPATH\" PLAYERARGS1" -a "PLAYERARGS2" ...
  ```
  Before (Powershell)
  ```pwsh
  streamlink -p "\`"PLAYERPATH\`" PLAYERARGS1" -a "PLAYERARGS2" ...
  ```
  After
  ```
  -p "PLAYERPATH" -a "PLAYERARGS1 PLAYERARGS2"
  ```
- Before (config file):
  ```
  player="PLAYERPATH" PLAYERARGS1
  player-args=PLAYERARGS2
  ```
  After
  ```
  player=PLAYERPATH
  player-args=PLAYERARGS1 PLAYERARGS2
  ```

## Current state of `PlayerOutput`

https://github.com/streamlink/streamlink/blob/78c319508c80a5c4cf72992e7b976d0ff741c81b/src/streamlink_cli/output/player.py

bastimeyer:
Resolves #5305 

Breaking change... Opening this as a draft for now, since I'm aiming for the next major version release when py37 support gets dropped in two months.

Please read #5305 for the full details about the motivation of these changes.

## Changes

1. Rewrite `find_default_player()`
2. Turn `--player` into a path-only argument and remove support for specifying a whole player command-line with additional arguments. Player arguments now must be set via the `--player-args` argument.
   As a result, `subprocess.Popen()` and `subprocess.call()` now receive an argv with a properly defined and safe player path with additional player arguments. The player detection also doesn't rely on a weird path basename workaround anymore.
3. Remove the old `--player-args` `{filename}` fallback variable. Only `{playerinput}` is now supported.
4. Refactor the `PlayerOutput`'s player detection and the player arguments builder for better code quality and easier extensibility.
5. Actually check/verify the player path via `shutil.which()`, so a better error can be raised when the path is incorrect, or when the player path is not executable, or when the single executable name can't be resolved from the system's `PATH` env var. Additionally raise a warning with more info when the player path begins with a single or double quotation mark, as this is now unsupported. Detecting player args in the `--player` value is not possible, so a warning can't be raised for those cases.
6. Add `{playertitleargs}` to `--player-args`, so the order of automatically generated player titles can be customized. By default, player title args come in front of the player args set by the user, so user data can override (if the player supports overriding previous args).

More details in the commit message bodies...

## Notes

- ~~Since custom player arguments now can only be set via `--player-args`, this makes the addition of player title arguments more strict. Previously it was possible to set arguments in front of the automated player title arguments, but this is not possible anymore because the order is always path -> title-args -> custom-args, where custom-args has support for the `{playerinput}` variable, so the input argument can be set freely here. I was thinking about adding another variable (`{playertitle}`) that replaces the title-args segment and that automatically gets added to the `--player-args` if it's missing, similar to `{playerinput}`. Then the player title args can also be set freely, like the player input arg. Thoughts?~~
- ~~The updated default player lookup doesn't have tests. I've tested it manually though...~~
- No changes to the `PlayerOutput` implementation itself (which I'm still unhappy about)
- The Streamlink config files will all be outdated with these changes. This is especially bad on Windows where the installer only writes the config file if it doesn't exist yet. This will require a special note, or maybe even a new option in the installer which lets it overwrite it with the default config.

----

Please give this branch a thorough test. It would be pretty annoying if something was missed in these changes.


bastimeyer:
Small refactor, added/improved some test cases and added `{playertitleargs}` var.

bastimeyer:
Rebased to master (#5317) and updated tests

bastimeyer:
I will add more documentation to the deprecations page once this got merged, among all the other breaking changes.

Not sure yet about the Windows installer and its config file. I'll have to figure this out later.

A bit sad that nobody bothered to give this branch a try and give feedback. Will merge now anyway...

PR #5321, opened 2023-05-02
stream.hls: reload playlist in strict intervals

### LINKED ISSUE #4646: stream.hls: smarter playlist reload times
The `HLSStreamWorker` is currently reloading playlists of live streams in a strict, continuous time interval. Depending on the options used ([`--hls-playlist-reload-time`](https://streamlink.github.io/cli.html#cmdoption-hls-playlist-reload-time)), the reload time is either the playlist's `targetduration` (default), or the duration of the last segment, or the sum of segment durations according to the `live-edge`, or a constant time value. Having strict and continuous playlist reload wait times causes a couple of problems.

The issue is that [HTTP request/response delays (including retries)](https://github.com/streamlink/streamlink/blob/4.2.0/src/streamlink/stream/hls.py#L283-L292), [execution delays](https://github.com/streamlink/streamlink/blob/4.2.0/src/streamlink/stream/hls.py#L383-L429) and [buffer availability wait times](https://github.com/streamlink/streamlink/blob/4.2.0/src/streamlink/stream/hls.py#L302) are [not taken into account when reloading](https://github.com/streamlink/streamlink/blob/4.2.0/src/streamlink/stream/hls.py#L425), so that over time, playlist reloads will get out-of-sync compared to the playlist's segment durations as more and more delay is added between reloads, and the point where the playlist gets reloaded gets shifted.

For most streams, this is no problem, but for low latency streams, where playlist reloads are supposed to only queue the next segment (as soon as possible), shifting the reload points is a problem because the writer's queue can become empty at some point and then it'll get filled with two segments on the following reload, which will unnecessarily introduce a streaming delay.

----

The reload times of the [`DASHStreamWorker`](https://github.com/streamlink/streamlink/blob/4.2.0/src/streamlink/stream/dash.py#L78-L87) already take execution times into consideration. The intention here is of course that [DASH segments can have an availability time](https://github.com/streamlink/streamlink/blob/4.2.0/src/streamlink/stream/dash.py#L36-L40), so being precise is important.

HLS streams however also can have segment time annotations via [`EXT-X-PROGRAM-DATE-TIME`](https://github.com/streamlink/streamlink/blob/4.2.0/src/streamlink/stream/hls_playlist.py#L355-L360). The problem though is that it's an optional tag and the time doesn't necessarily have to be precise. And since the server time is unknown and can only get determined by reading optional HTTP headers or custom playlist tags, calculating an offset to the system time is often impossible. If the time offsets of the segment times could be calculated reliably, then HLS playlist reload times could be calculated so that reloads happen as soon as possible when the next segment becomes available on the server. That would be a micro-optimization, albeit not an important one, as HLS low latency streams have prefetch segments (AHLS (draft) or custom ones (eg. Twitch)).

bastimeyer:
Make `HLSStreamWorker` reload the playlist in strict time intervals instead of just queuing the next playlist reload.

Previously, playlist fetch time and segment processing time always added to the overall reload time, resulting in an eventual shift of the expected reload time, which could lead to an empty segment queue in certain scenarios (live-edge=1 or Twitch low latency streaming).

When the reload time exceeds the expected reload interval, continue with the shifted interval, which matches the old behavior.

----

Resolves #4646 

> When the reload time exceeds the expected reload interval, continue with the shifted interval, which matches the old behavior.

That could be changed, so that if a playlist reload takes too long, the next one could start immediately. I am not sure though if that's a good idea, because depending on the response data and the timing of when the data was generated by the server, no new segments could be included in the next response. And if one request takes more than two intervals, then the next x playlist reloads will be done immediately one after another, and that doesn't sound right.

Opening this as a draft for now, because I want to test this a bit first...

Tests are a bit complicated, but that's because of the playlist reload iteration with the timing involved, and that's always messy. I tried to write this as comprehensible as possible.

----

The timestamps of the playlist reloads now don't get shifted (example with playlist reload time of 2s):

```
[13:32:39.553732][stream.hls][debug] Reloading playlist
...
[13:35:09.553753][stream.hls][debug] Reloading playlist
```

master:

```
[13:35:36.733978][stream.hls][debug] Reloading playlist
...
[13:37:56.415168][stream.hls][debug] Reloading playlist
```

### DIFF
--- a/src/streamlink/stream/hls.py
+++ b/src/streamlink/stream/hls.py
@@ -2,6 +2,7 @@ import logging
 import re
 import struct
 from concurrent.futures import Future
+from datetime import datetime, timedelta
 from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Union
 from urllib.parse import urlparse
 
@@ -22,6 +23,7 @@ from streamlink.stream.http import HTTPStream
 from streamlink.stream.segmented import SegmentedStreamReader, SegmentedStreamWorker, SegmentedStreamWriter
 from streamlink.utils.cache import LRUCache
 from streamlink.utils.formatter import Formatter
+from streamlink.utils.times import now
 
 
 log = logging.getLogger(__name__)
@@ -297,6 +299,7 @@ class HLSStreamWorker(SegmentedStreamWorker):
         self.playlist_end: Optional[int] = None
         self.playlist_sequence: int = -1
         self.playlist_sequences: List[Sequence] = []
+        self.playlist_reload_last: datetime = now()
         self.playlist_reload_time: float = 6
         self.playlist_reload_time_override = self.session.options.get("hls-playlist-reload-time")
         self.playlist_reload_retries = self.session.options.get("hls-playlist-reload-attempts")
@@ -412,6 +415,7 @@ class HLSStreamWorker(SegmentedStreamWorker):
         return default
 
     def iter_segments(self):
+        self.playlist_reload_last = now()
         try:
             self.reload_playlist()
         except StreamError as err:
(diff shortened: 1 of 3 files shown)


bastimeyer:
I want to publish a new release for the urllib3 issue first, without these changes here

Issue #5324, opened 2023-05-03
AttributeError: module 'urllib3.util.url' has no attribute 'PERCENT_RE'


botbahlul:
### Checklist

- [X] This is a bug report and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

C:\Program Files\Python310\lib\site-packages\streamlink\plugin\api\http_session.py line 53 typo?

I was trying to run my script with python 3.10.4 in Windows 10 and got that error, but trying to run it with python 3.8.6 is working fine, why?

### Debug log

```text
File "C:\Program Files\Python310\lib\site-packages\streamlink\plugin\api\http_session.py", line 53, in Urllib3UtilUrlPercentReOverride
    _re_percent_encoding: Pattern = urllib3.util.url.PERCENT_RE  # type: ignore[attr-defined]
AttributeError: module 'urllib3.util.url' has no attribute 'PERCENT_RE'. Did you mean: '_PERCENT_RE'?
```

bastimeyer:
Streamlink applies an override for a faulty url encoding in urllib3, as you can see here:
- https://github.com/streamlink/streamlink/blob/5.4.0/src/streamlink/plugin/api/http_session.py#L41-L62
- https://github.com/streamlink/streamlink/blob/5.4.0/tests/test_api_http_session.py#L12-L33

Apparently, urllib3 has renamed its `PERCENT_RE` constant to `_PERCENT_RE` in its `2.0.0` release:
https://github.com/urllib3/urllib3/blame/2.0.0/src/urllib3/util/url.py#L15

**On the 1.x branch, it's still `PERCENT_RE`:**
https://github.com/urllib3/urllib3/blame/1.26.15/src/urllib3/util/url.py#L17

**Streamlink does only support urllib3 1.x, so if you've installed urllib3 2.x in your python environment, then you're ignoring the dependency version constraints of streamlink and requests:**
- streamlink: https://github.com/streamlink/streamlink/blob/5.4.0/setup.cfg#L49-L50
[...]

PR #5325, opened 2023-05-04
build: update version range of urllib3

bastimeyer:
`urllib3` is a transitive dependency of `streamlink`, via `requests`.

We define it here too because of the main CLI module which lists all direct dependencies from the package's metadata (via `importlib.metadata`) when the log level is <= debug, and listing urllib3 is useful. Another reason is that we're applying a couple of urllib3 overrides.

We don't specify an upper version limit though. This is only done by requests, namely `<1.27`. Now that urllib3 has published its 2.0 release last week, there are some issues. For some reason, the version doesn't always get resolved correctly by pip. I just noticed that in the `get-dependencies.sh` script of the Windows builds which unnecessarily downloads and tries to build urllib3 2.x until it falls back to 1.x. And then there was #5324 yesterday.

So let's set the max version here too in order to avoid any confusion and to fix any dependency resolving issues. `requests` won't bump `urllib3` without having a major version bump on its own.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ install_requires =
   pycryptodome >=3.4.3,<4
   PySocks !=1.5.7,>=1.5.6
   requests >=2.26.0,<3.0
-  urllib3 >=1.26.0
+  urllib3 >=1.26.0,<2
   websocket-client >=1.2.1,<2.0
 
 [options.packages.find]


bastimeyer:
Apparently they won't bump the major version:
https://github.com/psf/requests/releases/tag/v2.30.0

This change is required now, otherwise we're having the breaking change described in #5324.

PR #5326, opened 2023-05-04
session: add support for urllib3 2.0

bastimeyer:
- Bump max version constraint from `<2` to `<3`
- Fix `_PERCENT_RE` constant name in URL percent encoding override
- Fix `StreamlinkOptions._set_http_disable_dh()`, remove the override of the old `urllib3.util._ssl.DEFAULT_CIPHERS` list and instead add `streamlink.plugin.api.http_session.TLSNoDHAdapter` which reads the ciphers returned by Python's default SSLContext, adds `:!DH` to the updated ciphers list, and finally replaces the HTTP session's `https://` adapter
- Use `urllib3.util.create_urllib3_context()` for creating `SSLContext` objects in custom `HTTPAdapter`s

----

Resolves #5324 

urllib3 migration guide here:
https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html

PR #5327, opened 2023-05-04
release: 5.5.0

bastimeyer:
Preliminary changelog for tomorrow (5.5.0 on May 5th :open_mouth:).

#5326 hasn't been merged yet. If there's been no review until tomorrow noon or afternoon, I'm going to merge it myself and publish this here afterwards, so we can have a stable release that's not broken when installing from PyPI without downgrading urllib3 to 1.26.15.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## streamlink 5.5.0 (2023-05-05)
+
+Release highlights:
+
+- Added: `--no-config` ([#5314](https://github.com/streamlink/streamlink/pull/5314))
+- Added: `--player-external-http-interface` ([#5295](https://github.com/streamlink/streamlink/pull/5295))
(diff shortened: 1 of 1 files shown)


gravyboat (approved):

bastimeyer:
Going to merge now...

I won't be bumping urllib3 to 2.0 in the Windows builds and Linux AppImages though because I don't want to introduce any potential issues. Support for urllib3 2.0 is there and any incompabilities are fixed now, so users who will install from PyPI or who will install from third party packages who've already upgraded urllib3 to 2.0 will then find out for us if there are any internal issues with the dependency.

Unfortunately we can't test this a bit longer ourselves, because we need to publish the release now in order to fix the dependency resolution issues because requests refused to bump its major version when they clearly introduce incompabilities in their own dependencies. We did all we could by limiting requests to `<3.0.0` in all our previous releases.

PR #5340, opened 2023-05-15
stream.dash: optional video/audio-only streams

bastimeyer:
Add `with_{video,audio}_only` params to `DASHStream.parse_manifest()` which allow video-only or audio-only streams to be included in the returned streams dict. Both parameters default to false, so only muxed streams with video and audio get returned, just like before.

This is a temporary workaround until a proper stream selection mechanism gets implemented at some point in the future.

----

Ref #5339 

I decided to add the `with_` prefix to both parameters, because it makes the intention clearer. Without the prefix, this would imply that muxed streams are not included in the returned streams dict.

```
$ streamlink 'dash://file://./tests/resources/dash/test_timeline_ids.mpd'
[cli][info] Found matching plugin dash for URL dash://file://./tests/resources/dash/test_timeline_ids.mpd
Available streams: 360p+a48k_alt (worst), 360p+a96k_alt, 360p+a48k, 360p+a96k (best)
```

```
$ streamlink 'dash://file://./tests/resources/dash/test_timeline_ids.mpd with_video_only=True'
[cli][info] Found matching plugin dash for URL dash://file://./tests/resources/dash/test_timeline_ids.mpd with_video_only=True
Available streams: 360p_alt (worst), 360p, 360p+a48k_alt, 360p+a96k_alt, 360p+a48k, 360p+a96k (best)
```

```
$ streamlink 'dash://file://./tests/resources/dash/test_timeline_ids.mpd with_audio_only=True'
[cli][info] Found matching plugin dash for URL dash://file://./tests/resources/dash/test_timeline_ids.mpd with_audio_only=True
Available streams: a48k (worst), a96k, 360p+a48k_alt, 360p+a96k_alt, 360p+a48k, 360p+a96k (best)
```

```
$ streamlink 'dash://file://./tests/resources/dash/test_timeline_ids.mpd with_video_only=True with_audio_only=True'
[cli][info] Found matching plugin dash for URL dash://file://./tests/resources/dash/test_timeline_ids.mpd with_video_only=True with_audio_only=True
Available streams: a48k (worst), a96k, 360p_alt, 360p, 360p+a48k_alt, 360p+a96k_alt, 360p+a48k, 360p+a96k (best)
```

https://deploy-preview-5340--streamlink.netlify.app/api.html#streamlink.stream.DASHStream.parse_manifest

### DIFF
--- a/src/streamlink/stream/dash.py
+++ b/src/streamlink/stream/dash.py
@@ -264,6 +264,8 @@ class DASHStream(Stream):
         session: Streamlink,
         url_or_manifest: str,
         period: int = 0,
+        with_video_only: bool = False,
+        with_audio_only: bool = False,
         **args,
     ) -> Dict[str, "DASHStream"]:
         """
@@ -272,6 +274,8 @@ class DASHStream(Stream):
         :param session: Streamlink session instance
         :param url_or_manifest: URL of the manifest file or an XML manifest string
         :param period: Which MPD period to use (index number) for finding representations
+        :param with_video_only: Also return video-only streams, otherwise only return muxed streams
+        :param with_audio_only: Also return audio-only streams, otherwise only return muxed streams
         :param args: Additional keyword arguments passed to :meth:`requests.Session.request`
         """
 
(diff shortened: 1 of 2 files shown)


bastimeyer:
Quick note that users do only have access to these params via the DASH plugin itself (`dash://` URL scheme or `.mpd` URL path suffix). Other plugins which return DASH streams need to set these params themselves.

We could however instead add parameters to the session and CLI (e.g. `--dash-with-{video,audio}-only`), but I'm not really sure if I want that because as said, it's much more of a workaround until a better stream selection gets implemented.

gravyboat:
No need to add params since it's a workaround any way.

PR #5343, opened 2023-05-16
tests: fix CLI versioncheck tests

bastimeyer:
- Fix incorrect `called_once` assertion of the `cache` fixture
- Refactor `pypi` and `cache` fixtures

----

py312 revealed that these tests were broken.

There are more broken tests which I will fix later. Most of them are related to the incorrect and now invalid usage of `Mock().called_once`
https://github.com/python/cpython/commit/1d4d677d1c90fcf4886ded0bf04b8f9d5b60b909
and some changes in regards to the `pathlib` module (haven't figured that out yet)

### DIFF
--- a/tests/cli/utils/test_versioncheck.py
+++ b/tests/cli/utils/test_versioncheck.py
@@ -17,12 +17,21 @@ def test_logger_name():
 
 class TestGetLatest:
     @pytest.fixture()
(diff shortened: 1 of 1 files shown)


bastimeyer:
Btw, I will add py312 test runners once it enters beta on 2023-05-22 and becomes available on GH

PR #5384, opened 2023-06-17
Erttv branch

un1c0rn-the-pwnie:
I developed a plugin for integrating to streamlink cli tool the ERT live TV channels owned by the Hellenic Broadcasting Corporation (ERT). ERT is backed by the government of Greece and it is the national television.

### DIFF
--- a/src/streamlink/plugins/erttv.py
+++ b/src/streamlink/plugins/erttv.py
@@ -1,0 +1,29 @@
+"""
+$description Greek national live TV channels owned by The Hellenic Broadcasting Corporation.
+$url ert.gr
+$type live
+"""
+
+import re
+
+from streamlink.plugin import Plugin, pluginmatcher
+from streamlink.plugin.api import validate
+from streamlink.stream.hls import HLSStream
+
+# example inputs urls:
+# https://www.ert.gr/webtv/ert/tv/live-glm/ert1.html
+# https://www.ert.gr/webtv/ert/tv/live-glm/ert2.html
+# https://www.ert.gr/webtv/ert/tv/live-glm/ert3.html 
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/erttv.py
```diff
@@ -0,0 +1,26 @@
...
+
+@pluginmatcher(re.compile(
+    r"https?://(?:www\.)?ert.gr/",
+))
+class ERTTV(Plugin):
+    def _get_streams(self):
+        re_streams = re.compile(r"var\s*streamww\s*\=\s*\'(https:\/\/.*playlist_dvr\.m3u8*)\'")
+        stream_url = self.session.http.get(self.url, schema=validate.Schema(
+            validate.transform(re_streams.findall),
+        ))[0]
```
bastimeyer, line 22: This would raise an `IndexError` if the regex doesn't match.

bastimeyer (commented):
Thanks for the PR, but we're not going to merge this. The HLS URLs appear to be all static, so a plugin is not needed. Apart from that, the content is also region-locked.

- `https://cbd537474fbad4634b64787657ff6456.msvdn.net/ert1/ert_ev1_main/mainabr/playlist_dvr.m3u8`
- `https://cbd537474fbad4634b64787657ff6456.msvdn.net/ert2/ert_ev2_main/mainabr/playlist_dvr.m3u8`
- `https://cbd537474fbad4634b64787657ff6456.msvdn.net/ert3/ert_ev3_main/mainabr/playlist_dvr.m3u8`

----

https://streamlink.github.io/latest/cli/protocols.html

```sh
streamlink 'hls://FULL_HLS_URL_HERE'
```

Issue #5380, opened 2023-06-15
Implement data extraction via web browser (Chrome devtools protocol)


bastimeyer:
Meta-thread for the implementation of data extraction via a web browser.

----

## Motivation

We need a way to extract data from websites which make it impossible to do this in a pure Python implementation, e.g. because of JS runtime and DOM requirements, etc.

There are Python implementations for executing JS code, and other projects also try to execute JS code in NodeJS, but none of this solves the problem we're facing with the Twitch plugin, not to mention that running untrusted JS code in NodeJS is insanely stupid.

See the following thread/comment which explains this in detail:

- https://github.com/streamlink/streamlink/issues/5370#issuecomment-1582806603

## Rationale

In the linked comment I came to the conclusion that adding `selenium` as a dependency for this task is impossible, because of the detection methods which certain sites implement when the web browser is launched via its `webdriver` interface, which is what Selenium does. It's not just that `Navigator.prototype.webdriver` returns `true` in that case, but there are many other ways for detecting this different environment the browser is run in.

This is a bit unfortunate, because the W3C `webdriver` and `bi-directional webdriver` specifications are the correct way to automate browsers, and it sadly limits us to chromium-based browsers and an implementation based on the Chrome Devtools Protocol (CDP) for remotely and bi-directionally controlling the browser process. The reason for this is that Chromium works just like normal when it's launched with the `--remote-debugging-port=...` parameter (unless the port is set to 0 for some reason) compared to using `chromedriver` (which itself is communicating via the CDP). Firefox also implements the CDP, but only a tiny subset which is unsuitable for our needs. Safari doesn't support this and other non-Chromium browsers are mostly irrelevant.

We can implement this ourselves fairly "easily" without having to rely on many additional dependencies. All that needs to be done is build APIs for launching the chromium-based web browser, the CDP websocket connection and session management, serialization/deserialization classes+methods for the devtools protocol, and finally a CDP client which plugins can use to write extraction logic.

On Streamlink's Matrix/Gitter channel I've already announced yesterday that I have the whole thing already working for the Twitch plugin, but in a very early stage without any tests and without proper separation in the published git branch. That's what I'm going to finish now.

## Roadmap

1. **CDP spec implementation**

   - [x] #5381

   Add `script/generate-cdp.py` which parses the JSON data of the CDP specification and which then generates {,de}serialization classes and methods for CDP commands and events with proper typing data. Those generated python modules will be committed into the code base and can be updated if needed.

[...]

PR #5386, opened 2023-06-17
webbrowser: implement Webbrowser / ChromiumWebbrowser

bastimeyer:
Part 2/4 of #5380 

Opening this as a draft was well until I have the rest finished.

----

This

1. Adds a utility executable resolver function with support for custom values, default names and fallback paths.
2. Adds a utility function for finding an available port on a specific network interface
   - No tests, because I don't think mocking everthing here makes sense, and neither does running this without mocking
3. Adds the `trio` dependency, as well as `pytest-trio` and `trio-typing`
   - `trio-typing` looks like it's going to get merged into `trio` soon
   - `pytest-trio` is used for initializing the runloop in async tests, similar to `pytest-asyncio`.
     Having implementations for both `asyncio` and `trio` seems to cause minor issues with trio's runloop in the tests, and on Windows it triggers a warning when asyncio was run prior to its own runloop. This is related to `signal.set_wakeup_fd()`. This can be resolved in the future by re-implementing the `utils.processoutput` based on trio instead of asyncio. The other asyncio stuff is purely concurrency related and not I/O related, so it's irrelevant which runloop we're using. The warning can be ignored for now.
4. Adds the `Webbrowser` base class with the basic launch implementation
   - I initially had custom process cancellation logic added, when the process doesn't end on SIGTERM on POSIX systems. The default behavior of trio waits for 5 seconds and then sends SIGKILL. My plan was to wait a bit less and log a warning instead of triggering a warning via the warnings system, but I ran into issues with the custom implementation while testing. Some weird internal issues with trio and the stdlib subprocess module. Doesn't matter though.
5. Adds the `ChromiumWebbrowser` class
   - This includes a list of all known Chromium/Chrome executable names (with Chromium having higher priority)
   - With fallback paths for Windows and macOS (because of `PATH` env var nonsense)
   - I carefully selected launch parameters for disabling lots of unneeded features and behaviors which we don't need or which are bad for our purposes. This is all annotated and linked. There's probably more that could be added.
   - ~~Headless mode is currently not added, but this can be done with a simple class argument. I can't get a valid client-integrity token on Twitch with headless mode, regardless of what others are saying, so I don't see the point right now.~~
   - Chromium-based browsers I tested (CLI arguments for custom paths are not included in this PR):
     - Chromium (Arch): worked
     - Google Chrome (Windows): worked
     - Brave (Windows): worked
     - Vivaldi (Windows): didn't work, despite a successful CDP session, didn't check further
     - Opera (Windows): didn't work, didn't check further

### DIFF
--- a/tests/webbrowser/test_webbrowser.py
+++ b/tests/webbrowser/test_webbrowser.py
@@ -1,0 +1,144 @@
+from __future__ import annotations
+
+from contextlib import AbstractContextManager, nullcontext
+from pathlib import Path
+from signal import SIGTERM
+from typing import List, Optional
+
+import pytest
+import trio
+
+from streamlink.compat import is_win32
+from streamlink.webbrowser.exceptions import WebbrowserError
+from streamlink.webbrowser.webbrowser import Webbrowser
+
+
+class _FakeWebbrowser(Webbrowser):
+    @classmethod
+    def launch_args(cls) -> List[str]:
+        return ["foo", "bar"]
+
+
+class TestInit:
+    @pytest.mark.parametrize(("executable", "resolve_executable", "raises"), [
+        pytest.param(
+            None,
+            None,
+            pytest.raises(WebbrowserError, match="Could not resolve web browser executable"),
+            id="Failure with default path",
+        ),
+        pytest.param(
+            "custom",
+            None,
+            pytest.raises(WebbrowserError, match="Could not resolve web browser executable: custom"),
+            id="Failure with custom path",
+        ),
+        pytest.param(
+            None,
+            "default",
+            nullcontext(),
+            id="Success with default path",
+        ),
+        pytest.param(
+            "custom",
+            "custom",
+            nullcontext(),
+            id="Success with custom path",
+        ),
+    ], indirect=["resolve_executable"])
+    def test_resolve_executable(self, resolve_executable, executable: Optional[str], raises: nullcontext):
+        with raises:
+            Webbrowser(executable=executable)
+
+    def test_arguments(self):
+        webbrowser = _FakeWebbrowser()
+        assert webbrowser.executable == "default"
+        assert webbrowser.arguments == ["foo", "bar"]
+        assert webbrowser.arguments is not _FakeWebbrowser.launch_args()
+
+
+class TestLaunch:
+    @pytest.mark.trio()
+    async def test_terminate_on_nursery_exit(self, caplog: pytest.LogCaptureFixture, webbrowser_launch):
+        nursery: trio.Nursery
+        process: trio.Process
+        async with webbrowser_launch() as (nursery, process):  # noqa: F841
+            assert process.poll() is None, "process is still running"
+
+        assert process.poll() == (1 if is_win32 else -SIGTERM), "Process has been terminated"
+        assert [(record.name, record.levelname, record.msg) for record in caplog.records] == [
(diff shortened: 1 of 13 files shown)


tests/webbrowser/test_webbrowser.py
```diff
@@ -0,0 +1,142 @@
...
+        async with webbrowser_launch(timeout=10) as (nursery, process):  # noqa: F841
+            assert process.poll() is None, "process is still running"
+            assert process.stdin
+            await process.stdin.send_all(str(exit_code).encode() + b"\r\n")
+            await trio.sleep(5)
+
+        assert process.poll() == exit_code, "Process has ended with the right exit code"
+        assert [(record.name, record.levelname, record.msg) for record in caplog.records] == [
+            ("streamlink.webbrowser.webbrowser", "warning", "Web browser process ended early"),
+        ]
```
bastimeyer, line 133: Apparently there's an issue with this test on Windows.

One CI job of the dependent PR #5388 which includes this branch has failed:
https://github.com/streamlink/streamlink/actions/runs/5335524587/jobs/9668866676?pr=5388#step:5:271

> ```
> >       assert process.poll() == exit_code, "Process has ended with the right exit code"
> E       AssertionError: Process has ended with the right exit code
> E       assert 1 == 0
> E        +  where 1 = <bound method Process.poll of <trio.Process ['C:\\hostedtoolcache\\windows\\Python\\3.7.9\\x64\\python.exe', '-c', 'import sys; sys.exit(int(sys.stdin.readline()))']: exited with status 1>>()
> E        +    where <bound method Process.poll of <trio.Process ['C:\\hostedtoolcache\\windows\\Python\\3.7.9\\x64\\python.exe', '-c', 'import sys; sys.exit(int(sys.stdin.readline()))']: exited with status 1>> = <trio.Process ['C:\\hostedtoolcache\\windows\\Python\\3.7.9\\x64\\python.exe', '-c', 'import sys; sys.exit(int(sys.stdin.readline()))']: exited with status 1>.poll
> ```

No idea what's going on here...

The fake web browser process is this (from the `webbrowser_launch` fixture in `conftest.py`)
```sh
python -c 'import sys; sys.exit(int(sys.stdin.readline()))'
```
and it's supposed to exit with code 0 when `0\r\n` is written on its stdin pipe within the 5 second window (`sys.stdin.readline()` blocks).

Why is the process exiting with code 1? This looks like a generic exit code on Windows. It's the same when terminating the process, hence why we can't check for `-SIGTERM` in the other tests. Did the pipe break or what's going on here?

I've been running the webbrowser tests on my Arch Linux system in a loop for a while, and there are no timing issues. Guess I will have to debug this race condition in my Windows VM. :vomiting_face: 

Help or hints are appreciated...

bastimeyer: I had the `tests/webbrowser` tests now running in a loop in my Windows VM for over 45 mins.... More than a thousand iterations without failures.
```fish
while pytest --disable-warnings -qq tests/webbrowser/; end
```
I'll ignore this failure for now and if it comes up again, I will ignore the test on Windows.

bastimeyer:
Rebased to current head of master, because this PR is the base branch for other PRs until it got merged into master.

Also disabled that one flaky test on Windows because I'm already annoyed by it.

bastimeyer:
Thinking about it, it would probably make sense adding MSEdge to the executables list and fallback path list on Windows, even with a higher priority, since it's installed by default and can't be removed on this OS. The old non-Chromium Edge is dead, right?

PR #5381, opened 2023-06-15
Add Chrome Devtools Protocol modules

bastimeyer:
Part 1/4 of #5380 

CDP documentation here:
https://chromedevtools.github.io/devtools-protocol/

Generator script borrowed from:
https://github.com/HyperionGray/python-chrome-devtools-protocol

----

This

1. Adds the unmodified script file with added license info
2. Fixes linting issues
3. Adds features, fixes bugs and fixes style of generated code
4. Adds the following CDP domains:
   `fetch`, `runtime`, `input`, and `dom`, as well as all their dependent domains and the mandatory `target` and `inspector` domains.
   Anything else is not worth adding in my opinion. As you can see, this already adds 17k lines of code...

Opening as a draft until I have the remaining stuff ready.

PR #5388, opened 2023-06-19
webbrowser.cdp: implement CDP connection

bastimeyer:
Part 3/4 of #5380

This currently includes the commits of #5381 (part 1) and #5386 (part 2), because this code is needed for the CDP connection stuff. Once those PRs have been merged into master, I will rebase this branch.

So the only relevant commits of this PR are the ones after the merge commit. See the check marks of the HEAD commits of those other branches.

----

This
1. Adds the `trio-websocket` and `typing-extensions` runtime dependencies.
   - A bit unfortunate that we have to add another dependency for websockets, but it is what it is... Async I/O via trio is a requirement for a proper implementation.
   - `typing-extensions` is required, because I needed `TypeAlias` for making the `CDPEventListener`'s typing work. This conveniently also adds `Self` and other things.
2. Implements the CDP connection and session management.
   - As mentioned in #5380, the basic design is borrowed from the `trio-chrome-devtools-protocol` project, which uses the MIT license. This is properly annotated with the included license text. I applied several improvements and fixes to the code. This is all fully tested with full coverage.
   - Trio 0.22 currently requires the `from __future__ import annotations` workaround so the memory channels can have proper typing information, which is important. This will be resolved in their next release (they are waiting for the EOL of py37), but I don't think it's worth removing the import and bumping the version requirement once it's released.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -378,6 +378,8 @@ runtime   `pycryptodome`_           Used for decrypting encrypted streams
 runtime   `PySocks`_                Used for SOCKS Proxies
 runtime   `requests`_               Used for making any kind of HTTP/HTTPS request
 runtime   `trio`_                   Used for async concurrency and I/O in some parts of Streamlink
+runtime   `trio-websocket`_         Used for WebSocket connections on top of the async trio framework
+runtime   `typing-extensions`_      Used for backporting runtime support of certain type hints on older Python versions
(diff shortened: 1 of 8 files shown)


bastimeyer:
Rebased to current head of master, because this PR is the base branch for other PRs until it got merged into master.

PR #5395, opened 2023-06-22
cli.argparser: allow disabling session options

bastimeyer:
`streamlink_cli.argparser` should be able to set session options which default to `True` to `False`. Fix that by checking whether the argument dest on the argparse namespace is not `None` (aka set).

Currently, there are no session options which default to `True`.

----

Required for being able to add `--webbrowser=false` with respective default value of `True` on the Streamlink session.


bastimeyer:
I did a mistake here. This unintentionally now always sets `--http-no-ssl-verify`, which is bad.

Please don't immediately merge PRs without a review. I sometimes make mistakes.

gravyboat:
I did review it, just didn't realize that the change would also impact `-http-no-ssl-verify`.

PR #5398, opened 2023-06-24
docs: split API docs into multiple documents

bastimeyer:
The API docs should be split up for better legibility (I intend on adding docs for the webbrowser API). This also removes an unnecessary layer in the sidebar on the right. Some references needed an update, so I checked for the rest as well, but I didn't check thoroughly...

Current mess:

New (mess):
https://deploy-preview-5398--streamlink.netlify.app/api.html

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,97 +1,12 @@
 API Reference
 =============
 
-This is a reference of all the available API methods in Streamlink.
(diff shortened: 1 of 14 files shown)


gravyboat:
The new (mess) is still better than the old mess!

gravyboat (approved):

PR #5410, opened 2023-07-02
webbrowser.cdp: implement CDP client

bastimeyer:
Part 4/4 of #5380

This currently includes the commits of #5381 (part 1), #5386 (part 2) and #5388 (part 3), because this code is needed for the CDP connection stuff. Once those PRs have been merged into master, I will rebase this branch. The base of the PR-branch is also outdated, hence the current merge-conflict (API docs related).

----

This

1. Adds various `webbrowser-...` session options and `--webbrowser-...` CLI arguments and finally implements the `CDPClient` class which plugins can use for extracting data from websites. It manages launching the Chromium-based web browser, connects to its remote debugging interface and establishes a CDP session instance. The API for performing the CDP session tasks is implemented in the `CDPClientSession` class, namely navigating, adding request/response handlers, managing requests/responses and evaluating JavaScript. I didn't cover all edge-cases and method parameter combinations in the tests, but it's still pretty extensive and has full coverage.

----

### Notes

- I struggled a bit with my API design choices, with lots of back-and-forth work. In the end I settled with the most simple approach which is also less noisy on the websocket connection. This version should be fine for most cases though. And certainly more than enough for the Twitch plugin.
- This also adds API documentation, but I didn't proof-read anything before opening this PR. I wrote this stuff last week when my API designs weren't final and probably has some mistakes in it and isn't as polished as it could be. Will re-check when rebasing.
- In the docs I added a note that all these APIs are considered unstable, but I will add another comment to the main py module of the webbrowser package when rebasing/updating the other PRs which this PR depends on.

### Concerns

There is still a few concerns that I have, and that is that plugins will need to choose this `CDPClient` explicitly. What I mean by that is that in the future, to be able to support more browsers, the W3C's webdriver-bidi interface could be implemented instead. That would make the `CDPClient` obsolete.

This however depends on whether the webdriver-bidi stuff doesn't interfere with any detection scripts which this whole webbrowser endeavor is meant to circumvent. Having multiple implementations is also problematic, because there's no mechanism that would let users choose.

And talking about choosing, the `CDPClient` obviously depends on Chromium-based web browsers. While Chromium and Google Chrome should work OOTB, other forks might cause problems or are simply incompatible. There's nothing like specific browser profiles which are implemented that could modify launch parameters for certain forks. If this was required to get some forks working, then this would cause design problems in the `CDPClient`'s launch logic.

All these concerns can be ignored for now though. I can't make everyone happy with this, and targeting the basics with the highest user share makes the most sense anyway.

### DIFF
--- a/src/streamlink/webbrowser/cdp/client.py
+++ b/src/streamlink/webbrowser/cdp/client.py
@@ -1,0 +1,424 @@
+import base64
+import re
+from contextlib import asynccontextmanager
+from dataclasses import dataclass
+from typing import Any, AsyncGenerator, Awaitable, Callable, Coroutine, List, Mapping, Optional, Set
+
+import trio
+
+from streamlink.session import Streamlink
+from streamlink.webbrowser.cdp.connection import CDPConnection, CDPSession
+from streamlink.webbrowser.cdp.devtools import fetch, network, page, runtime, target
+from streamlink.webbrowser.cdp.exceptions import CDPError
+from streamlink.webbrowser.chromium import ChromiumWebbrowser
+
+
+try:
+    from typing import Self, TypeAlias  # type: ignore[attr-defined]
+except ImportError:  # pragma: no cover
+    from typing_extensions import Self, TypeAlias
+
+
+TRequestHandlerCallable: TypeAlias = Callable[["CDPClientSession", fetch.RequestPaused], Awaitable]
+
+
+_re_url_pattern_wildcard = re.compile(r"(.+?)?(\\+)?([*?])")
+
+
+@dataclass
+class RequestPausedHandler:
+    async_handler: TRequestHandlerCallable
+    url_pattern: str = "*"
+    on_request: bool = False
+
+    def __post_init__(self) -> None:
+        self._re_url: re.Pattern = self._url_pattern_to_regex_pattern(self.url_pattern)
+
+    def matches(self, request: fetch.RequestPaused) -> bool:
+        on_request: bool = request.response_status_code is None and request.response_error_reason is None
+        return on_request is self.on_request and self._re_url.match(request.request.url) is not None
+
+    @staticmethod
+    def _url_pattern_to_regex_pattern(url_pattern: str) -> re.Pattern:
+        pos = 0
+        regex = ""
+
+        for match in _re_url_pattern_wildcard.finditer(url_pattern):
+            regex += re.escape(match[1]) if match[1] else ""
+            if match[2]:
+                if len(match[2]) % 2:
+                    regex += f"{re.escape(match[2][:-1])}\\{match[3]}"
+                else:
+                    regex += re.escape(match[2])
+                    regex += ".+" if match[3] == "*" else "."
+            else:
+                regex += ".+" if match[3] == "*" else "."
+
+            pos = match.end()
+
+        regex += re.escape(url_pattern[pos:])
+
+        return re.compile(f"^{regex}$")
+
+
+@dataclass
+class CMRequestProxy:
+    body: str
+    response_code: int
+    response_headers: Optional[Mapping[str, str]]
+
+
+class CDPClient:
+    """
+    The public interface around :class:`ChromiumWebbrowser <streamlink.webbrowser.chromium.ChromiumWebbrowser>`
+    and :class:`CDPConnection <streamlink.webbrowser.cdp.connection.CDPConnection>`.
+
+    It launches the Chromium-based web browser, establishes the remote debugging WebSocket connection using
+    the `Chrome Devtools Protocol <https://chromedevtools.github.io/devtools-protocol/>`_,  and provides
+    the :meth:`session()` method for creating a new :class:`CDPClientSession` that is tied to an empty new browser tab.
+
+    :class:`CDPClientSession` provides a high-level API for navigating websites, intercepting network requests and responses,
+    as well as evaluating JavaScript expressions and retrieving async results.
+
+    Don't instantiate this class yourself, use the :meth:`CDPClient.launch()` async context manager classmethod.
+
+    For low-level Chrome Devtools Protocol interfaces, please see Streamlink's automatically generated
+    ``streamlink.webbrowser.cdp.devtools`` package, but be aware that only a subset of the available domains is supported.
+    """
+
+    def __init__(self, cdp_connection: CDPConnection, nursery: trio.Nursery):
+        self.cdp_connection = cdp_connection
+        self.nursery = nursery
+
+    @classmethod
+    def launch(
+        cls,
+        session: Streamlink,
+        runner: Callable[[Self], Coroutine],
+        executable: Optional[str] = None,
+        timeout: Optional[float] = None,
+        cdp_host: Optional[str] = None,
+        cdp_port: Optional[int] = None,
+        cdp_timeout: Optional[float] = None,
+        headless: Optional[bool] = None,
+    ) -> Any:
+        """
+        Start a new :mod:`trio` runloop and do the following things:
+
+        1. Launch the Chromium-based web browser using the provided parameters or respective session options
+        2. Initialize a new :class:`CDPConnection <streamlink.webbrowser.cdp.connection.CDPConnection>`
+           and connect to the browser's remote debugging interface
+        3. Create a new :class:`CDPClient` instance
+        4. Execute the async runner callback with the :class:`CDPClient` instance as only argument
+
+        If the ``webbrowser`` session option is set to ``False``, then a :exc:`CDPError` will be raised.
+
+        Example:
+
+        .. code-block:: python
+
+            async def fake_response(client_session: CDPClientSession, request: devtools.fetch.RequestPaused):
+                if request.response_status_code is not None and 300 <= request.response_status_code < 400:
+                    await client_session.continue_request(request)
+                else:
+                    async with client_session.alter_request(request) as cmproxy:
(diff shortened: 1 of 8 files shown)


src/streamlink_cli/argparser.py
```diff
@@ -1272,6 +1272,73 @@ def build_parser():
...
+    )
+    webbrowser.add_argument(
+        "--webbrowser-cdp-port",
+        metavar="PORT",
+        type=num(int, ge=0, le=65535),
+        help="""
+        Port for the web browser's Chrome Devtools Protocol interface.
+
+        Tries to find a free port by default.
+        """,
```
gravyboat, line 1319: Hmm, not sure about allowing users to allocate the lower port values which may be used but I guess we'll let them shoot themselves in the foot here.

src/streamlink_cli/argparser.py
```diff
@@ -1272,6 +1272,73 @@ def build_parser():
...
+
+        Default is true.
+        """,
+    )
+    webbrowser.add_argument(
+        "--webbrowser-executable",
+        metavar="PATH",
+        help="""
+        Path to the web browser's executable.
+        """,
```
gravyboat, line 1292: I'm always torn on this but should there be an example in the help? I feel like we're going to get a lot of questions on this.

src/streamlink/webbrowser/chromium.py
```diff
@@ -0,0 +1,178 @@
...
+
+        if is_darwin:
+            return [
+                "/Applications/Chromium.app/Contents/MacOS/Chromium",
+                str(Path.home() / "Applications/Chromium.app/Contents/MacOS/Chromium"),
+                "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
+                str(Path.home() / "Applications/Google Chrome.app/Contents/MacOS/Google Chrome"),
+            ]
+
+        return []
```
gravyboat, line 52: What do you think about preferring Chromium here as the first choice over Google Chrome? I'd rather we try to default to the less evil option first if we can. Same note about the tests but I'm not going to mention it there.

src/streamlink/webbrowser/chromium.py
```diff
@@ -0,0 +1,178 @@
...
+            # Disables the service process from adding itself as an autorun process
+            #   This does not delete existing autorun registrations, it just prevents the service from registering a new one
+            "--no-service-autorun",
+            # Avoid potential instability of using Gnome Keyring or KDE wallet
+            "--password-store=basic",
+            # No initial CDP target (no empty default tab)
+            "--silent-launch",
+            # Use mock keychain on Mac to prevent the blocking permissions dialog asking:
+            #   Do you want the application "Chromium.app" to accept incoming network connections?
+            "--use-mock-keychain",
```
gravyboat, line 120: The rest of the options look good here, this is the only one I'm not sure about as it's a bit of a security bypass. Probably won't matter to 99.9% of users though.

src/streamlink/webbrowser/chromium.py
```diff
@@ -0,0 +1,178 @@
...
+
+            async with super()._launch(self.executable, arguments, timeout=timeout) as nursery:
+                yield nursery
+
+            # Even though we've awaited the process termination in the async generator above,
+            # the rmtree() call of the temp-dir's context manager can sometimes still fail.
+            # This is probably caused by filesystem commits of the OS, not sure,
+            # but we have to wait a bit in order to be able to gracefully remove the temp user data dir.
+            # A terrible solution to use a static timer :(
+            await trio.sleep(0.5)
```
gravyboat, line 164: Bummer :\

src/streamlink/webbrowser/cdp/client.py
```diff
@@ -0,0 +1,406 @@
...
+    ) -> Any:
+        """
+        Start a new :mod:`trio` runloop and do the following things:
+
+        1. Launch the Chromium-based web browser using the provided parameters or respective session options
+        2. Initialize a new :class:`CDPConnection` and connect to the browser's remote debugging interface
+        3. Create a new :class:`CDPClient` instance
+        4. Execute the async runner callback with the :class:`CDPClient` instance as only argument
+
+        If the ``webbrowser`` session option is set to ``False``, then a :exc:`CDPError` will be raised.
```
gravyboat, line 107: Might be some annoying comments about this because people are confused, should be added as a TODO for the release notes to clarify.

src/streamlink/webbrowser/cdp/client.py
```diff
@@ -0,0 +1,406 @@
...
+                            If unset, falls back to the ``webbrowser-executable`` session option.
+                            Otherwise, it'll be looked up according to the rules of the :class:`ChromiumBrowser` implementation.
+        :param timeout:     Optional global timeout value, including web browser launch time.
+                            If unset, falls back to the ``webbrowser-timeout`` session option.
+        :param cdp_host:    Optional remote debugging host.
+                            If unset, falls back to the ``webbrowser-cdp-host`` session option.
+                            Otherwise, ``127.0.0.1`` will be used.
+        :param cdp_port:    Optional remote debugging port.
+                            If unset, falls back to the ``webbrowser-cdp-port`` session option.
+                            Otherwise, a random free port will be chosen.
```
gravyboat, line 133: > Otherwise, a random free port will be chosen.

Any concerns about random not including a specified range of upper level typically unused ports?

gravyboat (commented):
I looked at the previous PRs a bit already and my comments are below. Here are some general notes:

* Slightly concerned about the maintenance overhead of the formatting modifications to the  generate-cdp.py file but not a huge deal. I've looked over it but I don't have any specific comments about it versus the original.
* I've looked over the code on a per commit basis, but the comments are all grouped here for easier reviewing.
* The API design seems fine, it's easier to understand and I'd prefer starting there. No reason for unnecessary complexity.

src/streamlink/webbrowser/chromium.py
bastimeyer: I am not aware of any Chromium builds on Windows. If there are any, what are their default paths?

bastimeyer:
> * but the comments are all grouped here for easier reviewing

All these diffs you're commenting on will disappear once I rebased the PR branch.

src/streamlink/webbrowser/chromium.py
gravyboat: There are some projects like https://chromium.woolyss.com/ which provide various "google free" builds but you're right there's nothing I can find that has a default path. That's what I get for not having used Chrome or Chromium in almost a decade I guess. 🤷‍♂️

gravyboat:
> All these diffs you're commenting on will disappear once I rebased the PR branch.

That's fine.

src/streamlink/webbrowser/cdp/client.py
bastimeyer: Doesn't matter, because it's chosen by the system and not actually "random".
https://github.com/streamlink/streamlink/blob/20c8b6be542539681c96708d4f2f3f993463312b/src/streamlink/utils/socket.py#L6-L20

src/streamlink/webbrowser/cdp/client.py
bastimeyer: I don't know what you mean. It's up to each plugin implementing the webbrowser API to handle errors. This is just the description of one of them.

Please take a look at my cdp-wip branch mentioned in the meta thread and check the Twitch plugin. Then you can play with the CLI args and see what happens. If it's disabled, then no client-integrity token will/can be retrieved. If this breaks the access token retrieval (which it currently does not), then the user will see why.

src/streamlink/webbrowser/chromium.py
bastimeyer: No idea what you mean here. See the links above that list where the parameter is described in more detail. This prevents annoying unrelated popups on macOS, nothing more. This has to run as transparently as possible. There is no security issue, because a completely new and empty user dir gets used.

src/streamlink/webbrowser/webbrowser.py
```diff
@@ -0,0 +1,115 @@
...
+        return []
+
+    @classmethod
+    def launch_args(cls) -> List[str]:
+        return []
+
+    def __init__(self, executable: Optional[str] = None):
+        resolved = resolve_executable(executable, self.names(), self.fallback_paths())
+        if not resolved:
+            raise WebbrowserError(f"Could not resolve web browser executable{f': {executable}' if executable else ''}")
```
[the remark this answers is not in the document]
bastimeyer: This can be fine-tuned later. This code isn't part of this PR anyway.

bastimeyer:
I updated the CLI argparser help texts and some docstrings. Here are the rendered results (the API reference isn't too much important)

- https://deploy-preview-5410--streamlink.netlify.app/cli.html#web-browser-options
- https://deploy-preview-5410--streamlink.netlify.app/api/webbrowser.html

I believe the rest should be fine for now. If there are any issues down the road, those can be fixed later. The whole thing is tagged as unstable anyway.

If there's a desire for a dedicated docs page, e.g. why this is necessary, I can also add this later.

----

As a way forward in regards to a potential webdriver-bidi implementation, this could be achieved by adding the `--webbrowser-implementation=...` option. This would default to `cdp` now, but since it'd only allow a single value, it doesn't make sense adding it now.

The other options should be generic enough for letting the user choose a different browser in that case, e.g. Firefox, or for choosing different timeout values. The host and port options could change maybe, but that could be done using fallback options, so no big deal here really.

gravyboat:
Docs look good to me!

PR #5416, opened 2023-07-05
webbrowser.chromium: add MSEdge fallback on win32

bastimeyer:
This adds Microsoft Edge to the Chromium fallback list on win32.

It actually prioritizes MS Edge over Google Chrome, which I think is a better solution because on Windows it's guaranteed that Edge is installed. Both are non-free browsers on a non-free OS, so I don't care one single bit about what people will say about edge vs chrome.

This does only affect the fallback paths, which means `chromium`, etc. is still looked up first in the `PATH` env var, even though this won't work on Windows unless people have special setups.

For those users of Windows who don't want to run Edge, they either won't notice it because it runs headlessly by default, or they can change it via the `--webbrowser-executable` option.

### DIFF
--- a/src/streamlink/webbrowser/chromium.py
+++ b/src/streamlink/webbrowser/chromium.py
@@ -26,7 +26,20 @@ class ChromiumWebbrowser(Webbrowser):
     @classmethod
     def fallback_paths(cls) -> List[Union[str, Path]]:
         if is_win32:
-            return [
+            ms_edge: List[Union[str, Path]] = [
+                str(Path(base) / sub / "msedge.exe")
+                for sub in (
+                    "Microsoft\\Edge\\Application",
+                    "Microsoft\\Edge Beta\\Application",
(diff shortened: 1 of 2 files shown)


gravyboat:
Is it even worth supporting 32 bit platforms? The Steam survey: https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam essentially shows the number of people running 32 bit as a rounding error and Windows 11 doesn't even have a 32 bit version. Doesn't seem worth the time to support users on dead platforms.

bastimeyer:
- https://superuser.com/questions/811262/why-chrome-64-bit-is-installed-in-32-bit-program-files
- https://techcommunity.microsoft.com/t5/discussions/the-edge-64-bit-browser-installation-path-is-still-in-the/m-p/3084404

gravyboat (approved):

PR #5419, opened 2023-07-05
ci.github: remove custom lxml cp312 wheel

bastimeyer:
lxml has published 4.9.3 with `cp312` wheels for `manylinux_2_28` and `win_amd64`
https://pypi.org/project/lxml/4.9.3/#files

~~This means our custom lxml wheel is not needed anymore and the Windows CI runner can be enabled. Tests will fail because of other issues which will get fixed by #5418.~~ rebased

### DIFF
--- a/script/install-dependencies.sh
+++ b/script/install-dependencies.sh
@@ -7,13 +7,6 @@ set -ex
 python -m pip install --disable-pip-version-check --upgrade pip setuptools
 python -m pip install --upgrade -r dev-requirements.txt
 
-# install a custom-built lxml wheel for the py312 CI runner until an official one is available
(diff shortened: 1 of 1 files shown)


bastimeyer:
Why does Windows fail?

- https://github.com/streamlink/streamlink/actions/runs/5469653062/jobs/9958828639?pr=5419#step:5:14
- https://github.com/streamlink/streamlink/actions/runs/5469653062/jobs/9958828639?pr=5419#step:6:16

Looks like a different dependency is failing on `cp312-win_amd64`

bastimeyer:
This is caused by pycryptodome when cffi is installed in the same environment. cffi is a transitive dependency of streamlink, required by trio. I've opened an issue on pycryptodome's issue tracker earlier today. Not sure whose fault this is. Could be cffi...

I've disabled Windows cp312 again, so this can get merged.

PR #5427, opened 2023-07-09
Removed unnecessary dependencies

pepsik-kiev:
Removed unnecessary dependencies. Simplified code to perform the same functionality


bastimeyer:
Well, thanks for the PR, but I don't understand the intention and motivation of this PR at all. This looks like random code changes which you didn't actually test.

First, what is the point of removing the 3rd party module `packages.requests_file` (which is a hardcoded fork of the [`requests-file`](https://github.com/dashea/requests-file) package on pypi with the added apache2 license) and copying a different `FileAdapter` with less functionality from online forums or from other projects without providing a license?
- https://stackoverflow.com/a/27786580
- https://github.com/search?q=return+400%2C+%22Path+Not+A+File%22&type=code

Second, you can't just simply break the Streamlink session API and remove certain methods and remove plugin overriding logic. Not to mention that this will break several dozens of test cases.

Going to close this. If you want to argue for implementing a simpler version of the FileAdapter, then please do this properly.

Issue #5401, opened 2023-06-27
release: 6.0.0 tracking issue


bastimeyer:
**Tracking issue of breaking changes of release 6.0.0**

Official support for py37 ends today, which means we're going to drop support for this old Python version in Streamlink as well.
After #5302 will have been merged into master, this will put the branch into a breaking change state, so other breaking changes will follow.

This doesn't have to be done today though. Streamlink doesn't use the git-flow development model or something similar of that kind, so having breaking changes on the only development+release branch is always risky when other changes aren't ready yet and when we may have to publish quick bugfix releases prior to the intended major version release. We don't want to branch off for patch releases and also want to avoid having unnecessary/ugly revert commits.

There's also the webbrowser implementation which isn't finished yet (I'm getting close though), and I'd like to have it included in 6.0.0, at least in an experimental state, so as long as it's not yet ready, merging breaking changes is probably not a good idea.

----

- [x] #5302 
  Ready
- [x] #5310
  Should be pretty much ready, but I think it could use some documentation (in addition to the announcements in the upcoming changelog)
- [x] #5033 
  Needs a rebase, linting fixes and merge conflict resolving.
- [x] #5402 
- [x] #5403 
- [x] #5404
- [x] #5429
- [x] #5449 

----

Anything else that is currently considered deprecated and could be removed? We don't have deadlines for anything on the deprecations list...
https://streamlink.github.io/deprecations.html

- `Plugin.can_handle_url()` / `Plugin.priority()` looks like a good candidate, especially with the other planned `Plugin` changes.
- `plugin.api.validate.text` as well
- Deprecated CLI args and config paths are rather irrelevant and would only bloat up the changelog for end users who won't read it anyway and confuse them.

bastimeyer:
I will probably also replace the currently used sdist signing key. I don't have the passphrase and in a short conversation with beardypig a couple of months ago, he said that he didn't have it anymore either, which means nobody except GitHub CI via its repo secrets has access to it. Running a special action that sends it to someone else is not a great idea, but could work. Creating a new one is much better though.

gravyboat:
> Creating a new one is much better though.

Agreed, there's no reason to risk it when a new one will get the job done. Let's make sure it is mentioned in the changelog when appropriate.

bastimeyer:
With the basic webbrowser API stuff being finished and submitted in PRs now, we can now go ahead and merge the breaking changes into master.

I will have a look at this later and update the stuff I talked about in the OP.

[...]

PR #5449, opened 2023-07-17
build: new Streamlink signing key 44448A298D5C3618

bastimeyer:
Replace old signing key:
E2B794C7C2C37162E5E2A097E3DB9E282E390FA0

With new one:
CDAC41B9122470FAF357A9D344448A298D5C3618

----

Public key:

```
-----BEGIN PGP PUBLIC KEY BLOCK-----

mDMEZLWhshYJKwYBBAHaRw8BAQdAu0sD5Ez8mfroVXpEohGHAeH1H2xduEHsYHkG
IciKHdy0MlN0cmVhbWxpbmsgc2lnbmluZyBrZXkgPHN0cmVhbWxpbmtAcHJvdG9u
bWFpbC5jb20+iJAEExYIADgWIQTNrEG5EiRw+vNXqdNERIopjVw2GAUCZLWhsgIb
AwULCQgHAgYVCgkICwIEFgIDAQIeAQIXgAAKCRBERIopjVw2GH2MAQCxW0AG6K8A
P7hUfQHU7eMCBlNxhr1JcCcxl2ajcaOojAD9E+klQNd7VIFH3NcVlw42ADfFGdph
t6uwU8886DclIgO4MwRktaJCFgkrBgEEAdpHDwEBB0CgjJ8c9OXf+00EpqZb6Oxk
hIDDbLJKG04d7eRigLUYt4j1BBgWCAAmFiEEzaxBuRIkcPrzV6nTRESKKY1cNhgF
AmS1okICGwIFCQPCZwAAgQkQRESKKY1cNhh2IAQZFggAHRYhBImk76VlO4meZhF5
mRrrZADton2pBQJktaJCAAoJEBrrZADton2pcMUA/jwTLdeanUFXpaEY1dOkHR3x
skagH/y1TT7r0jIkfRKZAP0dadPIYHcIwm5y8ijs2Th4FRfwTzmKGhLv4hfNRxVD
BK3gAQCl+067glcwcgeWHq0mxhoWTN1oaaMBbQHN423fTs7TjgD/e5Fra1oTLW+I
Acu3T/qLqTiOpSHzlNQOKogPilLU7Q4=
=IEx0
-----END PGP PUBLIC KEY BLOCK-----
```

----

Ref #5401 

The old signing key was created by @beardypig in 2018. Unfortunately, he never shared the passphrase or the unencrypted private key data with anybody else because we didn't set up proper communication channels back then. Since he apparently misplaced the passphrase and private key, nobody except GitHub is in control of the signing key now, so unless we do some stupid CI shenanigans, we'll need to generate a new one. This also unfortunately means that we can't revoke the old one.

The old one was an RSA4096 signing key and the new one is an ED25519 one with a dedicated sub-key for signing, with an initial expiration time set to 2 years:
`89A4EFA5653B899E661179991AEB6400EDA27DA9`

```
$ gpg --keyid-format long -k CDAC41B9122470FAF357A9D344448A298D5C3618 
pub   ed25519/44448A298D5C3618 2023-07-17 [SC]
      CDAC41B9122470FAF357A9D344448A298D5C3618
uid                 [ultimate] Streamlink signing key <streamlink@protonmail.com>
sub   ed25519/1AEB6400EDA27DA9 2023-07-17 [S] [expires: 2025-07-16]
```

I will publish the public key to a keyserver once I've performed a test release on my GH test account, and I will share the passphrases and private keys of the primary-key and sub-key with the other maintainers, so we won't lose them again. I will do this as soon as I get the time in the next couple of days.

----

Similar to the storage of the old signing key, the new one is encrypted symmetrically via AES256. The decrypted signing key itself doesn't have a passphrase, because initially its passphrase was the same as the primary-key (which we certainly don't want) and setting a different one is problematic too, because gpg uses a weaker encryption for key storage (which can't be changed for some reason) and there's no point using that or having two encryption layers.

The build-and-sign script has been updated and it now writes the decrypted signing key file to the system's tmpfs instead of storing it in the git repo, like the old script did. I don't think that was a good idea, because the decrypted signing key could be read on a developer's system by anyone else at any time after running the signing script locally once.

Since GitHub apparently doesn't like showing diffs when a binary file gets turned into a text file (it's now stored as base64 data), I renamed the encrypted signing key file from `signing.key.gpg` to `signing.key.enc`.

I also decided to use a different secret env var on GitHub actions, so the old one can be kept, just in case.
New: `SIGNING_KEY_PASSPHRASE`
Old: `RELEASE_KEY_PASSPHRASE`

----

Local tests worked fine, btw:

```
$ SIGNING_KEY_ID=1AEB6400EDA27DA9 SIGNING_KEY_PASSPHRASE='...' ./script/build-and-sign.sh
...

$ gpg --verify ./dist/streamlink-5.5.1+114.g86698437.tar.gz.asc 
gpg: assuming signed data in './dist/streamlink-5.5.1+114.g86698437.tar.gz'
gpg: Signature made 2023-07-18T00:12:18 CEST
gpg:                using EDDSA key 89A4EFA5653B899E661179991AEB6400EDA27DA9
gpg: Good signature from "Streamlink signing key <streamlink@protonmail.com>" [ultimate]
```

----

### TODO

- [x] Test release
- [x] Add `SIGNING_KEY_PASSPHRASE`
- [x] Publish public key to keyserver(s)
- [x] Share private key data


bastimeyer:
Should be good to go now.

1. test release didn't show any issues
2. new secret env is set
3. public key has been uploaded to keyserver.ubuntu.com and keys.openpgp.org
   - https://keyserver.ubuntu.com/pks/lookup?search=44448a298d5c3618&fingerprint=on&op=index
   - https://keys.openpgp.org/search?q=44448a298d5c3618
4. private keys shared

beardypig:
chagrin emoji

PR #5455, opened 2023-07-18
release: 6.0.0

### LINKED ISSUE #5401: release: 6.0.0 tracking issue

bastimeyer:
Closes #5401 

https://deploy-preview-5455--streamlink.netlify.app/changelog.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,38 @@
 # Changelog
 
+## streamlink 6.0.0 (2023-07-20)
+
+Breaking changes:
+
+- BREAKING: dropped support for Python 3.7 ([#5302](https://github.com/streamlink/streamlink/pull/5302))
+- BREAKING: [turned `--player` CLI argument into a player-path-only argument](https://streamlink.github.io/migrations.html#player-path-only-player-cli-argument) ([#5305](https://github.com/streamlink/streamlink/issues/5305), [#5310](https://github.com/streamlink/streamlink/pull/5310))  
+  Its value won't be interpreted as a command line string anymore, so paths with whitespace don't require additional quotation. Custom player arguments now always need to be set via `--player-args`.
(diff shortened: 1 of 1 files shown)


gravyboat (approved):

bastimeyer:
Going to merge this tomorrow. This will give me a bit more time thinking about stuff's that potentially missing.

PR #5457, opened 2023-07-19
docs: add client-integrity to Twitch plugin docs

bastimeyer:


gravyboat (approved):
👍 Explanation is thorough and clear.

Issue #5047, opened 2022-12-27
plugins.twitch: (some?) low-level socket/SSL errors are unexpectedly treated as NoStreamsError


PleasantMachine9:
### Checklist

- [X] This is a plugin issue and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

While trying to use the SSLKEYLOGFILE feature of python's SSL module, I set the value of this env variable to an invalid path string (I added quotes around the path which are not expected by the Windows API) which gets rejected by openssl/kernel layer. 

This raises a specific kind of error (obtained by running the process in Pdb):
```
Traceback (most recent call last):
  File "C:\Users\Admin\AppData\Local\Programs\Python\Python310\lib\site-packages\urllib3\connectionpool.py", line 701, in urlopen
    httplib_response = self._make_request(
  File "C:\Users\Admin\AppData\Local\Programs\Python\Python310\lib\site-packages\urllib3\connectionpool.py", line 384, in _make_request
    self._validate_conn(conn)
  File "C:\Users\Admin\AppData\Local\Programs\Python\Python310\lib\site-packages\urllib3\connectionpool.py", line 1013, in _validate_conn
    conn.connect()
  File "C:\Users\Admin\AppData\Local\Programs\Python\Python310\lib\site-packages\urllib3\connection.py", line 402, in connect
    self.ssl_context = create_urllib3_context(
  File "C:\Users\Admin\AppData\Local\Programs\Python\Python310\lib\site-packages\urllib3\util\ssl_.py", line 350, in create_urllib3_context
    context.keylog_filename = sslkeylogfile
OSError: [Errno 22] Invalid argument: '"C:\\Users\\Admin\\SSLKEYLOGFILE.txt"'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Admin\AppData\Local\Programs\Python\Python310\lib\site-packages\requests\adapters.py", line 439, in send
    resp = conn.urlopen(
[...]

PR #5472, opened 2023-08-01
tools: bump ruff to 0.0.282 and fix issues

bastimeyer:


src/streamlink/plugin/api/http_session.py
```diff
@@ -187,7 +187,7 @@ def request(self, method, url, *args, **kwargs):
                 if retries >= total_retries:
                     err = exception(f"Unable to open URL: {url} ({rerr})")
                     err.err = rerr
-                    raise err
+                    raise err from rerr
```
bastimeyer, line 190: Technically, this is the correct thing to do here, but the last time I had at look at this, I decided to ignore adding the reference of the original/parent exception, because this can change how the stack gets printed, and changing that behavior should ideally be done in a dedicated PR.

The line above was added during the early Livestreamer days when the `HTTPSession` subclass was added in 936e66dd90f67377451be27951a7d5553a09088c. I'm not sure, but I believe `raise from` was already a thing in Python 2.x, so this shouldn't have been added.

The whole implementation of the `HTTPSession.request()` method is old and pretty bad, so I really don't want to touch any of this.

Maybe I should `raise from None` here, to not introduce any changes.

bastimeyer: Related to #5047, btw

bastimeyer: Changed to `raise from None`. Don't want to introduce any changes.

gravyboat (approved):

PR #5489, opened 2023-08-07
chore: update GitHub issue templates

bastimeyer:
1. This makes the header text more clear
2. A specific version string is now required for plugin issues and bug reports instead of the dropdown, so people are asked to actually check their used version
3. Random cleanups and fixes

Checked my test repo for rendering issues. Everything looked fine...


gravyboat:
👏 Good changes, if this doesn't result in better issues at least it sets the standard we'll just close them.

Issue #5498, opened 2023-08-15
Refactor segmented streams and use a common BaseSegment


bastimeyer:
While talking about the extensibility and lack of full typing information of the segmented stream implementations in the past, I already mentioned that the base implementations of segmented streams, as well as HLS and DASH are in need of a huge code refactoring.

Now with my attempt at turning the `--hls-duration` and `--hls-start-offset` options into the generic `--stream-segmented-duration` and `--stream-segmented-start-offset` options, it became clear to me that it didn't make sense finishing that work with the current code.

See #5450 and https://github.com/streamlink/streamlink/pull/5493#issuecomment-1677680058.

----

The issue I'm intending to solve, ignoring all other flaws for now (there are a lot), is making both HLS, DASH and other subclasses of the segmented stream base classes share more common logic in regards to how segments are handled. I also want to fix the lack of typing information, so working with and understanding the segmented stream implementations is much easier.

Currently, each segmented stream implementation does its own thing with its segments, so implementing common logic for calculating a max stream duration or a stream start offset for example is impossible without writing lots of duplicate code with only small variations. This is also the reason why adding proper typing information can't be done. Writing logic for the same thing in each segmented stream type just doesn't make sense.

----

The big outlier here is HLS, where segments are wrapped in a `Sequence` named tuple. This is due to historic reasons when the segmented streams were initially implemented. Segments of each stream type were defined as named tuples which don't support subclassing or changing values after instantiation, so in order for the HLS stream's M3U8 parser to work properly, the generated segments needed to be wrapped, so a sequence number could be added after parsing the playlist as a whole (`EXT-X-MEDIA-SEQUENCE`).

While refactoring/rewriting the DASH implementation earlier this year, I already replaced its segments with ones that are based on the stdlib's `dataclasses`. Those fix all the issues that come with named tuples, but `dataclasses` come at the cost of a slight performance hit compared to named tuples. That is negligible though and is not important. It doesn't matter for Streamlink.

With dataclass segments, a `BaseSegment` type can be created which HLS, DASH and other can inherit from. The base `SegmentedStream{Worker,Writer,Reader}` classes can then properly handle the `BaseSegment` objects, and common base logic for max stream durations and start offsets could be added. Then fixing the lack of typing information would be much easier, too.

----

[...]

PR #5507, opened 2023-08-19
stream.segmented: fix and add more typing

bastimeyer:
- Turn `SegmentedStream{Writer,Worker,Reader}` into generic classes with segment and segment-result types
- Fix typing information of the writer's queue
- Add missing typing information
- Update segmented stream classes accordingly

----

#5498

Next step will be fixing the whole `Segment` mess by adding a common `BaseSegment`. This will require more changes due to how the current segments are implemented and structured.


bastimeyer:
Will fix the py38 errors tomorrow...

PR #5513, opened 2023-08-22
stream.hls: refactor tag parser mapping

bastimeyer:
- Declaratively define tag parsers explicitly using a decorator instead of relying on the method name format
- Build the tag-parser mapping when creating the parser class, and don't build it on the parser class when first instantiating it
- Update parser subclasses (Twitch)
- Update tests


bastimeyer:
???
Why did the Windows py311 CI runner fail after merging this?
https://github.com/streamlink/streamlink/actions/runs/5944995273/job/16123282553
This seems very very random and probably unrelated, but we haven't had any flaky tests in a long while now, so seeing this confuses me a lot.

I've been running this specific test in a loop for the past 10 minutes now on my Linux system, without any issues. Running all HLS tests also works fine. I don't even understand how these changes here could possibly affect this test.

Going to check my Windows VM now.

PR #5521, opened 2023-08-30
stream.hls: refactor M3U8Parser state attributes

bastimeyer:
Store the parser state in dedicated attributes with proper typing information instead of using a state hash map with arbitrary keys.

----

This was never changed since the initial implementation in 39008e9, probably because of the convenience of being able to read and unset certain data in a single call via `pop()`. That however doesn't justify the CPU and memory disadvantanges that come with this design choice, so let's finally fix it.


gravyboat (approved):
🥳

PR #5526, opened 2023-09-01
stream.hls: remove Sequence named tuple wrapper

bastimeyer:
Ref #5498 

1. **Turn `Segment` and `Playlist` into dataclasses, so they can be subclassed**
    - Make `TwitchSegment` inherit from `Segment`
    - Update `TwitchM3U8Parser` accordingly
    - Fix `ltv_lsm_lv` plugin (previously cloned the `Segment` named tuple, now simply updates the `uri` attribute)

2. **Remove `Sequence` named tuple wrapper, so `HLSStream{Writer,Worker}` operate on the `Segment` directly**
    - Add `num` attribute to `Segment` and initialize with `-1`
    - Move sequence number logic from `HLSStreamWorker` to `M3U8Parser`
    - Update various `Sequence` references in `HLSStream{Writer,Worker}`
    - Update generic typevar of `HLSStream{Writer,Worker,Reader}`
    - Update HLS subclasses accordingly in various plugins

    I was thinking about adding a `segment` compatibility property with a deprecation warning to `Segment`, so third-party plugin implementations which subclass `HLSStream{Writer,Worker}` don't break when they still implement logic for the `Sequence` wrapper (where the actual segment was accessed via the `segment` attribute), but I decided against it, because (3) changes the names and signatures of some methods anyway. As said in #5498, these are not public interfaces, so I don't care.

3. **Update class attributes and method names/signatures of `HLSStream{Writer,Worker}` after the removal of `Sequence`**
    - Change names from sequence(s) to segment(s)
    - Remove `sequences` arg from `HLSStreamWorker.process_segments()` as it's not needed anymore after the sequence number logic was moved to the parser
    - Keep "Sequence" names and log messages for everything that's about the order of segments in the playlist or the position in the stream
    - Update HLS subclasses accordingly in various plugins

4. **Refactor M3U8 playlist loading and update/fix typing**
    - Rename `stream.hls_playlist.load` to `stream.hls_playlist.parse_m3u8`
    - Replace `HLSStreamWorker._reload_playlist()` and `HLSStream._get_variant_playlist()` with direct `parse_m3u8()` calls, using dependency injection and the `HLSStream.__parser__` reference
    - Turn `M3U8` into a generic class and add covariant type-vars for the `Segment` and `Playlist` dataclasses
    - Turn `M3U8Parser` into a generic class and add covariant type-vars for the `M3U8` class, as well as `Segment` and `Playlist`
    - Update HLS Twitch subclasses accordingly

    There are still some typing issues left, but I didn't manage to fix it (yet). I believe there are currently some restrictions ([PEP 696 - default values for `TypeVar`s](https://peps.python.org/pep-0696/)), so unless the whole design of the parser and stuff gets rewritten, this can't be improved further. But I'm not 100% sure. Should be good enough though.

----

I was also thinking about finally moving the HLS and DASH stuff into their own sub-packages, so we don't have multiple modules lying around in the `stream` package. This would even make more sense with my plans for the `BaseSegment` stuff, so the HLS and DASH parsers don't have to import the segmented stream implementations, as the `BaseSegment` definition would have to go there otherwise. Those changes don't belong in this PR though.

----

Btw, I'm not expecting any reviews here, because it requires knowledge of the details of the HLS classes. I will merge this when I feel ready.

### DIFF
--- a/src/streamlink/plugins/abematv.py
+++ b/src/streamlink/plugins/abematv.py
@@ -31,7 +31,7 @@ log = logging.getLogger(__name__)
 
 class AbemaTVHLSStreamWriter(HLSStreamWriter):
     def should_filter_sequence(self, sequence):
-        return "/tsad/" in sequence.segment.uri or super().should_filter_sequence(sequence)
+        return "/tsad/" in sequence.uri or super().should_filter_sequence(sequence)
 
 
 class AbemaTVHLSStreamReader(HLSStreamReader):
--- a/src/streamlink/plugins/abematv.py
+++ b/src/streamlink/plugins/abematv.py
@@ -30,8 +30,8 @@ log = logging.getLogger(__name__)
 
 
 class AbemaTVHLSStreamWriter(HLSStreamWriter):
-    def should_filter_sequence(self, sequence):
-        return "/tsad/" in sequence.uri or super().should_filter_sequence(sequence)
+    def should_filter_segment(self, segment):
+        return "/tsad/" in segment.uri or super().should_filter_segment(segment)
 
 
 class AbemaTVHLSStreamReader(HLSStreamReader):
--- a/src/streamlink/plugins/afreeca.py
+++ b/src/streamlink/plugins/afreeca.py
@@ -16,8 +16,8 @@ log = logging.getLogger(__name__)
 
 
 class AfreecaHLSStreamWriter(HLSStreamWriter):
-    def should_filter_sequence(self, sequence):
-        return "preloading" in sequence.uri or super().should_filter_sequence(sequence)
+    def should_filter_segment(self, segment):
+        return "preloading" in segment.uri or super().should_filter_segment(segment)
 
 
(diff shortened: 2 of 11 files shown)


bastimeyer:
Rebased to `master` and fixed merge conflicts. Going to merge now, so I can finally open the follow-up PRs.

As explained, some HLS **internals** got renamed. Here's a quick summary of the relevant stuff which could potentially be accessed by 3rd party plugins:

- `hls_playlist.load()` -> `hls_playlist.parse_m3u8()`
- `hls.HLSStreamWriter.should_filter_sequence()` -> `hls.HLSStreamWriter.should_filter_segment()`
- `hls.HLSStreamWorker.playlist_sequences` -> `hls.HLSStreamWorker.playlist_segments`
- `hls.HLSStreamWorker.process_sequences()` -> `hls.HLSStreamWorker.process_segments()`
- `hls.HLSStreamWorker.valid_sequence()` -> `hls.HLSStreamWorker.valid_segment()`

PR #5535, opened 2023-09-04
cli: add --player-env argument

### LINKED ISSUE #5530: Allow setting environment variables when launching the player
### Checklist

- [X] This is a feature request and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [X] [I have checked the list of open and recently closed feature requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22feature+request%22)

### Description

There is currently the `--player-arg`  that allows to pass argument to the player.

It would be very handy to have another one, `--player-env` to enable setting environment variables when the player is launched.

I currently have only a single use case:

- I use pipewire, and I want to hint pipewire to connect a specific node, this is done via the `PIPEWIRE_NODE` environment variable.

I can set the environment variable before launching streamlink, but this requires editing all my launchers, instead of setting it in the config file.

bastimeyer:
Resolves #5530

This adds the `--player-env KEY=VALUE` CLI argument, which allows adding environment variables to the spawned player process (on top of the inherited ones). The CLI argument can be repeated for being able to add multiple env vars. As mentioned in #5530, the intended purpose here is to be able to set env vars from the config files.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -428,6 +428,18 @@ def build_parser():
         """,
     )
     player.add_argument(
+        "--player-env",
+        metavar="KEY=VALUE",
+        type=keyvalue,
+        action="append",
+        help="""
+        Add an additional environment variable to the spawned player process, in addition to the ones inherited from
+        the Streamlink/Python parent process. This allows setting player environment variables in config files.
+
+        Can be repeated to add multiple environment variables.
+        """,
+    )
+    player.add_argument(
         "-v", "--verbose-player",
         action="store_true",
         help="""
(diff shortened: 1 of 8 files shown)


gravyboat (approved):
I don't see a benefit to adding this as the use cases are extremely slim and environment variables are easily managed via rc files. If an application is setting a dynamic value for something such as a port (which is already bad behavior on the application's part as typically a default port would be configured via the application's config file) then the application should be setting the environment variable appropriately when the service spins up and the random port is allocated if that port is needed for whatever reason.

That being said as there are no real standards around environment variable usage or expectations and there's little negative impact on Streamlink by adding this. I'll mark it as approved with the caveat that we offer as little support for this feature as possible.

Issue #4902, opened 2022-10-24
Overhaul of the stream selection


bastimeyer:
I've already talked about this multiple times over the past years in various issues and pull requests, so let me finally open a dedicated thread for this.

## Motivation

The way streams are returned by plugins, how the stream naming works, how the stream weighting/ranking works, how stream sorting works, how stream deduplication works, and how streams are presented to the user via the CLI, including the stream selection, are far from ideal. With the advent of the AV1 codec in the near future, this will cause lots of problems because of the reliance on hardware accelerated decoders on the user's systems, because stream weights are not related to the used codec or other stream properties in any way. Once streaming sites start publishing the same stream in different codecs, there won't be a simple way for Streamlink to adapt to this unless we make changes to the individual plugin/stream implementations and carry on with the bad design.

## Background

The return value of [`Plugin.streams()`](https://github.com/streamlink/streamlink/blob/5.0.1/src/streamlink/plugin/plugin.py#L358-L487) has always been a dictionary of stream names and `Stream` instances, determined by plugin implementations in `Plugin._get_streams()`. When this was [implemented in Livestreamer more than 10 years ago](28d0110d44a9a78233dfc829d5510a2c02c264d1), the project was still in its infancy and there was only a total of [**three** plugins](https://github.com/streamlink/streamlink/tree/28d0110d44a9a78233dfc829d5510a2c02c264d1/src/livestreamer/plugins), so there was never any concern about different stream naming/weighting schemes that depended on various different stream types and properties. The whole idea behind having simple stream names as dictionary keys was being able to provide a simple interface on the CLI for stream selections.

Over the time, more and more stream implementations and more and more plugins were added with lots of different features on the CLI side. The design of this however was never changed, even after the Streamlink fork, leading to the current situation. And because of how this design was carried on and nobody dared changing/fixing it, changing all this now has become a really difficult task, especially when having a clean and easy migration path in mind.

## The issue

Stream names only include a bare minimum of information about the stream and its properties/metadata. This leads to a very poor way of weighting streams, selecting streams and filtering out duplicates or unwanted streams.

While stream names are determined by plugin implementations, most plugins simply forward the result of `HLSStream.parse_variant_playlist()` or `DASHStream.parse_manifest()`. Lots of others however don't do that and return progressive `HTTPStream`s or other custom stream implementations with custom non-descriptive stream names.

[...]

PR #5588, opened 2023-10-05
chore: fix typo

michaelarnauts:
Fix a small typo I've discovered when reading the source: "QUALITY_WEIGTHS_EXTRA" -> "QUALITY_WEIGHTS_EXTRA"

### DIFF
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -45,7 +45,7 @@ BIT_RATE_WEIGHT_RATIO = 2.8
 
 ALT_WEIGHT_MOD = 0.01
 
-QUALITY_WEIGTHS_EXTRA = {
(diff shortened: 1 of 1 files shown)


bastimeyer:
This has intentionally not been changed for the past several years, because the entire stream weighting/ranking/sorting logic is flawed and needs to be rewritten from scratch in a completely different way (see #4902).
https://github.com/streamlink/streamlink/blame/6.2.1/src/streamlink/plugin/plugin.py#L48

The plugin module doesn't clearly define which constants are part of the public interface, apart from re-exports in the package's `__init__` module, which are just "convenience exports" rather than a definition of a public interface, so renaming stuff could always be considered a breaking change. The risk here however is rather low, since it's essentially dead code anyway, so let's finally fix the typo.

bastimeyer:
Thanks

PR #5593, opened 2023-10-07
stream: switch to sub-packages

bastimeyer:
#5498

This moves the various stream implementations into dedicated sub-packages, for having a cleaner separation of different stream types, as well as for being able to define the `BaseSegment` in the upcoming PR in a dedicated module. This PR therefore basically just moves the module files and updates the import paths (and mock paths in tests). I decided against naming the individual modules of those sub-packages with a leading underscore character.

As a side effect of moving the modules, and I'm not sure whether I like it, the logger names have changed as well (due to the `__name__` reference). Example: `stream.hls` -> `stream.hls.hls` and `stream.hls_playlist` -> `stream.hls.m3u8`.

I believe it'd be better to override/revert the redundant logger names, which are `stream.segmented.segmented`, `stream.hls.hls` and `stream.dash.dash`. The logger names of the parser modules (`hls.m3u8` and `dash.manifest`) are fine IMO and should be kept. They will be different no matter what due to the modules being renamed (previously `hls_parser` and `dash_manifest`).

In regards to the exports of the sub-packages, there will be no difference for the segmented stream base classes. The HLS sub-package also exports most of the m3u8 module, which I don't think is necessary, but I added those exports anyway since there are lots of custom HLS implementations. The DASH sub-package however only exports the relevant stuff. If there's a need for plugins to subclass DASH internals for custom implementations, then any missing stuff can be added later.


bastimeyer:
I've updated the logger names to reduce unnecessary redundancy.

gravyboat (approved):
Looks good. I agree with you in regards to avoiding the redundant loggers 👍

PR #5607, opened 2023-10-12
stream.hls: warn about skipped segments

### LINKED ISSUE #5603: stream.hls: Add warning when segments expire from playlist before being queued
### Checklist

- [X] This is a feature request and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [X] [I have checked the list of open and recently closed feature requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22feature+request%22)

### Description

Today my internet briefly went out while using streamlink to record an HLS stream, but came back up about a minute later. When it came back, it started downloading segments again, but I was surprised to see in the log that it simply skipped from segment 143 to 174 without any kind of log message indicating any issue. I had actually thought that streamlink output a warning when things like that occurred, but I realize now I was probably thinking of ffmpeg.
ffmpeg has a warning for when this situation occurs here:
https://github.com/FFmpeg/FFmpeg/blob/c06d3d24047206f9c11bfc5849544b960e5d68eb/libavformat/hls.c#L1518-L1523

It would be nice if streamlink could output a similar warning if it encounters non-consecutive sequence numbers. This wouldn't just be useful for internet issues but also potentially as an indicator that the `--hls-playlist-reload-time` needs to be set. I have occasionally encountered misbehaving HLS producers with wildly incorrect `EXT-X-TARGETDURATION` values causing segments to expire between playlist refreshes, but wouldn't necessarily have noticed any issue if I was just looking at the current log output.

bastimeyer:
Warn about skipped segments when reloading the playlist and queuing new segments with a sequence number offset that's greater than one.

----

Resolves #5603


bastimeyer:
I think it would probably make more sense warning about which segments were skipped and not just how many. Will change this later...

PR #5613, opened 2023-10-14
ci.github: update py312 runners

bastimeyer:
Apparently, the runner images have removed py312 again 4 days ago in
https://github.com/actions/runner-images/commit/929931f61727947eb60ba4ecae57fe82202728ed
so `actions/setup-python` will download the binaries every time, but that doesn't matter apart from slightly longer setup times


bastimeyer:
I will switch the branch protection rules from 3.11 to 3.12 once the runner images have added 3.12 again.

PR #5649, opened 2023-11-03
docs: move API guide pages into second layer

bastimeyer:
I'm in the process of finally adding API docs for the `streamlink.plugin.api.validate` module and also writing an API guide with a couple of (real-world) usage examples. This isn't done yet (it's way more than I expected), but these changes here are a requirement because I'll need to add another API guide page for the validate module, so the current one needs to be moved one level down, so we can have an TOC index in its current place. The current API guide is therefore now called "Quickstart".

I could've included these changes in a single PR, but I wanted to get them out already today, and it's probably better to do it this way anyway... No changes to the quickstart docs apart from a missing comma, one grammar error and different headline levels (so that the sections don't appear in the main menu). The guide could've been updated and refined, but I didn't feel like it. It's not my focus right now.


bastimeyer:
https://deploy-preview-5649--streamlink.netlify.app/api_guide

gravyboat (approved):

Issue #5543, opened 2023-09-10
ci: docs preview builds on netlify currently limited to sphinx 7.1.2


bastimeyer:
Netlify is still stuck on Python 3.8:

- https://docs.netlify.com/configure-builds/available-software-at-build-time/
- https://answers.netlify.com/t/python-3-11-or-newer/84566

Since Sphinx 7.2.0 has already dropped support for Python 3.8, this means that all docs preview builds on netlify are built with Sphinx 7.1.2 instead of the latest one.

- https://www.sphinx-doc.org/en/master/changes.html#release-7-2-0-released-aug-17-2023
- https://github.com/streamlink/streamlink/blob/4f65b9a97d80b7ba45e6175d4a09d4c707035238/docs-requirements.txt

The docs dependencies must not be bumped when bumping would drop support for Python 3.8, or we would break the preview builds.

bastimeyer:
`myst-parser >=4` has dropped support for py38 and py39. Netlify is still stuck on py38 and there's no upgrade in sight. They simply don't care about Python. Being stuck on `myst-parser >=1.0.0,<4` could introduce problems in the future though if they introduce breaking changes which affect our documentation sources. Then Streamlink packagers who build the docs and man pages can't use the latest `myst-parser` versions. I doubt that there will be any breaking changes which will affect us, but it's possible.
- https://github.com/streamlink/streamlink/blob/6.8.3/docs-requirements.txt#L10
- https://github.com/executablebooks/MyST-Parser/releases/tag/v4.0.0

Previews for GitHub-pages builds are still not supported, same as deploying to a different repo, so dropping Netlify in favor of GH actions for docs previews is currently impossible:
- https://github.com/actions/deploy-pages/issues/337
- https://github.com/actions/deploy-pages/issues/308

PR #5652, opened 2023-11-05
docs: add plugin.api.validate API documentation

bastimeyer:
This finally adds the API documentation for `plugin.api.validate`. An API guide with real-life examples is not included yet and will be added later.

This was much more work than expected. Especially because of several sphinx-autodoc issues which required me to reformat and move everything multiple times.

I've tried to make the style as consistent as possible, but there are some inconsistencies between the `validate` docs, schema docs and utility function docs. These are just minor differences though and you won't notice them. I won't even tell you.

As explained in the top comment of `validate.rst`, the layout was chosen deliberately because of certain autodoc quirks. Unfortunately, it's not ideal, but it's the best I could come up with.

----

My first attempt was documenting all overloading function of `validate()` (base schema types, as well as custom schemas), but then I noticed that autodoc doesn't support `singledispatch` properly.

Manually importing the functions using `autofunction` also was a bad idea, because this would require duplicate docs, on both the schema validation functions and the schema classes themselves.

So I ended up with the current appoach of adding custom function docs in the `validate.rst` file for the basic schema validation functions like `type`, `abc.Callable`, `list` (etc), `dict` and `re.Pattern`, and importing the rest with proper docstrings in the python modules.

### DIFF
--- a/docs/api/validate.rst
+++ b/docs/api/validate.rst
@@ -239,9 +239,11 @@ Validation schemas
 .. automodule:: streamlink.plugin.api.validate
     :imported-members:
     :exclude-members: Schema, SchemaContainer, validate
+    :member-order: bysource
 
 .. automodule:: streamlink.plugin.api.validate._schemas
     :exclude-members: SchemaContainer
+    :member-order: bysource
     :no-show-inheritance:
 
 .. autoexception:: streamlink.plugin.api.validate._exception.ValidationError
--- a/docs/api/validate.rst
+++ b/docs/api/validate.rst
@@ -1,0 +1,247 @@
+Validation schemas
+------------------
+
+.. ..
+    Sphinx's autodoc doesn't properly document imported module members and it just outputs "alias of" for re-exported classes.
+    This means we'll have to run `automodule` twice if we want to document the original classes:
+    1. the main public interface (which contains aliases, like `all` for `AllSchema` for example)
+    2. the original schema classes with their full signatures and docstrings
+    .
+    Ignore unneeded classes like `SchemaContainer` which are not useful for the API docs.
+    .
+    Ignore `validate` as well, as `functools.singledispatch` functions are not fully supported by autodoc.
+    Instead, manually document `validate` and its overloading functions for base schema types here at the top,
+    just below the manually imported `Schema` (the main validation schema interface).
+    The documentations for any custom schemas like `AllSchema` for example is done on the schemas themselves.
+    .
+    Ideally, we'd just run autodoc on the main module and configure the order of items. :(
+
+.. autoclass:: streamlink.plugin.api.validate.Schema
+    :members:
+    :undoc-members:
+
+.. py:function:: validate(schema, value)
+    :module: streamlink.plugin.api.validate
+
+    The core of the :mod:`streamlink.plugin.api.validate` module.
+
+    It validates the given input ``value`` and returns a value according to the specific validation rules of the ``schema``.
+
+    If the validation fails, a :exc:`ValidationError <_exception.ValidationError>` is raised with a detailed error message.
+
+    The ``schema`` can be any kind of object. Depending on the ``schema``, different validation rules apply.
+
+    Simple schema objects like ``"abc"`` or ``123`` for example test the equality of ``value`` and ``schema``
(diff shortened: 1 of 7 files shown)


bastimeyer:
I haven't done a final proof-read yet, so there are probably a couple of issues left.

----

One typing issue because of a tiny change and also a netlify build issue because of Sphinx 7.2.0 and `:no-index:` :rage: #5543
https://www.sphinx-doc.org/en/master/changes.html#release-7-2-0-released-aug-17-2023

```
5:06:14 PM: Failed during stage "building site": Build script returned non-zero exit code: 2
5:06:14 PM: reading sources... [100%] thirdparty
5:06:14 PM: Warning, treated as error:
5:06:14 PM: /opt/build/repo/docs/api/validate.rst:18:Error in "module" directive:
5:06:14 PM: unknown option: "no-index".
5:06:14 PM: .. module:: streamlink.plugin.api.validate
5:06:14 PM:     :no-index:
5:06:14 PM: make: *** [Makefile:46: html] Error 2
5:06:14 PM: make: Leaving directory "/opt/build/repo/docs"
```

bastimeyer:
https://deploy-preview-5652--streamlink.netlify.app/api/validate

docs/api/validate.rst
```diff
@@ -0,0 +1,246 @@
...
+
+.. py:function:: _validate_sequence(schema, value)
+    :module: streamlink.plugin.api.validate
+
+    :class:`list <builtins.list>`, :class:`tuple`, :class:`set` and :class:`frozenset` validation.
+
+    Each item of ``value`` gets validated against **any** of the items of ``schema``.
+
+    Please note the difference between :class:`list <builtins.list>`
+    and the :class:`ListSchema <_schemas.ListSchema>` validation.
```
gravyboat, line 138: Potential slight chance of confusion here as people may think you are referring to `streamlink.plugin.api.validate.list` which itself is an alias of `ListSchema` if they aren't familiar with the doc formatting. Not a big deal but it's easier to understand the difference for something like https://github.com/streamlink/streamlink/pull/5652/files#diff-010b6199b0fc1ebe5df5bdb5ee86c9d024fea4dca02b63f388fe11b1d5156599R216 because of the reference to `re.pattern`.

I wish there was a way `<builtins.list>` could easily link to the [python docs](https://docs.python.org/3/library/functions.html#func-list).

gravyboat (approved):
Took a bit but I looked at everything. I don't see any typos or grammar issues. I've noted one small point about the doc formatting for `builtins` but it's not a show stopper nor does it need to be considered here. Looks really good!

docs/api/validate.rst
bastimeyer: > people may think you are referring to `streamlink.plugin.api.validate.list`

Maybe, but if they check `validate.list`, they'll see that it's an alias for `ListSchema`, which means that this line compares the same things, which doesn't make much sense and it must be two different things. All members of the `validate` module are always referenced with their module name, never on their own, apart from schema class names.

> I wish there was a way `<builtins.list>` could easily link to the [python docs](https://docs.python.org/3/library/functions.html#func-list).

It's currently disabled. For a good reason, because it's way too noisy with external links everywhere...
https://github.com/streamlink/streamlink/blob/6.3.1/docs/conf.py#L108-L111

bastimeyer:
I'll fix the order of items later today or tomorrow and then merge. Any follow-up issues can be fixed afterwards.

docs/api/validate.rst
gravyboat: Works for me.

bastimeyer:
Just to document this:

The reason why autodoc doesn't fully document the schema classes in the main module and instead outputs "alias of" is that the classes' `__name__` attribute is different from the names of the `streamlink.plugin.api.validate` module members (`"AllSchema" != "all"`).

This could be fixed by overriding the `__name__` attribute of each class like this in `__init__.py`:

```py
def _(items):
    for name, obj in items:
        if type(obj) is type and obj.__module__.startswith(f"{__name__}."):
            obj.__name__ = name


_(globals().items())
del _
```

But this would change the validation error message and it would require fixing the tests. I'll have a look at this in the future, because it also requires some changes in the docstrings in regards to the type annotations (the list issue mentioned above now being an actual issue).

PR #5653, opened 2023-11-07
plugins.telemadrid: fix xpath

pablohoffman:
Fix broken XPath in Telemadrid plugin


bastimeyer (approved):
Thanks. Works... :+1:

Issue #5129, opened 2023-01-29
ci: switch to token-based PyPI upload (again)


bastimeyer:
Opening this as a reminder for myself, to (once again) change the PyPI upload from a password-based one to a token-based one.

- #5081

Release 5.2.0 this week unfortunately had a failed attempt due to the changed 2FA requirement on the streamlink PyPI package, even though a token with full access was generated for the streamlink PyPI account.

- https://github.com/streamlink/streamlink/pull/5111#issuecomment-1401057624
- https://github.com/streamlink/streamlink/actions/runs/3990883680/jobs/6845136470#step:10:27

----

- https://pypi.org/help/#apitoken
- https://pypi.org/help/#twofa

> Users who have chosen to set up two factor authentication will be asked to provide their second method of identity verification during the log in process. This only affects logging in via a web browser, and not (yet) package uploads.

:confused:
No idea what they are thinking... NPM does a way better job here, with "automation tokens" purely for CI package updates.

When I get the time, I will take a look at the PyPI test index, so another failed release can be avoided.

bastimeyer:
2FA will be mandatory beginning in January. No idea if this will affect user+pass publishing via twine, but it probably will. The token-based upload already failed while I had 2FA enabled on the streamlink PyPI account.

I'll have a look at switching to "trusted publishing" via
- https://github.com/pypa/gh-action-pypi-publish
- https://docs.pypi.org/trusted-publishers/
- https://docs.pypi.org/trusted-publishers/adding-a-publisher/

This will probably require a bit of refactoring of the CI workflows though.

PR #5668, opened 2023-11-15
ci.github: rewrite workflows

bastimeyer:
- Split up workflows from `main` into `test`, `docs` and `release`
- Update `lint` workflow
- Remove `install-dependencies.sh` script
- Remove script dependencies from `dev-requirements.txt`

----

A bit of CI workflow cleanup, which properly separates the docs+release jobs from pushes to master and PRs. It is also kind of a requirement for the switch towards password-less/token-less PyPI uploads: #5129 

New workflows are tested here on my test-repo:

- https://github.com/bastimeyer-test/streamlink-test/actions
- https://github.com/bastimeyer-test/streamlink-test/pull/13
- https://github.com/bastimeyer-test/streamlink-test/releases/tag/999.0.0

Don't get confused about the test-PR. It lists everything for the same commit ID, push events and pull_request events, so there's a bit of duplication here. It also contains two more commits, one which removes repo name-checks of some CI jobs and which fakes successful deployments, and a second one with a fake 999.0.0 release.

----

- `test`
  Runs on each push (excluding `automated/**` branches), as well as on each pull request and on a scheduled timer, similar to before. It runs pytest (coverage data is disabled on scheduled runs) and the docs tests (only enabled on pull requests and scheduled runs).
- `docs`
  Builds and deploys the docs. It is run on each push to master or when pushing tags.
- `release`
  Only runs when pushing tags. It is now split up into two separate jobs: build, and then deploy, with build-artifacts shared between those jobs.

----

Notes:

- ~~The branch protection rules will need to be updated because of these changes, which will require all open PRs to be rebased.~~
  Apparently, GitHub only checks the job name and not the name of its workflow?! Current protection rules somehow apply to the new `test` workflow...

- The `release` workflow now immediately starts with building and deploying. There is no test step anymore because the test steps are only run on the `test` workflow. This is fine though, because we push tags after pushing branches. And since we need to wait for the `docs` workflow to complete anyway, so that there's no merge-conflict when building dev and prod/release docs in parallel, this is okay.

- Once merged, I will have a look at publishing a test release to PyPI-test using the [`pypa/gh-action-pypi-publish`](https://github.com/pypa/gh-action-pypi-publish) action.
  https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-pypi
  
  PyPI's docs optionally recommend setting up deloyment environments, but we don't need this here. It would only add another approval requirement before those CI jobs can be run, which is unnecessary in our case. The build and deploy jobs are already split up now, so the build job doesn't have access to the id-token permission, according to [`pypa/gh-action-pypi-publish`](https://github.com/pypa/gh-action-pypi-publish). Adding a deployment environment would also mean splitting up the deploy job of the docs. I don't think it's worth doing that.


bastimeyer:
Rebased to master, with the coverage data fix #5669 

Old commit id for comparison with my test-PR: [ece947d](https://github.com/streamlink/streamlink/commit/ece947da8c6736ec629a75ad00f16b295343e6d9)

bastimeyer:
Ooops... I accidentally set all test runners to 3.11 after rewriting the workflows... No wonder why there were indirect coverage changes.

gravyboat (approved):
Really nice clean up!

PR #5677, opened 2023-11-20
Add colors to HTML docs for player compatibility table

rpdelaney:
Changes proposed in this pull request:

- Add custom css for compatibility colors
- Add colors to compatibility table

Looks like this on my system (LibreWolf / Firefox 118.0.2-1):

PR #5678, opened 2023-11-20
docs: clean up players page

bastimeyer:
- Remove unmaintained/obsolete and proprietary players
  - Daum PotPlayer
  - MPC-HC
  - OMXPlayer
  - QuickTime
- Add new players
  - IINA
- Change player order
- Add OS and license columns
- Add colored icons for supported transport methods
- Fix footnotes and known issues section

----

Closes #5677
This here uses color-coded icons from the `sphinx-design` dependency instead of "yes" and "no"...

----

Dead players like MPC-HC and OMXPlayer should be removed. And while we're at it, let's also remove proprietary players. There isn't even much info about PotPlayer available, and Quicktime doesn't work anyway...

No idea if my data about IINA is correct. The `--stdin` requirement definitely is, I even checked it just now in my macOS VM. I couldn't manage to get the other methods working though, so I assume it's correct.

### DIFF
--- a/docs/players.rst
+++ b/docs/players.rst
@@ -8,15 +8,26 @@ There are three different modes of transporting the stream to the player.
 
 .. list-table::
     :header-rows: 1
+    :class: sd-w-100
 
     * - Name
       - Description
     * - Standard input pipe
-      - This is the default behaviour when there are no other options specified.
+      - This is the default behavior when there are no other options specified.
     * - Named pipe (FIFO)
-      - Use the :option:`--player-fifo` option to enable.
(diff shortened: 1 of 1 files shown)


bastimeyer:
I also had a look at mplayer, and neither stdin, nor fifo or http worked. What's going on here? It only let me play back videos from the filesystem.

Also had a look at other players, but none seemed worth adding.

bastimeyer:
https://deploy-preview-5678--streamlink.netlify.app/players

gravyboat (approved):

PR #5679, opened 2023-11-20
docs: remove common issues page

bastimeyer:
Follow-up of #5678 

If there is a need for player-specific arguments, then those can be added to the players page in the future. The common issues page in its current state is useless, wrong and outdated.


gravyboat (approved):
Agreed.

Issue #4310, opened 2022-01-25
stream.hls: about the output of media initialization section


thisisgk:
### Checklist

- [X] This is a bug report and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

For HLS playlist which has EXT-X-MAP tag (media initialization section), streamlink currently prepends the media initialization section to each fMP4 segment and make an output.
From the explanation of PR #3828, this seems to be done intentionally, but it results in the output stream/file containing multiple 'ftyp' boxes and 'moov' boxes, which should not be correct for ISOBMFF.
It seems that the media initialization section should only be output at the beginning of the stream/file.

The output file can be handled by ffmpeg, and seems to be played in mpv and vlc, though. However, ffmpeg and mpv show the warning "Found duplicated MOOV Atom. Skipped it", and vlc shows "mp4 demux: Fragment sequence discontinuity detected 1 != 0".

I generated and tested the following HLS playlist and DASH manifest using the same media files, so as to make it easy to compare the streamlink outputs.
```
$ mkdir /tmp/test
$ ffmpeg -f lavfi -i testsrc \
-use_template 0 \
-init_seg_name 'init$RepresentationID$.$ext$' \
-media_seg_name 'seg$RepresentationID$-$Number%02d$.$ext$' \
-dash_segment_type mp4 \
-hls_playlist 1 -hls_master_name master.m3u8 \
-f dash -t 30 /tmp/test/manifest.mpd

$ ls -1 /tmp/test
init0.m4s
manifest.mpd
master.m3u8
media_0.m3u8
seg0-01.m4s
seg0-02.m4s
seg0-03.m4s
```
Then ran streamlink.
```
$ streamlink hls://file:///tmp/test/master.m3u8 best -o output-hls.mp4
$ streamlink dash://file:///tmp/test/manifest.mpd best -o output-dash.mp4
```
output-hls.mp4 with the HLS playlist matches `cat init0.m4s seg0-01.m4s init0.m4s seg0-02.m4s init0.m4s seg0-03.m4s`.
[...]

PR #5689, opened 2023-11-26
stream.hls: fix segment maps

bastimeyer:
Only write segment maps to the output once

----

Fixes: #4310

### DIFF
--- a/src/streamlink/stream/hls/hls.py
+++ b/src/streamlink/stream/hls/hls.py
@@ -155,17 +155,18 @@ class HLSStreamWriter(SegmentedStreamWriter[HLSSegment, Response]):
             self.queue(None, None)
             return
 
-        # always queue the segment's map first if it exists
+        # queue segment-map first
         if segment.map is not None:
-            cached_map_future = self.map_cache.get(segment.map.uri)
-            # use cached map request if not a stream discontinuity
-            # don't fetch multiple times when map request of previous segment is still pending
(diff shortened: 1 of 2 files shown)


bastimeyer:
Copying this from my post of the 6.4.2 release pull request:

----

Old recordings with the faulty output which FFmpeg 6.1 now rejects need to be remuxed using an older FFmpeg version.

```sh
#!/usr/bin/env bash
for file in $(find . -type f -name "*.mp4" -print); do
  ffprobe -v quiet "${file}" && continue  # ffprobe >=6.1 returns status code >0 for those faulty mp4 containers
  echo "Remuxing ${file}"
  /path/to/older/ffmpeg -v error -i "${file}" -c copy -f mp4 "${file}.new" \
    && touch -c -r "${file}" "${file}.new" \
    && rm "${file}" \
    && mv "${file}.new" "${file}"
done
```

Where you can get an older FFmpeg build depends on your OS, but here are some relevant links:

- Windows (`x86_64`) and Linux (`x86_64` / `aarch64`)
  https://github.com/BtbN/FFmpeg-Builds/releases (rolling releases with different FFmpeg versions)
- Windows (`x86` / `x86_64`)
  https://github.com/streamlink/FFmpeg-Builds/releases/tag/20230930-1 (FFmpeg 6.0)
- macOS
  `brew install ffmpeg@6`

PR #5701, opened 2023-12-04
plugins.btv: fix 

SMNn1:
Fix to plugin btv.tv which was broken though changes on stream provider.
Plugin is now parsing the hls URL and starting it directly, instead of only throwing error.

### DIFF
--- a/src/streamlink/plugins/btv.py
+++ b/src/streamlink/plugins/btv.py
@@ -11,7 +11,6 @@ import re
 from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
 from streamlink.stream.hls import HLSStream
-from streamlink.plugin.plugin import LOW_PRIORITY, parse_params
 from streamlink.utils.url import update_scheme
 
 
--- a/src/streamlink/plugins/btv.py
+++ b/src/streamlink/plugins/btv.py
@@ -11,6 +11,8 @@ import re
 from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
 from streamlink.stream.hls import HLSStream
+from streamlink.plugin.plugin import LOW_PRIORITY, parse_params
+from streamlink.utils.url import update_scheme
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


src/streamlink/plugins/btv.py
```diff
@@ -60,8 +61,10 @@ def _get_streams(self):
         if stream_url == "geo_blocked_stream":
             log.error("The content is not available in your region")
             return
-
-        return HLSStream.parse_variant_playlist(self.session, stream_url)
+        
+        hls_url = update_scheme("hls://https://", stream_url, force=False)
```
bastimeyer, line 65: `update_scheme()` doesn't make any sense here and needs to be removed:

1. `hls://` is the explicit scheme of Streamlink's HLS **plugin**, so that direct HLS playlist URL inputs can be handled whose path don't end with `.m3u8`.
   See https://github.com/streamlink/streamlink/blob/6.4.2/src/streamlink/plugins/hls.py#L13-L15
2. `update_scheme()` does only apply the scheme to the stream URL if it's missing, which is not the case here, hence the reason why you didn't notice any issues with `hls://`. If you'd actually add `hls://` to the stream URL, then `requests` couldn't find a matching adapter for this URL.

src/streamlink/plugins/btv.py
```diff
@@ -60,8 +61,10 @@ def _get_streams(self):
...
+        streams = HLSStream.parse_variant_playlist(self.session, hls_url)
+        return streams or {"live": HLSStream(self.session, hls_url)}
```
bastimeyer, line 67: As you can see in #5698, the site started returning HLS multivariant playlists instead of HLS media playlists, so a fix was committed accordingly. The URL test cases returned valid streams (`./script/test-plugin-urls.py btv`), and they still do that now, I just checked it again.

Does your PR mean that this is not the case 100% of the time, and they return both kinds? You're saying that Streamlink is throwing errors. That shouldn't happen when trying to parse a multivariant playlist on a media playlist. The result should be an empty stream list.

bastimeyer:
Closing, since there hasn't been any response to my comments from two weeks ago.

The plugin is still working fine and I don't see any reason why an HLS media playlist fallback should be added when the site has switched to multivariant playlists

```
$ ./script/test-plugin-urls.py btv
:: Finding streams for URL: https://btvplus.bg/live
:: Found streams: 720p, 1080p, worst, best
:: Finding streams for URL: https://btvplus.bg/live/
:: Found streams: 720p, 1080p, worst, best
:: Finding streams for URL: https://www.btvplus.bg/live/
:: Found streams: 720p, 1080p, worst, best
```

Issue #5710, opened 2023-12-07
plugins.wasd: service gone


Mehavoid:
### Checklist

- [X] This is a [plugin issue](https://streamlink.github.io/plugins.html) and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

6.4.2

### Description

A few days ago, the service [gone](https://mts.ru/personal/novosti/2023-12-05/vstrechajte-polzovatelskuyu-videoplatformu-nuum). Now this [nuum.ru](https://nuum.ru).

Though we could easily replace the plugin, but I'm not sure it's worth adding it to upstream, because it's a beta version.
<details>

```diff
diff --git a/src/streamlink/plugins/wasd.py b/src/streamlink/plugins/wasd.py
index 7d61304e..656a16eb 100644
--- a/src/streamlink/plugins/wasd.py
+++ b/src/streamlink/plugins/wasd.py
@@ -16,7 +16,7 @@ log = logging.getLogger(__name__)
 
 
 @pluginmatcher(re.compile(
-    r"https?://(?:www\.)?wasd\.tv/(?P<nickname>[^/]+)/?$",
+    r"https?://(?:www\.)?nuum\.ru/channel/(?P<nickname>[^/]+)/?$",
 ))
 class WASD(Plugin):
     _media_schema = validate.Schema({
@@ -53,11 +53,11 @@ class WASD(Plugin):
 
     def _get_streams(self):
         nickname = self.match.group("nickname")
-        res = self.session.http.get(f"https://wasd.tv/api/channels/nicknames/{nickname}")
+        res = self.session.http.get(f"https://nuum.ru/api/channels/nicknames/{nickname}")
         channel_id = self.session.http.json(res, schema=self._api_nicknames_schema)
 
         res = self.session.http.get(
-            "https://wasd.tv/api/v2/media-containers",
+            "https://nuum.ru/api/v2/media-containers",
             params={
                 "media_container_status": "RUNNING",
                 "limit": "1",
```
</details>

PR #5712, opened 2023-12-07
plugins.wasd: remove plugin

Mehavoid:
Closes: #5710


bastimeyer:
Unlucky timing with both PRs being pushed at the same time... Thanks though :sweat_smile: 
btw, I would've preferred merging yours, since you opened the issue, but you've already closed this one and deleted the branch...

PR #5722, opened 2023-12-14
tests: remove unneeded pycountry warning filters

bastimeyer:
Fixed in `pycountry 23.12.7`:
https://github.com/pycountry/pycountry/blob/23.12.7/HISTORY.txt#L13

Filters were added in #5230, #5393

Setting the min-version requirement of pycountry is not necessary...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -126,8 +126,6 @@ testpaths = [
 ]
 filterwarnings = [
   "always",
-  "ignore:::pkg_resources",
-  "ignore::DeprecationWarning:pycountry",
 ]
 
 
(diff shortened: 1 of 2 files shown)


bastimeyer:
https://github.com/pycountry/pycountry/blob/23.12.11/HISTORY.txt#L7

This new `UserWarning` when a country's `official_name` is missing is unnecessary, so the second commit will now ignore it.

For example, "Canada" doesn't have an `official_name` attribute. Streamlink's `utils.l10n.Country` implementation however has always implemented the `name` fallback, so the warning is very much unnecessary.

PR #5779, opened 2024-01-16
plugin: add str to pluginargument type argument

bastimeyer:
Follow-up of #5715

This adds a `@pluginargument()` `type` function registry, so the `@pluginargument()` data can be fully JSON-serialized.

Built-in plugins now must always reference the `type` argument of each `@pluginargument()` (if set) via the registered function name. This is optional for third-party plugins.

### DIFF
--- a/docs/api/plugin.rst
+++ b/docs/api/plugin.rst
@@ -13,3 +13,5 @@ Plugin decorators
 .. autodecorator:: pluginmatcher
 
 .. autodecorator:: pluginargument
+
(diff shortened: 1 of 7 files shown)


bastimeyer:
This unfortunately can't be done better, but it's good enough:
https://deploy-preview-5779--streamlink.netlify.app/api/plugin#streamlink.plugin.plugin._PLUGINARGUMENT_TYPE_REGISTRY

PR #5793, opened 2024-01-21
build: add plugins JSON build script

bastimeyer:
Ref #4741
Part 1/x

----

This adds the plugins JSON build script + setuptools hook, and a CI test job for it.

The build script parses each plugin file and reads the pluginmatcher and pluginargument data by walking the module's abstract syntax tree. The result is the JSON-serialized pluginmatcher and pluginargument data of each plugin without having to install Streamlink and import any plugin modules. There are many cases though where parsing could fail, depending on the plugin implementation, for example because object references are not supported. That's why the pluginmatcher and pluginargument data is now defined very strictly.

The next step (different PR) will be implementing the session and CLI logic for reading the JSON and loading only the matching plugin module based on that.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,6 +68,34 @@ jobs:
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
+  test_build:
+    name: Test build
+    if: github.event_name != 'push'
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


bastimeyer:
https://github.com/streamlink/streamlink/actions/runs/7604463573/job/20707382744?pr=5793#step:6:1

bastimeyer:
Rebased to master... Going to open the lazy plugins loader PR now with these changes included. Once everything's validated and potential issues resolved, this PR here will get merged first, then the lazy plugins loader PR will be rebased and merged after that.

PR #5801, opened 2024-01-28
plugins.rtbf: remove plugin

bastimeyer:
The plugin's matcher pattern needs to be fixed, because it's defining a verbose pattern while not escaping the `#` character (which is interpreted as a comment), so it breaks the pattern's non-capture group when minifying the pattern in the plugins JSON build script (#5793) where it removes line breaks and leading whitespace of each line. The verbose-pattern issue was introduced by me in 98428d663e1cac084c4597419445042e6a81f9fd three years ago.

**As far as I can tell though, the plugin is completely broken and doesn't even match the URLs the site uses now.** The last time it was updated was 6 years ago:
https://github.com/streamlink/streamlink/commits/6.5.1/src/streamlink/plugins/rtbf.py

The site now also requires a login when accessing it, but I'm not sure if that's only the case when accessing it from outside of France/Belgium. It's also required when using proxies/VPNs. The DASH streams are working fine though without a proxy when copying the URL from the web browser's dev tools.

**If it's broken, then it should be removed instead. Can someone clarify?**

Pinging @melmorabity, who's added the plugin and who's updated it last.


bastimeyer:
Rebased... Going to remove the plugin instead...

PR #5816, opened 2024-02-07
ci.github: bump setup-python action to v5

bastimeyer:
Fixes the NodeJS 16 deprecation warning on the CI runners.


bastimeyer:
Will bump codecov/codecov-action separately. Breaking changes here apparently... Need to have a look later...

PR #5819, opened 2024-02-08
docs: fix custom ext_argparse Sphinx extension

bastimeyer:
- Move the `parser_helper()` function from `streamlink_cli.main` into the `streamlink_cli._parser` module
- Refactor `ext_argparse` and set default import paths
- Update import path in `build-shell-completions.sh`

Unfortunately, the `streamlink_cli._parser` module can't be excluded from the sdist/wheel distributions, as packagers might build shell completions from the installed `streamlink_cli` package instead of using the pre-built shell completions included in the sdist or building them from an editable install.

----

This helper function was added in #1638 and serves no purpose in the CLI's main module itself. It should therefore not be a part of it.

### DIFF
--- a/docs/_man.rst
+++ b/docs/_man.rst
@@ -31,8 +31,6 @@ Options
 -------
 
 .. argparse::
-    :module: streamlink_cli.main
-    :attr: parser_helper
 
 
 Bugs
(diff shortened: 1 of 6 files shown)


bastimeyer:
https://github.com/streamlink/streamlink/blob/6.5.1/script/build-shell-completions.sh#L26
will fix later...

gravyboat (approved):

PR #5838, opened 2024-02-17
Fix plugin arguments docs / man page / help text

bastimeyer:
#5837

### DIFF
--- a/docs/ext_argparse.py
+++ b/docs/ext_argparse.py
@@ -157,7 +157,7 @@ class ArgparseDirective(Directive):
         for group in parser.NESTED_ARGUMENT_GROUPS[parent]:
             is_parent = group in parser.NESTED_ARGUMENT_GROUPS
             # Exclude empty groups
-            if not group._group_actions and not is_parent:
+            if not is_parent and not any(action.help != argparse.SUPPRESS for action in group._group_actions or []):
                 continue
             title = group.title
             yield ""
--- a/src/streamlink/options.py
+++ b/src/streamlink/options.py
@@ -1,3 +1,4 @@
+import argparse
 from typing import (
     Any,
     Callable,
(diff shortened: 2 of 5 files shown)


bastimeyer:
1 and 2:
`twitch` arguments from the `--help` outpuf with plugins data from JSON. Correct order and no suppressed arguments.
```
  Twitch:
    --twitch-disable-ads
    Skip embedded advertisement segments at the beginning or during a stream.
    Will cause these segments to be missing from the output.
    
    --twitch-low-latency
    Enables low latency streaming by prefetching HLS segments.
    Sets --hls-segment-stream-data to true and --hls-live-edge to 2, if it is higher.
    Reducing --hls-live-edge to `1` will result in the lowest latency possible, but will most likely cause buffering.
    
    In order to achieve true low latency streaming during playback, the player's caching/buffering settings will
    need to be adjusted and reduced to a value as low as possible, but still high enough to not cause any buffering.
    This depends on the stream's bitrate and the quality of the connection to Twitch's servers. Please refer to the
    player's own documentation for the required configuration. Player parameters can be set via --player-args.
    
    Note: Low latency streams have to be enabled by the broadcasters on Twitch themselves.
    Regular streams can cause buffering issues with this option enabled due to the reduced --hls-live-edge value.
    
    --twitch-api-header KEY=VALUE
    A header to add to each Twitch API HTTP request.
    
    Can be repeated to add multiple headers.
    
    Useful for adding authentication data that can prevent ads. See the plugin-specific documentation for more information.
    
    --twitch-access-token-param KEY=VALUE
    A parameter to add to the API request for acquiring the streaming access token.
    
    Can be repeated to add multiple parameters.
    
    --twitch-purge-client-integrity
    Purge cached Twitch client-integrity token and acquire a new one.
```

3:
`sbscokr` is not listed anymore in the CLI docs:
- https://deploy-preview-5838--streamlink.netlify.app/cli#raiplay
- https://github.com/streamlink/streamlink/blob/6.6.0/src/streamlink/plugins/sbscokr.py#L26-L29

gravyboat (approved):

PR #5933, opened 2024-04-09
tests: fix test_onbuild when executed from build directory

amurzeau:
When executing tests from the build directory (generated by bdist), the location of the Streamlink source is directly at the root directory and not in a `src` subdirectory.

This is what Debian does when building python packages to ensure newly built files are tested instead of source files.

### DIFF
--- a/build_backend/test_onbuild.py
+++ b/build_backend/test_onbuild.py
@@ -38,7 +38,13 @@ def build(request: pytest.FixtureRequest, tmp_path: Path, template_fields: dict)
     (pkg_dir / "streamlink").mkdir(parents=True)
     shutil.copy(PROJECT_ROOT / "pyproject.toml", tmp_path / "pyproject.toml")
     shutil.copy(PROJECT_ROOT / "setup.py", tmp_path / "setup.py")
-    shutil.copy(PROJECT_ROOT / "src" / "streamlink" / "_version.py", pkg_dir / "streamlink" / "_version.py")
+
+    # Lookup for the path from the root source directory
+    if Path(PROJECT_ROOT / "src" / "streamlink" / "_version.py").exists():
+        shutil.copy(PROJECT_ROOT / "src" / "streamlink" / "_version.py", pkg_dir / "streamlink" / "_version.py")
+    else:  # pragma: no cover
+        # We didn't find it, use the build location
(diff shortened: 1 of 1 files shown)


build_backend/test_onbuild.py
```diff
@@ -38,7 +39,13 @@ def build(request: pytest.FixtureRequest, tmp_path: Path, template_fields: dict)
     (pkg_dir / "streamlink").mkdir(parents=True)
     shutil.copy(PROJECT_ROOT / "pyproject.toml", tmp_path / "pyproject.toml")
     shutil.copy(PROJECT_ROOT / "setup.py", tmp_path / "setup.py")
-    shutil.copy(PROJECT_ROOT / "src" / "streamlink" / "_version.py", pkg_dir / "streamlink" / "_version.py")
+
+    # Lookup for the path from the root source directory
+    if os.path.isfile(PROJECT_ROOT / "src" / "streamlink" / "_version.py"):
```
bastimeyer, line 44: `pathlib`'s equivalent of `os.path.isfile()` doesn't require an additional import and is simpler:
https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

build_backend/test_onbuild.py
```diff
@@ -38,7 +39,13 @@ def build(request: pytest.FixtureRequest, tmp_path: Path, template_fields: dict)
...
+        shutil.copy(PROJECT_ROOT / "src" / "streamlink" / "_version.py", pkg_dir / "streamlink" / "_version.py")
+    else:
```
bastimeyer, line 46: Please disable the branch coverage on the else-block:
`  # pragma: no cover`
https://coverage.readthedocs.io/en/latest/excluding.html#branch-coverage

The codecov config is currently not set up correctly (oops), which is why the PR passed...
https://github.com/streamlink/streamlink/blob/6.7.2/.codecov.yml#L47-L51
https://github.com/streamlink/streamlink/pull/5933/checks?check_run_id=23636174400

bastimeyer (commented):
Thanks!

build_backend/test_onbuild.py
amurzeau: Thanks, I've replaced it with Path.exists() instead of is_file(), it is more in line with what I wanted to do.

PR #5947, opened 2024-04-16
docs: clean up contributing guidelines

bastimeyer:
Some clean up of the contributing guidelines. This intentionally ignores the plugin requests for now.

Regarding the removal of `KNOWN_ISSUES.md`: it's completely useless. Known issues are those which are opened on the issue tracker.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,6 +14,7 @@ indent_size = 2
 
 [*.{md,markdown}]
 trim_trailing_whitespace = false
+max_line_length = off
 
 [docs/Makefile]
 indent_size = 8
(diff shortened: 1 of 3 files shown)


mkbloke:
Is it worth adding some advice to paragraph two of `Isolate the demonstrable problem` regarding checking that a stream is working correctly in a browser?  I know occasionally it has happened where people have reported a plugin issue, only to find it's actually a site issue.  It doesn't happen that often though, admittedly...

mkbloke (approved):

bastimeyer:
> checking that a stream is working correctly in a browser

Good point. Should be added to the plugin issue template though IMO

mkbloke:
> Should be added to the plugin issue template though IMO

Yes, that could be a better place.

gravyboat (approved):
Looks good and addresses a few annoying problems we've been seeing a lot lately. 👍

Issue #5904, opened 2024-03-25
Plugin-request rule update discussion


bastimeyer:
Let's talk about updating the plugin request/inclusion rules... The last time these rules were updated was in [October 2020](https://github.com/streamlink/streamlink/pull/3244) and [June 2018](https://github.com/streamlink/streamlink/pull/1838).

While not being mentioned directly anywhere except in some buried discussions from years ago, there are still plugins on the master branch despite them violating the (current) rules. We never deleted any of those plugins unless they broke and required major changes in order to fix them. Others however were removed with the addition of certain rules.

----

## Current plugin request rules

https://github.com/streamlink/streamlink/blob/6.7.2/CONTRIBUTING.md#plugin-requests

> 1. Any kind of streaming service that uses DRM protection
> 2. Sites which are hosting stolen content as their main source of content
> 3. Sites which are primarily rehosting content that is available from a legitimate source (TV shows, sports, news, etc.)
> 4. Sites which require any sort of cable login or subscription
> 5. NSFW sites of a pornographic nature (cam sites, porn sites, etc.)
> 6. Sites which don't provide any real live streaming content, eg. only VODs or VODs being rebroadcasted
> 7. Sites which don't provide any video streaming content, eg. radio stations
> 8. Sites where the maintainer has requested we not add their site to Streamlink
> 9. Sites which are unmaintained, are in beta or are undergoing heavy amounts of development and may change rapidly
> 10. Sites which have no way to determine viewership numbers
> 11. Sites which are static cameras of a physical location

## Rule "categories"

There are mainly two different kinds of rules, excluding three of the eleven rules:

1. Ones which prevent plugins from being added due to the site's content
    - 1 (DRM content)
    - 2 (stolen content)
    - 3 (rehosted content)
    - 5 (NSFW content)
    - 6 (only VOD content)
2. Ones which were added in order to keep the number of plugins low
    - 7 (no video streams)
    - 10 (unknown number of viewers)
    - 11 (static cameras)

The other three rules are this:

- 4 (cable logins / subscriptions)
- 8 (site asked for the plugin to be removed)
- 9 (unmaintained / in-development sites)

----

Category 1 is pretty much self-explanatory.

Rules of category 2 were added either because of maintenance cost or due to plugins deemed being "not useful" or "too niche" and unnecessarily causing bloat in Streamlink's code base and session memory during runtime. Another reason for adding those rules were people who made daily low-effort requests for sites they didn't even use themselves and probably just found via Google.

[...]

PR #5949, opened 2024-04-17
docs: rephrase and clarify plugin rules

bastimeyer:
See #5904
Follow-up of #5947 

This doesn't update any of the current plugin rules and just rephrases and clarifies them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,31 +45,64 @@ Feature requests are welcome, but take a moment to find out whether your idea fi
 
 ## Plugin requests
 
-Plugin submissions and requests are a great way to improve Streamlink. Requests should be as detailed as possible and dedicated to only a single streaming service. Information about the service as well as explicit URLs for live streams are required, as are details regarding the website, what it is, who runs it, how it adds to Streamlink, etc. This allows for easier plugin review and implementation. We reserve the right to refuse to implement, develop, or update any plugin. In addition we may remove existing plugins at our own discretion.
+Plugin submissions and requests are a great way to improve Streamlink. Requests should be as detailed as possible and dedicated to only a single streaming website/service.
 
-Plugins which fall under the following categories will not be implemented or considered and the request will be closed:
+Information about the website, as well as explicit URLs from your web browser's URL address bar are required, as are details regarding the website, what it is, who it is run by, and more. This allows for easier plugin review and implementation. We reserve the right to refuse to implement, develop, or update any plugin. In addition, we may remove existing plugins at our own discretion at any time without explanation.
 
-1. Any kind of streaming service that uses DRM protection
+Before you request a plugin, please be aware that this will merely be a suggestion for the project maintainers (or anyone else) to have a look at the website and implement a new plugin for it. In the vast majority of cases, no one except you will initially have an interest in such a plugin, so you should think about whether it will be useful for other users of Streamlink.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -45,31 +45,64 @@ Feature requests are welcome, but take a moment to find out whether your idea fi
...
+Plugin submissions and requests are a great way to improve Streamlink. Requests should be as detailed as possible and dedicated to only a single streaming website/service.
 
-Plugins which fall under the following categories will not be implemented or considered and the request will be closed:
+Information about the website, as well as explicit URLs from your web browser's URL address bar are required, as are details regarding the website, what it is, who it is run by, and more. This allows for easier plugin review and implementation. We reserve the right to refuse to implement, develop, or update any plugin. In addition, we may remove existing plugins at our own discretion at any time without reasoning.
```
gravyboat, line 50: > we may remove existing plugins at our own discretion at any time without reasoning.

maybe instead say:

> we may remove existing plugins at our own discretion at any time without explanation.

Some people may try to argue with us on this point if they believe we can be reasoned with.

CONTRIBUTING.md
```diff
@@ -45,31 +45,64 @@ Feature requests are welcome, but take a moment to find out whether your idea fi
...
+1. **Any kind of streaming service that uses DRM protection**
 
-8. Sites where the maintainer has requested we not add their site to Streamlink
+    Streamlink won't implement or break any DRM schemes.
 
-9. Sites which are unmaintained, are in beta or are undergoing heavy amounts of development and may change rapidly
+2. **Sites which are hosting stolen content as their main source of content**
 
-10. Sites which have no way to determine viewership numbers
+    Plugins must only support websites with authentic content.
```
gravyboat, line 69: How are we defining "authentic" here? This was an annoying spot we got in to in the past when people asked us to implement some weird TV station that was running news broadcasts or other local content. Does this need to be more explicit? I don't think a nightly news broadcast for a town of 20 people really counts as authentic or non-niche.

CONTRIBUTING.md
```diff
@@ -45,31 +45,64 @@ Feature requests are welcome, but take a moment to find out whether your idea fi
...
+
+    Plugins must only implement the original/primary website where the content is hosted or shown.
+
+4. **Sites which require paid logins or subscriptions**
+
+    Plugin implementations and maintenance must be possible by everyone. Paid access prevents this.
+
+5. **NSFW sites of a pornographic nature**
+
+    This also includes borderline NSFW content, as it lowers the bar for follow-up plugin requests.
```
gravyboat, line 81: 👍 Good addition to note here regarding borderline content.

gravyboat (approved):
Really good changes. I've added a minor suggestion and brought up one point around content which could put us in a sticky spot regarding which might allow people to start asking for their local TV station websites again. Take or leave whatever makes sense to you.

CONTRIBUTING.md
bastimeyer: I believe you're confusing two rules here. Niche content is already covered by rule 10.

gravyboat: No I'm aware of rule 10 but my assumption is most people aren't going to read all the rules so I'm not sure if we should be more explicit in this one. It's not a major deal or anything, I just don't want the headache of arguing with people about what the word "authentic" means so I figured I'd mention it. We can revise later if it becomes a problem.

PR #5958, opened 2024-04-21
cli: remove console.exit()

bastimeyer:
Second attempt: #5044

Had these changes stashed for a while and forgot about it... This cleans up the random `console.exit()` calls and instead raises a `StreamlinkCLIError` exception, which is then caught in the `main()` function.

There's still a lot to do here in the future. For example, the entire `ConsoleOutput` implementation is pretty bad, especially in combination with the regular logging system and the download progress output, where it interweaves the progess and log lines. In the previous PR I also rewrote some tests. I didn't do this here and just did some bare minimum fixing of the existing tests, because it doesn't make sense writing/updating tests for the flawed CLI main module. Hopefully at some point in the future, the entire CLI will be rewritten with a clean design.

This newly added `StreamlinkCLIError` exception class allows adding proper exit codes. Currently everything's set to exit code 1, just like it is now (except the untouched parts where 128+2 is returned on SigInt/SigTerm). Adding specific exit codes for different kinds of errors can be figured out for the next major version release.

The error behavior and error output format should be the same compared to master. Lots of tests are missing though, so there's potential for me to have changed something by accident.

### DIFF
--- a/src/streamlink_cli/console.py
+++ b/src/streamlink_cli/console.py
@@ -82,16 +82,5 @@ class ConsoleOutput:
         msg = dumps(out, cls=JSONEncoder, indent=2)
         self.output.write(f"{msg}\n")
 
-        if isinstance(out, dict) and out.get("error"):
-            sys.exit(1)
-
-    def exit(self, msg: str) -> None:
-        if self.json:
-            self.msg_json(error=msg)
-        else:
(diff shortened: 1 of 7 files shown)


gravyboat (approved):
The changes look good and are much easier to understand. If something did get missed I'm not seeing it. Everything looks like it should operate the same as before.

Issue #5732, opened 2023-12-20
Should the "32-bit" Windows builds and Linux AppImages be dropped?


bastimeyer:
*Asking this on the main repo for the sake of visiblity... Since it affects both the Windows builds and Linux AppImages, it doesn't make sense opening an issue on each of the issue trackers of the respective repos.*

I could also simply ignore user feedback, but I'm interested to hear about **actual** use cases.

----

## Issue

Both the `x86` Windows builds and `i686` Linux AppImages are currently stuck on Python 3.11 and can't be upgraded to Python 3.12 due to missing wheels of some of Streamlink's dependencies. This blocks the upgrade of other architectures (`x86_64` and `aarch64`) from Python 3.11 to 3.12, because all builds are supposed to be kept on the same version of Python, to avoid unnecessary maintenance headaches and to avoid unnecessary end-user confusion. New dependency issues have also already come up...

- streamlink/windows-builds#7
- streamlink/streamlink-appimage#3

The wheels which are missing could theoretically be built manually in separate repos here using GitHub CI, so the wheels could then be downloaded by the Windows builds and Linux AppImages from GitHub release assets instead of PyPI (where they are missing), but this requires maintenance of lots of build scripts for those third-party dependencies and their own dependencies. I absolutely don't want to spend any time on that, so this is not going to happen. I also don't want to have to deal with more dependency overrides in the build configs.

lxml [`4.9.3`](https://pypi.org/project/lxml/4.9.3/#files) which was released in July still hasn't published py312 wheels for `win32` and for `i686` (on `manylinux2014` aka. `manylinux_2_17`), and their most recent release from yesterday ([`4.9.4`](https://pypi.org/project/lxml/4.9.4/#files)) has exactly the same issue. I had already pointed this out on their issue tracker in October when py312 was released, but this issue didn't get any attention by the project's developer until yesterday when I asked again while pointing out new issues. I didn't get a reply, but some of the pointed out issues like a PyPI deploy issue appear to be resolved now or are being worked on according to the project's commit log. Not sure though when all the missing wheels will be fully resolved by a new release.

As you can see, it's been almost three months now and the issue still remains.

## Reasoning

So why keep 32-bit Windows builds and Linux AppImages, even if those dependency issues may get fixed in the future?

[...]

PR #5981, opened 2024-05-08
docs: update Windows builds flavors

bastimeyer:
In preparation for #5732 which will remove the py38 and x86 builds. The table has to go, hence the switch to the cards layout. Once only x86_64 builds are availble, the inner grid will be removed. Should there ever be ARM builds for Windows, then a simple badge can be added to the x86_64 one.

- https://sphinx-design.readthedocs.io/en/latest/grids.html
- https://sphinx-design.readthedocs.io/en/latest/cards.html

### DIFF
--- a/docs/_static/styles/custom.css
+++ b/docs/_static/styles/custom.css
@@ -296,7 +296,13 @@ table.table-custom-layout.table-custom-layout-dependencies tbody td {
   font-size: 5em;
 }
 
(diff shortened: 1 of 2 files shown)


bastimeyer:
https://deploy-preview-5981--streamlink.netlify.app/install#windows-binaries

gravyboat (approved):

Issue #5379, opened 2023-06-13
plugins.tf1: Permission insuffisante


campones:
### Checklist

- [X] This is a plugin issue and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

Latest stable release

### Description

Hi
Not sure exactly what means "permissions insuffisante" but sounds like something changed. Only LCI is still working

### Debug log

```text
/usr/local/bin/streamlink --loglevel debug "http://www.tf1.fr/tmc/direct" best                                                                                                         [cli][info] streamlink is running as root! Be careful!
[cli][debug] OS:         Linux-5.15.0-73-generic-x86_64-with-glibc2.29
[cli][debug] Python:     3.8.10
[cli][debug] Streamlink: 5.5.1
[cli][debug] Dependencies:
[cli][debug]  certifi: 2019.11.28
[cli][debug]  isodate: 0.6.0
[cli][debug]  lxml: 4.8.0
[cli][debug]  pycountry: 22.3.5
[cli][debug]  pycryptodome: 3.9.9
[cli][debug]  PySocks: 1.7.1
[cli][debug]  requests: 2.27.1
[cli][debug]  urllib3: 1.26.13
[cli][debug]  websocket-client: 1.3.2
[cli][debug] Arguments:
[cli][debug]  url=http://www.tf1.fr/tmc/direct
[cli][debug]  stream=['best']
[cli][debug]  --loglevel=debug
[cli][info] Found matching plugin tf1 for URL http://www.tf1.fr/tmc/direct
[plugins.tf1][debug] Found channel tmc (L_TMC)
[plugins.tf1][error] Permission insuffisante
error: No playable streams found on this URL: http://www.tf1.fr/tmc/direct
```

novazur972:
confirmed for me.
probably related to the creation of the subscription to tf1max (none free)

campones:
I dont think so because apparently tf1 max's purpose is to propose an ad-free streaming experience as well as HD streaming.
I could watched just fine on their site with just a free login

BellezaEmporium:
[...]

PR #5982, opened 2024-05-09
plugins.tf1 add authentication part

W4RdZ:
Hi,

To be transparent, I'm not a dev, so i took this modified code from BellezaEmporium code to make a pull request : https://raw.githubusercontent.com/BellezaEmporium/streamlink/master/src/streamlink/plugins/tf1.py

Also this code seems to works before lazy update.

there is an issue opened here : https://github.com/streamlink/streamlink/issues/5379

The actual issue on the master version is that there is no authentication part to access the streams. Part which is required by the host.

Kind regards.

### DIFF
--- a/src/streamlink/plugins/tf1.py
+++ b/src/streamlink/plugins/tf1.py
@@ -7,11 +7,13 @@ $type live
 $region France
 """
 
+import json
 import logging
 import re
 
 from streamlink.plugin import Plugin, PluginError, pluginmatcher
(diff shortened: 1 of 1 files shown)


bastimeyer:
That is not what I meant with "please open a PR". I can pull code from other people myself just fine. Opening a PR however means that you have to actively work with the maintainers with your own code contribution, which is not the case here and won't happen, because as you just said, you have no idea about what the code does.
Closing...

Issue #6001, opened 2024-05-23
plugins.chzzk: new live detail response found


Clap2rap:
### Checklist

- [X] This is a [plugin issue](https://streamlink.github.io/plugins.html) and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [X] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Streamlink version

6.7.4

### Description

Found counterexample: https://chzzk.naver.com/live/2f6ac1461216969b428ea31e83853d42
Response: `{"code":200,"message":null,"content":null}`

Some channels have no content value (which means they are currently not live).
I think those channels haven't broadcasted recently (or never) so there's no content value.

Current `_query_api` validation is as follows:
```python
    def _query_api(self, url, *schemas):
        return self._session.http.get(
            url,
            acceptable_status=(200, 404),
            schema=validate.Schema(
                validate.parse_json(),
                validate.any(
                    validate.all(
                        {
                            "code": int,
                            "message": str,
                        },
                        validate.transform(lambda data: ("error", data["message"])),
                    ),
                    validate.all(
                        {
                            "code": 200,
                            "content": dict,
                        },
                        validate.get("content"),
                        *schemas,
                        validate.transform(lambda data: ("success", data)),
                    ),
                ),
            ),
        )
```
so the response above makes a validation error.

### Debug log

```text
streamlink --loglevel=debug --output=~/Downloads/test.ts https://chzzk.naver.com/live/2f6ac1461216969b428ea31e83853d42

[session][debug] Loading plugin: chzzk
[cli][debug] OS:         macOS 14.5
[cli][debug] Python:     3.11.8
[cli][debug] OpenSSL:    OpenSSL 3.0.13 30 Jan 2024
[cli][debug] Streamlink: 6.7.4
[cli][debug] Dependencies:
[cli][debug]  certifi: 2024.2.2
[cli][debug]  isodate: 0.6.1
[cli][debug]  lxml: 5.2.1
[cli][debug]  pycountry: 23.12.11
[cli][debug]  pycryptodome: 3.20.0
[...]

PR #6002, opened 2024-05-23
plugins.chzzk: live detail api response validation error fix

Clap2rap:
Fixes: #6001

Any better suggestions are welcome.

### DIFF
--- a/src/streamlink/plugins/chzzk.py
+++ b/src/streamlink/plugins/chzzk.py
@@ -40,6 +40,13 @@ class ChzzkAPI:
                     validate.all(
                         {
                             "code": 200,
+                            "content": None,
+                        },
+                        validate.transform(lambda _: ("success", None)),
+                    ),
+                    validate.all(
+                        {
+                            "code": 200,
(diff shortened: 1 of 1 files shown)


bastimeyer (approved):
Thanks for the PR. Changes look fine.

However, should there be a need for extending the validation schemas even further, then I'd say that the error handling code should be moved from the `Plugin` subclass (`_get_live()` and `_get_video()`) to the `API` class. The error handling is currently done in the `Plugin` subclass, so that it can simply return `None` instead of having to raise a `NoStreamsError` in the `API` class. Irrelevant here though, just a quick thought...

Talking about error handling, are VODs also affected by a potential `None` return type?

Clap2rap:
Not affected, based on VODs I've tested. All the VODs have non-null `content` if they exist. Otherwise, the API returns a `code` field with the value 404 (or other error codes) and a `message` field indicating that the VOD does not exist (or other error messages).

And I agree with what you said. Raising `NoStreamsError` would be a more elegant way to handle it.

PR #6007, opened 2024-05-28
docs: add note about Ubuntu to install page

bastimeyer:
Since people are always confused about Ubuntu and don't read the package availability section in the install docs, let's add a note for Ubuntu, to explain the situation about the package always being outdated in their repos.

Ubuntu was removed from the install docs in #4670 / #4681


bastimeyer:
https://deploy-preview-6007--streamlink.netlify.app/install#linux-and-bsd

gravyboat (approved):

PR #6023, opened 2024-06-06
session: log sha256 of overriding plugin module

bastimeyer:


gravyboat:
Is there an issue attached to this? What is it needed for?

bastimeyer:
I want to be sure that when users sideload plugins when providing PR feedback that the plugin is indeed sideloaded correctly. The added SHA256 checksum in the log message will confirm that.

gravyboat (approved):
Okay sounds good, approved.

PR #6033, opened 2024-06-13
chore: add collab checkboxes to issue templates

bastimeyer:


.github/ISSUE_TEMPLATE/3_plugin_request.yml
```diff
@@ -33,32 +33,51 @@ body:
           required: true
         - label: "[I have checked the list of open and recently closed plugin requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+request%22)"
           required: true
+  - type: checkboxes
+    attributes:
+      label: Collaboration
+      options:
+        - label: My request is genuine, in the best interest of the project, and I have access to the stream(s) in my web browser
```
gravyboat, line 40: Does this need to be expanded to "without the usage of a VPN"?

gravyboat (approved):
Changes look good. One discussion point about VPN usage since it seems like we've been getting a lot more sites recently that only work with VPNs but up to you. 👍

.github/ISSUE_TEMPLATE/3_plugin_request.yml
bastimeyer: I was thinking about it, but there's the access restriction question down below

gravyboat: Works for me! Can always update it if needed.

PR #6051, opened 2024-06-25
plugins.twitch: log ad breaks with duration

bastimeyer:
This logs Twitch's ad breaks and their respective durations, which is especially useful if `--twitch-disable-ads` is set, so users will know how long the output will be paused for.

Everyone's invited to give this a test:
https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-request-feedback

----

This implementation caches the ad-break's `X-TV-TWITCH-AD-COMMERCIAL-ID` metadata in order to avoid logging multiple parts of the same ad break. This is because multiple `EXT-X-DATERANGE` tags can be inserted into the HLS playlists one after another, e.g. when multiple "ad-break-in-progress" placeholder loops are being shown in a row, or distinct "actual ads" are shown if enabled via `--twitch-access-token-param=playerType=web` (we default to "embed" so we only see the placeholders). Those ad-part date-ranges always share the ID of the whole ad break and always include the full duration via the `X-TV-TWITCH-AD-POD-FILLED-DURATION` metadata, so when the full ad break is 90s, it only logs it once instead of logging "30s" three times when there are three `EXT-X-DATERANGE` tags. There's also metadata for the number of parts in a break and the current part number, but this is irrelevant since we're only interested in the full duration which we only want to log once.

----

Prerolls + midrolls
```
[plugins.twitch][info] Will skip ad segments
[plugins.twitch][info] Low latency streaming (HLS live edge: 2)
[plugins.twitch][info] Waiting for pre-roll ads to finish, be patient
[plugins.twitch][info] Detected advertisement break of 15 seconds
[stream.hls][info] Filtering out segments and pausing stream output
[stream.hls][info] Resuming stream output
[plugins.twitch][info] Detected advertisement break of 90 seconds
[stream.hls][info] Filtering out segments and pausing stream output
[stream.hls][warning] Encountered a stream discontinuity. This is unsupported and will result in incoherent output data.
[stream.hls][info] Resuming stream output
```

No prerolls
```
[plugins.twitch][info] Will skip ad segments
[plugins.twitch][info] Low latency streaming (HLS live edge: 2)
[plugins.twitch][info] Detected advertisement break of 120 seconds
[stream.hls][info] Filtering out segments and pausing stream output
[stream.hls][warning] Encountered a stream discontinuity. This is unsupported and will result in incoherent output data.
[stream.hls][info] Resuming stream output
[plugins.twitch][info] Detected advertisement break of 120 seconds
[stream.hls][info] Filtering out segments and pausing stream output
[stream.hls][warning] Encountered a stream discontinuity. This is unsupported and will result in incoherent output data.
[stream.hls][info] Resuming stream output
```

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -15,14 +15,16 @@ $notes Acquires a :ref:`client-integrity token <cli/plugins/twitch:Client-integr
 import argparse
 import base64
 import logging
+import math
 import re
 import sys
+from collections import deque
 from contextlib import suppress
 from dataclasses import dataclass, replace as dataclass_replace
 from datetime import datetime, timedelta
 from json import dumps as json_dumps
 from random import random
-from typing import ClassVar, Mapping, Optional, Tuple, Type
+from typing import ClassVar, Deque, List, Mapping, Optional, Tuple, Type
 from urllib.parse import urlparse
 
 from requests.exceptions import HTTPError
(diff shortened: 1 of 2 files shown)


gravyboat (approved):
👍 Good idea. This is a nice quality of life change.

PR #6052, opened 2024-06-26
docs: drop i686 Linux AppImages and x86+py38 Windows builds

### LINKED ISSUE #5732: Should the "32-bit" Windows builds and Linux AppImages be dropped?

bastimeyer:
Closes #5732 

I'm going to merge this PR once the Linux AppImage and Windows builds have been updated. I don't know though when this will happen, because it's not particularly urgent to update the AppImages right now. I'll probably do this on Streamlink's next release with a minor version bump (`6.9.x`).

----

### Linux AppImages

CentOS 7 will reach its EOL at the end of this week on 2024-07-01, which is what the `pypa/manylinux2014` docker images are based on, which in turn is what Streamlink's AppImages are currently based on. This means that the AppImages will have to switch to newer manylinux images, namely `manylinux_2_28` (AlmaLinux 8 based with glibc 2.28), which will be the end for "32 bit" i686 AppImages.

- https://endoflife.date/centos
- https://github.com/pypa/manylinux/blob/095daf5fc97b98aa0f1e48c3a594df654caa09ee/README.rst#manylinux2014-centos-7-based-glibc-217
- https://peps.python.org/pep-0599/#rationale

### Windows builds

Since it's less work for me to do both things at the same time, I'm also going to drop the "32 bit" x86 builds for Windows, as well as the py38-based builds for Windows 7, even though py38 will see its EOL at the end of October (13 weeks from now). Streamlink's extended support for W7 via additional py38 builds has been going on for long enough, so I absolutely don't care about anyone using this OS in 2024 who won't be able to upgrade after this. Streamlink 7.0.0 won't support py38 anyway.

- https://endoflife.date/windows

And before anyone asks about ARM64/aarch64 builds for Windows (I left space for that in the docs layout), I've recently explained it here:
https://github.com/streamlink/windows-builds/issues/10

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -571,76 +571,36 @@ Windows binaries
     .. grid-item-card::
         :padding: 3
         :class-header: sd-text-center
-        :class-footer: sd-text-center sd-bg-transparent sd-border-0
+        :class-body: sd-py-0
+        :class-footer: sd-text-center sd-bg-transparent sd-border-0 sd-pt-0 sd-pb-3
 
         :fas:`gears` **Installer**
         ^^^
 
+        - Windows 10+
         - Adds itself to the system's ``PATH`` env var
         - Automatically creates a :ref:`config file <cli/config:Configuration file>`
         - Sets :option:`--ffmpeg-ffmpeg` in config file
 
         +++
-        .. grid:: 2
-            :gutter: 1
-            :padding: 0
-
-            .. grid-item::
-                :class: sd-text-right
-
-                Windows 10+
-
-            .. grid-item::
-
-                :bdg-link-success-line:`x86_64 <https://github.com/streamlink/windows-builds/releases>`
-                :bdg-link-primary-line:`x86 <https://github.com/streamlink/windows-builds/releases>`
-
-            .. grid-item::
-                :class: sd-text-right
-
-                Windows 7 (py38)
-
-            .. grid-item::
-
-                :bdg-link-secondary-line:`x86_64 <https://github.com/streamlink/windows-builds/releases>`
-                :bdg-link-secondary-line:`x86 <https://github.com/streamlink/windows-builds/releases>`
+        :bdg-link-success-line:`x86_64 <https://github.com/streamlink/windows-builds/releases>`
 
     .. grid-item-card::
         :padding: 3
         :class-header: sd-text-center
-        :class-footer: sd-text-center sd-bg-transparent sd-border-0
+        :class-body: sd-py-0
+        :class-footer: sd-text-center sd-bg-transparent sd-border-0 sd-pt-0 sd-pb-3
 
         :fas:`file-zipper` **Portable archive**
         ^^^
 
+        - Windows 10+
         - No :ref:`config file <cli/config:Configuration file>` created automatically
         - :option:`--ffmpeg-ffmpeg` must be set manually
         - No pre-compiled Python bytecode
 
         +++
-        .. grid:: 2
-            :gutter: 1
-            :padding: 0
-
-            .. grid-item::
-                :class: sd-text-right
-
-                Windows 10+
-
-            .. grid-item::
-
-                :bdg-link-success-line:`x86_64 <https://github.com/streamlink/windows-builds/releases>`
-                :bdg-link-primary-line:`x86 <https://github.com/streamlink/windows-builds/releases>`
-
-            .. grid-item::
-                :class: sd-text-right
-
-                Windows 7 (py38)
-
-            .. grid-item::
-
-                :bdg-link-secondary-line:`x86_64 <https://github.com/streamlink/windows-builds/releases>`
-                :bdg-link-secondary-line:`x86 <https://github.com/streamlink/windows-builds/releases>`
+        :bdg-link-success-line:`x86_64 <https://github.com/streamlink/windows-builds/releases>`
 
 **Contents**
 
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -715,7 +715,7 @@ Linux AppImages
 
 **Architectures**
 
-.. grid:: 3
+.. grid:: 2
     :padding: 0
 
     .. grid-item-card::
@@ -729,12 +729,6 @@ Linux AppImages
         :text-align: center
 
         :bdg-link-success-line:`aarch64 <https://github.com/streamlink/streamlink-appimage/releases>`
-
-    .. grid-item-card::
-        :padding: 3
-        :text-align: center
-
-        :bdg-link-primary-line:`i686 <https://github.com/streamlink/streamlink-appimage/releases>`
 
 **Contents**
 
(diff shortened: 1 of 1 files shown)


bastimeyer:
Rebased to master and updated the AppImage how-to section because of the glibc bump and the updated AppImage file names.

bastimeyer:
- https://streamlink--6052.org.readthedocs.build/install.html#windows-binaries
- https://streamlink--6052.org.readthedocs.build/install.html#linux-appimages

Issue #5859, opened 2024-02-24
Rumble.com


bermudi:
### Checklist

- [X] This is a plugin request and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+request%22)

### Description

Rumble is an online video platform founded in October 2013 by Chris Pavlovski as an alternative to YouTube for independent vloggers and smaller content creators.

According to the platform's terms and conditions, Rumble forbids pornography, harassment, racism, antisemitism, and copyright infringement. The platform also prohibits illegal content.

As of August 15, 2022, Rumble reported 78 million monthly active users (MAU). According to an August 2022 Reuters article, Rumble is 'better-funded' and 'more mainstream' than its competitors BitChute and Odysee.

According to a May 2022 Pew Research Center study, 20% of American adults have heard of Rumble, while 2% regularly got their news from Rumble.

In early 2023, Rumble began hosting live broadcasts for sports leagues owned by Thrill One Sports & Entertainment such as Nitrocross, Street League Skateboarding, and Power Slap.

### Input URLs

https://rumble.com/v4do2zn-2024-sls-paris-womens-and-mens-knockout-rounds.html

bastimeyer:
https://github.com/streamlink/streamlink/pull/6077#issuecomment-2220858720

PR #6077, opened 2024-07-09
plugins.rumble: new plugin

v2wy:
support rumble

#5859

### DIFF
--- a/src/streamlink/plugins/rumble.py
+++ b/src/streamlink/plugins/rumble.py
@@ -1,5 +1,5 @@
 """
-$description Rumble is an online video platform founded in October 2013 by Chris Pavlovski as an alternative to YouTube for independent vloggers and smaller content creators.
+$description Rumble is an online video platform founded in October 2013 by Chris Pavlovski as an alternative to YouTube.
 $url rumble.com
 $type live
 $metadata id
@@ -66,7 +66,8 @@ class Rumble(Plugin):
                 schema=validate.Schema(
                     validate.parse_html(),
                     validate.xml_xpath_string(
-                        "/html/body/main/section/ol/div[1]/div[2][@class='videostream__footer videostream__footer--live']/a/@href",
+                        "/html/body/main/section/ol/div[1]/div[2][@class='videostream__footer "
+                        + "videostream__footer--live']/a/@href",
                     ),
                 ),
                 timeout=5,
--- a/src/streamlink/plugins/rumble.py
+++ b/src/streamlink/plugins/rumble.py
@@ -1,0 +1,147 @@
+"""
+$description Rumble is an online video platform founded in October 2013 by Chris Pavlovski as an alternative to YouTube for independent vloggers and smaller content creators.
+$url rumble.com
+$type live
+$metadata id
+$metadata author
+$metadata title
+"""
+
+import logging
+import re
+from typing import Dict
+
+from streamlink.plugin import Plugin, pluginmatcher
+from streamlink.plugin.api import validate
+from streamlink.stream import HLSStream, HTTPStream
+
+
+log = logging.getLogger(__name__)
+
+
+@pluginmatcher(
+    re.compile(
+        r"https?://rumble\.com/c/(?P<channel>[^/?]+)",
+    ),
+    name="channel",
+)
+@pluginmatcher(
+    re.compile(
+        r"https?://rumble\.com/(?P<room_id>[^/.]+)\.html",
+    ),
+    name="live",
+)
+class Rumble(Plugin):
+    QUALITY_WEIGHTS: Dict[str, int] = {}
+
+    _URL_WEB_LIVE = "https://rumble.com/c/{channel}/livestreams"
+    _URL_API_LIVE_DETAIL = "https://rumble.com/embedJS/u3/"
+
+    _PARAM_VER = 2
+    _PARAM_REQUEST = "video"
+
+    _SCHEMA_STREAM = {
+        str: validate.all(
+            {
+                "url": validate.url(),
+                "meta": {"bitrate": int},
+            },
+            validate.union_get("url", ("meta", "bitrate")),
+        ),
+    }
+
+    @classmethod
+    def stream_weight(cls, key):
+        weight = cls.QUALITY_WEIGHTS.get(key)
+        if weight:
+            return weight, key
+
+        return super().stream_weight(key)
+
+    def _get_streams(self):
+        if self.matches["channel"]:
+            channel = self.match["channel"]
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/rumble.py
```diff
@@ -0,0 +1,148 @@
...
+                    validate.xml_xpath_string(
+                        "/html/body/main/section/ol/div[1]/div[2][@class='videostream__footer "
+                        + "videostream__footer--live']/a/@href",
+                    ),
+                ),
+                timeout=5,
+            )
+            if not live_uri:
+                return
+            room_url = f"https://rumble.com{live_uri}"
```
bastimeyer, line 77: That's an unsafe URL concatenation which must be avoided when random data is involved.
See `streamlink.utils.url.url_concat` or `urllib.parse.{urlunparse,urljoin}`.

src/streamlink/plugins/rumble.py
```diff
@@ -0,0 +1,148 @@
...
+            room_url = self.url
+        else:
+            return
+        embed_url = self.session.http.get(
+            room_url,
+            schema=validate.Schema(
+                validate.parse_html(),
+                validate.xml_xpath_string(
+                    '//script[@type="application/ld+json"]',
+                ),
```
bastimeyer, line 88: That XPath query will return the contents of all matching DOM nodes, or `None` if none was found. Limit the query to the first matching one, then also get its text contents explicitly.
`.//script[@type="application/ld+json"][1]/text()`

Due to the potential `None` return value, `validate.none_or_all(...)` must follow, or a `TypeError` will be shown to the user.

src/streamlink/plugins/rumble.py
```diff
@@ -0,0 +1,148 @@
...
+                                },
+                            },
+                            [],
+                        ),
+                        validate.optional("mp4"): validate.any(self._SCHEMA_STREAM, []),
+                        validate.optional("webm"): validate.any(self._SCHEMA_STREAM, []),
+                    },
+                    "title": str,
+                    "author": {"name": str},
+                },
```
bastimeyer, line 128: The HLS multivariant playlist URLs seem to be static:
`https://rumble.com/live-hls/{self.id}/playlist.m3u8`

Is there a need for mp4/webm containers via `HTTPStream`?

bastimeyer (commented):
Thanks for the PR.

No idea how I feel about merging a plugin for this site, considering the primary content that's hosted there.
https://en.wikipedia.org/wiki/Rumble_(company)

Just because there are requests open on the issue tracker doesn't mean that a working plugin implementation is guaranteed be merged, regardless the plugin request rules, so please be aware of that, so you don't potentially waste your time.

Nevertheless, let me give you a quick code review.

src/streamlink/plugins/rumble.py
v2wy: If the video is in live broadcast mode, then HLS is available. Once the live broadcast ends, HLS is no longer accessible, but MP4 can still be obtained, for example on this page
https://rumble.com/v5680wc-the-rob-carson-show-07092024-hour-1-newsmax-podcasts.html 
However, I have not yet discovered any instances of the use of WebM

v2wy:
I have read the plugin rules, but I'm not sure which rule this website has violated?
@bastimeyer

src/streamlink/plugins/rumble.py
v2wy: > The HLS multivariant playlist URLs seem to be static: `https://rumble.com/live-hls/{self.id}/playlist.m3u8`
> 
> Is there a need for mp4/webm containers via `HTTPStream`?

You are right, currently HLS follows this rule, as long as you have obtained self.id, you can already get the HLS URL. I didn't notice this when I was developing.

bastimeyer:
> I have read the plugin rules, but I'm not sure which rule this website has violated?

I didn't say that it violates any of the listed rules. I raised some concerns considering the site's content, which is my personal opinion. But as the project maintainer, this also matters, because I don't want the project to be involved in this, and neither do I want to be associated with it.
https://github.com/streamlink/streamlink/blob/6.8.2/CONTRIBUTING.md#plugin-requests

Apart from their tiny and basically irrelevant gaming category, all I can see on this site is political content with lots of misinformation/propaganda, conspiracy theories, hate speech, and similar content from channels which have been banned on every other big and established video/streaming or social-media platform, because of their content. This site is anything but neutral in regards to the content that's available there.
https://rumble.com/browse/live
https://rumble.com/category/gaming

And on top of that, the content that's supposedly live is for the most part just a re-broadcast. The viewer numbers are also completely fake and don't match the viewer chat activity at all, which is typical for sites like this.

gravyboat:
@bastimeyer I agree with you. At some point as project maintainers we have to take a stand. I do not support merging a plugin for a site where the primary focus is hate and misinformation. As you noted even if that factor is excluded it's just a bunch of garbage re-broadcasts with faked viewer numbers. We wouldn't merge this any way.

bastimeyer:
Going to close this PR now, as we're not going to merge a plugin for that site. Plugins can always be sideloaded though.
Thanks for the PR nonetheless, @v2wy.

bermudi:
> @bastimeyer I agree with you. At some point as project maintainers we have to take a stand. I do not support merging a plugin for a site where the primary focus is hate and misinformation. As you noted even if that factor is excluded it's just a bunch of garbage re-broadcasts with faked viewer numbers. We wouldn't merge this any way but that should not be our main reason for not merging this. The content is despicable and that should be more than enough justification.

I can't believe politics is driving this decision. I'm as far left as they come and I am interested in this for research purposes but whenever politics starts clouding the decisions it affects everyone, not just the people you hate and want to exact some kind of revengeance on. But anyhow, just my 2¢

gravyboat:
@bermudi To specifically say politics instead of what I focused on in my comment that you've quoted (hate and misinformation) is disingenuous and @bastimeyer touched on other valid concerns in their comment. If you want to use the plugin side load it or fork the project and then you can add whatever plugins you want.

bermudi:
> @bermudi To specifically say politics instead of what I focused on in my comment that you've quoted (hate and misinformation) is disingenuous 

you seriously trying to say that excluding a plugin because of the type of content is not political decision?

gravyboat:
> you seriously trying to say that excluding a plugin because of the type of content is not political decision?

Absolutely. We have excluded many plugins based on the type of content they provide.

bermudi:
> We have excluded many plugins based on the type of content they provide.

How could that possibly be not political? Would you describe it as technical decision then? This is unbelievable

PR #6089, opened 2024-07-20
tools: add explicit ruff format config

bastimeyer:
Except for `preview` (and `exclude`), all these config values are already the default, but let's define them explicitly.
- https://docs.astral.sh/ruff/settings/#format
- https://docs.astral.sh/ruff/formatter/#preview-style

Adding the formatter config doesn't change anything, as we still don't enforce a strict code style. But as said in #6087, I'm going to finally change that after dropping py38.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -177,6 +177,16 @@ extend-exclude = [
   "src/streamlink_cli/packages/",
 ]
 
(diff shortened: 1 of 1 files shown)


gravyboat:
👍 Good idea.

PR #6112, opened 2024-08-04
session: move plugin.api.useragents

bastimeyer:
Move `plugin.api.useragents` to `session.http_useragents` so that the HTTPSession doesn't depend on the plugin API.

Done in two commits, so that the file's history is retained.

----

This change is necessary for the webbrowser API in #6111 (which I will update afterwards) to also not depend on the plugin API, same as the HTTPSession.


bastimeyer:
Plugins should keep importing from `plugin.api.useragents` if custom UAs are required.

PR #6114, opened 2024-08-04
webbrowser.cdp.client: update UA in headless mode

bastimeyer:
- Add `headless` keyword and attribute to `CDPClient`
- If in headless mode, read user-agent from CDP, remove headless hint, and then update UA via `Network.setUserAgentOverride()`
- Remove `--user-agent=...` Chromium launch argument if in headless mode as it also overrides the UA platform

----

Replaces #6111

See #6113 
/ping @Hakkin 

No changes to any plugins / CDPClient implementation yet...

Overriding the UA string via `Network.setUserAgentOverride()` worked in my AWS-WAF test. I don't know the difference between this and `Emulation.setUserAgentOverride()` though. The method from the CDP emulation domain might do more than overriding network request headers.
- https://chromedevtools.github.io/devtools-protocol/1-3/Network/#method-setUserAgentOverride
- https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setUserAgentOverride

I also avoided any other params like `userAgentMetadata` because I don't think it's necessary, as we're only interested in removing the "Headless" part from the UA string, and not change any platform values. So the defaults will be kept by Chromium.

----

So in order to test this in the Twitch plugin, the forced `headless=False` parameter needs to be removed and then specific GQL API endpoints need to be checked with the CI token included. As said in #6113, the streaming access tokens seem to work regardless the (now hidden) `is_bad_bot` status, while some GQL API endpoints allegedly fail. I haven't tested this yet and will check tomorrow. That's it for me today.

PR #6116, opened 2024-08-05
Change default value of --webbrowser-headless to False

bastimeyer:
See https://github.com/streamlink/streamlink/pull/6114#issuecomment-2269268484

Since headless mode can potentially be detected, we shouldn't set it as the default mode. Plugins should always work reliably in their default configuration without users having to change Streamlink's default options should there be a change of headless mode detection by the website a plugin covers with Streamlink's webbrowser API. It doesn't matter that a short web browser window might appear on the user's screen (in the default configuration).

So, this changes the default value of `--webbrowser-headless` from `True` to `False` and now also logs which mode Chromium is being launched in. I don't consider this a "breaking change" (wouldn't be a real one anyway as it's only a change in behavior), as the entire API is marked as unstable, and we've also only been shipping it in the Twitch plugin which so far (still) overrides the value and sets it to `False` anyway. Recently we've updated the vtvgo plugin with a web browser requirement, but this hasn't made it into a stable release yet.

I'm going to remove the override from the Twitch plugin in a follow-up PR, which means that with the new defaults, nothing will change here and users will be able to set headless mode via `--webbrowser-headless=True`. Users who insisted on having a client-integrity token included in Twitch API requests so far needed to have a patched plugin version anyway, because Twitch removed the requirement after a couple of days, so the webbrowser API was never actually used by default, but with the recent addition of `--twitch-force-client-integrity`, users will now be able to get CI tokens with and without headless mode, or not (as long as Twitch keeps the requirement disabled).

This change of default value also means that the CLI arg needs to be set explicitly in environments without a desktop (actual headless requirement).

----

Opinions on the log message and updated argument help text? Should this be made more clear? I don't really want this to be verbose, because it's logged on the info level every time the process is launched.

### DIFF
--- a/src/streamlink/session/options.py
+++ b/src/streamlink/session/options.py
@@ -296,7 +296,7 @@ class StreamlinkOptions(Options):
           - The maximum amount of time for waiting on a single CDP command response
         * - webbrowser-headless
           - ``bool``
-          - ``True``
+          - ``False``
           - Whether to launch the webbrowser in headless mode or not
     """
 
@@ -340,7 +340,7 @@ class StreamlinkOptions(Options):
             "webbrowser-cdp-host": None,
             "webbrowser-cdp-port": None,
             "webbrowser-cdp-timeout": 2.0,
-            "webbrowser-headless": True,
+            "webbrowser-headless": False,
         })
         self.session = session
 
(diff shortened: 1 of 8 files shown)


gravyboat (approved):
> Opinions on the log message and updated argument help text? Should this be made more clear?

I think the log message and updated argument help text is fine. Most people aren't going to use this any way so if someone is they should know what they're doing. As you said it's logged every time at info level so anything more would be pretty spammy.

If it does become a problem or people are confused it can be updated at that time. Until then this should be plenty.

Issue #4535, opened 2022-05-19
Support for EXT-X-DISCONTINUITY tag


Grabien:
### Checklist

- [X] This is a feature request and not a different kind of issue
- [X] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [X] [I have checked the list of open and recently closed plugin requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22feature+request%22)

### Description

I'm trying to record an M3U8 stream from this page: https://www.wthitv.com/livestream/

Streamlink records the stream for some time, but then stops recording. When I look at the M3U8 source, I see in it EXT-X-DISCONTINUITY tags, and after each such tag, the TS fragments numeration starts again from "1". It looks like Streamlink stops recordings because of this.

Is it possible to add proper processing of this tag?

It's an example of M3U8:

```
#EXTM3U
#EXT-X-TARGETDURATION:12
#EXT-X-VERSION:3
#EXT-X-MEDIA-SEQUENCE:54
#EXT-X-START:TIME-OFFSET=0
## 2022-05-19 09:09:52.619518339 +0000 UTC
#EXT-X-DISCONTINUITY-SEQUENCE:1
#EXTINF:6.00,
https://wthi-cdn.bitcentral.com/sef/c_bdb309a0-87cf-402a-b8e3-d54922e91ec2_v_3500K720AWS_s_00085.ts?s=7b0d8466-3194-4409-85eb-b4dd8a0763d1&c=bdb309a0-87cf-402a-b8e3-d54922e91ec2&i=84&v=3500K720AWS&u=https:%2F%2Ffuel-streaming-prod01.fuelmedia.io&m=53&ch=3a9b6ddf-a8ad-422d-adf7-0a26f09c685d&d=6.00
#EXTINF:6.00,
https://wthi-cdn.bitcentral.com/sef/c_bdb309a0-87cf-402a-b8e3-d54922e91ec2_v_3500K720AWS_s_00086.ts?s=7b0d8466-3194-4409-85eb-b4dd8a0763d1&c=bdb309a0-87cf-402a-b8e3-d54922e91ec2&i=85&v=3500K720AWS&u=https:%2F%2Ffuel-streaming-prod01.fuelmedia.io&m=54&ch=3a9b6ddf-a8ad-422d-adf7-0a26f09c685d&d=6.00
#EXTINF:6.00,
[...]

PR #6123, opened 2024-08-11
plugins.pickleballtv: new plugin

kmac:
Add plugin for https://pickleballtv.com

### DIFF
--- a/src/streamlink/plugins/pickleballtv.py
+++ b/src/streamlink/plugins/pickleballtv.py
@@ -1,0 +1,63 @@
+"""
+$description pickleballtv.com
+$url pickleballtv.com
+$type live
+"""
+
+import logging
+import re
+
+from streamlink.plugin import Plugin, pluginmatcher
+from streamlink.plugin.api import validate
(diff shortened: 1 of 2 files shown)


kmac:
Note: this plugin suffers from https://github.com/streamlink/streamlink/issues/4535 on recording to file whenever pbtv goes to commercial. I just handle that with a wrapper script to handle writing to numbered files.

bastimeyer:
Sorry, but I'm afraid we're not going to merge a plugin for this site. See rule 10 (and maybe also 6):
https://github.com/streamlink/streamlink/blob/6.8.3/CONTRIBUTING.md#plugin-rules

This is way too niche for a plugin to be considered useful. I even had to look up what Pickleball is, because I've never heard of it. Their social media presence is another indicator for this.

As stated in the contribution guidelines, please open plugin requests on the issue tracker next time before implementing new plugins, so you don't waste your time. You can still [sideload](https://streamlink.github.io/cli/plugin-sideloading.html) your own plugin implementations, even if this one won't be merged into Streamlink's master branch.

Thanks for the PR though.

PR #6133, opened 2024-08-18
build: add Python 3.13 to package classifiers

bastimeyer:
No reason to merge this just yet


gravyboat (approved):
I'll mark it as approved so it's ready when you are.

PR #5508, opened 2023-08-20
plugins.piaulizaportal: new plugin

pzhlkj6612:
### Summary

This is a very simple extractor for video platform "[ULIZA](https://uliza.jp)" operated by the [PIA Corporation](https://corporate.pia.jp/).
Tickets purchased at "[PIA LIVE STREAM](https://t.pia.jp/streaming)" are used for "ULIZA".

log of PIA LIVE STREAM

### Bug

There are two public videos:

- [プレゼンテーションプレイヤーのサンプル](https://ulizaportal.jp/pages/005f18b7-e810-5618-cb82-0987c5755d44): A video for testing functionality (with delicious food).
- [【確認用】視聴サンプルページ（ULIZA）](https://ulizaportal.jp/pages/005e1b23-fe93-5780-19a0-98e917cc4b7d?expires=4102412400&signature=f422a993b683e1068f946caf406d211c17d1ef17da8bef3df4a519502155aa91&version=1): A video for testing device and network.

The first one works well with this plugin; the second one, however, makes VLC player end **immediately**.

M3U8 playlists and logs are as follows:

<details open>
<summary>M3U8 playlists and logs</summary>

The master M3U8 is:

```
#EXTM3U
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=4600000
https://d3b2nkk5uuz3xf.cloudfront.net/5165744460395/20210114121519_7aa7c28d5006907933df1d4a495206decd8dc14bf808af21c757a7dc16aa9999_1.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=2300000
https://d3b2nkk5uuz3xf.cloudfront.net/5165744460395/20210114121519_8877176d61b6d0ea48f3296e2408c660f5370758ce59e89ba4bcd7dcfbea291d_2.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=1200000
https://d3b2nkk5uuz3xf.cloudfront.net/5165744460395/20210114121519_3485c570e7b228104d6620427c38782397b310ed6123ce74cd316aec6da90d0c_3.m3u8

```

The ```BANDWIDTH=1200000``` M3U8 is:

```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-MEDIA-SEQUENCE:0
#EXT-X-ALLOW-CACHE:YES
#EXT-X-TARGETDURATION:11
[...]

PR #6141, opened 2024-08-26
plugins.pialive: new plugin

ChocoLZS:
Add support for the website https://pia-live.jp. This website sells tickets for 5 streaming platforms ( [`ULIZA`](https://uliza.jp/), [`VIMEO`](https://vimeo.com/), [`Smart STREAM`](https://www.smartstream.ne.jp/), [`Theater ZA`](https://za.theater/), [`Live Extreme`](https://www.live-extreme.net/) ) ( https://pia-live.jp/requirement?lang=en ).

image

In this PR, I currently **only add support** for [`ULIZA`](https://uliza.jp/), which has already been implemented in `plugins/piaulizaportal.py` #5508 . Because I don't know how to call another plugin from the current plugin,  part of the logic in that file has been reused.

By the way, I've also found the [`VIMEO`](https://vimeo.com/) plugin in the `plugins/vimeo.py` file, but unfortunately there isn't any url for testing now.

e.g:
- [こながめでたい日２０２４の視聴ページ](https://player.pia-live.jp/stream/4JagFBEIM14s_hK9aXHKf3k3F3bY5eoHFQxu68TC6krUDqGOwN4d61dCWQYOd6CTxl4hjya9dsfEZGsM4uGOUdax60lEI4twsXGXf7crmz8Gk__GhupTrWxA7RFRVt76)  This video will expire on 2024/9/1.

### DIFF
--- a/src/streamlink/plugins/pialive.py
+++ b/src/streamlink/plugins/pialive.py
@@ -67,8 +67,8 @@ class PiaLive(Plugin):
             ),
         )["data"]["movie_one_tag"]
         player_url = validate.Schema(
-                validate.regex(re.compile(r"\s+src=([\"'])(?P<player_url>.*?)\1")),
-            ).validate(player_script_tag)["player_url"]
+            validate.regex(re.compile(r"\s+src=([\"'])(?P<player_url>.*?)\1")),
+        ).validate(player_script_tag)["player_url"]
 
         if not player_url:
             log.error("Player URL not found")
@@ -81,9 +81,7 @@ class PiaLive(Plugin):
         if player_url.startswith(self._ULIZA_URL_PLAYER_DATA):
             m3u8_url = self.session.http.get(
                 player_url,
-                headers={
-                    "Referer": self._URL_BASE,
-                },
+                headers={"Referer": self._URL_BASE},
                 schema=validate.Schema(
                     re.compile(rf"""{re.escape(self._ULIZA_URL_PLAYLIST)}[^"']+"""),
                     validate.none_or_all(
--- a/src/streamlink/plugins/pialive.py
+++ b/src/streamlink/plugins/pialive.py
@@ -21,7 +21,7 @@ log = logging.getLogger(__name__)
 
 
 @pluginmatcher(re.compile(
-    r"https://player\.pia-live\.jp/stream/(?P<video_key>[\w-]+)",
+    r"https?://player\.pia-live\.jp/stream/(?P<video_key>[\w-]+)",
 ))
 class PiaLive(Plugin):
     _URL_BASE = "https://player.pia-live.jp/"
--- a/src/streamlink/plugins/pialive.py
+++ b/src/streamlink/plugins/pialive.py
@@ -1,0 +1,102 @@
+"""
+$description Japanese live-streaming and video hosting platform owned by PIA Corporation.
+$url ulizaportal.jp
+$type live, vod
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/pialive.py
```diff
@@ -0,0 +1,102 @@
...
+                validate.regex(re.compile(r'\s+src=(["\'])(?P<player_url>.*?)\1')),
+            ).validate(player_script_tag)["player_url"]
+
+        if not player_url:
+            log.error("Player URL not found")
+            return None
+
+        return self.handle_embed_player(player_url)
+
+    def handle_embed_player(self, player_url):
```
ChocoLZS, line 79: I guess these five platforms actually return HTML tags through the same logic, consistent with the [test pages](https://pia-live.jp/requirement?lang=en). They are either a `script` or an `iframe`, and they all have a `src` attribute. The actual m3u8 is obtained through the specific URL inside.

pzhlkj6612:
For maintainers:

> This is a very simple extractor for video platform "[ULIZA](https://uliza.jp/)" operated by the [PIA Corporation](https://corporate.pia.jp/).
Tickets purchased at "[PIA LIVE STREAM](https://t.pia.jp/streaming)" are used for "ULIZA".
> 
> *from: #5508*

I must have written some wrong content. There is no proof that ULIZA is being operated by the PIA Corporation. Maybe we need to rename the "piaulizaportal.py" file to "ulizaportal.py" (or even "uliza.py")?

bastimeyer:
Apologies for the late response, but I had to wait until today to give this a review. I won't give this a code-review at the moment though.

----

Please, before opening a pull request, have a look at the [contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md), especially when it's about adding a new plugin. New plugins should always have a plugin request opened first on the issue tracker, with all the necessary details about the website / streaming platform, so this can be discussed before a pull request is opened. All the information which is asked for in the plugin request template/form is missing here.

----

Also, similar to the `piaulizaportal` plugin, this plugin actually violates rule 4:

> 4\. Sites which require paid logins or subscriptions
>
> Plugin implementations and maintenance must be possible by everyone. Paid access prevents this.

In #5508 I ignored that fact, so `piaulizaportal` was merged, because it looked like a video hosting site paid by other companies for their own (private) content, similar to other plugins which were already merged. At the time, the plugin rules were unclear and a bit muddy in regards to paid access to content, so this slipped through. There was also the fact that two freely accessible videos were linked.

To be honest, neither this PR nor #5508 have any good descriptions about the websites. This is all a bit confusing. Please don't expect people from a completely different part of the earth to know video platforms from your local region. There isn't even a Wikipedia page for me to look this up, which is already a bad sign. This needs to be clarified.

Also, what's the point of having a new plugin if this content is (apparently) already accessible via the other plugin?

And in regards to redirecting to another plugin because of embedded remote video players, this violates rule 3 and has only been implemented in some plugins in very rare cases as an exception because of special circumstances:

> 3\. Sites which are primarily rehosting content
>
> Plugins must only implement the original/primary website where the content is hosted or shown.

ChocoLZS:
Thank you for your patience in replying; I have learned a lot from this.

I sincerely apologize for not carefully reviewing the `contribution guidelines`.

As a result, I will close this pull request.

I am deeply sorry for any inconvenience I have caused you.

PR #6176, opened 2024-09-09
cli.main: refactor run()

bastimeyer:
This refactors `streamlink_cli.main.run()`, adds a few missing tests for better code coverage, and refactors existing `streamlink_cli.main` tests. The goal was to remove application logic from `run()`, and to simplify it for a (potential) future rewrite of the CLI module (one step at a time).

- ~The added `handle_url_wrapper()` function currently does not have tests, but the others do. Those missing tests should be added once the `handle_url()` function itself gets refactored and simplified (some day in the future).~

- `--version-check` now finally returns exit code 1 if the version is outdated, in addition to the log message, which should've been fixed ages ago. `--auto-version-check=True` is unaffected by this change (of course).

- Tests which call `streamlink_cli.main.main()` or `streamlink_cli.main.setup()` or `streamlink_cli.main.run()` have now been moved into the `tests/cli/main` package, with shared test fixture setups. Other `streamlink_cli.main` tests have not been touched. The old and (very much) dated `tests/cli/test_cmdline{,_title}` test modules will need to be updated and moved there in the future. They are still based on unittest.TestCase and not pytest, so it doesn't make sense moving them now.

- `handle_url()` tests have been rewritten (and `handle_url_wrapper()` tests been added). `KeyboardInterrupt`s during `output.close()` are also now handled properly.


gravyboat (approved):
This looks really good. The simplification of `run()` makes it a lot easier to understand what is happening and why. Nice job.

PR #6190, opened 2024-09-16
cli.argparser: update _match_argument() override

bastimeyer:
Streamlink overrides `ArgumentParser._match_argument()` in order to be able to parse `--arg -v` arguments. By default `-v` is interpreted as an argument-parser argument, and not as a value. This was fixed in #2333, but the method of the stdlib has received a change after our override was merged. This PR fixes that, fixes the code formatting, and adds missing tests for full code coverage.

- https://github.com/python/cpython/blame/v3.13.0rc2/Lib/argparse.py#L2227-L2247
- https://github.com/python/cpython/commit/be5c79e0338005d675a64ba6e5b137e850d556d1

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -60,13 +60,15 @@ class ArgumentParser(argparse.ArgumentParser):
         else:
             yield f"{prefix}{name}"
 
+    # noinspection PyProtectedMember,PyUnresolvedReferences,PyTypeChecker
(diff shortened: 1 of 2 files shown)


bastimeyer:
https://www.githubstatus.com/

> Update - Actions is experiencing degraded availability. We are continuing to investigate.
> Sep 16, 2024 - 21:37 UTC

gravyboat (approved):

PR #6192, opened 2024-09-18
tests: rewrite cli.output.file tests

bastimeyer:
Continuing with some test rewrites, this PR rewrites the `FileOutput` tests based on pytest. The only `unittest.TestCase`-based tests remaining are now HLS integration tests.

### DIFF
--- a/src/streamlink_cli/output/file.py
+++ b/src/streamlink_cli/output/file.py
@@ -26,7 +26,7 @@ class FileOutput(Output):
     def _open(self):
         if self.filename:
             self.filename.parent.mkdir(parents=True, exist_ok=True)
-            self.fd = open(self.filename, "wb")
+            self.fd = self.filename.open("wb")
 
         if self.record:
             self.record.open()
--- a/tests/cli/output/test_file.py
+++ b/tests/cli/output/test_file.py
@@ -1,83 +1,103 @@
-import os
-import unittest
+from contextlib import contextmanager
+from io import BufferedRandom, BufferedWriter
 from pathlib import Path
-from unittest.mock import Mock, call, patch
+from typing import Iterator
 
 import pytest
 
 from streamlink_cli.output import FileOutput
 
 
-@patch("streamlink_cli.output.file.stdout")
-class TestFileOutput(unittest.TestCase):
-    @staticmethod
-    def subject(filename, fd):
-        fo_record = FileOutput(fd=fd)
-        fo_main = FileOutput(filename=filename, record=fo_record)
+@contextmanager
+def _create_fd(root: Path, name: str) -> Iterator[BufferedRandom]:
+    fd = (root / name).open("w+b")
+    try:
+        yield fd
+    finally:
+        fd.close()
 
-        return fo_main, fo_record
 
-    def test_init(self, mock_stdout: Mock):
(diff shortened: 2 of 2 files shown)


bastimeyer:
One thing to mention though is that `FileOutput.open()` still includes [this line on Windows](https://github.com/streamlink/streamlink/blob/6.10.0/src/streamlink_cli/output/file.py#L34-L35):
> `msvcrt.setmode(self.fd.fileno(), O_BINARY)`

which was initially implemented in ac006c29ea56af11b1fdb9ef57882220affc687e on Jun 22, 2012 during the Livestreamer days.

I really don't think this is needed/relevant (anymore), but I didn't want to remove it just yet. The intention is to flag the output stream as binary. `FileOutput` however already writes to `sys.stdout.buffer`, which is the stdout binary stream. Removing it had no effects during my tests in my W10 VM.

**Was this a W7 or py2x thing? Does anyone know more about this?**

----

https://docs.python.org/3/library/sys.html#sys.stdout

> To write or read binary data from/to the standard streams, use the underlying binary [buffer](https://docs.python.org/3/library/io.html#io.TextIOBase.buffer) object. For example, to write bytes to [stdout](https://docs.python.org/3/library/sys.html#sys.stdout), use `sys.stdout.buffer.write(b'abc')`.

https://docs.python.org/3/library/msvcrt.html#msvcrt.setmode

> `msvcrt.setmode(fd, flags)`
> Set the line-end translation mode for the file descriptor `fd`. To set it to text mode, `flags` should be [`os.O_TEXT`](https://docs.python.org/3/library/os.html#os.O_TEXT); for binary, it should be [`os.O_BINARY`](https://docs.python.org/3/library/os.html#os.O_BINARY).

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setmode?view=msvc-170

> The `_setmode` function sets to `mode` the translation mode of the file given by `fd`. Passing `_O_TEXT` as `mode` sets ANSI text (that is, translated) mode. Carriage return-line feed (CR-LF) combinations are translated into a single line feed character on input. Line feed characters are translated into CR-LF combinations on output. Passing `_O_BINARY` sets binary (untranslated) mode, in which these translations are suppressed.
>
> `_setmode` is typically used to modify the default translation mode of `stdin` and `stdout`, but you can use it on any file. If you apply `_setmode` to the file descriptor for a stream, call `_setmode` before you perform any input or output operations on the stream.

bastimeyer:
Another thing I'll be looking into some time in the future is refactoring the way the `Output` classes are implemented in regards to the `record` attributes.

Currently, additional `FileOutput` objects are added to the main `PlayerOutput` or `FileOutput` objects, and those then forward `open()`, `write()` and `close()` calls to their respective `record` objects, if they exist. This was implemented when the `--record` and `--record-and-pipe` options were added, which is is far from ideal.

The CLI module should handle **a list of `Output` objects** and write to them directly in the `StreamRunner`. This would make `--record-and-pipe` superfluous, as it's basically the same logic as `--record=file --stdout`. Maybe `--record` could be removed as well in favor of simply having `--player=... --output=...`. That would be a breaking change though, because currently, `--output=...` suppresses the `PlayerOutput`. In order to prevent the player from being launched, `--player` would have to be set to an empty value.

Just thinking here... This can all be figured out later.

PR #6208, opened 2024-09-29
Rewrite asyncio-based `utils.processoutput` code based on trio and remove `pytest-asyncio` dev-requirement

bastimeyer:
The `utils.processoutput` utility is currently used by `FFMPEGMuxer` to validate the ffmpeg binary by reading its version string and compile flags.

This was previously done using `asyncio` from the stdlib, hence the requirement of `pytest-asyncio` in the tests. The only other tests making use of the `asyncio` run loop were the CLI's `StreamRunner` tests (just for convenience though) and the `handshake` test utilities for making assertions about thread synchronization.

Since Streamlink's webbrowser API is based on `trio`, it doesn't make sense mixing async run loops from two different libs. On Windows, this also raises a warning when running Streamlink's testsuite.

The rewritten `utils.processoutput` module and its tests are now much cleaner and easier to read due to trio's structured concurrency. No other changes were made here apart from gracefully terminating the spawned subprocess now, instead of just killing it when all tasks were completed (in case it was still running).

A nice side effect of the `pytest-asyncio` removal is that it reduces the run time of the whole testsuite on my Arch Linux host system from 9s to 7s, and in my W10 VM from 35s to 29s.

I will have another look at the diff later before merging, in case I missed something.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,7 +5,6 @@ versioningit >=2.0.0,<4  # required for being able to run editable installs
 
 # tests
 pytest >=8.0.0
-pytest-asyncio >=0.23.5
 pytest-trio
 requests-mock
 freezegun >=1.0.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -132,9 +132,6 @@ testpaths = [
 filterwarnings = [
   "always",
 ]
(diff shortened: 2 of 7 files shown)


bastimeyer:
Looks like there's a race-condition when running the rewritten tests on Windows. Will fix later...

PR #6226, opened 2024-10-07
cli.argparser: re-organize argument groups

bastimeyer:
This moves `--version-check` and `--auto-version-check` right next to `--version`,
and it adds the following argument groups and split them from the "General arguments" group  
- Logging arguments
  - `--loglevel`
  - `--logformat`
  - `--logdateformat`
  - `--logfile`
  - `--quiet`
  - `--json`
- Network arguments
  - `--interface`
  - `--ipv4`
  - `--ipv6`


bastimeyer:
Should `--json` actually be part of the logging group? Any opinions on this?

mkbloke:
> Should --json actually be part of the logging group? Any opinions on this?

Seems like as good a place as any to me.

PR #6233, opened 2024-10-09
cli: remove now unused DeprecatedPath

bastimeyer:
For some reason, `LOG_DIR` was set as a `DeprecatedPath` on macOS. This didn't have any effect, but it's weird that it was set here...

### DIFF
--- a/src/streamlink_cli/compat.py
+++ b/src/streamlink_cli/compat.py
@@ -1,7 +1,6 @@
 import sys
 from os import devnull
-from pathlib import Path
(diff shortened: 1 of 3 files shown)


bastimeyer:
Accidental change here:
https://github.com/streamlink/streamlink/commit/79a4232e02e7d6eca46df5e164064663abf14ced#diff-3f9b9219afe6191925ba344252b5b589436134deb8ffd6e97236e9a322e2b51fR51

Issue #6243, opened 2024-10-13
streamlink==6.11.0 requires trio>=0.23 / py313: requires trio>=0.25


yashikada:
### Checklist

- [x] This is a bug report and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [x] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [x] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [x] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Collaboration

- [x] [I will provide feedback should a pull request be opened with a fix for the problem](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-request-feedback)

### Streamlink version

[cli][info] Your Streamlink version (6.11.0) is up to date!

### Description

I'm using the latest stable version installed via pip3 install on Linux Debian 12.
On my workstation I installed the packages distributed by Debian

```
ii python3-trio 0.22.0-0.2 all Python async concurrency and I/O library
ii python3-trio-websocket 0.9.2-2 all Server and client Python library of the WebSocket protocol (Python3 version)
```

Until version 6.10.0 I had no problems but with version 6.11.0 there are problems with the trio version.

If I install streamlink with pip package trio is already satisfied, but it's wrong, it should be a more updated version.
While I'm opening the issue the latest version of trio 0.26.2 has no issues

```
Collecting streamlink
  Using cached streamlink-6.11.0-py3-none-any.whl (530 kB)
Requirement already satisfied: certifi in /usr/lib/python3/dist-packages (from streamlink) (2022.9.24)
Requirement already satisfied: isodate in /usr/lib/python3/dist-packages (from streamlink) (0.6.1)
Requirement already satisfied: lxml<6,>=4.6.4 in /usr/lib/python3/dist-packages (from streamlink) (4.9.2)
Requirement already satisfied: pycountry in /usr/lib/python3/dist-packages (from streamlink) (22.3.5)
Collecting pycryptodome<4,>=3.4.3
  Using cached pycryptodome-3.21.0-cp36-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (2.3 MB)
[...]

PR #6244, opened 2024-10-13
build: bump trio dependency to >=0.25,<1

bastimeyer:
See https://github.com/streamlink/streamlink/issues/6243#issuecomment-2408988087 (second issue mentioned)

There appear to be runtime issues with `trio<0.25` on Python 3.13. It's not the case on Python 3.12 and below, but since we now officially support 3.13, we'll have to bump the version.

Alternatively, we could add environment markers to the dependencies list, so that we could remain on `trio>=0.22,<1` on Python 3.12 and below, but that can cause issues with pip's dependency resolver in certain situations, which I hate. On the other hand, `trio-websocket` still hasn't published a new release with their compatibility fixes of `trio>=0.25`, so making the `trio>=0.25,<1` requirement now would probably introduce more packaging issues for some systems.

I'll have to think about it...

----

```
$ pip install trio==0.24
$ streamlink
  File "/home/basti/venv/streamlink-313/bin/streamlink", line 5, in <module>
    from streamlink_cli.main import main
  File "/home/basti/repos/streamlink/src/streamlink_cli/main.py", line 24, in <module>
    from streamlink.stream.stream import Stream, StreamIO
  File "/home/basti/repos/streamlink/src/streamlink/stream/__init__.py", line 1, in <module>
    from streamlink.stream.dash import DASHStream
  File "/home/basti/repos/streamlink/src/streamlink/stream/dash/__init__.py", line 1, in <module>
    from streamlink.stream.dash.dash import DASHStream, DASHStreamReader, DASHStreamWorker, DASHStreamWriter
  File "/home/basti/repos/streamlink/src/streamlink/stream/dash/dash.py", line 20, in <module>
    from streamlink.stream.ffmpegmux import FFMPEGMuxer
  File "/home/basti/repos/streamlink/src/streamlink/stream/ffmpegmux.py", line 19, in <module>
    from streamlink.utils.processoutput import ProcessOutput
  File "/home/basti/repos/streamlink/src/streamlink/utils/processoutput.py", line 10, in <module>
    import trio
  File "/home/basti/venv/streamlink-313/lib/python3.13/site-packages/trio/__init__.py", line 75, in <module>
    from ._path import Path as Path
  File "/home/basti/venv/streamlink-313/lib/python3.13/site-packages/trio/_path.py", line 203, in <module>
    class Path(metaclass=AsyncAutoWrapperType):
    ...<252 lines>...
                async def with_segments(self, *pathsegments: StrPath) -> Path: ...
  File "/home/basti/venv/streamlink-313/lib/python3.13/site-packages/trio/_path.py", line 147, in __init__
    type(cls).generate_forwards(cls, attrs)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/home/basti/venv/streamlink-313/lib/python3.13/site-packages/trio/_path.py", line 164, in generate_forwards
    raise TypeError(attr_name, type(attr))
TypeError: ('parser', <class 'module'>)
```

```
$ pip install trio==0.25
$ pytest >/dev/null && echo OK
OK
```


bastimeyer:
I decided to add environment markers for py>=313 and py<313 instead of bumping trio to >=0.25 on all Python versions.

The reason for that is that trio>=0.25 still hasn't been packaged on many Linux distros where Streamlink is packaged. Considering that those distros also (obviously) haven't been updated to Python 3.13, only bumping trio to >=0.25 on py313+ is fine.

The environment markers don't cause any issues when building Streamlink's Windows builds / Linux AppImages, so it's fine. But that's only because we set a lower version boundary, not an upper one. Pip still ignores environment markers when resolving packages on foreign platforms (e.g. building for Windows on Linux).

PR #6260, opened 2024-10-21
chore: reformat everything using ruff format

bastimeyer:
This reformats the entire codebase using `ruff format`, which is almost identical to `black`'s formatting:
https://docs.astral.sh/ruff/formatter/

For now, I've split the reformat changes into multiple commits, but I will squash them before merging, so the one giant commit can be added to `.git-blame-ignore-revs` afterwards. Having multiple commits is easier to review, because the diff shortstat is currently this: `257 files changed, 9880 insertions(+), 7909 deletions(-)`

Instead of just running `ruff format`, the reformatting was done by using PyCharm with ruff's lsp set up, so there's probably some additional PyCharm reformatting logic included, not sure, but doesn't matter.

I've also had to skip the formatting in lots of places (`# fmt:skip` annotation after a whole statement), because tooling is usually not smart and the formatting often was chosen deliberately in order to highlight certain circumstances or to make it more legible.

Most of the diff is only indentation related though. For example, almost all of the pytest parametrization didn't align its parameters according to the expected style, in order to prevent one indentation level. That's all changed and trivial to review.

Another huge part of the diff are plugins of course. I tried to keep a common style of the pluginmatchers and their non-verbose regexes, but the annoying things here are HTTP requests with validation schemas, which almost all of them required a reformatting due to the parameter style of function calls with multi-line parameter values. Most of that is of course also only indentation, with new trailing commas added.

Apart from that, I believe I also changed/rearranged a few minor things here and there. Can't remember.

----

This PR also adds `ruff format --diff` to the linting CI workflow, so using the correct coding style is now mandatory. The docs were updated in this regard.

One thing that needs to be mentioned though is that there are two *linting* rules which we use which are incompatible with the *formatter's* **auto-fix logic**, namely
- ["missing-trailing-comma" `COM812`](https://docs.astral.sh/ruff/rules/missing-trailing-comma/)
- ["single-line-implicit-string-concatenation" `ISC001`](https://docs.astral.sh/ruff/rules/single-line-implicit-string-concatenation/)

This doesn't affect its check logic (`--check` / `--diff`) and triggers errors in the linter, which is therefore fine.

Another thing worth mentioning is that the formatter uses ruff's "preview" rules. Those will only be changed in their minor releases though. And since we're pinning the ruff version, there's nothing to worry anyway.

----

I will go through the diff tomorrow again, in case I made some mistakes, and then merge.


bastimeyer:
Btw, GitHub's diff view is terrible for these kinds of changes where mainly whitespace was changed. Other tools have much better diff highlighting.

gravyboat:
I'm about 50% through reviewing this locally. I'll try to get the rest done tonight or tomorrow.

gravyboat (approved):
Alright I'm going to mark this as approved. As you noted please do your secondary check just in case I also missed something. I don't see anything problematic currently however. Nice job.

bastimeyer:
Thanks for taking the time having a second pair of eyes on the changes...
I force pushed a few more cleanups and squashed.

Changes should be fine now (for the most part). Anything else can be changed/fixed later.

Going to merge now.

PR #6264, opened 2024-10-24
cli: minor argparse improvements

bastimeyer:


bastimeyer:
I decided against updating the CLI's argparser and setting the `exit_on_error=False` parameter which was added in py39, at least for now, because it would require updating the error message when catching `ArgumentError` during argument parsing, and it would basically just re-implement the stdlib logic, just so that we're able to exit the program in `streamlink_cli.main.main()` instead of `argparse.ArgumentParser.error()`.

PR #6266, opened 2024-10-26
docs: rewrite plugin sideloading page

bastimeyer:
- https://streamlink--6266.org.readthedocs.build/cli/plugin-sideloading.html
- https://streamlink--6266.org.readthedocs.build/cli/config.html

### DIFF
--- a/docs/cli/plugin-sideloading.rst
+++ b/docs/cli/plugin-sideloading.rst
@@ -1,7 +1,48 @@
 Plugin sideloading
 ==================
 
-Streamlink will attempt to load standalone plugins from these directories:
+Streamlink supports overriding its :ref:`built-in plugins <plugins:Plugins>` or loading custom third party plugins
+without having to modify its sources or its built and installed `Python distribution`_. This is called plugin sideloading.
+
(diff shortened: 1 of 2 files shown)


docs/cli/plugin-sideloading.rst
```diff
@@ -1,54 +1,69 @@
 Plugin sideloading
 ==================
 
-Streamlink will attempt to load standalone plugins from these directories:
+Streamlink supports overriding its :ref:`built-in plugins <plugins:Plugins>` or loading custom third party plugins
+without having to modify its sources or its built and installed `Python distribution`_. This is called plugin sideloading.
 
-.. rst-class:: table-custom-layout table-custom-layout-platform-locations
+Those additional plugin modules will automatically be loaded `from the paths listed down below <Sideloading locations_>`_,
```
mkbloke, line 7: I would suggest removing 'down' here, as it's redundant.

mkbloke (approved):
Other than one very minor thing, it looks good to me.

bastimeyer:
Thanks. Updated...

Btw, I was thinking about moving the page from the "CLI" section, but that would break existing links, so it's probably not worth it.

PR #6270, opened 2024-10-29
ci.github: update fetch tags commands

bastimeyer:
See https://github.com/streamlink/streamlink/pull/6269#issue-2621679025

The command currently used fetches all tags, despite the `--depth=300` parameter:
https://github.com/streamlink/streamlink/actions/runs/11579714247/job/32236578782

PR #6289, opened 2024-11-11
ci: update git fetch commands

bastimeyer:
https://github.com/streamlink/streamlink/pull/6270#issuecomment-2468511007

This adds a newline character to the output of awk. Not sure if this will fix the issue. Let's see. Otherwise we'll have to disable the `pipefail` option or explicitly catch the 141 error code and ignore it.


bastimeyer:
Still failing
https://github.com/streamlink/streamlink/actions/runs/11847912682/job/33018510274

PR #6298, opened 2024-11-18
options: make Arguments inherit from dict

bastimeyer:
Instead of using an internal `arguments` dict attribute, make `Arguments` inherit from `dict[str, Argument]` directly, so it's shown as a dict type in the rendered docs.

Follow-up of #6297


bastimeyer:
Before:
> arguments: ClassVar[Arguments] = <streamlink.options.Arguments object>

After:
> arguments: ClassVar[Arguments] = {}

PR #6335, opened 2024-12-12
cli: improve --logformat/--logdateformat help text

refack:
Small phrasing change and explicit mentioning of the two possible defaults

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -353,7 +353,13 @@ def build_parser():
             See the Python standard library's `logging.Formatter` docs for more information about the logging format
             and the available `LogRecord` attributes. Streamlink's formatter uses the curly brace style.
 
-            The default format depends on the chosen log level (may include the `asctime` attribute).
+            The default format depends on the chosen log level.
+
+            For verbose levels (`trace` and `all`):
+
+            Default is "[{asctime}][{name}][{levelname}] {message}".
+
+            Otherwise:
 
             Default is "[{name}][{levelname}] {message}".
         """,
@@ -366,7 +372,13 @@ def build_parser():
 
             This formats the `LogRecord`'s `asctime` attribute via `strftime()`.
 
-            The default date format depends on the chosen log level (may include fractions).
+            The default date format depends on the chosen log level.
+
+            For verbose levels (`trace` and `all`):
+
+            Default is "%%H:%%M:%%S.%%f".
+
+            Otherwise:
 
             Default is "%%H:%%M:%%S".
         """,


bastimeyer (commented):
Thanks for the PR.

The CLI argument help texts require special care, because they are used in the rendered HTML docs, the man page, and also in the `--help` text output. While the `--help` output is simple plain text, the HTML docs and man page both use reST syntax, hence why the help text strings are transformed by Streamlink's custom `ext_argparse` Sphinx extension:
https://github.com/streamlink/streamlink/blame/ac030af5e0fb5b5ec20b8efdaca665b557da4209/docs/ext_argparse.py#L65

Your changes currently don't work because of the line breaks in reST, and they also break the default value highlighting via the `Default is XYZ.` line:
https://streamlink--6335.org.readthedocs.build/cli.html#logging-arguments

```
$ man ./docs/_build/man/streamlink.1 | grep -FA16 -e --logformat
       --logformat FORMAT
              Set a custom logging format.

              See  the  Python  standard  library's  logging.Formatter docs for more information about the logging format and the available Lo‐
              gRecord attributes. Streamlink's formatter uses the curly brace style.

              The default format depends on the  chosen  log  level:  "[{name}][{levelname}]  {message}"  for  the  lower  levels,  or  "[{asc‐
              time}][{name}][{levelname}] {message}" for verbose levels ("trace", "all").

       --logdateformat DATEFORMAT
              Set a custom logging date format.

              This formats the LogRecord's asctime attribute via strftime().

              The  default  date  format  depends on the chosen log level: "%H:%M:%S" for the lower levels, or "%H:%M:%S.%f" for verbose levels
              ("trace", "all").
```

refack:
> use reST syntax

Got it, on it.

bastimeyer:
I guess this PR has been abandoned?!

gravyboat:
> I guess this PR has been abandoned?!

We should assume as much. It's been a month for a 6 line change.

refack:
I just lost focus... ADHD + old age lul...
I've actually been elbows deep in the codebase as my method of grokking is refactoring.
I had a few Ideas I'll maybe bring forward
o7

refack:
Is this acceptable?

image

I don't 'love' the excessive line spacing, but it's the best way I found to get everything lined up

bastimeyer (changes requested):
Help texts must not contain any markup (reST), as explained. While the current changes may also work on the man page, the `--help` output looks bad due to the usage of `**`, so it must be avoided.

Btw, screenshots of the rendered HTML docs are not necessary, because we have docs preview builds (see the CI statuses).
https://streamlink--6335.org.readthedocs.build/cli.html#logging-arguments

Please also don't introduce new help text format styles and stick to the format used in the help text of other CLI args. [The `Default is (.+)\.\n` format](https://github.com/streamlink/streamlink/blame/90fc82b2025f3e660732d0f35905dabd2faf89ca/docs/ext_argparse.py#L24) isn't ideal - I'm aware - but if we want to change it, we should do it everywhere all at once.

Also, the "verbose" log level is not explained anywhere, so being specific about the actual log level makes more sense. This can be highlighted using backticks, which we allow in the help texts, as it's similar to regular quotation marks.

```diff
diff --git a/src/streamlink_cli/argparser.py b/src/streamlink_cli/argparser.py
index ddb8c0a7..876adbb8 100644
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -353,11 +353,15 @@ def build_parser():
             See the Python standard library's `logging.Formatter` docs for more information about the logging format
             and the available `LogRecord` attributes. Streamlink's formatter uses the curly brace style.
 
-            The default format depends on the chosen log level:
+            The default format depends on the chosen log level.
 
-            When using the default log level, the default is **"[{name}][{levelname}] {message}"**.
+            When using a log level above `trace`:
 
-            When using the verbose log level, the default is **"[{asctime}][{name}][{levelname}] {message}"**.
+            Default is "[{name}][{levelname}] {message}".
+
+            Otherwise:
+
+            Default is "[{asctime}][{name}][{levelname}] {message}".
         """,
     )
     logging.add_argument(
@@ -370,9 +374,13 @@ def build_parser():
 
             The default date format depends on the chosen log level.
 
-            When using the default log level, the default is **"%%H:%%M:%%S"**.
+            When using a log level above `trace`:
+
+            Default is "%%H:%%M:%%S".
+
+            Otherwise:
 
-            When using the verbose log level, the default is **"%%H:%%M:%%S.%%f"**.
+            Default is "%%H:%%M:%%S.%%f".
         """,
     )
     logging.add_argument(
```

refack:
> Help texts must not contain any markup

Ahh, I didn't grok this. Will fix.


> Please also don't introduce new help text format

Ok.


> Also, the "verbose" log level is not explained anywhere so being specific about the actual log level makes more sense

👍

refack:
> 
> ```diff
> diff --git a/src/streamlink_cli/argparser.py b/src/streamlink_cli/argparser.py
> --- a/src/streamlink_cli/argparser.py
> +++ b/src/streamlink_cli/argparser.py
> -            When using a log level above `trace`:
> +            For verbose levels (`trace` and `all`):
> ...
> +            Otherwise:
> ```
[^1]

I went one step more explicit, but in this case flipping the order and finishing with the default-deafult made more sense.



[^1]: Just in case you were not aware, the "Allow edits by maintainers" checkbox gives you write permission on the branch I used for the PR, and since you went to all the trouble of preparing a patch, it is ok by me if you pushed it as a commit and I could have confirm it  ¯\_(ツ)_/¯
image

bastimeyer (approved):
Thanks, looking good now.

Yes... I'm very much aware of how PR branches work. I only (force-)push onto PR branches of other users if I'm certain that the PR author would need further help (code- or git-related), or if I think that my corrections will lead to a faster merge. This one here is so trivial that fetching the PR branch, editing and running a quick `git diff | wl-copy` is more than enough, especially if my intention is to make a suggestion rather than a final decision.

refack:
Cool.
This PR was a very nice experience, you are very nice to work with. Hope to collaborate again soon.

PR #6347, opened 2024-12-15
stream.hls: refactor playlist reload

### LINKED ISSUE #6340: stream.hls: refactor polling-related logic
Move logic related to polling new data (live playlists) into a new intermediate base class, so it can be re-used in other segmented stream types, like DASH for example.

The `HLSStreamWorker` currently implements its own

- playlist reload timings logic
- logic for stopping the stream early if no new segments were queued until a certain deadline has passed
- checks for gaps in the segment queue between reloads

Since the `DASHStreamWorker` is in need for a full refactor, rewriting its reload timing logic separately in addition to all the `HLSStreamWorker` logic doesn't make sense. Having another shared base class, namely `PollingSegmentedStreamWorker` where reload timings and other stuff can be implemented makes much more sense.

One thing that needs to be kept in mind are session options and CLI arguments for configuring specific values, as those will have to be renamed and updated as well with a deprecation phase.

bastimeyer:
Resolves #6340

Opening this as a draft for now, as it's WIP and unfinished. This simply moves stuff into a new intermediate abstract base class, `PollingSegmentedStreamWorker`, and it simplifies attribute and method names.

I currently don't know if this requires further changes, as I haven't had a look at the `DASHStreamWorker` yet, which is supposed to utilize the shared code here in the future.

Another thing that's missing, but could be changed later, is renaming the `hls-segment-queue-threshold` session option and CLI argument, as it's now part of the shared base class. A good opportunity for choosing a better name, e.g. `segmented-queue-deadline` or something like that.

### DIFF
--- a/src/streamlink/plugins/twitch.py
+++ b/src/streamlink/plugins/twitch.py
@@ -168,11 +168,11 @@ class TwitchHLSStreamWorker(HLSStreamWorker):
         self.logged_ads: deque[str] = deque(maxlen=10)
         super().__init__(reader, *args, **kwargs)
 
-    def _playlist_reload_time(self, playlist: TwitchM3U8):  # type: ignore[override]
+    def _get_reload_time(self, playlist: TwitchM3U8):  # type: ignore[override]
         if self.stream.low_latency and playlist.segments:
             return playlist.segments[-1].duration
 
-        return super()._playlist_reload_time(playlist)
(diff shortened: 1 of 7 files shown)


bastimeyer:
Closing this, as those changes have been / will be implemented differently in other multiple PRs.

PR #6425, opened 2025-02-06
ci.github: add preview-build workflow

### LINKED ISSUE #6424: Replace nightly builds of Windows installer+portable and Linux AppImage
I'm going to have a look at replacing the automated nightly builds of Streamlink's Windows installer and portable builds, as well as Streamlink's Linux AppImage. This will probably happen this weekend or so.

The goal is to replace timed/scheduled builds with event-based builds (GH actions `workflow_dispatch`) that are triggered from commits to `master` here when something in `pyproject.toml`, `setup.py` or `src/**` changed, meaning that only relevant code-changes will be built. This will remove lots of unnecessary duplicates if nothing was changed, and users also won't have to wait anymore until the next build is scheduled if something was just merged.

The new `preview-build` workflow could also optionally be extended to allow (maintainer-requested) builds from PRs. Those however could also be triggered manually on the `streamlink/windows-builds` and `streamlink/streamlink-appimage` repos by setting the input `ref` value to `refs/pull/$ID/head` (workflows don't exist yet).

- https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch
- https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event
- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

bastimeyer:
Closes #6424 

This new `preview-build` workflow triggers the `preview-build` workflows on the `streamlink/windows-builds` and `streamlink/streamlink-appimage` repos whenever something was pushed to `master` and either `pyproject.toml`, `setup.py`, `MANIFEST.in` or `src/**` was modified:
- https://github.com/streamlink/windows-builds/actions/workflows/preview-build.yml
- https://github.com/streamlink/streamlink-appimage/actions/workflows/preview-build.yml

These two links are added to the run's summary markdown text, so users can find the build artifacts easily when clicking the workflow here. Unfortunately, direct links in the summary can't be included, as the workflow here merely triggers the workflows on the other repos. GitHub doesn't include a workflow run ID in its API response.

The docs have also been updated accordingly. I will remove the nightly workflows from the repos after this PR here was merged. The recent runs and their build artifacts should remain for 90 days (until expiration) when deleting workflow files.

I've checked and triggered the workflows using a GH PAT, but I don't know if `secrets.GITHUB_TOKEN` with `permissions.actions=write` also has the permissions to trigger a workflow on other repos. If it doesn't, then I will have to add a PAT for the @streamlinkbot and use that instead.

Does the `on.push` config here look fine? Not 100% sure... Tags pushed should be ignored and the paths should be correct.
- https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions
- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -1,0 +1,46 @@
+name: Preview build
+run-name: "Preview build - ${{ github.ref }}"
+
+on:
+  push:
+    branches:
+      - "master"
+    tags-ignore:
+      - "**"
+    paths:
+      - "pyproject.toml"
+      - "setup.py"
+      - "MANIFEST.in"
+      - "build_backend/**"
+      - "src/**"
+
+jobs:
+  trigger:
+    name: "Preview build - ${{ github.ref }}"
+    if: github.repository == 'streamlink/streamlink'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Windows builds
+        run: |
+          curl -sfL \
+            -X POST \
+            -H 'Accept: application/vnd.github+json' \
+            -H 'X-GitHub-Api-Version: 2022-11-28' \
+            -H 'Authorization: Bearer ${{ secrets.STREAMLINKBOT_PREVIEW_BUILD }}' \
+            -d '{"ref":"master","inputs":{"ref":"${{ github.ref }}"}}' \
+            https://api.github.com/repos/streamlink/windows-builds/actions/workflows/preview-build.yml/dispatches
+
+          echo -e '## Windows builds\n\nhttps://github.com/streamlink/windows-builds/actions/workflows/preview-build.yml\n' \
+            >> "${GITHUB_STEP_SUMMARY}"
+      - name: Linux AppImage
+        run: |
+          curl -sfL \
+            -X POST \
+            -H 'Accept: application/vnd.github+json' \
+            -H 'X-GitHub-Api-Version: 2022-11-28' \
+            -H 'Authorization: Bearer ${{ secrets.STREAMLINKBOT_PREVIEW_BUILD }}' \
+            -d '{"ref":"master","inputs":{"ref":"${{ github.ref }}"}}' \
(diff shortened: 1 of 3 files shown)


gravyboat:
> Does the on.push config here look fine?

Does that syntax work? I thought you still had to provide it a list of object for both the branch and the tags so something like:

```yml
on:
  push:
    branches:
      - master
    tags-ignore:
      - '**'
    paths:
      - "pyproject.toml"
      - "setup.py"
      - "MANIFEST.in"
      - "build_backend/**"
      - "src/**"
```

per the documentation here: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-including-branches-and-tags

bastimeyer:
Thanks. True. No idea what I was thinking there...

I also temporarily added a PR trigger to test the API permissions, and it looks like the permissions won't work (403 - "Resource not accessible by integration").
https://github.com/streamlink/streamlink/actions/runs/13197133321/job/36840773490?pr=6425#step:3:19

Going to have to use a PAT

bastimeyer:
PAT + write permissions added for @streamlinkbot (copy sent to streamlink@protonmail.com)

Tested by running curl locally and building AppImages from this PR branch (we're only triggering workflows from the master branch however):
https://github.com/streamlink/streamlink-appimage/actions/runs/13197448703

Going to merge this now. Should be fine...

Afterwards I'm going to delete the nightly build workflows from both repos. As said, old runs with their build artifacts should be kept until they expire.

Btw, there's a known issue with GH's new public ubuntu-24.04-arm runners and their docker service, which sometimes fails to initialize and runners will sometimes fail randomly.

PR #6441, opened 2025-02-19
plugins.nicolive: set stream cookies from WS data

bastimeyer:
Fixes #6440 

@cykac04 please have a look at this PR...
https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-request-feedback

After browsing lots of streams, I was finally able to find one live stream where cookies were required. The one you posted was offline unfortunately.

HLS streams where cookies are required all appear to be multivariant streams with an external audio stream, so muxing via FFmpeg is required. I'm not sure, but I think there's an issue with the audio stream synchronization. That is unrelated to the plugin however, as it has nothing to do with the HLS protocol or the stream data. It may be an issue with the stream itself, because audio synchronization depends on the individual bitstreams and their presentation timestamps, not on the HLS metadata (this is not a packed audio stream where this would be the case). It might be necessary to set `--ffmpeg-copyts` in order to fix audio sync issues, which could also be forced by the plugin code.

Audio sync issues can happen occasionally - they don't have to appear all the time, depending on the initial sub-stream access timing. **So I'd like to ask you to check and see if there are audio sync issues on the streams you're trying to watch without the `--ffmpeg-copyts` parameter being set and whether setting the `--ffmpeg-copyts` parameter fixes them reliably 100% of the time.**

Btw, the fact that they're using a newer version of the HLS protocol (despite not annotating it correctly) which is still not yet released/finalized is irrelevant, even if Streamlink doesn't support it yet, as everything's backwards compatible.

```
$ streamlink -l debug https://live.nicovideo.jp/watch/lv347077742 best
[cli][debug] OS:         Linux-6.13.3-1-git-x86_64-with-glibc2.41
[cli][debug] Python:     3.13.2
[cli][debug] OpenSSL:    OpenSSL 3.4.1 11 Feb 2025
[cli][debug] Streamlink: 7.1.3+3.g50de81b0
[cli][debug] Dependencies:
[cli][debug]  certifi: 2024.12.14
[cli][debug]  isodate: 0.7.2
[cli][debug]  lxml: 5.3.0
[cli][debug]  pycountry: 24.6.1
[cli][debug]  pycryptodome: 3.21.0
[cli][debug]  PySocks: 1.7.1
[cli][debug]  requests: 2.32.3
[cli][debug]  trio: 0.27.0
[cli][debug]  trio-websocket: 0.11.1
[cli][debug]  urllib3: 2.3.0
[cli][debug]  websocket-client: 1.8.0
[cli][debug] Arguments:
[...]

PR #6445, opened 2025-02-20
stream.hls: reset key state on discontinuity

bastimeyer:
Don't try to decrypt segments when a discontinuity occurs after a sequence of encrypted segments

----

See https://github.com/streamlink/streamlink/pull/6441#issuecomment-2672176738

The added test fails without the fix.

### DIFF
--- a/src/streamlink/stream/hls/m3u8.py
+++ b/src/streamlink/stream/hls/m3u8.py
@@ -324,6 +324,7 @@ class M3U8Parser(Generic[TM3U8_co, THLSSegment_co, THLSPlaylist_co], metaclass=M
         """
         self._discontinuity = True
         self._map = None
+        self._key = None
 
     @parse_tag("EXT-X-KEY")
     def parse_tag_ext_x_key(self, value: str) -> None:
--- a/tests/stream/hls/test_hls.py
+++ b/tests/stream/hls/test_hls.py
@@ -936,6 +936,38 @@ class TestHLSStreamEncrypted(TestMixinStreamHLS, unittest.TestCase):
         assert data == self.content([segments[1]], prop="content_plain")
         assert mock_log.error.mock_calls == [call("Error while decrypting segment 0: PKCS#7 padding is incorrect.")]
 
+    def test_hls_encrypted_aes128_discontinuity(self):
+        aesKey1, aesIv1, key1 = self.gen_key()
+        aesKey2, aesIv2, key2 = self.gen_key()
+        discontinuity = Tag("EXT-X-DISCONTINUITY")
+
+        segments = self.subject([
+            Playlist(
+                0,
+                [
+                    key1,
+                    SegmentEnc(0, aesKey1, aesIv1),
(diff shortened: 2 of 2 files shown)


bastimeyer:
I'm a bit hesitant with merging this yet, because [RFC 8216](https://datatracker.ietf.org/doc/html/rfc8216) doesn't explicitly mention key resets on discontinuities. Discontinuities merely describe that the stream data between two segments is incoherent, hinting at the parser and decoder to reset their states. It *may* be possible that both of the two different segments use the same encryption, but I don't think this would make much sense.

https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.3

> The EXT-X-DISCONTINUITY tag MUST be present if there is a change in any of the following characteristics:
> -  file format
> - number, type, and identifiers of tracks
> - timestamp sequence
>
> The EXT-X-DISCONTINUITY tag SHOULD be present if there is a change in any of the following characteristics:
> - encoding parameters
> - encoding sequence

https://datatracker.ietf.org/doc/html/rfc8216#section-6.3.3

> The client MUST be prepared to reset its **parser(s)** and decoder(s) before playing a Media Segment that has an EXT-X-DISCONTINUITY tag applied to it; otherwise, playback errors can occur.

bastimeyer:
Also no mention of this in the most recent HLS2 draft
https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-17

gravyboat:
I'm against doing this if it isn't the standard established in the RFC. It would be an unexpected behavior and this seems like a problem with how the site is implementing their streaming service rather than an error we should address. Nicolive has been nothing but a headache for years due to their sloppy work so I don't see why we should continue to support it.

bastimeyer:
My motivation for this change is not fixing any plugins. This is about the HLS spec, nothing else. If this interpretation is wrong, then I'm going to close this, simple as that.

Whether it's correct however, I'm not so sure anymore, because I've checked every relevant part of the HLS spec, and it looks like that the discontinuity tag should not reset the key state.

https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.4

> Media Segments MAY be encrypted. The EXT-X-KEY tag specifies how to decrypt them. **It applies to every Media Segment** and to every Media Initialization Section declared by an EXT-X-MAP tag **that appears between it and the next EXT-X-KEY tag** in the Playlist file with the same KEYFORMAT attribute (**or the end of the Playlist file**).

Discontinuity tags are usually used in live playlists to switch between the regular content and other content like ads. If the server encrypts segments of the regular content and then switches to ads and those are not encrypted (e.g. because encryption happens on the live stream's ingestion point whereas ads are read from offline storage), then it should notify the client about that and set the encryption METHOD to NONE. Switching between the AES-128 and NONE methods is handled properly by Streamlink (we don't have tests for that, but it's trivial to add one).

So I'm going to close this now, as I don't trust these changes (unless proven otherwise). If this is a problem for certain sites and plugins, then these need their own HLS parser subclasses where those quirks can be applied.

PR #6451, opened 2025-02-28
build: add decompress extras marker

bastimeyer:
Instead of defining brotli and zstandard directly, rely on urllib3's version requirements and use its brotli and zstd extras markers.

brotli or zstandard may be required by some websites which ignore the HTTP client's decompression capabilities set in the Accept-Encoding HTTP header. Both dependencies are picked up automatically if available.

----

https://github.com/urllib3/urllib3/blob/2.3.0/pyproject.toml#L41-L48

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,7 +55,7 @@ jobs:
       - name: Install Python dependencies
         run: >
           python -m pip install -U
-          -e .
+          -e .[decompress]
           -r dev-requirements.txt
         continue-on-error: ${{ matrix.continue || false }}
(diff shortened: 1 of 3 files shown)


bastimeyer:
Before I'm going to merge this, keep the "decompress" name, or rename to something else?

gravyboat (approved):
> Before I'm going to merge this, keep the "decompress" name, or rename to something else?

I'm fine with decompress. I can't think of a better word that easily explains what is happening based on your notes here and in #6452. If it becomes a problem we can always rename, no reason to over complicate it now though.

Issue #6449, opened 2025-02-26
cli: write download progress to the same stream as the logger/consoleoutput, keep sticky status line at the bottom, make other messages flow above


bastimeyer:
The CLI's download progress status is currently always written to `stderr`. This was already the case when it was implemented first during the Livestreamer era:
- https://github.com/streamlink/streamlink/commit/c389337237f280370a3d5635c9e8e393453c8029

And it was kept this way after the Streamlink fork and when it was refactored multiple times:
- https://github.com/streamlink/streamlink/commit/424aac28f05806d8e54413956a95f4bb957acb49
- https://github.com/streamlink/streamlink/commit/2699ca90246abbf061af044287ab5f3c7234dab3

Since the default logger/console output stream is `stdout` (unless `--stdout` is set, which makes it switch to `stderr`), this means that the rendered text in a terminal application will always be interweaved if messages are written to both streams at the same time. This is especially noticable and annoying for log levels below `info`, like `debug` or `trace`, and reading/copying the entire output is ugly. Merging both streams doesn't fix this, even if messages are always written as entire text lines, so simply switching to `stdout` by default or `stderr` on `--stdout` won't work.

The download progress should be kept as a sticky line at the bottom while the remaining content should flow above. This requires additional stream message write/merge logic, but additional ANSI escape codes / control sequence commands are not necessary for a single sticky line at the bottom and flowing text above. I therefore also don't think that a curses-like dependency or something like [`rich`](https://github.com/Textualize/rich) is necessary and that the logic for writing messages is still trivial. It however means that the `logging.StreamHandler` setup, `ConsoleOutput`, `Progress`, `StreamRunner` and main CLI module all need to be refactored and logic for writing log-, console- and progress-messages needs to be properly reorganized and restructured.

I already had a look at this, but then realized that this is a little bit more effort than initially thought. I'm therefore opening this issue as a reminder. It's just a minor issue, but it's at a point now where it has annoyed me way too much, so it should be fixed eventually.

bastimeyer:
> I already had a look at this, but then realized that this is a little bit more effort than initially thought.

> and that the logic for writing messages is still trivial

Unfortunately, it's not as trivial as thought if we want to implement this properly.

We've been using a cheap workaround in the current implementation, and that is simply filling the entire line with space characters for overriding the previous line after the carriage return, since we know the terminal's width (Python's stdlib implements this). This wasn't a problem because the data was written to stderr anyway.

[...]

PR #6457, opened 2025-03-02
cli: make --quiet also suppress ConsoleOutput

bastimeyer:
#6449 can't be resolved without prior changes to the setup of Streamlink's text output streams. This is the first step of the required setup changes.

----

This PR fixes the behavior of `--quiet`. Previously, it was treated as a simple alias of `--loglevel=none`. This however meant that the `ConsoleOutput` still remained on `stdout` (or `stderr` if `--stdout` is set), so non-log messages (especially most error messages) were still written to the output. A `--quiet` CLI argument should suppress all kinds of text output though, because that's what it implies, no text output at all.

The whole text stream setup is unfortunately a bit messy. Not only do we have two separate ways of writing text to the output, namely `logging.StreamHandler` and `ConsoleOutput`, where `ConsoleOutput` always inherits the output stream of the root logger (which also seems wrong), but `--logfile` and also potentially missing `stdout`/`stderr` file descriptors can affect the chosen output stream.

Another thing that was previously not handled correctly were user input prompts when the text output was a file stream, when setting the `--logfile` CLI argument for example, as `ConsoleOutput` does always inherit the logger's stream, or when the text output was unavailable (no file descriptors). This is now fixed and works correctly when `--quiet` is set, as it now writes to devnull.

----

Before this can be merged, I will have to check a couple of things first, because it's possible that I've broken something with these changes.

```
$ touch /tmp/foo

# Default log output and console output, with prompt
$ streamlink httpstream://file:///dev/zero best -o /tmp/foo
[cli][info] Found matching plugin http for URL httpstream://file:///dev/zero
[cli][info] Opening stream: live (http)
[cli][info] Writing output to
/tmp/foo
File /tmp/foo already exists! Overwrite it? [y/N]

# Without log output, but with console output and prompt
$ streamlink httpstream://file:///dev/zero best -o /tmp/foo --loglevel=none
File /tmp/foo already exists! Overwrite it? [y/N]

# --logfile sets stream of log output and console output, prompt not possible, immediate exit
$ streamlink httpstream://file:///dev/zero best -o /tmp/foo --logfile=/dev/null
$ streamlink httpstream://file:///dev/zero best -o /tmp/foo --logfile=/tmp/bar

# Weird special case, because isatty()==True on /dev/stdout. Broken... Needs further fixes...
$ streamlink httpstream://file:///dev/zero best -o /tmp/foo --logfile=/dev/stdout
[cli][info] Found matching plugin http for URL httpstream://file:///dev/zero
[cli][info] Opening stream: live (http)
[cli][info] Writing output to
/tmp/foo

File /tmp/foo already exists! Overwrite it? [y/N]

# missing stdout, prompt not possible, immediate exit
$ streamlink httpstream://file:///dev/zero best -o /tmp/foo 1&>-

# no idea why missing stderr behaves the same, will have to check this again
$ streamlink httpstream://file:///dev/zero best -o /tmp/foo 2&>-
```

The `--logfile` CLI arg affecting the `ConsoleOutput` stream can be changed later. It's unrelated to this PR.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -412,7 +412,7 @@ def build_parser():
         "--quiet",
         action="store_true",
         help="""
-            Suppress console and log output.
+            Suppress console and log output, and disable user input prompts.
         """,
     )
     logging.add_argument(
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -412,9 +412,7 @@ def build_parser():
         "--quiet",
         action="store_true",
         help="""
-            Hide all log output.
-
-            Alias for --loglevel=none.
+            Suppress console and log output.
         """,
     )
     logging.add_argument(
--- a/src/streamlink_cli/compat.py
+++ b/src/streamlink_cli/compat.py
@@ -1,18 +1,24 @@
 import sys
-from os import devnull
+from atexit import register as _atexit_register
+from os import devnull as _devnull
 from typing import BinaryIO
 
 
-try:
-    stdout: BinaryIO = sys.stdout.buffer
-except AttributeError:  # pragma: no cover
-    from atexit import register as _atexit_register
+devnull_bin = open(_devnull, "wb")
+_atexit_register(devnull_bin.close)
+devnull_txt = open(_devnull, "w", encoding=None)
+_atexit_register(devnull_txt.close)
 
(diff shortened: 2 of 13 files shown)


bastimeyer:
Going to close this for now with a different approach in the next couple of days.

The log file stream and consoleoutput stream need to be separated first.

PR #6461, opened 2025-03-06
cli.console: refactor and split into two streams

bastimeyer:
Replaces #6457
In preparation for #6449

This PR fixes a couple of console output issues and logger setup issues, and also splits the console output into two streams. The separation of console output streams is necessary in order for the download progress to be fixed later on. Progress output should never be written to a file, but since we want some console output to be written to log files as well, especially error messages, we need to use two streams.

Currently on master, the `ConsoleOutput` always inherits the stream of Streamlink's root logger. This is problematic because `--logfile=...` sets the logger stream to a file, which means all `ConsoleOutput` including user prompts are written to the log file instead of being written to a TTY where the user has to write their input.

Also, `--quiet` currently does only affect the log level, as it's treated as an alias of `--loglevel=none`, which means the inherited output stream is still `stdout`, so all console messages and error messages are printed regardless. `--quiet` should mean no output at all.

Another problem is that if there's no TTY output for user prompts, those prompts currently don't fail. If the user doesn't have a chance to know what's prompted, then the prompt must fail. And failure is currently also not handled correctly and `None` is returned instead, which is wrong.

----

I've decided to use one single PR branch for these changes, because they are all a bit intertwined. Splitting this up would make this unnecessarily complicated for me.

----

So, a quick overview of the individual commits:

1. Fix the method name of the `ConsoleOutput` password prompt according to Streamlink's public interface of the `UserInputRequester`
2. Fix the prompts, raise `OSError` on failure, and most importantly, check for valid input and output TTY streams (must exist and must not be file streams).
3. Update the error handling of the one function in Streamlink's CLI that doesn't use the `Streamlink` session's `UserInputRequester` and that instead calls the `ask()` method of the `ConsoleOutput` directly.
4. Refactor `ConsoleOutput` and use two separate streams, one for console output and one for file output. Both are optional, as `stdout`/`stderr` can be missing, or there's no log file stream. Also ensure that the file output is not a TTY, so that `--logfile=/dev/tty` doesn't duplicate console messages (as `msg()` and `msg_json()` write to both streams).
5. Refactor the setup of `ConsoleOutput` and Streamlink's root logger and actually use two separate streams. Make the root logger inherit the primary stream of the `ConsoleOutput` (not the other way around). If there's a log file, attach the file stream to the `ConsoleOutput` instance.
6. Add lazily opened file descriptors for `devnull` to the CLI's `compat` module and rename existing attributes accordingly. Changes are required for the next commit.
7. Check if the output stream for Streamlink's root logger actually exists. Previously, the `logging` module of Python's stdlib switched to `stderr` if the passed stream was `None`, and we also had to check if `stderr` existed and use an additional fallback to devnull as a `FileHandler` in that case. Simply don't attach any log handlers to the root logger if there's the passed stream is `None`.
8. Make `ConsoleOutput` fall back from `stdout` to `stderr` if `stdout` doesn't exist, to align with the previous fallback mechanic of the root logger that was removed in the previous commit. Since the primary `ConsoleOutput` stream is inherited by the logger now, the old behavior is restored. But now, both console and log output have a fallback.

----

Considering that there are a lot of changes, this needs a bit of additional testing before I'm going to merge this. As always, I've added tests with full coverage for the changed code parts, but this doesn't cover all potential combinations of CLI arguments and environments.

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -412,9 +412,7 @@ def build_parser():
         "--quiet",
         action="store_true",
         help="""
-            Hide all log output.
-
-            Alias for --loglevel=none.
+            Suppress all console and log output, and also disable user prompts.
         """,
     )
     logging.add_argument(
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -86,12 +86,12 @@ def check_file_output(path: Path, force: bool) -> Path:
     log.debug("Checking file output")
 
     if realpath.is_file() and not force:
-        if sys.stdin and sys.stdin.isatty():
+        try:
             answer = console.ask(f"File {path} already exists! Overwrite it? [y/N] ")
-            if not answer or answer.lower() != "y":
-                raise StreamlinkCLIError()
-        else:
+        except OSError:
             log.error(f"File {path} already exists, use --force to overwrite it.")
+            raise StreamlinkCLIError() from None
+        if not answer or answer.lower() != "y":
             raise StreamlinkCLIError()
 
     return realpath
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -852,7 +852,9 @@ def setup_console() -> None:
     global console
 
     console_output: TextIO | None
-    if args.stdout or args.output == "-" or args.record == "-" or args.record_and_pipe:
+    if args.quiet:
+        console_output = None
+    elif args.stdout or args.output == "-" or args.record == "-" or args.record_and_pipe:
         # Console output should be on stderr if we are outputting a stream to stdout
         console_output = sys.stderr
     else:
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -15,7 +15,7 @@ from contextlib import closing, suppress
 from gettext import gettext
 from pathlib import Path
 from time import sleep
-from typing import Any
+from typing import Any, TextIO
 
 import streamlink.logger as logger
 from streamlink import NoPluginError, PluginError, StreamError, Streamlink, __version__ as streamlink_version
@@ -716,6 +716,8 @@ def setup_args(
     if not args.url and args.url_param:
         args.url = args.url_param
 
+    args.silent_log = any(getattr(args, attr) for attr in QUIET_OPTIONS)
+
 
 def setup_config_args(parser, ignore_unknown=False):
     if args.no_config:
@@ -846,17 +848,26 @@ def log_current_arguments(session: Streamlink, parser: argparse.ArgumentParser):
             log.debug(f" {name}={value if name not in sensitive else '*' * 8}")
 
 
-def setup_logger_and_console(
-    stream=sys.stdout,
-    level: str = "info",
-    fmt: str | None = None,
-    datefmt: str | None = None,
-    file: str | None = None,
-    json=False,
-):
+def setup_console() -> None:
     global console
 
-    verbose = level in ("trace", "all")
+    console_output: TextIO | None
+    if args.stdout or args.output == "-" or args.record == "-" or args.record_and_pipe:
+        # Console output should be on stderr if we are outputting a stream to stdout
+        console_output = sys.stderr
+    else:
+        console_output = sys.stdout
+
+    console = ConsoleOutput(console_output=console_output, json=args.json)
+
+
+def setup_logger() -> None:
+    level: str = args.loglevel if not args.silent_log else logging.getLevelName(logger.NONE)
+    file: str | None = args.logfile if level != logging.getLevelName(logger.NONE) else None
+    fmt: str | None = args.logformat
+    datefmt: str | None = args.logdateformat
+
+    verbose = level in (logging.getLevelName(logger.TRACE), logging.getLevelName(logger.ALL))
     if not fmt:
         if verbose:
             fmt = "[{asctime}][{name}][{levelname}] {message}"
@@ -885,13 +896,14 @@ def setup_logger_and_console(
             datefmt=datefmt,
             style="{",
             level=level,
-            stream=stream,
+            stream=console.console_output,
             capture_warnings=True,
         )
     except Exception as err:
         raise StreamlinkCLIError(f"Logging setup error: {err}") from err
 
-    console = ConsoleOutput(console_output=streamhandler.stream, json=json)
+    if isinstance(streamhandler, logging.FileHandler):
+        console.file_output = streamhandler.stream
 
 
 def setup(parser: ArgumentParser) -> None:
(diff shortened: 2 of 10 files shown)


src/streamlink_cli/argparser.py
```diff
@@ -412,9 +412,7 @@ def build_parser():
         "--quiet",
         action="store_true",
         help="""
-            Hide all log output.
-
-            Alias for --loglevel=none.
+            Suppress all console and log output, and also disable user prompts.
```
gravyboat, line 415: Is this going to be clear enough to users? I feel like this option is typically used by power users so it shouldn't be a problem but do we need to specify "all" user prompts? Because of the way the sentence is structured someone could think it's all console and log output, but only some user prompts. I know this is pedantic but we get a lot of silly questions so it's worth noting.

src/streamlink_cli/main.py
```diff
@@ -846,17 +849,28 @@ def log_current_arguments(session: Streamlink, parser: argparse.ArgumentParser):
...
-):
+def setup_console() -> None:
     global console
 
-    verbose = level in ("trace", "all")
+    console_output: TextIO | None
+    if args.quiet:
+        console_output = compat.devnull_txt
+    elif args.stdout or args.output == "-" or args.record == "-" or args.record_and_pipe:
+        # Console output should be on stderr if we are outputting a stream to stdout
```
gravyboat, line 859: It's up to you but I don't think we need this comment. It's pretty clear to me what is happening here thanks to the rewritten code.

tests/cli/main/test_check_file_output.py
```diff
@@ -130,22 +101,44 @@ def test_exists(
...
+            ],
+            id="none",
+        ),
+        pytest.param(
+            {"ask": OSError()},
+            pytest.raises(StreamlinkCLIError),
+            [
+                ("streamlink.cli", "info", "Writing output to\n/path/to/file"),
+                ("streamlink.cli", "debug", "Checking file output"),
+                ("streamlink.cli", "error", "File file already exists, use --force to overwrite it."),
```
gravyboat, line 139: Are we actually testing `--force` somewhere in terms of the file output to ensure it doesn't (or does) properly overwrite?

tests/cli/test_argparser.py
```diff
@@ -296,18 +296,11 @@ class StopTest(Exception):
...
-        monkeypatch.setattr("sys.stdin", mock_stdin)
-
-        return mock_stdin
-
     @pytest.fixture()
     def console(self):
         return Mock(
             ask=Mock(return_value="answer"),
-            askpass=Mock(return_value="password"),
+            ask_password=Mock(return_value="password"),
```
gravyboat, line 303: Good name change throughout. 👍

tests/cli/test_compat.py
```diff
@@ -22,9 +41,52 @@ def test_no_stdout(monkeypatch: pytest.MonkeyPatch):
...
-    assert module.sys.stdout is None
-    assert isinstance(module.stdout, BufferedWriter)
-    assert module.stdout.name == devnull
 
-    assert mock_atexit_register.call_args_list == [call(module.stdout.close)]
-    module.stdout.close()  # close manually, since we've mocked the atexit.register() call
+    try:
+        yield module
+    finally:
+        # close manually, since we've mocked the atexit.register() call
```
gravyboat, line 48: Another comment I don't think you need personally. The updated code is pretty clear but it's up to you.

tests/cli/test_console.py
```diff
@@ -127,55 +235,93 @@ def test_msg_json_merge_list(self, output: TextIOWrapper):
...
-
-    def test_askpass_no_stdin(self, monkeypatch: pytest.MonkeyPatch, output: TextIOWrapper):
-        monkeypatch.setattr("sys.stdin", None)
+    @pytest.mark.parametrize(
+        ("stdin", "console_output", "expected"),
+        [
+            pytest.param({"exists": False}, {}, "^No input TTY available$", id="no-stdin"),
+            pytest.param({"isatty": False}, {}, "^No input TTY available$", id="stdin-no-tty"),
+            pytest.param({}, {"exists": False}, "^No output TTY available$", id="no-output"),
+            pytest.param({}, {"isatty": False}, "^No output TTY available$", id="output-no-tty"),
```
gravyboat, line 309: For consistency with the two `stdin` ids should the id values here be `no-stdout` and `stdout-no-tty`?

tests/test_plugin_userinput.py
```diff
@@ -8,51 +8,67 @@
...
+def test_user_input_not_implemented(session: Streamlink, testplugin: _TestPlugin):
+    session.set_option("user-input-requester", None)
+
+    with pytest.raises(FatalPluginError) as cm:
+        testplugin.input_ask("test")
+    assert str(cm.value) == "This plugin requires user input, however it is not supported on this platform"
+
+    with pytest.raises(FatalPluginError) as cm:
+        testplugin.input_ask_password("test")
+    assert str(cm.value) == "This plugin requires user input, however it is not supported on this platform"
```
gravyboat, line 54: Could we possibly use a better name for this? The name `test_user_input_not_implemented` plus the file name alluding to testing user input makes it seem like the plugin is at fault for somehow not accepting the input but the message then says the platform is at fault. Can we clarify it in some way to make it more consistent that the problem is system/platform input and not a problem with the plugin itself?

tests/test_plugin_userinput.py
```diff
@@ -8,51 +8,67 @@
...
+
+    assert testplugin.input_ask_password("password") == "password"
+    assert console.ask_password.call_args_list == [call("password: ")]
+
+
+@pytest.mark.parametrize("console", [{"failure": True}], indirect=True)
+def test_user_input_console_no_tty(console: Mock, testplugin: _TestPlugin):
+    with pytest.raises(FatalPluginError) as cm:
+        testplugin.input_ask("username")
+    assert str(cm.value) == "User input error: No input TTY available"
```
gravyboat, line 70: This one feels a lot more consistent in what it is communicating/testing than `test_user_input_not_implemented`. 👍

gravyboat (approved):
Thanks for breaking this up in to several commits, made it much easier to review each chunk. Made a couple non-blocking notes. Feel free to address or not address whatever you prefer.

src/streamlink_cli/argparser.py
bastimeyer: I think this is fine. "disable user prompts" should be pretty clear that no input will be possible when the program asks for input.

tests/cli/main/test_check_file_output.py
bastimeyer: See `test_exists`. I'll have a look later again, because some parametrization data was removed for this test because of the changed error handling, and the current setup is a bit weird now.

tests/cli/test_console.py
bastimeyer: `ConsoleOutput` can receive both `stdout` or `stderr`, depending on the used parameters and availability of `stdout`. Hence why it's internally also using the `console_output` name and why it's saying "output" on error.

tests/test_plugin_userinput.py
bastimeyer: I didn't touch any Plugin-specific APIs. These were left intact, and if they require changes, then in a different PR.

The error message you're commenting on comes from the `Plugin` class which uses the `UserInputRequester` interface and thus needs to do error handling, in which case it is raising its own `FatalPluginError` exception. This has nothing to do with any changes here.

I don't see any problem with the error message, but if you have a suggestion for an improvement, then please post it.

tests/test_plugin_userinput.py
bastimeyer: The difference here is that this error message is propagated from the set-up `ConsoleUserInputRequested`. The other one you've commented on comes from a situation where no `UserInputRequester` was set on the `Streamlink` session object, which can't be the case for `streamlink_cli` anyway because we always set it. It's for third parties which may implement user inputs differently (which I doubt anyone does anyway, but this API exists, so these two different tests exist, because they are supposed to cover all cases)

tests/test_plugin_userinput.py
gravyboat: Sounds good, let's leave it.

tests/test_plugin_userinput.py
gravyboat: Makes sense to me, I'll mark as resolved.

bastimeyer:
f2b7764 is probably unnecessary if we re-arrange the commits. `--quiet` should set `None` as the `console_output` stream, not `devnull_txt`. This was necessary because of the previous `logger` setup that defaulted to `stderr` on `None`, as the console output was inherited here.

I'll have a look at this again some time next week.

PR #6464, opened 2025-03-09
docs: improve API quickstart guide

bastimeyer:
Leaving this up for review, in case something's not as clear as it should be (which I think it is now).

Ref #6463 

See the rendered docs:
https://streamlink--6464.org.readthedocs.build/api_guide/quickstart.html

### DIFF
--- a/docs/api_guide/quickstart.rst
+++ b/docs/api_guide/quickstart.rst
@@ -17,7 +17,7 @@ The simplest use of the Streamlink API looks like this:
 
 This simply attempts to find a plugin and use it to extract streams from
 the URL. This works great in simple cases but if you want more
-fine tuning you need to use a `session object`_ instead.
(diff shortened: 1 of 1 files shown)


docs/api_guide/quickstart.rst
```diff
@@ -80,19 +80,55 @@ when extracting streams more than once. You start by creating a
...
+
+The first example will automatically try to find a matching plugin and available streams from the input URL:
 
 .. code-block:: python
 
     >>> streams = session.streams("URL")
 
+See :py:meth:`Streamlink.streams() <streamlink.session.Streamlink.streams>` for more.
+
+It however doen't allow you to set any plugin options which might be necessary in order to access streams, e.g. when
```
gravyboat, line 108: I don't think "It" should be used here to avoid confusion. The exact name should be used. "`session.streams` doesn't allow you to..."

gravyboat (approved):
One minor nit, rest looks good. Nice job!

PR #6476, opened 2025-03-22
plugins.nicolive: filter out blank segments

bastimeyer:
Ref https://github.com/streamlink/streamlink/pull/6441#issuecomment-2745230175

@secondaryH please test...
https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-request-feedback

Since there has been no follow up to my request in the linked thread, I haven't tested these changes myself. These were purely done by reading the posted log output.

### DIFF
--- a/src/streamlink/plugins/nicolive.py
+++ b/src/streamlink/plugins/nicolive.py
@@ -9,6 +9,8 @@ $account Required by some streams
 $notes Timeshift is supported
 """
 
+from __future__ import annotations
+
 import logging
 import re
 from threading import Event
@@ -17,7 +19,7 @@ from urllib.parse import urljoin
 from streamlink.plugin import Plugin, pluginargument, pluginmatcher
 from streamlink.plugin.api import useragents, validate
 from streamlink.plugin.api.websocket import WebsocketClient
-from streamlink.stream.hls import HLSStream, HLSStreamReader
+from streamlink.stream.hls import HLSSegment, HLSStream, HLSStreamReader, HLSStreamWriter
 from streamlink.utils.parse import parse_json
 from streamlink.utils.url import update_qsd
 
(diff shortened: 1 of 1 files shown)


SecondaryH:
Hi, I've tested the new nicolive.py and my issue has been resolved. Audio&video are good and no sync issues.

Command used: streamlink -o "outputs/{author}/{title}-{time:%Y-%m-%d-%H-%M-%S}.ts" --ffmpeg-copyts --logfile log-streamlink.txt --loglevel all https://live.nicovideo.jp/watch/lv347277552 best

Version: streamlink 7.1.3+26.g6e8d7ce

Attached logfile if interested. Thankyou!

[log-streamlink.txt](https://github.com/user-attachments/files/19404945/log-streamlink.txt)

nnn-revo2012:
I'm using ChatGPT to translate from Japanese to English, so I apologize if the grammar is incorrect.

For the nicolive.py that was submitted as this pull request, there were broadcasts where the video would not play correctly unless ``--hls-start-offset 6`` was specified.
Conversely, for some broadcasts, specifying ``--hls-start-offset 6`` would result in the following error:

```[cli][error] Try 1/1: Could not open stream <MuxedHLSStream ['hls-multi', 'https ://assetdelivery.dlive.nicovideo.jp/...m3u8']> (No data returned from stream)```

Additionally, I confirmed that all broadcasts, such as official broadcasts that do not require ``--hls-start-offset 6``, could be downloaded normally without specifying ``--hls-start-offset``.

Checked broadcasts:
- https://live.nicovideo.jp/watch/lv347086431
- https://live.nicovideo.jp/watch/lv347108476
- https://live.nicovideo.jp/watch/lv347335723

bastimeyer:
> I've tested the new nicolive.py and my issue has been resolved.

:+1: 

According to the log output, the `/blank/` segments at the beginning are filtered out correctly and thus don't disturb the actual content output.

But please don't overwrite/replace the plugin file directly. Have a look at the PR feedback link I've posted earlier. You're supposed to install+run the PR branch or sideload the modified plugin file. You're doing neither.

> Audio&video are good and no sync issues.

It's a VOD, so there's no sync issue to be expected because both playlists will start at the same time, regardless of access time, unlike live playlists, as explained in the other PR. You also don't need the `--ffmpeg-copyts` parameter, because it's now forcefully set by the plugin when muxing the audio and video streams.

----

> there were broadcasts where the video would not play correctly unless `--hls-start-offset 6` was specified

All of these links are restricted and can't be accessed.

`--hls-start-offset` is not meant to be used on live streams. It's for VODs, to select a specific range of segments instead of all segments. For live streams where there's a live-edge and only a small window of available segments, this doesn't make sense, which is why you're seeing no data being written to the buffer when setting a nonsensical value.

But without any useful debug logs, none of your claims can be checked.

Btw, this is totally unrelated to this pull request. If there is a plugin issue that can be reliably reproduced, then please open a plugin issue thread and don't comment here.

PR #6541, opened 2025-05-26
session.http: proxy matchers with wildcards

### LINKED ISSUE #6509: Add support for host-specific proxies with wildcards (`--http-proxy "[HOST ]...PROXY"`)
### Checklist

- [x] This is a feature request and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [x] [I have checked the list of open and recently closed feature requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22feature+request%22)

### Collaboration

- [x] My request is genuine and in the best interest of the project
- [x] [I will provide feedback should a pull request be opened with a new feature implementation](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-request-feedback)

### Description

Streamlink currently only allows setting one global proxy server (HTTP/HTTPS/SOCKS4/SOCKS5) for all of its HTTP/HTTPS connections via `--http-proxy=PROXY` (`--https-proxy` is the deprecated alias argument).

This isn't ideal, because bypassing geo-restrictions on web-API endpoints while using the local connection for downloading stream contents is impossible this way (when this is not restricted by the streaming provider). Some plugins therefore need to be forked and sideloaded, which is unnecessary.

----

The proxy server config is applied by adding the `http` and `https` keys to the `HTTPSession.proxies` dict
- https://github.com/streamlink/streamlink/blame/7.3.0/src/streamlink/session/options.py#L372-L376

which makes `requests` always select the set proxy. `requests` however does support more fine-grained control over this by allowing users to specify a **scheme** and a **hostname**, with a matching scheme+hostname having a higher priority:
- https://requests.readthedocs.io/en/latest/api/#requests.Session.proxies

The problem with this though is that while the scheme can be set to `all://` (which matches both `http` and `https`), hostnames must always match exactly for a proxy to be selected. This is implemented in `requests.utils.select_proxy()` here:
- https://github.com/psf/requests/blame/v2.32.3/src/requests/utils.py#L838-L861

For an application like Streamlink this doesn't make sense, as the relevant backend hostnames of most of the large streaming providers are dynamic, due to load-balancing reasons. `requests` unfortunately doesn't support hostname wildcards, as shown in the linked implementation. So in order for Streamlink to add support for host-based proxy configurations, `requests`'s proxy selection needs to be overridden with a function that implements a hostname wildcard selection.

I'm not sure how I feel about this, because `urllib3`+`requests` are clearly a dead end for many many years now, and adding more features to Streamlink that depend on low-level overrides of both libs is obviously bad if the eventual goal is to migrate to different networking libs. `httpx` (#4915) on the other hand does support hostname wildcards in its proxy selection, so it's actually not a bad thing to implement an override for `requests` now that works similar to `httpx`, because it could be removed easily when migrating:
- https://github.com/encode/httpx/blame/0.28.1/httpx/_utils.py#L120-L226

----

On the Streamlink's CLI side of things, `--http-proxy`/`--https-proxy` should be turned into a list (`action="append"`), so that it can be repeated multiple times, and on the session's `StreamlinkOptions` side of things, the setter function for the `http-proxy`/`https-proxy` keys needs to allow passing a `list[str] | str` value in order to stay backwards compatible.

The value of each proxy config string should be in the form of `[HOST ]...PROXY`, so basically a space-separated list of multiple optional hostnames (with wildcard support) and the proxy address at the end. From that, the `HTTPSession.proxies` dict can be built with all the host->proxy mappings. The scheme for each host selection schould be set to `all://`.

----

```
streamlink --http-proxy "*.123.foo *.456.bar proxy1" --http-proxy "proxy2" URL stream
```

This would use `proxy1` for all requests made to `*.123.foo` or `*.456.bar`, and `proxy2` for everything else.

bastimeyer:
Resolves #6509

----

- **Why the over-complicated CLI argument format?**
  We can't use the regular key=value syntax for repeated CLI arguments to construct the `HTTPSession.proxies` dict, because `=` can be part of URLs, so a space separator is needed. We also want the keys to be optional in order to keep backwards compatibility where the `http` and `https` protocol matchers are set. And since we already have a custom format, we can already add support for adding multiple proxy matchers at once, so the CLI arg doesn't have to be set multiple times with the same proxy server URL.
- **Why the `no_proxy` logic in `--http-proxy`?**
  This is a consequence of `requests` supporting this key in its `proxies` dict:  
  https://github.com/psf/requests/blob/v2.32.3/src/requests/utils.py#L864-L888
  https://github.com/psf/requests/blob/v2.32.3/src/requests/utils.py#L765-L823
  We could add a separate CLI argument for that though, like `curl`'s `--noproxy`.

----

## TODO

- [x] **Add support for wildcards in URL-matchers**

  ~~I had already implemented URL-wildcard-matching for proxies, but then I noticed that my approach wasn't working properly when trying to integrate it into the `HTTPSession` subclass of `requests`'s `Session`, because the proxy config is merged with parameters of individual requests and the system environment. I have another idea though, but this is unfinished.~~

- [ ] **Update the proxies docs page**

  The `--http-proxy` help text now has more information than the dedicated docs page

## MAYBE

- Replace the `http` and `https` proxy matchers that are set by default with `all` (the lowest priority). This may result in behavior changes though in regards to the proxy env vars.

### DIFF
--- a/src/streamlink/session/http.py
+++ b/src/streamlink/session/http.py
@@ -15,6 +15,7 @@ import streamlink.session.http_useragents as useragents
 from streamlink.exceptions import PluginError, StreamlinkDeprecationWarning
 from streamlink.packages.requests_file import FileAdapter
 from streamlink.utils.parse import parse_json, parse_xml
+from streamlink.utils.url import select_proxy
 
 
 try:
@@ -79,6 +80,10 @@ class Urllib3UtilUrlPercentReOverride:
 
 # urllib3>=2.0.0: _PERCENT_RE, urllib3<2.0.0: PERCENT_RE
 urllib3.util.url._PERCENT_RE = urllib3.util.url.PERCENT_RE = Urllib3UtilUrlPercentReOverride  # type: ignore[attr-defined]
+
+
+requests.utils.select_proxy = select_proxy  # type: ignore[assignment]
+requests.adapters.select_proxy = select_proxy  # type: ignore[attr-defined]
 
 
 # requests.Request.__init__ keywords, except for "hooks"
--- a/src/streamlink/utils/url.py
+++ b/src/streamlink/utils/url.py
@@ -1,5 +1,94 @@
+from __future__ import annotations
+
 import re
+from enum import IntEnum
+from functools import cache
+from typing import Mapping
 from urllib.parse import parse_qsl, quote_plus, urlencode, urljoin, urlparse, urlunparse
+
+
+_re_scheme = re.compile(r"""^[a-z0-9][a-z0-9.+-]*://""", re.IGNORECASE)
+
+
+class _ProxyMatcherPriority(IntEnum):
+    ALL_SCHEME_WITHOUT_NETLOC = 1
+    ALL_SCHEME_WITH_NETLOC = 2
+    SPECIFIC_SCHEME_WITHOUT_NETLOC = 3
+    SPECIFIC_SCHEME_WITH_NETLOC = 4
+
+
+@cache
+def _proxymatcher_to_prio_and_pattern(matcher: str) -> tuple[int, re.Pattern] | None:
+    if not matcher or matcher == "all":
+        return _ProxyMatcherPriority.ALL_SCHEME_WITHOUT_NETLOC, _re_scheme
+
+    if matcher in ("http", "https"):
+        return _ProxyMatcherPriority.SPECIFIC_SCHEME_WITHOUT_NETLOC, re.compile(f"^{matcher}://", re.IGNORECASE)
+
+    if not _re_scheme.match(matcher):
+        matcher = f"all://{matcher}"
+
+    try:
+        parsed = urlparse(matcher)
+        hostname: str = parsed.hostname or ""
+        portnumber: int = parsed.port or 0
+    except ValueError:
+        return None
+
+    netloc = parsed.netloc
+    if not hostname or hostname == "*":
+        netloc = ".+?"
+        hostname = ""
+    elif netloc.startswith("*."):
+        netloc = f".+\\.{re.escape(netloc[2:])}"
+    elif netloc.startswith("*"):
+        netloc = f"(?:.+\\.)?{re.escape(netloc[1:])}"
+    else:
+        netloc = re.escape(netloc)
+
+    if not hostname and portnumber > 0:
+        netloc += f":{portnumber}"
+    elif parsed.scheme == "http":
+        netloc += "(?::80)?"
+    elif parsed.scheme == "https":
+        netloc += "(?::443)?"
+
+    if parsed.scheme == "all":
+        scheme = "[a-z0-9][a-z0-9.+-]*"
+        priority = _ProxyMatcherPriority.ALL_SCHEME_WITH_NETLOC
+    else:
+        scheme = re.escape(parsed.scheme)
+        priority = _ProxyMatcherPriority.SPECIFIC_SCHEME_WITH_NETLOC
+
+    pattern = re.compile(f"^{scheme}://{netloc}(?:/.*)?$", re.IGNORECASE)
+
+    return priority, pattern
+
+
+def select_proxy(url: str, proxies: Mapping[str, str]) -> str | None:
+    """
+    Replacement function for `requests.utils.select_proxy`, with support for wildcards in proxy URL matchers.
+    Support for wildcards means that the lookup time now scales linearly with the size of `proxies`.
+    """
+
+    proxies = proxies or {}
+    urlparts = urlparse(url)
+    if urlparts.hostname is None:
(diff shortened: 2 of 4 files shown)


Hakkin:
Tested with a variety of hosts and it seems to work as expected for the most part, but I can't seem to get `no_proxy` to work correctly, unless I'm misunderstanding how it's suppose to work.

I would assume this command would do "route all http(s) traffic via the proxy, except for host gql.twitch.tv".
```shell
$ streamlink@proxy --loglevel=debug --http-proxy "socks5h://127.0.0.1:8889" --http-proxy "no_proxy gql.twitch.tv" "https://www.twitch.tv/twitchmedia21" best
...
[cli][debug]  --http-proxy=['socks5h://127.0.0.1:8889', 'no_proxy gql.twitch.tv']
```

yet `gql.twitch.tv` is still routed via the proxy.

```shell
$ export NO_PROXY=gql.twitch.tv
$ streamlink@proxy --loglevel=debug --http-proxy "socks5h://127.0.0.1:8889" "https://www.twitch.tv/twitchmedia21" best
...
[cli][debug]  --http-proxy=['socks5h://127.0.0.1:8889']
```

`gql.twitch.tv` is still routed via the proxy.

```shell
$ export ALL_PROXY=socks5h://127.0.0.1:8889
$ streamlink@proxy --loglevel=debug --http-proxy "no_proxy gql.twitch.tv" "https://www.twitch.tv/twitchmedia21" best
...
[cli][debug]  --http-proxy=['no_proxy gql.twitch.tv']
```

works as expected, `gql.twitch.tv` request bypasses the proxy while the rest go through it.

```shell
$ export NO_PROXY=gql.twitch.tv
$ export ALL_PROXY=socks5h://127.0.0.1:8889
$ streamlink@proxy --loglevel=debug "https://www.twitch.tv/twitchmedia21" best
```

works as expected, `gql.twitch.tv` request bypasses the proxy while the rest go through it.

So it seems like `--http-proxy` with the `no_proxy` matcher can't be used in conjunction with any other `--http-proxy` arguments? Although it does work when used with env variables. Not sure if that's the intended behavior or not.

bastimeyer:
There's indeed an issue with `no_proxy` here...

This PR simply lets the user set up the [`proxies`](https://requests.readthedocs.io/en/latest/api/#requests.Session.proxies) dict that is set on the HTTPSession object (this is also where the `no_proxy` data is stored). Previously, Streamlink only set the `http` and `https` protocol/scheme keys on the `proxies` dict and that was it.
- https://requests.readthedocs.io/en/latest/api/#requests.Session.proxies

It also replaces the `select_proxy()` function that's called down the stack in `HTTPAdapter.send()`
- https://github.com/psf/requests/blob/v2.32.3/src/requests/adapters.py#L633-L635
- https://github.com/psf/requests/blob/v2.32.3/src/requests/adapters.py#L466
- https://github.com/psf/requests/blob/v2.32.3/src/requests/utils.py#L838-L861

----

When making an HTTP request, the request's `proxies` dict is first merged with the `proxies` session data and with the environment vars:
- https://github.com/psf/requests/blob/v2.32.3/src/requests/sessions.py#L579-L581
- https://github.com/psf/requests/blob/v2.32.3/src/requests/sessions.py#L750-L779
- https://github.com/psf/requests/blob/v2.32.3/src/requests/utils.py#L826-L835
- https://github.com/psf/requests/blob/v2.32.3/src/requests/utils.py#L765-L823
- https://github.com/python/cpython/blob/v3.13.3/Lib/urllib/request.py#L2532-L2564

But when getting the env vars, it also evaluates the `no_proxy` data in `requests.utils.should_bypass_proxies()` (depending on the OS, not just the `NO_PROXY` env var is taken into account - via the stdlib's `urllib.request.proxy_bypass`), and if the request's host matches one of the comma-separated hosts, `should_bypass_proxies()` returns True and thus `get_environ_proxies()` returns an empty dict that is then merged with the other `proxies` data from the specific request and the session.

Later on when we arrive at `select_proxy()`, the `proxies` data is either the session+request data when `should_bypass_proxies()` returned True, or the `proxies` dict includes an additional `no` key with the `NO_PROXY` env var data.

I'm not fully sure if this is a bug in `requests`, because this could also be intentional. If proxies data exists on the session object or if it's passed on a specific request, then the env var data is ignored. The `no` key however doesn't make any sense and is likely a bug.

This also suggests that this was already broken on master.

I'll have another look at this later...

bastimeyer:
Going to close this. Will attempt this again in a different PR once it's ready.

I'll add a second session option + CLI arg for the no-proxy stuff, to simplify user inputs.
Fixing the no-proxy logic will require further overrides of the requests functions, which is annoying, especially considering that the long-term goal is to move away from requests.
What I also forgot to change here is updating the WebsocketClient, which currently relies on the old http-proxy logic.

PR #6547, opened 2025-06-03
plugins.blasttv: new plugin

Lense:
Blast is a video game (mainly Counter-Strike) tournament organizer who streams their matches on their website alongside the usual Twitch and YouTube. The streams on blast.tv are usually in 4K as opposed to 1080p on other platforms.

I've been using this plugin locally for a couple years. Over time, I've  recorded different API (`https://api.blast.tv/v1/broadcasts/live`) responses in comments in the plugin code to make development and testing easier, but I'm happy to move those to a more appropriate place.

There are some caveats:
- Streams don't run 24/7--they'll only run, idk, 8-12 hours a day, 10 weeks a year, which I imagine would be annoying for reviewers. I'm creating the PR now because there should be regular daytime CST streams from now through June 22.
- The stream used to occasionally (once every hour or so, but with high variance) exit early when `streamlink` was run with default options. Adding `--hls-segment-queue-threshold 8` completely eliminated the issues. My understanding is that plugins can't do anything about this.

### DIFF
--- a/src/streamlink/plugins/blasttv.py
+++ b/src/streamlink/plugins/blasttv.py
@@ -1,0 +1,114 @@
+"""
+$description Esports tournaments run by BlastTV, based in Denmark.
+$url blast.tv
+$type live, vod
+$metadata id
+$metadata title
+"""
+
+import re
+
+from streamlink.plugin import Plugin, pluginmatcher
+from streamlink.plugin.api import validate
+from streamlink.stream.hls import HLSStream
+
+
+@pluginmatcher(
+    name="live",
+    pattern=re.compile(r"^https?://(?:www\.)?blast\.tv(?:/?$|/live(?:/(?P<channel>[a-z0-9_-]+))?)"),
+)
+@pluginmatcher(
+    name="vod",
+    pattern=re.compile(r"^https?://(?:www\.)?blast\.tv/(?P<game>[^/]+)/tournaments/[^/]+/match/(?P<shortid>\w+)/"),
+)
+class BlastTv(Plugin):
+    _URL_API_LIVE = "https://api.blast.tv/v1/broadcasts/live"
+    _URL_API_MATCHES = "https://api.blast.tv/v2/games/{game}/matches/{shortid}"
+    _URL_API_REWATCH = "https://api.blast.tv/v1/videos/rewatch/{id}"
+
+    def _get_live(self):
+        channel = self.match["channel"]
+
+        live_channels = self.session.http.get(
+            self._URL_API_LIVE,
+            schema=validate.Schema(
+                validate.parse_json(),
+                [
+                    {
+                        "id": str,
+                        "slug": str,
+                        "priority": int,
+                        "title": str,
+                        "videoSrc": validate.any("", validate.url(path=validate.endswith(".m3u8"))),
+                        "videoAlternativeSrc": validate.any("", validate.url(scheme="http")),
+                    },
+                ],
+            ),
+        )
+
+        for live_channel in sorted(live_channels, key=lambda x: x["priority"]):
+            if channel and channel != live_channel["slug"]:
+                continue
+
+            if live_channel["videoSrc"]:
+                self.id = live_channel["id"]
+                self.title = live_channel["title"]
+                return HLSStream.parse_variant_playlist(self.session, live_channel["videoSrc"])
+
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/blasttv.py
```diff
@@ -0,0 +1,74 @@
...
+                validate.filter(lambda v: v.get("videoSrc")),
+                [
+                    {
+                        "id": str,
+                        "slug": str,
+                        "title": str,
+                        "videoSrc": validate.url(),
+                    },
+                ],
+            ),
```
bastimeyer, line 54: Parsed JSON responses require a proper data structure before you can iterate the data (validate.filter). If they don't return an array, then this will raise an ugly error.

bastimeyer (commented):
Thanks for the PR, appreciated.

Adding a plugin for this site should be fine. But next time, please open a plugin request first, so you don't risk wasting your time should a PR you've worked on be rejected.

Before this can be merged, a couple of things need to be changed, fixed and also cleaned up.

**Are you fine with me pushing onto your PR branch with my changes?** I had a look at the site and modified your plugin accordingly, and I also added support for VODs. Both live streams and VODs can also be hosted externally, e.g. on YouTube or Twitch, so the plugin should redirect towards that automatically. I also fixed that.

I really don't feel like annotating code in PRs for non-trivial changes that I have already fixed locally, but let me comment on some of the stuff anyway.

Results from my local changes:

```
$ ./script/test-plugin-urls.py blasttv
:: https://blast.tv
::  160p, 360p, 480p, 720p, 1080p, 1440p, 2160p, worst, best
:: https://blast.tv/
::  160p, 360p, 480p, 720p, 1080p, 1440p, 2160p, worst, best
:: https://blast.tv/cs/tournaments/open-2025-season-1/match/24ff9d6c/vitality-mouz
::  160p, 360p, 480p, 720p, 1080p, 1440p, worst, best
:: https://blast.tv/cs/tournaments/rivals-2025-season-1/match/bfaaa42e/falcons-vitality
::  160p, 360p, 480p, 720p, 1080p, 1440p, worst, best
:: https://blast.tv/dota/tournaments/blast-slam-iii/match/acabb915/falcons-tundra
::  audio, worst, best, 160p, 360p, 480p, 720p60, 1080p60
:: https://blast.tv/live
::  160p, 360p, 480p, 720p, 1080p, 1440p, 2160p, worst, best
:: https://blast.tv/live/
::  160p, 360p, 480p, 720p, 1080p, 1440p, 2160p, worst, best
:: https://blast.tv/live/a
::  160p, 360p, 480p, 720p, 1080p, 1440p, 2160p, worst, best
:: https://blast.tv/live/americas1
!! No streams found
:: https://blast.tv/live/co-stream
!! No streams found
:: https://blast.tv/live/f
::  worst, best, 144p, 240p, 360p, 480p, 720p, 1080p
```

Lense:
> **Are you fine with me pushing onto your PR branch with my changes?**

Please go ahead, thank you.

bastimeyer:
Had a second look and I think the plugin implementation should be fine. If you want, you can have a look too and see if there are any streams that don't work. Otherwise, I'm going to merge later this evening.

Thanks.

Lense:
Everything seems to work great, thank you so much!

bastimeyer:
> * The stream used to occasionally (once every hour or so, but with high variance) exit early when `streamlink` was run with default options. Adding `--hls-segment-queue-threshold 8` completely eliminated the issues. My understanding is that plugins can't do anything about this.

Forgot to comment on this...

It's the server's responsibility according to the HLS protocol to provide the client with new media segments within the playlist's [`targetduration`](https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.3.1). Streamlink can end streams early in two different ways: first, via the global buffer read-timeout, which is a time value for how long the stream's consumer can wait for new data, and second, via the `hls-segment-queue-threshold` option (to be renamed in the future) that's a multiplication factor (default is 3) of the `targetduration` time, which is already a rather liberal value.

The intention is to close streams early which don't announce the stream's end properly without causing a buffer read timeout to trigger, as it's not a graceful end of the stream.

Plugins can always change session options, but it doesn't make any sense increasing this value, because it's an issue with the server. Setting the value to 8 means that the stream is allowed to stall for eight times its actual targetduration.

PR #6551, opened 2025-06-07
release: 7.4.0

bastimeyer:
Going to publish this evening today...

https://streamlink--6551.org.readthedocs.build/changelog.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## streamlink 7.4.0 (2025-06-08)
+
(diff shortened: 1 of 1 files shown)


bastimeyer:
The Streamlink AppImages also had a significant update since the last release, but I'm not going to list these changes here, even though we've had changelogs for the AppImage / Windows builds in Streamlink's 7.0.0 release due to the x86 removal. That was a breaking change that needed to be highlighted.

Regarding the AppImage update: with the switch to the "AppImage type2 runtime", `libfuse2` is not needed anymore on the host system in order to run the AppImages. Gzip compression has also been replaced with zstd, which is much faster.
https://github.com/streamlink/streamlink-appimage/commit/ce23c2bce7db97bab5052a058cd05c2d6422a8d4

PR #6589, opened 2025-06-27
ci.github: add py314 Linux+Windows test runners

bastimeyer:
The Python 3.14 Linux runner was temporarily removed in #6561 because pytest hang for over an hour and then the runner got killed, as it expired. Let's see if it's working again.

Also let's add the Windows runner, with matching temp wheels of Streamlink's binary dependencies.

### DIFF
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -380,15 +380,16 @@ class TestPluginArguments:
             namespace = parser.parse_args(args)
             assert namespace.myplugin_foo == expected
 
-    def test_decorator_typeerror(self):
-        with patch("builtins.repr", Mock(side_effect=lambda obj: obj.__name__)):
-            with pytest.raises(TypeError) as cm:
-                # noinspection PyUnusedLocal
-                @pluginargument("foo")
-                class Foo:
-                    pass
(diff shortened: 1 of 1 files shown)


bastimeyer:
Both are failing... great...

https://github.com/streamlink/streamlink/actions/runs/15934327543/job/44950603303?pr=6589#step:7:55
https://github.com/streamlink/streamlink/actions/runs/15934327543/job/44950603286?pr=6589#step:7:54

> ```
> tests/test_plugin.py ..............................FFFFFFEEEEEFFFFFFFFFF [ 12%]
> FFFFFFFF                                                                 [ 12%]
> tests/test_plugin_userinput.py EEEE                                      [ 12%]
> tests/test_streamlink_api.py FFFFF                                       [ 12%]
> tests/test_validate.py FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF [ 13%]
> ```

bastimeyer:
The Windows runner printed an error log after it got cancelled...

Looks like the issue is the monkeypatching of `builtins.repr` that breaks and which doesn't get reset, breaking all following tests:
https://github.com/streamlink/streamlink/blob/0555139a068e7f4588a34b83d5649528bd540d95/tests/test_plugin.py#L383-L391

PR #6592, opened 2025-06-29
ci.github: add free-threaded test runners

bastimeyer:
See #6591

There could possibly be some test failures. I didn't see anything on my local system ever since lxml (built with recent Cython) added support for free-threaded mode.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -3,7 +3,7 @@ codecov:
     require_ci_to_pass: true
     # wait until at all test runners have uploaded a report (see the test job's build matrix)
     # otherwise, coverage failures may be shown while some reports are still missing
-    after_n_builds: 10
+    after_n_builds: 12
 comment:
   # this also configures the layout of PR check summaries / comments
   layout: "reach, diff, flags, files"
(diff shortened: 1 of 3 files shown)


bastimeyer:
Apparently, there are no pre-built wheels available and the automated build fails. Will have a look at this tomorrow...

bastimeyer:
The Python setup of the free-threaded runner on Windows takes a ridiculous amount of time for some reason. The whole run takes 5 times longer than the non-free-threaded ones on the Linux runners (~1min vs ~5min), and the free-threaded Linux runner only takes half a min longer due to its setup and slightly longer test run time (which is expected).

Also no idea about the required test fix. It failed only on Windows... And that was because the `threading.Timer` thread was still running after releasing its inner `threading.Event` by calling its respective inner `threading.Condition.notify_all()`. Considering that this test didn't touch any low level stuff and only accessed some private class attributes, I can't explain this.

But with the longer overall run time, I don't want to merge this yet, I think. I'm going to keep this PR open and will then re-evaluate the situation after a while. I'll probably cherry-pick the test fix though...

PR #6605, opened 2025-07-10
plugins.kick: low latency

lbrooney:
Adds low latency mode to the Kick plugin. Implemented by porting over the functionality from the Twitch plugin without the ad filtering parts. I also ported what I felt were the relevant tests for this feature from the Twitch tests.

### DIFF
--- a/src/streamlink/plugins/kick.py
+++ b/src/streamlink/plugins/kick.py
@@ -41,12 +41,11 @@ LOW_LATENCY_MAX_LIVE_EDGE = 2
 
 @dataclass
 class KickHLSSegment(HLSSegment):
-    prefetch: bool
+    prefetch: bool = False
 
 
 class KickM3U8(M3U8[KickHLSSegment, HLSPlaylist]):
-    def __init__(self, *args, **kwargs) -> None:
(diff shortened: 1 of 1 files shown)


bastimeyer (commented):
Thanks. Apart from some minor nitpicks, this looks good.

Having duplicate code for this isn't ideal considering that Kick is basically using the same custom HLS additions as Twitch, as they're using Amazon's infrastructure, but I guess there's not much that can be done here. Both plugins relying on a common HLS implementation doesn't make sense either.

The plugin file could have a low latency note added to its metadata in the header, but it's not necessary, as the plugin argument already covers this. The Twitch plugin has a note there because it links to the dedicated Twitch plugin page in the docs.

lbrooney:
Thanks for the suggestions!

When I was porting over the feature, I did wonder if there were a way to make a generic Amazon IVS solution, but that is far out of my knowledge or expertise to comment deeply on.

I was unsure whether this warranted a seperate page, but as you said the plugin argument likely sufficiently covers this.

bastimeyer:
Thanks!

PR #6612, opened 2025-07-17
tests: req pytest>=8.4.0, load plugins explicitly

bastimeyer:
Add the `--disable-plugin-autoload` pytest option (pytest >=8.4.0) and explicitly load pytest plugins which are required by Streamlink's test suite.

By default, pytest plugins are discovered automatically by checking all installed Python distributions in the current Python environment, which may lead to unwanted pytest plugins being loaded and executed.

----

Random pytest plugins shouldn't be loaded when running the test suite.

1. Some Streamlink packagers have set incorrect test dependencies in their packages (we've removed pytest-asyncio in the past), resulting in those plugins being loaded automatically.
2. Streamlink could be installed in an env that's not clean and where lots of other pytest plugins are installed. For example, I've been casually working on the transition from requests+urllib3 to httpx, and this causes plugins of both to be loaded by default.

Before merging, I'll need to have a look at which pytest versions are packaged in the main Linux distros where Streamlink is packaged. Pytest 8.4.0 is fairly recent, so this might be a problem. I'll do this later...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -67,7 +67,7 @@ jobs:
         shell: bash
         run: |
           export PYTHON_GIL=$(python -c 'import sysconfig;print(1 - sysconfig.get_config_var("Py_GIL_DISABLED"))')
-          pytest -r a --cov --cov-branch --cov-report=xml --durations 10
+          pytest --trace-config -r a --cov --cov-branch --cov-report=xml --durations 10
       - name: Upload coverage data
         if: github.event_name != 'schedule'
         continue-on-error: ${{ matrix.continue || false }}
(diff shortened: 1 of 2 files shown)


bastimeyer:
Going to merge this now after 3 months of waiting...

I'm not familiar with Fedora's release schedule, but they are still holding on to pytest 8.3.5 for some reason, probably because they consider it a big breaking change (upstream 8.4.0 lists breaking changes that should be irrelevant for almost all projects using pytest, unless neglected):

- https://bugzilla.redhat.com/show_bug.cgi?id=2369863
- https://docs.pytest.org/en/stable/changelog.html#pytest-8-4-0-2025-06-02

The other distros are small ones which have lots of outdated Python packages.

PR #6621, opened 2025-07-30
Added DRM Support!

kumaraguru1735:
This pull request introduces two new arguments to enable encrypted DASH media stream decryption using Common Encryption (CENC) keys via FFmpeg:

* `-decryption_key`:
  Accepts a hexadecimal string representing a CENC decryption key to decrypt media received from DASH streaming.
  If only one key is provided, it will be applied to both video and audio tracks.
  Example usage:

  ```bash
  -decryption_key "0123456789abcdef0123456789abcdef"
  ```

* `-decryption_key_2`:
  Optional second CENC decryption key, intended for the secondary media track (typically audio).
  This allows different decryption keys to be used for video and audio tracks when needed.
  Example usage:

  ```bash
  -decryption_key_2 "fedcba9876543210fedcba9876543210"
  ```

These additions enhance the flexibility of handling encrypted DASH streams and make FFmpeg integration with DRM-enabled media workflows easier.

#### Motivation

Some DASH streams encrypt video and audio tracks using different CENC keys. Previously, there was no option to specify a separate key for audio. This update resolves that limitation and allows full decryption support for both tracks.

#### Notes

* Both keys must be valid 128-bit hexadecimal strings.
* These arguments will be ignored if the stream is not encrypted or if no keys are provided.

### DIFF
--- a/src/streamlink/session/options.py
+++ b/src/streamlink/session/options.py
@@ -200,6 +200,21 @@ class StreamlinkOptions(Options):
           - ``[]``
           - Select a specific audio source or sources when multiple audio sources are available,
             by language code or name, or ``"*"`` (asterisk)
+        * - decryption_key
+          - ``str``
+          - ``None``
+          - Use a CENC decryption key to decrypt media streams received by FFmpeg from DASH streaming.
+          If only one key is provided, it will be used for both video and audio tracks.
+          Use a hexadecimal string as the key.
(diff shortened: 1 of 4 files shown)


bastimeyer:
https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#plugin-rules

> Streamlink won't implement or break any DRM schemes.

Issue #6624, opened 2025-08-01
Building the man page doesn't work with docutils 0.22


heirecka:
### Checklist

- [x] This is a bug report and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [x] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [x] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [x] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Collaboration

- [x] [I will provide feedback should a pull request be opened with a fix for the problem](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-request-feedback)

### Streamlink version

streamlink 7.5.0

### Description

` make --directory=docs man` fails to build with docutils 0.22:

```

Traceback
=========

      File "/usr/lib/python3.13/site-packages/docutils/nodes.py", line 736, in __radd__
        return other + self.children
               ~~~~~~^~~~~~~~~~~~~~~
    TypeError: unsupported operand type(s) for +: 'NoneType' and 'list'
```

The full traceback is attached below.

I thought to report this to Sphinx, but if I comment out streamlink's ext_argparse and ext_plugins extensions in `doc/conf.py` it works (with obvious warnings due to the missing extensions), so I choose to report it here first.
It happens both with Sphinx 8.2.3 + backports for docutils 0.22 and current master of Sphinx.

### Debug/trace log

```text
Versions
========

* Platform:         linux; (Linux-6.15.1-x86_64-with-glibc2.42)
* Python version:   3.13.5 (CPython)
* Sphinx version:   8.3.0
* Docutils version: 0.22
* Jinja2 version:   3.1.6
* Pygments version: 2.19.2

Last Messages
=============

    writing output...

    building [man]: all manpages
    updating environment:
    [new config]
    34 added, 0 changed, 0 removed

    reading sources... [  3%]
    _man

Loaded Extensions
=================

* sphinx.ext.mathjax (8.3.0)
* alabaster (1.0.0)
* sphinxcontrib.applehelp (2.0.0)
* sphinxcontrib.devhelp (2.0.0)
* sphinxcontrib.htmlhelp (2.1.0)
* sphinxcontrib.serializinghtml (2.0.0)
[...]

PR #6699, opened 2025-10-22
Fix docutils 0.22.2 errors

amurzeau:
Hi,

Debian Unstable has now docutils 0.22.

But when building Streamlink with it, there are some errors causing the build to fail:
```
/<<PKGBUILDDIR>>/docs/api/validate.rst:72: WARNING: Transition must be child of <document> or <section>. [docutils]
/<<PKGBUILDDIR>>/docs/install.rst:253: ERROR: Indirect hyperlink target (id="linux-appimages") refers to target "linux appimages", which is a duplicate, and cannot be used as a unique reference. [docutils]
/<<PKGBUILDDIR>>/docs/install.rst:262: ERROR: Indirect hyperlink target "appimages" (id="appimages") refers to target "linux appimages", which is a duplicate, and cannot be used as a unique reference. [docutils]
/<<PKGBUILDDIR>>/docs/install.rst:143: ERROR: Duplicate target name, cannot be used as a unique reference: "linux appimages". [docutils]
/<<PKGBUILDDIR>>/docs/install.rst:145: ERROR: Duplicate target name, cannot be used as a unique reference: "linux appimages". [docutils]
/<<PKGBUILDDIR>>/docs/install.rst:253: ERROR: Duplicate target name, cannot be used as a unique reference: "linux appimages". [docutils]
/<<PKGBUILDDIR>>/docs/install.rst:262: ERROR: Duplicate target name, cannot be used as a unique reference: "linux appimages". [docutils]
```

This can be reproduced by using `pip install docutils==0.22.2` and then building the documentation.

This PR fix the errors and the warning.
For the horizontal line, I think the offending commit on docutils side are:
- Added an assert: https://sourceforge.net/p/docutils/code/9702/
- Changed assert to the error I get with Streamlink: https://sourceforge.net/p/docutils/code/10128/

I'm not sure this is an intended forbidden case according to the commit messages, but now it is released in docutils 0.22 :(

### DIFF
--- a/docs/api/validate.rst
+++ b/docs/api/validate.rst
@@ -69,8 +69,6 @@ for an introduction to this API and a list of examples.
     :func:`validate()` is implemented using the stdlib's :func:`functools.singledispatch` decorator, where more specific
     schemas overload the default implementation with more validation logic.
 
-    ----
-
     By default, :func:`validate()` compares ``value`` and ``schema`` for equality. This means that simple schema objects
     like booleans, strings, numbers, None, etc. are validated here, as well as anything unknown.
 


docs/api/validate.rst
```diff
@@ -69,8 +69,6 @@ for an introduction to this API and a list of examples.
     :func:`validate()` is implemented using the stdlib's :func:`functools.singledispatch` decorator, where more specific
     schemas overload the default implementation with more validation logic.
 
-    ----
-
```
bastimeyer: I don't get why this is suddenly an issue according to their dtd. Stupid and sad...

bastimeyer (commented):
Thanks.

[As commented on in a related issue one and a half months ago](https://github.com/streamlink/streamlink/issues/6624#issuecomment-3145189738), `docutils >=0.22` is currently not supported by the latest versions of `sphinx` and `myst-parser` (still no new versions since then), hence why it's not pulled by Python package management tools like pip. If you're using docutils 0.22 with these other docs dependencies, then you might break other things as well, so please be aware. I have not checked this myself yet because of these version restrictions.

This doesn't mean though that we shouldn't fix issues for this soon-to-be-used docutils version in our docs.

bastimeyer:
Thanks, appreciated!

PR #6716, opened 2025-11-10
cli: add --ffmpeg-validation-timeout option

oliviernes:
# Add --ffmpeg-validation-timeout CLI option to control FFmpeg version validation

## Summary
This PR adds a new CLI argument `--ffmpeg-validation-timeout` that allows users to configure the timeout used when validating the FFmpeg binary with `ffmpeg -version`.

On slower systems, such as low-power SBCs (e.g., Nanopi-NEO) or heavily loaded machines, the default 4.0-second timeout can cause false validation failures and warnings such as:

[stream.ffmpegmux][error] Could not validate FFmpeg!
[stream.ffmpegmux][warning] No valid FFmpeg binary was found.
[stream.ffmpegmux][warning] Muxing streams is unsupported!

These failures can result in audio not being muxed properly in recorded streams.

## Changes in this PR
- Added new CLI argument `--ffmpeg-validation-timeout`.
- Mapped it to the session option `ffmpeg-validation-timeout`.
- Updated `FFMPEGMuxer._resolve_command()` to use the session-provided timeout.
- Added **tests** verifying both default (4.0s) and custom timeout behavior.
- Added a **changelog entry** in `docs/changelog.md`.

## Default behavior
If the option is not specified, the timeout remains at 4.0 seconds.

## Checklist
- [x] Added new CLI argument
- [x] Mapped argument to session options
- [x] Updated FFMPEG validation logic to use session-provided timeout
- [x] Verified proper debug output and audio muxing behavior
- [x] Added tests for default and custom timeout behavior
- [x] Added changelog entry in docs/changelog.md
- [x] Maintains backward compatibility

### DIFF
--- a/src/streamlink_cli/argparser.py
+++ b/src/streamlink_cli/argparser.py
@@ -1301,6 +1301,18 @@ def build_parser():
             Enable the `-start_at_zero` FFmpeg option when using --ffmpeg-copyts.
         """,
     )
+    transport_ffmpeg.add_argument(
+        "--ffmpeg-validation-timeout",
+        type=float,
+        metavar="SECONDS",
+        help="""
+            Timeout in seconds for FFmpeg version validation.
+
+            Default is 4.0.
+
+            Increase this on low-power systems if FFmpeg startup is slow.
+        """,
+    )
 
     http = parser.add_argument_group("HTTP options")
     http.add_argument(
@@ -1545,6 +1557,7 @@ _ARGUMENT_TO_SESSIONOPTION: list[tuple[str, str, Callable[[Any], Any] | None]] =
     ("webbrowser_cdp_port", "webbrowser-cdp-port", None),
     ("webbrowser_cdp_timeout", "webbrowser-cdp-timeout", None),
     ("webbrowser_headless", "webbrowser-headless", None),
+    ("ffmpeg_validation_timeout", "ffmpeg-validation-timeout", None),
 ]
 
 
--- a/tests/stream/test_ffmpegmux.py
+++ b/tests/stream/test_ffmpegmux.py
@@ -121,6 +121,35 @@ class TestCommand:
             ("ffmpegmux", "debug", " bar"),
         ]
 
+    @pytest.mark.parametrize(
+        ("timeout_value", "expected_timeout"),
+        [
+            pytest.param(None, 4.0, id="default"),
+            pytest.param(9.5, 9.5, id="custom"),
+        ],
+    )
+    def test_validate_timeout(self, session: Streamlink, timeout_value, expected_timeout):
+        session.options.update({
+            "ffmpeg-no-validation": False,
+            "ffmpeg-validation-timeout": timeout_value,
+        })
+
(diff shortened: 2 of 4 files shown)


src/streamlink_cli/argparser.py
```diff
@@ -1545,6 +1554,7 @@ def build_parser():
     ("webbrowser_cdp_port", "webbrowser-cdp-port", None),
     ("webbrowser_cdp_timeout", "webbrowser-cdp-timeout", None),
     ("webbrowser_headless", "webbrowser-headless", None),
+    ("ffmpeg_validation_timeout", "ffmpeg-validation-timeout", None),
```
bastimeyer: `ffmpeg-validation-timeout` needs to be added to the actual Streamlink session options at `streamlink.session.options.StreamlinkOptions`.

tests/stream/test_ffmpegmux.py
```diff
@@ -121,6 +121,48 @@ def run(self):
...
+        with (
+            patch("streamlink.stream.ffmpegmux.which", return_value="/usr/bin/ffmpeg"),
+            patch("streamlink.stream.ffmpegmux.FFmpegVersionOutput", side_effect=MyFFmpegVersionOutput) as mock_versionoutput,
+        ):
+            result = FFMPEGMuxer.command(session)
+
+        assert result == "/usr/bin/ffmpeg"
+        assert mock_versionoutput.call_args_list == [
+            call(["/usr/bin/ffmpeg", "-version"], timeout=9.5),
+        ]
```
bastimeyer: Instead of adding a second test case, different timeout values (default and custom) should be parametrized (`@pytest.mark.parametrize(...)`).

Btw, there's an additional blank line below that causes a linting error.

bastimeyer (changes requested):
Thanks for the PR. Appreciated...

I've added some comments where changes are required.

Is it intentional that the commit does have different author and committer data? Both seem to be yours, but with different GH accounts attached.

oliviernes:
Thank you for the review and feedback. I’ve fixed the author/committer metadata and updated the code based on your suggestions.

tests/stream/test_ffmpegmux.py
```diff
@@ -121,6 +121,33 @@ def run(self):
...
+    @pytest.mark.parametrize("timeout_value,expected_timeout", [
+        (None, 4.0),  # default
+        (9.5, 9.5),   # custom
+    ])
+    def test_validate_timeout(self, session: Streamlink, timeout_value, expected_timeout):
+        session.options.update({
+            "ffmpeg-no-validation": False,
+        })
+        if timeout_value is not None:
+            session.options.set("ffmpeg-validation-timeout", timeout_value)
```
bastimeyer: That if-block is unnecessary. Just set the `ffmpeg-validation-timeout` session option in the dict above.

oliviernes:
Thanks for catching that — updated as suggested

oliviernes:
I’ve applied the final suggested cleanup: removed the redundant if timeout is None check and set the default directly in the _resolve_command signature.

Thanks again for the review.

PR #6722, opened 2025-11-11
stream.hls: warn on playlist_sequence access

bastimeyer:
`HLSStreamWorker.playlist_sequence` was moved to `SegmentedStreamWorker.sequence` in 46565ca3.

Add backwards compatibility with warning messages, in case third-party plugins override *private* HLSStream interfaces like the worker.


bastimeyer:
This is not going to have a deprecation phase and will be removed at any time.

Issue #6548, opened 2025-06-04
Allow specifying Content-Type headers for HTTP streams


Lense:
### Checklist

- [x] This is a feature request and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [x] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [x] [I have checked the list of open and recently closed feature requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22feature+request%22)

### Collaboration

- [x] My request is genuine and in the best interest of the project
- [x] [I will provide feedback should a pull request be opened with a new feature implementation](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-request-feedback)

### Description

Use case: I use streamlink to play videos on my TV by creating an HTTP server with `--player-external-http` and then sending a UPnP message to the TV telling it to render the stream. This does not work out of the box because the TV rejects HTTP servers with Content-Type headers that it doesn't support.

I have gotten streams I care about to work with my TV by hard-coding `video/mp4`:
```
--- src/streamlink_cli/output/http.py
+++ src/streamlink_cli/output/http.py
@@ -90,7 +90,7 @@ class HTTPOutput(Output):
         try:
             conn.send(b"HTTP/1.1 200 OK\r\n")
             conn.send(b"Server: Streamlink\r\n")
-            conn.send(b"Content-Type: video/unknown\r\n")
+            conn.send(b"Content-Type: video/mp4\r\n")
             conn.send(b"\r\n")
         except OSError as err:
             raise OSError("Failed to write data to socket") from err
```

But obviously that's not a general solution.

It would be great to have a CLI option to set the Content-Type.

bastimeyer:
The issue is that Streamlink (and its CLI) doesn't know anything about the stream's format (MPEG transport stream, MPEG-4 container, etc) or any video/audio codecs being used. This is also the reason why there's no `{type}`/`{extension}` variable for the `--output` path template.

This isn't directly related to #4902, but this issue also talks about the missing stream metadata.

An option for the CLI's HTTP server doesn't make sense IMO. We shouldn't add tons and tons of CLI arguments. Changing the hardcoded value is also not a solution. The MIME-type must be determined by the stream format.

[...]

PR #6731, opened 2025-11-18
Add permissive CORS header to http stream

rhelmot:
I am trying to build an app on top of streamlink which uses a javascript library to stream onto a webpage with some context-specific graphics and controls. Currently, it doesn't work because of CORS! This fixes that.


bastimeyer:
I'm sorry, but I don't want to merge this.

Streamlink-CLI's `--player-external-http` is not meant for front-facing websites. It's just another player transport mechanism, like `--player-http` or the other transport methods. Only one request can be served at a time, which is intentional. The `Content-Type` header also doesn't reflect the actual content mime-type (see #6548). And on top of this, the media formats that are output are usually not compatible with regular web browsers.

If you still want to use Streamlink for your web application, then you should put it behind a local forward-proxy which adds your desired headers, or better, implement your own HTTP server logic with a Streamlink backend that reads from its stdout or fifo output.

PR #6747, opened 2025-12-15
stream.hls: Apply segment ignore_name regardless of extension

qwer-lives:
Minor change to match ignore_name patterns regardless of segment extension (before this it only matched .ts segments).

### DIFF
--- a/src/streamlink/stream/hls/hls.py
+++ b/src/streamlink/stream/hls/hls.py
@@ -96,8 +96,7 @@ class HLSStreamWriter(SegmentedStreamWriter[HLSSegment, Response]):
         ignore_names = {*options.get("hls-segment-ignore-names")}
         if ignore_names:
             segments = "|".join(map(re.escape, ignore_names))
-            # noinspection RegExpUnnecessaryNonCapturingGroup
-            self.ignore_names = re.compile(rf"(?:{segments})\.\w+", re.IGNORECASE)
+            self.ignore_names = re.compile(segments, re.IGNORECASE)
 
     @staticmethod
     def num_to_iv(n: int) -> bytes:
--- a/tests/mixins/stream_hls.py
+++ b/tests/mixins/stream_hls.py
@@ -74,17 +74,16 @@ class Tag(HLSItemBase):
 
 
 class Segment(HLSItemBase):
-    def __init__(self, num, title=None, duration=None, path_relative=True, ext=None):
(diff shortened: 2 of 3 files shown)


bastimeyer (commented):
Thanks for the PR.

The `--hls-segment-ignore-names` stuff is rather old (added in 39d81e3327057684846e5523a11c707c30fbe0e5) and predates the segment filtering implementation. I've talked about it a couple of times already on the issue tracker that I'd like to replace or remove it, because it's pretty bad. The reason why this ignore-names implementation is not ideal is that it receives user input from the CLI via a comma-separated string, and it's also very limiting for selecting specific segments because of this regex idea. All of that is irrelevant for your PR though.

What is relevant though is that both the current implementation and your PR are not correct and require a change. The HLS specification (RFC 8216) doesn't say anything about media segment URIs/URLs. It only requires HLS multivariant+media playlists to either have an `.m3u` or `.m3u8` URL path suffix, or the HTTP server must send the `application/vnd.apple.mpegurl`/`audio/mpegurl` Content-Type response header. Media segments are unspecified, meaning the regex should not contain a suffix, not even a "wildcard" one.

There's also an issue with a changed test which I've annotated.

qwer-lives:
Thanks for the comment - I didn't notice that the ".*" was actually useful :) I applied the changes.

Also, I didn't realize this feature was that old. This PR is mostly for my convenience since I use this to filter out segments when using the HLS plugin (for platforms that don't have their own specific plugin). Feel free to ignore it if you're going to rewrite / remove the feature in the near future.

bastimeyer:
Apologies for the late response.

I'm a little bit torn here, because changing the regex and removing the incorrect suffix results in some unintended side-effects that change the behavior of this feature, because the regex is applied to the whole `segment.uri` attribute. Previously, this **implicitly** matched the URL path due to the `.ts` suffix, but without a forced suffix, it can also match the hostname for example, or the scheme if the user just sets the `https` value.

As said, this feature is implemented poorly. A proper filtering mechanism shouldn't be based on the `segment.uri` attribute alone (and as a whole). The segment filtering should allow users to select specific segments based on various segment attributes, as well as individual attributes of the parsed URI. This would need to be implemented as a new feature, and the current one would need to be deprecated. But all of this is out of scope of this PR.

Since this is such a niche feature and one that should be replaced anyway, I think I'm going to merge your changes, but I will have another thought about it later today.

PR #6754, opened 2025-12-22
build: remove wheel from build-system.requires

bastimeyer:
Setuptools vendors `wheel` logic:
https://setuptools.pypa.io/en/latest/history.html#v70-1-0

`wheel` doesn't belong into the `build-system.requires` list, and also not into `dependency-groups.build`.

This should've been removed ages ago...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,7 +131,6 @@ build = [
   { include-group = "dev" },
   "shtab",
   "build",
-  "wheel",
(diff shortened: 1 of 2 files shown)


bastimeyer:
:heavy_check_mark: https://github.com/streamlink/streamlink/actions/runs/20442319623/job/58737950039?pr=6754#step:7:1

PR #6764, opened 2026-01-02
plugins.unwebtv: new plugin

Exafto:
# plugin: webtv.un.org
Adds support for the United Nations webTV site, where a 24/7 livestream is hosted and important events (like general assemblies) are streamed and later stored in VODs.

## Introduction:
Hello and happy new year! This is my second pull request, and I tried to follow community standards. Please let me know if I missed anything.

## Community guidelines compliance:
As stated in (https://webtv.un.org/en/copyright_use), and later in (https://www.un.org/en/about-us/terms-of-use):
"The United Nations grants permission to Users to visit the Site and to download and copy the information, documents and materials (collectively, “Materials”) from the Site for the User’s personal, non-commercial use (...)"

## Execution examples:
- `streamlink "https://webtv.un.org/en" best` plays the 24/7 livestream located on the homepage.
- `streamlink "https://webtv.un.org/en/asset/k14/k143ucuid9" best` plays a specific VOD.

## Notes for Reviewers:

This site uses a Kaltura backend which requires a valid Session Key (KS) to unlock streams. Without generating this key via the startWidgetSession API, valid HLS URLs return 404 Not Found. The plugin implements the following logic:
1.  **Generating a Session Key:** It uses the UN's Kaltura Partner ID (`2503451`) to request a session token (KS) from the API.
2.  **Auto-Detecting Live Streams:** The homepage contains multiple video IDs. The plugin scrapes all of them and tries them in order until the active livestream is found.
3.  **VOD Support:** It scrapes the page source to find the specific Entry ID needed to generate the manifest.

## Validation:

- [x]Code style checked with ruff.
- [x]Plugin tests passed locally.

### DIFF
--- a/src/streamlink/plugins/unwebtv.py
+++ b/src/streamlink/plugins/unwebtv.py
@@ -1,0 +1,115 @@
+import logging
+import re
+from streamlink.plugin import Plugin, pluginmatcher
+from streamlink.stream.hls import HLSStream
+
+log=logging.getLogger(__name__)
+
+@pluginmatcher(re.compile(r"https?://webtv\.un\.org/"))
+class UNWebTV(Plugin):
+    """
+    $description Global video streaming network of the United Nations.
+    $url webtv.un.org
+    $type live, vod
+    $notes Content is free and not geo-restricted.
+    """
+    
+    def _get_ks(self,partner_id):
+        """
+        Generates a Kaltura Session (KS) key.
+        This is required to unlock the stream URL.
+        """
+        log.debug(f"Generating Widget Session for Partner ID: {partner_id}")
+        try:
+            # widgetId is typically "_{partner_id}" for public sessions
+            url = "https://cdnapisec.kaltura.com/api_v3/service/session/action/startWidgetSession"
+            params = {
+                "widgetId": f"_{partner_id}",
+                "format": "1"  # JSON response
+            }
+            # calling the Kaltura API
+            res = self.session.http.get(url, params=params).json()
+            ks = res.get("ks")
+            if ks:
+                log.debug(f"Generated KS: {ks[:10]}...")
+                return ks
+            else:
+                log.warning("Kaltura API did not return a session key (KS).")
+        except Exception as e:
+            log.warning(f"Failed to generate session key: {e}")
+        return None
+
+    def _get_streams(self):
+        log.debug(f"Fetching page: {self.url}")
+        res = self.session.http.get(self.url)
+
+        # 1. extract partner ID (default: 2503451)
+        partner_id="2503451"
+        p_id_match=re.search(r'partnerId["\']?\s*[:=]\s*["\']?(\d+)', res.text)
+        if p_id_match:
+            partner_id=p_id_match.group(1)
+
+        # 2. get the Session Key (KS)
+        # without this, even valid IDs will return 404 Not Found
+        ks=self._get_ks(partner_id)
+
+        # 3. find ALL video (entry) IDs
+        # we look for standard Kaltura IDs (1_xxxx or 0_xxxx)
+        # we use a set to avoid duplicates
+        entry_ids=[]
+        seen_ids=set()
+
+        # find all 1_xxxx or 0_xxxx IDs in the source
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/unwebtv.py
```diff
@@ -0,0 +1,115 @@
...
+                log.debug(f"Generated KS: {ks[:10]}...")
+                return ks
+            else:
+                log.warning("Kaltura API did not return a session key (KS).")
+        except Exception as e:
+            log.warning(f"Failed to generate session key: {e}")
+        return None
+
+    def _get_streams(self):
+        log.debug(f"Fetching page: {self.url}")
```
bastimeyer: This log message is not useful. Please remove it.

src/streamlink/plugins/unwebtv.py
```diff
@@ -0,0 +1,115 @@
...
+        ks=self._get_ks(partner_id)
+
+        # 3. find ALL video (entry) IDs
+        # we look for standard Kaltura IDs (1_xxxx or 0_xxxx)
+        # we use a set to avoid duplicates
+        entry_ids=[]
+        seen_ids=set()
+
+        # find all 1_xxxx or 0_xxxx IDs in the source
+        matches=re.findall(r'["\']([0-1]_[a-zA-Z0-9]{8,})["\']', res.text)
```
bastimeyer: Same here with the validation schema.

src/streamlink/plugins/unwebtv.py
```diff
@@ -0,0 +1,115 @@
...
+            log.error("No Entry IDs found to play.")
+            return
+
+        log.info(f"Found {len(entry_ids)} potential video IDs. Trying them...")
+
+        # 4. try to play the found IDs using brute force
+        # on the homepage, there are many IDs. The main player (for the 24/7 livestream) is usually the first valid one.
+        # on an asset page, the specific video ID is usually in this list too.
+        
+        for entry_id in entry_ids:
```
bastimeyer: That doesn't seem like a good idea. A plugin should never guess stream URLs. This delays the stream start tremendously, because lots of HTTP requests will be made sequentially.

The used `entryId` appears to be included in some additional resource URLs in the HTML response, for example `<link rel="image_src" href="..."/>`, which can also easily be extracted via XPath.

src/streamlink/plugins/unwebtv.py
```diff
@@ -0,0 +1,115 @@
...
+            # construct the API URL
+            #we append /ks/{ks} if we have one
+            kaltura_url = (
+                f"https://cdnapisec.kaltura.com/p/{partner_id}/sp/{partner_id}00/"
+                f"playManifest/entryId/{entry_id}/format/applehttp/protocol/https/a.m3u8"
+            )
```
bastimeyer: This should also be a class member, because URL constants like this don't belong into functions/methods.

tests/plugins/test_unwebtv.py
```diff
@@ -0,0 +1,23 @@
...
+        "https://webtv.un.org/fr/asset/k1x/k1x2345",
+        "https://webtv.un.org/es/asset/k1y/k1y6789",
+        # Some URLs might have query parameters
+        "https://webtv.un.org/en?foo=bar",
+    ]
+
+    should_not_match = [
+        "https://www.un.org/",
+        "https://media.un.org/",
+        "https://webtv.other.org/",
```
bastimeyer: Plugins already have a generic negative test. This one is not useful.

Exafto:
Thank you for the detailed review!

I understand that I might have caused some confusion, and I apologise. To clarify, I used gen-AI to help me "scaffold" project structure (i.e. where to set up new files, which streamlink tools to use etc), since I am not familiar with the codebase, nor with contributing to open-source. In hindsight, I should have been more careful. My commit message is an oversight; I initially wrote the plugin to support live content and later added VOD support, which is why I used "rewrite".

**Next steps:** I will get started on implementing your notes. Regarding the plugin request issue, I will definitely keep the contributing guidelines in mind in the future.

I appreciate your insight!

bastimeyer:
Since there hasn't been any sign of activity, I'm going to close this.
You can try in another PR if you want, but please respect the gen-AI rules we've added.

PR #6792, opened 2026-01-23
session.http: add --http-cookies-file

### LINKED ISSUE #3370: Support Cookies.txt
## Feature Request

In order to prevent blocking by Youtube, .. support for a cookie jar would be helpful.

- [ ] This is not a bug report, feature request, or plugin issue/request.
- [X] I have read the contribution guidelines.

### Description

Youtube-DL supports this and has a class to load&save cookies in the "cookies.txt" format. It's derived from MozillaCookieJar:
https://github.com/ytdl-org/youtube-dl/blob/f9f9699f2fa87ec54d59b1834cc56adb6147fec6/youtube_dl/utils.py#L2739

### Implementation

The initialization is pretty easy:
```
self.cookiejar = YoutubeDLCookieJar(cookie_file_path)
if os.access(cookie_file_path, os.R_OK):
  self.cookiejar.load(ignore_discard=True, ignore_expires=True)
```

In order to add the cookies to a request, just use
`self.cookiejar.add_cookie_header(req)`

bastimeyer:
Resolves #3370

This implements `--http-cookies-file`, which can be set multiple times.

The supported format is the Netscape HTTP Cookie File format which is also used by curl for example. This is implemented via the stdlib's `http.cookiejar.MozillaCookieJar`.
https://docs.python.org/3/library/http.cookiejar.html#http.cookiejar.MozillaCookieJar

### DIFF
--- a/src/streamlink_cli/main.py
+++ b/src/streamlink_cli/main.py
@@ -983,7 +983,10 @@ def setup(parser: ArgumentParser) -> None:
     log_current_versions()
     log_current_arguments(streamlink, parser)
 
-    setup_session_options(streamlink, args)
+    try:
+        setup_session_options(streamlink, args)
+    except Exception as err:
+        raise StreamlinkCLIError from err
 
     setup_signals()
 
--- a/tests/cli/main/test_setup_session_options.py
+++ b/tests/cli/main/test_setup_session_options.py
@@ -1,0 +1,31 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import pytest
+
+import streamlink_cli.main
+
+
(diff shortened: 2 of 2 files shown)


bastimeyer:
I intentionally avoided cookiejar in the CLI argument name, to avoid confusion with curl's `--cookie-jar` option where data is written to after running curl. Curl treats `--cookie` special when no `=` is found and then interprets the value as a file path. We instead use a separate parameter for this.

bastimeyer:
For some strange reason, the Windows py 3.14 runner fails immediately upon VM boot. Will give this another go later today. If it keeps failing, I'll open a new PR.

PR #6804, opened 2026-01-30
tests.stream: refactor ffmpegmux tests

bastimeyer:
No idea why, but the Ubuntu py3.14 test runner kept failing in PR #6803 which has completely unrelated changes:
https://github.com/streamlink/streamlink/actions/runs/21511543737/job/61979791686

This looks like the `caplog.records` data wasn't reset by pytest from other test cases run prior to that specific test. Otherwise it doesn't make sense, as there's no segmented stream being run in that test, and the failure is about log records with "Closing writer thread" debug messages being incorrectly captured.

We've had some flaky tests before, but I haven't seen this one yet...

This replaces unittest.mock.patch with pytest.MonkeyPatch and also clears the caplog.records list in the failing test before it makes an assertion about no additional new log records.

### DIFF
--- a/tests/stream/test_ffmpegmux.py
+++ b/tests/stream/test_ffmpegmux.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import subprocess
 from typing import TYPE_CHECKING
-from unittest.mock import call, patch
+from unittest.mock import Mock, call
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


bastimeyer:
Still failing on master:
https://github.com/streamlink/streamlink/actions/runs/21512512695/job/61982776579

PR #6806, opened 2026-01-30
tests.stream: fix thread termination

bastimeyer:
See #6804 
https://github.com/streamlink/streamlink/actions/runs/21512512695/job/61983560185

I'm not 100% sure yet where the flaky tests are coming from. The issue is that pytest's caplog fixture captures debug log messages from other test cases, or that it doesn't clear the records properly.

The record clearing is rather unlikely. If that's the case, then it's a bug in pytest. But considering that we've never had flaky tests like this, it's even more unlikely.

To me, this looks like we don't correctly terminate and await the termination of the worker and writer threads of HLS stream tests that are run prior to the FFMPEGMux tests where the log record assertions fail. The worker/writer threads appear to be bleeding into the failing tests where they pollute the log records. This is strange though, because we call `Thread.join(TIMEOUT_AWAIT_CLOSE)` on each test case where HLS streams are run. DASH streams are not actually run, because we simply iterate segments there from the main thread.

I've now added explicit `close()` calls and also added explicit assertions about the thread state, so that the tests should fail where they're supposed to fail if the worker/writer threads are still alive.

### DIFF
--- a/tests/mixins/stream_hls.py
+++ b/tests/mixins/stream_hls.py
@@ -249,7 +249,12 @@ class TestMixinStreamHLS(unittest.TestCase):
         thread.reader.writer.join(timeout)
         thread.reader.worker.join(timeout)
         thread.join(timeout)
+        assert not thread.reader.writer.is_alive()
+        assert not thread.reader.worker.is_alive()
+        assert not thread.is_alive()
         assert self.thread.reader.closed, "Stream reader is closed"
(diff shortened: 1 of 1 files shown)


bastimeyer:
Five CI reruns without issues... Going to merge now.

Should the tests still fail, then we'll at least know where with these changes.

It would also be good to know the exact conditions for these test failures, but it's hard to tell with GH CI runner VMs and their host systems.

PR #6810, opened 2026-02-02
plugin: move match check to url setter, fix typing

bastimeyer:
Raise `PluginError` when the `url` attr is set and it doesn't match any of the plugin's matchers, not just once upon class initialization.

Remove `None` type from `Plugin.matcher` and `Plugin.match`: Since the input URL must always match, it is pointless to annotate these attributes as potentially being `None`. This causes lots of headaches in plugin implementations because it requires type-narrowing. Initialize them with `None`, so plugin tests without matchers don't have to call `hasattr()`.

----

This fixes 120 typing errors in Streamlink's plugins, which are not checked by mypy by default, as this requires typing annotations (unless a custom config is set). ty checks this by default because of its Unknown type.

### DIFF
--- a/src/streamlink/plugin/plugin.py
+++ b/src/streamlink/plugin/plugin.py
@@ -196,7 +196,7 @@ def parse_params(params: str | None = None) -> dict[str, Any]:
 
 
 class Matcher(NamedTuple):
-    pattern: re.Pattern
+    pattern: re.Pattern[str]
     priority: int
     name: str | None = None
 
@@ -231,7 +231,7 @@ class Matchers(_MCollection[Matcher]):
 
 
 class Matches(_MCollection[Union[re.Match, None]]):
-    def update(self, matchers: Matchers, value: str) -> tuple[re.Pattern | None, re.Match | None]:
+    def update(self, matchers: Matchers, value: str) -> tuple[re.Pattern[str] | None, re.Match[str] | None]:
         matches = [(matcher, matcher.pattern.match(value)) for matcher in matchers]
 
         self.clear()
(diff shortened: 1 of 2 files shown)


bastimeyer:
These plugins update the `url` attribute dynamically:

```
$ git grep -nFE 'self.url = ' -- src/streamlink/plugins
src/streamlink/plugins/atresplayer.py:30:        self.url = update_scheme("https://", f"{self.url.rstrip('/')}/")
src/streamlink/plugins/bbciplayer.py:101:        self.url = urlunparse(urlparse(self.url)._replace(scheme="https"))
src/streamlink/plugins/dogan.py:108:            self.url = default_service_url
src/streamlink/plugins/filmon.py:189:            self.url = urlunparse(parsed._replace(path=parsed.path.replace("/channel/", "/tv/")))
src/streamlink/plugins/mangomolo.py:42:        self.url = update_scheme("https://", player_url)
src/streamlink/plugins/okru.py:70:        self.url = url
src/streamlink/plugins/oneplusone.py:68:        self.url = url
src/streamlink/plugins/steam.py:230:            self.url = self._watch_broadcast_url.format(steamid=steamid)
src/streamlink/plugins/streann.py:151:            self.url = iframes[0].attrib.get("src")
src/streamlink/plugins/twitcasting.py:180:        self.url = update_qsd(url, params or {})
src/streamlink/plugins/vk.py:62:            self.url = f"{parsed_url.scheme}://{parsed_url.netloc}/{true_path}"
src/streamlink/plugins/vk.py:69:            self.url = self.session.http.get(
src/streamlink/plugins/youtube.py:91:            self.url = urlunparse(parsed._replace(scheme="https", netloc="www.youtube.com"))
src/streamlink/plugins/youtube.py:93:            self.url = self._url_canonical.format(video_id=self.match["video_id"])
src/streamlink/plugins/youtube.py:95:            self.url = self._url_canonical.format(video_id=self.match["video_id"])
src/streamlink/plugins/youtube.py:97:            self.url = self._url_channelid_live.format(channel_id=self.match["live"])
src/streamlink/plugins/youtube.py:99:            self.url = urlunparse(parsed._replace(scheme="https"))
src/streamlink/plugins/youtube.py:379:            self.url = self._url_canonical.format(video_id=video_id)
```

PR #6819, opened 2026-02-06
utils: deprecate package exports

bastimeyer:
This deprecates all (re-)exports in the `streamlink.utils` package, for multiple reasons:

1. only a subset of stuff from the submodules is exported, which is inconsistent and nonsensical
2. this can lead to cyclic imports, as importing `streamlink.utils.A` for example can automatically import `streamlink.utils.B` if `B` was imported and re-exported in `streamlink.utils`. So if one of the submodules imports a module up the stack while it is being executed by the importlib, for example `streamlink.logger`, then we have a cyclic import. I've just run into this while refactoring the logger module.
3. re-exporting means that unused stuff will always be loaded, which is unnecessary. All modules in the entire `streamlink` and `streamlink_cli` packages already import from the full specs, as it should be. I'm deprecating this instead of removing it, in case this breaks third-party plugins. No docs for this deprecation, as it's something internal.

### DIFF
--- a/src/streamlink/utils/__init__.py
+++ b/src/streamlink/utils/__init__.py
@@ -1,24 +1,24 @@
-from streamlink.utils.cache import LRUCache
-from streamlink.utils.data import search_dict
-from streamlink.utils.module import load_module
-from streamlink.utils.named_pipe import NamedPipe
-from streamlink.utils.parse import parse_html, parse_json, parse_qsd, parse_xml
-from streamlink.utils.url import absolute_url, prepend_www, update_qsd, update_scheme, url_concat, url_equal
+# ruff: noqa: RUF067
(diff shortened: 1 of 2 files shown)


bastimeyer:
```
$ git checkout 9180b8fc
$ python -c 'import sys,streamlink;print(f"{len(sys.modules)=}");streamlink.streams("twitch.tv/CHANNEL");print(f"{len(sys.modules)=}")'
len(sys.modules)=487
len(sys.modules)=795
```

```
$ git checkout ed4cae63
$ python -c 'import sys,streamlink;print(f"{len(sys.modules)=}");streamlink.streams("twitch.tv/CHANNEL");print(f"{len(sys.modules)=}")'
len(sys.modules)=493
len(sys.modules)=795
```

PR #6834, opened 2026-02-21
ci.github: get run-urls of preview-builds

bastimeyer:
GitHub now supports getting the run-url of dispatched workflows from their API, so let's update the preview-build workflow and include the URLs of the specific runs that include the preview-build artifacts, instead of just listing all runs.

----

- https://github.blog/changelog/2026-02-19-workflow-dispatch-api-now-returns-run-ids/
- https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event


bastimeyer:
Untested changes, but I doubt there will be an issue :crossed_fingers: 
We only run this workflow when certain files were changed on master, so let's see.

PR #6862, opened 2026-03-12
session.http: support SO_BINDTODEVICE on *nix

### LINKED ISSUE #6861: http_session: interface option binds to an ip address instead of a network interface
### Checklist

- [x] This is a bug report and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [x] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [x] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Collaboration

- [x] [I will provide feedback should a pull request be opened with a fix for the problem](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-request-feedback)

### Streamlink version

streamlink 8.2.1+2.g0144b824

### Description

When using the --interface cli option with an actual interface (ex. `streamlink --interface eno1 ...`) it just hangs and will return a name resolution error.

When setting the interface here
 https://github.com/streamlink/streamlink/blob/0144b824f51b5099e83bf9a9155078e447e5e007/src/streamlink/session/http.py#L120
the code actually sets the source address, and not the interface.
Setting the interface would look something like this:
```py
adapter.poolmanager.connection_pool_kw.update(socket_options=[(socket.SOL_SOCKET, socket.SO_BINDTODEVICE, interface.encode())])
```

I encountered this because I have two interfaces with the same IP address, and therefore cannot tell streamlink to bind to one of them.

I can open a PR with this change if needed, but was more wondering if this was intentional, or maybe a `--address` cli option should be added to keep the option to bind to a specific IP.

I also have not tested if this work on other platforms than Linux.

### Debug/trace log

```text
Not really needed.
```

bastimeyer:
In addition to binding to IP addresses, allow non-Windows systems to select the network interface by name. Also add support for prefixes, similar to curl's --interface syntax.

----

Resolves #6861 

@Nogesma please check and see if this successfully resolves your issue.

### DIFF
--- a/src/streamlink/session/http.py
+++ b/src/streamlink/session/http.py
@@ -5,6 +5,7 @@ import ssl
 import time
 import warnings
 from http.cookiejar import MozillaCookieJar
+from ipaddress import ip_address
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast
 
@@ -12,9 +13,11 @@ import urllib3
 import urllib3.util.connection as urllib3_util_connection
 from requests import Request, Session
 from requests.adapters import HTTPAdapter
+from urllib3.connection import HTTPConnection
 from urllib3.util import create_urllib3_context  # type: ignore[attr-defined]
 
 import streamlink.session.http_useragents as useragents
+from streamlink.compat import is_win32
 from streamlink.exceptions import PluginError, StreamlinkDeprecationWarning
 from streamlink.packages.requests_file import FileAdapter
 from streamlink.utils.parse import parse_json, parse_xml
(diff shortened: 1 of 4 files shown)


bastimeyer:
Quick check:
[`urllib3.connection.HTTPConnection.default_socket_options`](https://github.com/urllib3/urllib3/blob/2.0.0/src/urllib3/connection.py#L107-L111) exists on `urllib3==2.0.0`, our min version requirement (I previously only checked the latest version).

Nogesma:
Yes this works! Thanks a lot

bastimeyer:
> Yes this works!

Could you please check with and without the prefixes and report back? I currently don't have more than one network interface set up, so testing it myself is a bit annoying. Thanks.
https://streamlink--6862.org.readthedocs.build/cli.html#cmdoption-interface

Nogesma:
Yep it all works fine, with the exception of host! with an ip that's valid on two interfaces, where it will just hang after this `[plugins.twitch][debug] Getting live HLS streams for ...`. But this matches the behavior before this PR, and using if! or ifhost! to specify the interface works fine.

PR #6867, opened 2026-03-19
plugins.youtube: Refactor into package with Deno solver

fl-Henry:
## YouTube Plugin refactoring into a package with Deno solver

The plugin converts the original single-file `youtube.py` into a Python package (`src/streamlink/plugins/youtube/`). The new architecture introduces an extractor chain and a Deno JavaScript solver for YouTube's n-parameter challenge.

---

### Package Structure

```
youtube/
├── __init__.py          # Exports __plugin__ = Youtube
├── plugin.py            # Entry point, URL matching, extractor orchestration
├── extractors.py        # Extractors for different entities
├── deno.py              # Deno subprocess solver for n-challenges
├── structures.py        # Shared dataclasses, protocols, enums, Context singleton
├── youtube.py           # Original plugin kept as fallback
└── solver/
    ├── __init__.py      # Reads bundled JS files as strings
    ├── lib.min.js       # Bundled JS library for the solver (from yt-dlp-ejs)
    └── core.min.js      # Bundled JS core solver logic (from yt-dlp-ejs)
```

---

### n-Parameter Challenge Solver

YouTube embeds a token (`n` parameter) in HLS manifest URL paths (`/n/<token>/`). If unsolved, YouTube returns 403 errors for video segments. The solver works as follows:

1. `VideoExtractor._extract_hls()` detects `/n/<token>/` in each manifest URL path
2. It calls `DenoJCP.solve()`, which fetches the YouTube player JS and builds Deno stdin that inlines `lib.min.js` and `core.min.js` (from [yt-dlp-ejs](https://github.com/yt-dlp/ejs/tree/main))
3. Starts a `deno run` subprocess with the prebuilt stdin
4. Deno runs with network access, npm, remote imports, module cache, and config disabled (`--no-remote`, `--no-npm`, `--cached-only`, `--no-config`, etc.) - only bundled scripts and player source execute
5. Parses JSON output into an `NChallengeOutput` mapping original token -> solved token
6. Returns `NChallengeOutput` to `VideoExtractor`, which replaces the original token in the manifest URL

If [Deno](https://deno.com/) is not installed, the error message explicitly directs the user to the Deno installation page.

---

### Stream Probing and Fallback (`plugin.py`)

After collecting HLS manifest URLs, `_check_streams()` parses each as a variant playlist via `HLSStream.parse_variant_playlist()` and probes reachability by opening the first variant and reading 64 bytes (2-second timeout). Incorrect n-challenge solutions trigger 403 for every video segment. Unreachable streams are discarded with a warning.

If the entire extractor chain fails, the plugin falls back transparently to the original `youtube.py` (`YtOriginal`), stripping the `/streams` suffix from the URL if present.

---

### Extractors (`extractors.py`)

Contains three extractors: `StreamsExtractor`, `LiveExtractor`, `VideoExtractor`. Each implements an `_extract(url)` method returning either a redirect to the next extractor or a list of HLS URLs.

#### 1. `StreamsExtractor`
- Matches: `.../streams` URLs
- Parses active (non-upcoming) streams from `ytInitialData` -> `richGridRenderer`
- Selects a stream based on the `--youtube-stream` plugin-specific option
- Returns a `NextExtractor` with the resolved watch URL

#### 2. `LiveExtractor`
- Matches: `.../live` URLs
- Fetches video ID from `ytInitialData` -> `currentVideoEndpoint.watchEndpoint.videoId`
- Returns a `NextExtractor` with the resolved watch URL

#### 3. `VideoExtractor`
- Matches: `.../watch?v=ID`
- Fetches player responses using `web_safari` and `android_vr` clients
- Collects `hlsManifestUrl` values from `streamingData` in each player response
- Solves n-parameter challenges in each manifest URL via `DenoJCP`
- Returns an `ExtractorResult` with the final list of solved HLS manifest URLs

---

### Plugin-specific option `--youtube-stream`

| Value         | Behavior                                                        |
|---------------|-----------------------------------------------------------------|
| `popular`     | Picks the stream with the highest viewer count                  |
| `first`       | Picks the first stream in the listing                           |
| `last`        | Picks the last stream in the listing                            |
| `N` (integer) | Picks the Nth stream (1-based, clamped to last if out of range) |

Invalid values fall back to `popular` with a warning.

---

### Supported URL Formats

Handles the same formats as the original plugin, plus `/streams` tabs using the `--youtube-stream` option.


fl-Henry:
@bastimeyer  
Yes, I used AI to generate docstrings, comments, and a pull request description. However, I manually disassembled and tested the yt-dlp library's code for some time. I also encountered some recent problems because I had been working on the plugin for a long time when YouTube updated its system and yt-dlp didn't work for a while and had to issue several patches, which I also tracked and applied to the plugin.

>This is a modified copy of yt-dlp's code that even uses the same "extractor" naming scheme, with additional minified and thus unmaintainable JS code that bundles third-party JS-parser dependencies

Yes, that's correct. The idea is to stabilize the plugin by incorporating yt-dlp's approach in addition to the original plugin. I disassembled their code and incorporated only the logic that solves n-challenges in HLS URLs for YouTube. 

>The plugin was moved into a package, similar to what yt-dlp has done with its youtube "extractor" last year, but you're for some reason also moving the old module into the new package and using it as a fallback in your new implementation, essentially duplicating plugin logic in the worst way possible.

Yes, as I said, I got their code intentionally. I tried to keep everything in one file, "youtube.py," but it grew, so I decided it would be better to separate the logic into several modules as yt-dlp did.  
Regarding the fallback to the original plugin: For now, I decided it would be a good idea to keep it here so that the maintainer can be more confident that he has a reliable original plugin. If the suggested solution proves to be reliable, we can remove the fallback in the future.

>Your new implementation attempts to open and read streams while it's fetching them, sequentially. Plugins must never open and read streams. If there are 4xx HTTP responses which need to be checked, then this must be done differently.

I'll change it. Other checks took more time, so I decided this would be a quicker solution.

>Since this is an undisclosed LLM PR with an undisclosed PR body message, I'm going to close this now.

I understand. I'm sorry for that. I totally forgot to specify it. I read the contribution requirements and only used AI for documents, comments, and other types of descriptions. I also used it to provide me with suggestions. However, I composed the plugin manually.

>As I have said in a couple of YouTube-plugin issue threads in the past, I personally am not interested in maintaining this YouTube nonsense, because it's an uphill battle with YouTube/Google that can only be lost, and it's not worth my time. This must be properly maintained by someone else who doesn't just copy stuff or use LLMs.

That's also the reason for the PR with the fallback to the original plugin. In this case, the community and I could focus on the JS challenge solution, and the maintainer could still support the original plugin

>You are welcome to discuss implementing this in a proper way, but I'm not going to waste my time rewriting and cleaning up untested gen-AI code.

Yes, I would really appreciate it if we could find a proper way to implement the JS challenge solving.
Once again, I apologize for not mentioning the AI aspect, even though it is clearly present. I just forgot about it


You said that using Deno is "not the best idea, even though a web browser is overkill for executing JavaScript."  In this case, you would prefer to use Python libraries that embed engines? Could you please tell me what you think would be a suitable solution to the issue?

bastimeyer:
I'm sorry, but it's impossible to distinguish generated code from code with generated comments and what not, especially when the PR description has also been generated. Considering that yt-dlp "extractor" naming schemes and similar APIs were used instead of Streamlink's "plugin" naming schemes and our APIs, this is highly suspicious. As said, I'm not opposed to improving the plugin, but it needs to be done the proper way.

Let me further explain some of the pointed out issues:

- Streamlink doesn't support plugin packages, so it's currently not possible to split up plugins into multiple modules. This is because during build-time (sdist/wheel), Streamlink needs to generate an AST from the plugin module in order to statically analyze plugin matcher data that's then written into a JSON file, so the Streamlink session doesn't have to load all available plugins into memory during run-time at the same time. In an editable install (`pip install -e .`), this is irrelevant and therefore works with plugin packages. Streamlink's docs do the same thing with plugin-matcher and optional plugin-argument data, as well as plugin metadata in the module header docstring. And then there are also our plugin tests that need to do the same thing.
- Similar to the `streamlink.webbrowser` package, a Deno API should be implemented in its own sub-package or in a `streamlink.utils.deno` module, so it can be reused, which means it should not be tied to the YouTube plugin alone. The YT-specific JS payloads can be placed somewhere else. This would need to be figured out.
- Speaking of the YT-specific JS code, this should not be included from a manually-built or externally copied source that's unknown to other people. Vendored code like this must have an accompanied build script in the scripts directory, similar to our CDP build script. Since this is essentially bundled code from NPM repos or git repos directly, it should be trivial to write a simple shell script for acquiring the sources and bundling them, with whatever JS tool is suited best for it. The built JS code should then be committed together with its build script, and it should not be minified due to legibility/comprehension and git diffing reasons. Dependency versions should be pinpointed with a lockfile, so running the build script twice should result in the same output.
- As explained, Streamlink already has its `streamlink.webbrowser` API for running Chromium-based web browsers headlessly and for controlling them via the CDP (Chrome DevTools Protocol). See the Twitch plugin for example. This means that JS code can also be executed in this environment without having to necessarily add a new Deno dependency. This should be kept in mind when implementing the YT JS challenge solver, but since it's an overkill running a full web browser for this task (no DOM requirement, etc.), it should not be included in an initial implementation. Using Deno is the more sensible idea. The webbrowser API should however be made available afterwards as a fallback for users who don't have Deno installed. The yt-dlp project also supports multiple JS runtime providers.
- Regarding the maintenance, it totally makes sense not re-inventing the wheel and instead (re-)using available code from the yt-dlp project. However, Streamlink-specific APIs must be used in the plugin implementation, similar to what all other plugins are doing. The "extraction" + JS-challenge logic and whatnot can easily live in a single module. Once again, see the Twitch plugin and its size for the all the stuff it has to do. That's not a problem.

fl-Henry:
>Streamlink doesn't support plugin packages... Once again, see the Twitch plugin and its size for the all the stuff it has to do. That's not a problem.

I got it. I will repack it into one .py file, especially since we're moving the deno part outside of the package.

>Similar to the streamlink.webbrowser package, a Deno API should be implemented in its own sub-package or in a streamlink.utils.deno module, so it can be reused, which means it should not be tied to the YouTube plugin alone

Sure, I will move it to `streamlink.utils.deno` and will make it reusable.

>Speaking of the YT-specific JS code, this should not be included from a manually-built or externally copied source that's unknown to other people. Vendored code like this must have an accompanied build script in the scripts directory, similar to our CDP build script. Since this is essentially bundled code from NPM repos or git repos directly, it should be trivial to write a simple shell script for acquiring the sources and bundling them, with whatever JS tool is suited best for it. The built JS code should then be committed together with its build script, and it should not be minified due to legibility/comprehension and git diffing reasons. Dependency versions should be pinpointed with a lockfile, so running the build script twice should result in the same output.

Yt-dlp-ejs provides the [Python script](https://github.com/yt-dlp/ejs/blob/main/hatch_build.py) that uses NPM+Deno to build the necessary JavaScript (JS) scripts. As I understand it, I have to place the Python or Bash script in the `./script` directory, next to the `generate-cdp.py` file. Should the Bash script remove the created files after it finishes working? Where should we store the` yt.solver.core.js`, `yt.solver.lib.js`, and `yt.solver.deno.lib.js` scripts? I don't see where to logically place them in the current file structure. For now, I'll place them in `streamlink/solvers/`.

>Regarding the maintenance, it totally makes sense not re-inventing the wheel and instead (re-)using available code from the yt-dlp project. However, Streamlink-specific APIs must be used in the plugin implementation, similar to what all other plugins are doing. The "extraction" + JS-challenge logic and whatnot can easily live in a single module. Once again, see the Twitch plugin and its size for the all the stuff it has to do. That's not a problem.

Got it. Could you please clarify what we're going to do about the fallback to the original plugin? Should I leave it there? In which order should the plugins run: original first, then JS-Challenge as a fallback, or the opposite?

bastimeyer:
> For now, I'll place them in ...

This can be decided at the end. I'm not sure yet. It'll be a dedicated path for plugin data files, which we currently don't have. This will need to be added to the setuptools config:
https://github.com/streamlink/streamlink/blob/b56cfa08d9e9b535ff65a2f77f99ba857714f11a/pyproject.toml#L160-L166
And a respective function needs to be added that resolves paths relative to Streamlink's Python distribution (the package path in the current Python environment).

The Twitch plugin gets away without external plugin asset/resource files, because its JS payload for calculating and intercepting the client-integrity token is embedded in the plugin's Python code itself. This could theoretically also be done here by adding specific markers for the build script to find, but that'd be a terrible idea for larger JS modules like that.

> Could you please clarify what we're going to do about the fallback to the original plugin?

The plugin needs to be rewritten and replaced. The current implementation nowadays is basically just the HLS playlist retrieval from their v1 API using an older Android client version string, which itself was just a fallback when everything was just accessible in the initial-player-response JSON data. Then there's some video-ID translation stuff and metadata retrieval from the JSON data. The rest is regular plugin boilerplate code.

With a proper n-parameter calculation, the API call can be turned into a fallback again. The rest needs to be refactored. I can assist with that once the rest is finished.

But I'm not fully aware of all the details here, because the YouTube plugin is a total mess.

fl-Henry:
@bastimeyer 
I've made the necessary changes and committed them to the same [branch](https://github.com/Grabien-Inc/streamlink/tree/youtube), but this PR is closed, so I decided that it's better the clarify how to submit the changers beforehand

Should I open a new PR, or can we continue here?

---
Here's a short summary of what was changed:

Structure:
- The package layout (youtube/ directory with __init__.py, plugin.py, extractors.py, deno.py, structures.py, youtube.py, solver/) was abandoned.
- Everything was collapsed back into a single streamlink/plugins/youtube.py file.
- The only extraction from the package was the Deno subprocess handler, which was moved to streamlink/utils/deno.py as a reusable utility.
- The solver JS scripts (lib.min.js, core.min.js) and their loader moved to streamlink/solvers/youtube/.

Implementation:
- All dataclasses, enums, extractors, and solver classes that lived across structures.py, extractors.py, and plugin.py are now defined directly in youtube.py.
- DenoJCP was renamed to DenoSolver; it now inherits from both Solver and Deno (the new utils class) instead of being a standalone class.
- CDPSolver was added as a second solver type alongside DenoSolver, with _extract_hls trying both in sequence.
- The Solver base class (ABC) with shared logic (_get_player, _construct_stdin, validate_output, _get_script) was introduced to avoid duplication between DenoSolver and CDPSolver.
- Stream probing in _check_streams was simplified — instead of reading 64 bytes with a 2-second timeout, it now does a HEAD request on the first segment URL.
- The fallback is no longer a separate class — the fallback logic (_get_streams_fallback) lives directly in the YouTube plugin class.

I also added [`build-solvers.sh`](https://github.com/fl-Henry/streamlink/blob/main/script/build-solvers.sh) to the scripts directory and included the `*.js` files in the `streamlink.solvers.youtube` package data, so the solver scripts are bundled with the built distribution. I tested the build in a Docker container and it didn't throw any errors. I also edited and corrected the code style according to the mypy and ruff rules.

> **Note**
> Today Deno solving fails because the YouTube player JS that was fetched doesn't contain the expected n-parameter solver function. However, CDP can evaluate the script correctly.

bastimeyer:
Thanks for the update.

A PR branch should not include multiple commits with reverted changes. Changes should be committed in logical chunks, meaning the `utils.deno` module should be its own commit, or even better, its own PR that gets merged before the plugin PR is submitted. The module should also have accompanied tests, but as said, I can assist with stuff like that. Regarding its code though, I've already seen something about the subprocess implementation which would need to be changed first, but I'm not going to review this here yet.

After thinking about the ejs / JS solver code again, it'd probably be better to have it as an optional dependency, [as it's packaged on PyPI](https://pypi.org/project/yt-dlp-ejs), and a couple of Linux distros also already package it (e.g. [Arch](https://archlinux.org/packages/?sort=&q=yt-dlp-ejs&maintainer=&flagged=) or [Alpine](https://pkgs.alpinelinux.org/packages?name=yt-dlp-ejs&branch=edge&repo=&arch=x86_64&origin=&flagged=&maintainer=)), so it's better to rely on that instead of having to bundle this much code (using the "unlicense" license) in a new sub-package and having to maintain it here. Distros which don't package it yet will then have an incentive to do that, or the plugin will have to use the fallback code.

The script you've added for fetching new sources wasn't exactly what I had in mind, but by adding the optional `yt-dlp-ejs` dependency, there's no need for an update script.

Regarding the plugin changes: I had a very brief glimpse over the code and there's a lot that would need to be changed here as well, especially the "extraction" and "stream selection" logic, because this is not how Streamlink is supposed to work. Some old plugin code can also be removed (once the basic functionality has been figured out and implemented). And then there's some refactoring to do, too, I think.

I'm not a fan of implementing HLS segment checks in the plugin itself. `HLSStream.parse_variant_playlist()` has a `check_streams` argument which currently only does a GET request on the media playlists. This can independently be extended with segment-check logic. Unless checking the streams is 100% necessary though (I'm not sure if it is), this should not be included in the first iteration of the plugin rewrite, because it makes it unnecessarily complicated.

AI-generated comments/docstrings should also not be included, because they are unnecessary.

The inclusion of the CDPSolver logic is a good thought, but if this causes any friction in terms of code review, then this should be added in a later stage as an addition, as said.

----

So let's do this: rebase your changes and move the Deno module changes into a separate commit in a new PR. I can assist with adding tests. Once that's done, we can either merge into master, or leave it open and use the PR's branch as the base of the PR branch for the plugin changes, so we stay flexible, in case further changes are required. If you're not familar with rebasing (I'm not sure you are because you've opened two similar PRs and also added partial reverting changes to the linked branch here), then this will probably cause a bit of trouble, but once again, I'm happy to help.

fl-Henry:
Thank you for you reply 


>A PR branch should not include multiple commits with reverted changes. Changes should be committed in logical chunks, meaning the utils.deno module should be its own commit, or even better, its own PR that gets merged before the plugin PR is submitted. The module should also have accompanied tests, but as said, I can assist with stuff like that. Regarding its code though, I've already seen something about the subprocess implementation which would need to be changed first, but I'm not going to review this here yey

Ok, then I'll create a PR for deno first

>After thinking about the ejs / JS solver code again, it'd probably be better to have it as an optional dependency, [as it's packaged on PyPI](https://pypi.org/project/yt-dlp-ejs), and a couple of Linux distros also already package it (e.g. [Arch](https://archlinux.org/packages/?sort=&q=yt-dlp-ejs&maintainer=&flagged=) or [Alpine](https://pkgs.alpinelinux.org/packages?name=yt-dlp-ejs&branch=edge&repo=&arch=x86_64&origin=&flagged=&maintainer=)), so it's better to rely on that instead of having to bundle this much code (using the "unlicense" license) in a new sub-package and having to maintain it here. Distros which don't package it yet will then have an incentive to do that, or the plugin will have to use the fallback code.

In this case, I'll remove the bash script that builds the JS files and add checks for the yt-dlp-ejs import before trying to solve the challenges.

>I'm not a fan of implementing HLS segment checks in the plugin itself. HLSStream.parse_variant_playlist() has a check_streams argument which currently only does a GET request on the media playlists. This can independently be extended with segment-check logic. Unless checking the streams is 100% necessary though (I'm not sure if it is), this should not be included in the first iteration of the plugin rewrite, because it makes it unnecessarily complicated.

I understand, but I implemented the check because Streamlink tries to load every video segment, even if the previous attempt failed when the N-Challenge was solved incorrectly. This can take quite a while. During this time, we cannot fall back to the original plugin logic. 
We can get the N-Challenge token from the manifest URL and replace it with the solved token. However, YouTube returns the manifest content regardless of whether the challenge was solved (the same token is present in the inner URLs, both formats, and video segments), and YouTube only returns a 403 error for the video segments. This means we cannot check if the challenge was solved in the first two layers (manifest and formats) and can only check it using the direct video segment URL. 

> If you're not familar with rebasing (I'm not sure you are because you've opened two similar PRs and also added partial reverting changes to the linked branch here), then this will probably cause a bit of trouble, but once again, I'm happy to help.

You're right. I mainly use Git for work, performing simple actions. I'll do some research on this so that we waste less time on such things.

fl-Henry:
@bastimeyer 

I've rebased the branch and created [PR](https://github.com/streamlink/streamlink/pull/6893) only for the `deno.py`.  
I also have to implement tests for the module, so I would like to ask you to tell me what test cases should it cover?

fl-Henry:
@bastimeyer 

I've created a PR from a branch that uses [193da07](https://github.com/streamlink/streamlink/commit/193da07bfa1daea70aa870fb8351149bf370a431) as the base. I edited the original plugin's method `_data_video_id` because YouTube recently changed the initial JSON format.

I also assume we need to add/edit the test for the plugin. If so, please tell me what should be corrected/added.

PR #6878, opened 2026-03-31
stream.hls: update check_streams param

bastimeyer:
Add support for checking segments in media playlists, in addition to just checking the accessibility of a multivariant playlist's media playlists.

Turn the `check_streams` param of `HLSStream.parse_variant_playlist()` from a simple boolean into a param that also accepts the "playlists" and "segments" literals, with "playlists" being an alias for True, where only media playlists are checked.

If the value is set to "segments", only one segment of the first available media playlist is checked, to optimize loading times, as HTTP requests are made sequentially.

Choose the middle segment, so unreliable segments like ones in a live-stream just being discarded or pre-fetch segments at the edge of a live-stream can be avoided.


bastimeyer:
Opening this as a draft for now, as this delays the streams on Twitch, as the plugin sets the check_streams param. I think I will add another param for checking segments instead. Maybe also only a check for the first media playlist, as there won't be a difference between media playlists when access is restricted on the segment-level.

I'll change this later...

bastimeyer:
In the future, this new `_check_streams()` helper classmethod can be rewritten to check streams in parallel, once the stream selection has been rewritten.

PR #6893, opened 2026-04-09
Add Deno JavaScript runtime execution utility

fl-Henry:
Introduce a Deno wrapper class for executing JavaScript via the Deno runtime as a subprocess. Supports configurable CLI flags with secure defaults (no remote, no npm, no config, cached-only) and pipes stdin to the process, returning stdout or raising on failure.

----

### DIFF
--- a/src/streamlink/utils/deno.py
+++ b/src/streamlink/utils/deno.py
@@ -1,0 +1,68 @@
+import logging
+import os
+import shlex
+import subprocess
+
+from streamlink.utils.path import resolve_executable
+
+
+log = logging.getLogger(__name__)
+
+
+class Deno:
+    def __init__(self, params: dict | None = None):
+        self._DEFAULT_PARAMS = {
+            "--ext": "js",
+            "--no-code-cache": True,
+            "--no-prompt": True,
+            "--no-remote": True,
+            "--no-lock": True,
+            "--node-modules-dir": "none",
+            "--no-config": True,
+            "--no-npm": True,
+            "--cached-only": True,
+        }
+
+        self._exec_path = resolve_executable("deno")
+        if not self._exec_path:
+            raise FileNotFoundError("Deno not found. Please install Deno from the official website")
+        self._params = {**self._DEFAULT_PARAMS, **(params or {})}
+
+    def _build_cmd(self) -> list[str]:
+        cmd = [str(self._exec_path), "run"]
+        for flag, value in self._params.items():
+            if value is True:
+                cmd.append(str(flag))
+            elif value is not False and value is not None:
+                cmd.append(f"{flag}={value}")
+        cmd.append("-")
+        return cmd
+
+    def execute(self, stdin) -> str:
+        cmd = self._build_cmd()
+        log.debug("Executing Deno: %s", shlex.join(cmd))
+
+        proc = subprocess.Popen(
+            cmd,
+            text=True,
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            env=os.environ.copy(),
+            encoding="utf-8",
+        )
+        try:
+            stdout, stderr = proc.communicate(stdin)
+        except BaseException:
+            proc.kill()
+            proc.wait(timeout=0)
+            raise
+
+        if proc.returncode or stderr:
+            msg = f"Deno process failed (returncode: {proc.returncode})"
+            if stderr:
+                msg = f"{msg}: {stderr.strip()}"
+            raise Exception(msg)
+
+        log.debug("Deno process completed successfully")
+        return stdout
--- a/src/streamlink/utils/deno.py
+++ b/src/streamlink/utils/deno.py
@@ -1,0 +1,69 @@
+from __future__ import annotations
+
+from functools import lru_cache
+from typing import ClassVar
+
+from streamlink.logger import getLogger
+from streamlink.utils.path import resolve_executable
+from streamlink.utils.processoutput import ProcessOutput
+
+
+log = getLogger(__name__)
+
+
+class DenoProcessor(ProcessOutput):
+    """
+    Utility class for finding and launching Deno on the user's system and processing JavaScript via stdin/stdout.
+    """
+
+    # https://docs.deno.com/runtime/reference/cli/run/#usage
+    _RUN_FLAGS: ClassVar[frozenset[str]] = frozenset([
+        "--ext=js",
+        "--no-code-cache",
+        "--no-prompt",
+        "--no-remote",
+        "--no-lock",
+        "--node-modules-dir=none",
+        "--no-config",
+        "--no-npm",
+    ])
+
+    def __init__(
+        self,
+        executable_path: str | None = None,
+        **kwargs,
+    ) -> None:
+        if (exec_path := self.resolve_path(executable_path)) is None:
+            raise FileNotFoundError("Deno not found. Please install the Deno JavaScript runtime on your system.")
+
+        kwargs.pop("stdin", None)
+        super().__init__([exec_path, "run", *self._RUN_FLAGS, "-"], **kwargs)
+        self._output_lines: list[str] = []
+
+    # noinspection PyMethodOverriding
+    async def arun(self, stdin: bytes) -> bool:  # type: ignore[override, ty:invalid-method-override]
+        log.trace("Running Deno with a stdin payload of %d bytes: %r", len(stdin), self.command)
+        return await super().arun(stdin=stdin)
+
+    @property
+    def output(self) -> str:
+        return "\n".join(self._output_lines)
+
+    def onstdout(self, idx: int, line: str) -> bool | None:
+        self._output_lines.append(line)
+
+    def onstderr(self, idx: int, line: str) -> bool | None:
+        raise RuntimeError(f"Deno error: {line}")
+
+    @staticmethod
+    @lru_cache
+    def resolve_path(executable_path: str | None = None) -> str | None:
+        if not executable_path:
+            exec_path = resolve_executable(names=["deno"])
+        else:
+            exec_path = executable_path
+
+        if not exec_path:
+            return None
+
+        return str(exec_path)
--- a/tests/utils/test_deno.py
+++ b/tests/utils/test_deno.py
@@ -1,0 +1,84 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+from unittest.mock import AsyncMock, Mock
+
+import pytest
+
+from streamlink.utils.deno import DenoProcessor
+
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+
+@pytest.fixture(autouse=True)
+def _clear_resolve_path_cache():
+    yield
+    DenoProcessor.resolve_path.cache_clear()
+
+
+@pytest.fixture()
+def fake_exe(tmp_path):
+    exe = tmp_path / "deno.exe"
+    exe.write_bytes(b"")
+    return exe
+
+
+class TestResolvePath:
+    @pytest.fixture()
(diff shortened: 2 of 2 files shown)


src/streamlink/utils/deno.py
```diff
@@ -0,0 +1,68 @@
+import logging
+import os
+import shlex
+import subprocess
+
+from streamlink.utils.path import resolve_executable
+
+
+log = logging.getLogger(__name__)
```
bastimeyer, line 9: The logger object must be retrieved from the `streamlink.logger.getLogger` function. This is because Streamlink extends the default `logging.Logger` class with custom log levels and methods, so we need to make sure that the right logger class is used, which is not guaranteed via the stdlib function.

src/streamlink/utils/deno.py
```diff
@@ -0,0 +1,68 @@
...
+            if value is True:
+                cmd.append(str(flag))
+            elif value is not False and value is not None:
+                cmd.append(f"{flag}={value}")
+        cmd.append("-")
+        return cmd
+
+    def execute(self, stdin) -> str:
+        cmd = self._build_cmd()
+        log.debug("Executing Deno: %s", shlex.join(cmd))
```
bastimeyer, line 43: It's better to use `%r` and just print the representation of the cmd list, so that the argv is unmistakably clear. The FFMpegMuxer does the same for example, and it uses the `trace` log level:
https://github.com/streamlink/streamlink/blob/0ce2f5e0d97bf05e586d84dab592f38274368a7f/src/streamlink/stream/ffmpegmux.py#L137

src/streamlink/utils/deno.py
```diff
@@ -0,0 +1,68 @@
...
+            encoding="utf-8",
+        )
+        try:
+            stdout, stderr = proc.communicate(stdin)
+        except BaseException:
+            proc.kill()
+            proc.wait(timeout=0)
+            raise
+
+        if proc.returncode or stderr:
```
bastimeyer, line 61: Streamlink has the `streamlink.utils.processoutput` module for running subprocesses and reading their stdout/stderr streams and checking the return value. Using the `subprocess` stdlib module however also works.

src/streamlink/utils/deno.py
```diff
@@ -0,0 +1,68 @@
...
+
+        proc = subprocess.Popen(
+            cmd,
+            text=True,
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            env=os.environ.copy(),
+            encoding="utf-8",
+        )
```
bastimeyer, line 53: The `env` will already be inherited by the parent process, so explicitly copying the env mapping is not necessary.

src/streamlink/utils/deno.py
```diff
@@ -0,0 +1,68 @@
...
+        try:
+            stdout, stderr = proc.communicate(stdin)
```
bastimeyer, line 55: You're not setting the `timeout` parameter here, meaning that if the `deno` process doesn't terminate, it'll hang. The `deno` executable can be anything arbitrary on the user's system.

src/streamlink/utils/deno.py
```diff
@@ -0,0 +1,68 @@
...
+        except BaseException:
+            proc.kill()
+            proc.wait(timeout=0)
+            raise
```
bastimeyer, line 59: `wait()` after a `kill()` is unnecessary, because the process has already received a `SIGKILL` signal on POSIX systems or `TerminateProcess()` call on Windows. Considering that the BaseException is re-raised, `poll()` is not needed, as it would only guarantee that `returncode` is set.

src/streamlink/utils/deno.py
```diff
@@ -0,0 +1,68 @@
...
+
+        if proc.returncode or stderr:
+            msg = f"Deno process failed (returncode: {proc.returncode})"
+            if stderr:
+                msg = f"{msg}: {stderr.strip()}"
+            raise Exception(msg)
```
bastimeyer, line 65: `OSError` or `RuntimeError`

src/streamlink/utils/deno.py
```diff
@@ -0,0 +1,68 @@
...
+            stdout, stderr = proc.communicate(stdin)
+        except BaseException:
+            proc.kill()
+            proc.wait(timeout=0)
+            raise
+
+        if proc.returncode or stderr:
+            msg = f"Deno process failed (returncode: {proc.returncode})"
+            if stderr:
+                msg = f"{msg}: {stderr.strip()}"
```
bastimeyer, line 64: Since we're dealing with potentially arbitrary executables, stderr also need to be truncated to a max length.

src/streamlink/utils/deno.py
```diff
@@ -0,0 +1,68 @@
...
+
+from streamlink.utils.path import resolve_executable
+
+
+log = logging.getLogger(__name__)
+
+
+class Deno:
+    def __init__(self, params: dict | None = None):
+        self._DEFAULT_PARAMS = {
```
bastimeyer, line 14: Could you please add
https://docs.deno.com/runtime/reference/cli/run/#usage
above the dict as a comment, so these options are properly linked for easier lookup and research?

bastimeyer (commented):
Thanks for the PR!

I've annotated a couple of things that need to be changed.

> I also have to implement tests for the module, so I would like to ask you to tell me what test cases should it cover?

At least the `_build_cmd()` part should be covered. The `execute()` stuff would require mocking the APIs, because spawning a subprocess in the tests just for validating that stdout/stderr + returncode are read correctly is unnecessary. I've added a comment about the `ProcessOutput` util, but I'm not sure which one would be easier to work with. Keep the `execute()` optional.

As said, I can have a look at adding the tests myself if you're uncomfortable with that. I would then force-push onto your PR branch...

fl-Henry:
@bastimeyer

I've rebased the branch and changed the usage so that it uses ProcessOutput on initialization, but I've kept the execute() method as a class method so that you can execute the run command via subprocess.Popen. I've also added tests for the module. I would appreciate it if you could take a look at the changes. 

Thank you.

bastimeyer (commented):
Apologies for the late review.

> I've rebased the branch and changed the usage so that it uses ProcessOutput on initialization, but I've kept the execute() method as a class method so that you can execute the run command via subprocess.Popen.

What you've done unfortunately doesn't make sense. The whole point of the `ProcessOutput` class is to (optionally asynchronously) run a process while offering you a simple interface with three different callback methods to read stdout/stderr lines and the exit code, so you can make a decision about the validity of the process's execution, without necessarily having to wait for it to succeed or terminate on its own. Instead, you subclass it and still for some reason call the stdlib's `subprocess` APIs in the `execute` method, so you're basically adding your custom stuff on top of the `ProcessOutput` subclass.

I merely mentioned `ProcessOutput`, because it's a high-level utility class which might be easier to work with in tests. Whether you use is was totally optional.

fl-Henry:
@bastimeyer  
Thank you for your feedback

>Instead, you subclass it and still for some reason call the stdlib's subprocess APIs in the execute method, so you're basically adding your custom stuff on top of the ProcessOutput subclass.

I made this intentionally because, as it turned out, I didn't get you right

>I've added a comment about the ProcessOutput util, but I'm not sure which one would be easier to work with. Keep the execute() optional.

What did you mean here? I thought that you wanted me to keep `execute()` to work via stdlib's `subprocess` while implementing another (main) solution as a subclass of `ProcessOutput`. 

So in the current version, we have 2 solutions that can be used separately.
```python
# Example: run via ProcessOutput 
dp = DenoProcessor(["run", *DenoProcessor.DEFAULT_PARAMS], stdin=b"console.log('Hello World');")
dp.run()
print(dp.output)  # output='Hello world'

# Example of using optional `execute()` classmethod via stdlib's subprocess
stdout = DenoProcessor.execute("console.log('Hello World');")
print(stdout)  # output='Hello world'
```

I tried to implement the main solution similar to `FFmpegVersionOutput(ProcessOutput)` while keeping `execute()` optional.


Anyway, since I did it wrong, could you please tell me what you see as the correct solution for this PR:  
- inheritance of `ProcessOutput`
- rollback to the stdlib's `subprocess`
- or both are incorrect

Thank you

bastimeyer:
To avoid further confusions and unnecessary waste of time on both of our sides, let me clean up your PR later when I get the time. It's not a big deal.

Any utils modules, especially ones with code that execute arbitrary data on the user's system, need to be solid, so please understand that I'm a bit more nit-picky with that, as I want to get it right without having to make further adjustments afterwards.

Once I've updated your PR, you can use it as a base for your YouTube plugin changes. I would only like to merge this one here if the YouTube plugin changes are working out as expected, because if they don't, then we would've merged a useless utils module. If that causes any issues with your workflow, I'm happy to help you resolve that.

bastimeyer:
> $ git push -f Grabien-Inc utils/deno/init:youtube
> Enter passphrase for key ...:
> ERROR: Permission to Grabien-Inc/streamlink.git denied to bastimeyer.
> fatal: Could not read from remote repository.
> 
> Please make sure you have the correct access rights
> and the repository exists.

I don't have access to push onto your PR branch with my changes. You must enable the checkbox in the PR options.

Here's the updated branch on my fork:
https://github.com/streamlink/streamlink/compare/master...bastimeyer:streamlink:utils/deno/init

```sh
$ python -c 'from streamlink.utils.deno import DenoProcessor;dp=DenoProcessor(None,stdin="console.log(\"🐻\")\n");dp.run();print(dp.output)'
🐻
```

fl-Henry:
>You must enable the checkbox in the PR options

Unfortunately, I don't see such an option. I did some research and found out that GitHub disables this option for organization-owned forks.

I think I could copy the solution from your fork and push it myself. Would that work?

bastimeyer:
> Would that work?

Absolutely. Just add my fork as a remote, fetch the branch, push it onto your fork and submit it again. As you can see, I kept the commit authorship and added myself as a co-author.

Regarding the changes, I cleaned up a few things and also changed the signature of the class constructor, since it's only about passing `stdin` data to the `deno run` command.

fl-Henry:
I have rebased the branch with your updates. I believe we're ready to accept the PR now. If not, please let me know what else I can do. I'm online all day today and will make the changes as soon as possible.

bastimeyer:
> we're ready to accept the PR now

Yes, this one is ready, thanks. As said though, I don't want to merge it right now, because merging a new util into master without actual usage doesn't make sense.

Please just use 193da07bfa1daea70aa870fb8351149bf370a431 as the base of your planned YouTube changes. You can easily include it in any PR. It doesn't matter. Take your time with the YT changes.

For purposes of this kind, GitHub has a new feature upcoming with stacked PRs, where PRs can depend on other still unmerged PRs, so that the reviewing and discussion process can be kept clean. It's not publicly available yet, but as you can see, for a simple double PR chain it's not really necessary.

fl-Henry:
>Yes, this one is ready, thanks. As said though, I don't want to merge it right now, because merging a new util into master without actual usage doesn't make sense.

Perfect, thank you for your help.

I have the branch called `youtube` for the current PR, because we rebased from the original PR. I assume the branch name is not really important. 

>For purposes of this kind, GitHub has a new feature upcoming with stacked PRs, where PRs can depend on other still unmerged PRs, so that the reviewing and discussion process can be kept clean. It's not publicly available yet, but as you can see, for a simple double PR chain it's not really necessary.

Since we don't have the feature yet, I would like to clarify that I can continue pushing into the current branch/PR? Or I should create the new branch that is based on the specified commit?

bastimeyer:
Create a new one with 193da07bfa1daea70aa870fb8351149bf370a431 as the base, not `master`.
When submitting a PR against `master`, `193da07bfa1daea70aa870fb8351149bf370a431` will be included, but this can be removed later once that one is merged into master by a simple rebase.

PR #6896, opened 2026-04-13
Session Option: 'stream-passthrough-encrypted'

jordandalley:

----

Introduces two new session options, `hls-no-check-formats` and `dash-no-check-formats`, which mirror the behavior of yt-dlp’s --no-check-formats flag. These options allow users to bypass standard stream validation and content-protection checks to output raw data segments.

While standard users typically benefit from format validation, these options are critical for:

- Plugin Development: Building utilities that require access to unvalidated stream data.
- Research & Debugging: Investigating raw stream chunks and manifest structures without interference from the internal processing engine.
- Edge-case Handling: Accessing data in streams where standard detection might otherwise fail or block the pipeline.

No AI was used for this commit.

### DIFF
--- a/src/streamlink/stream/dash/dash.py
+++ b/src/streamlink/stream/dash/dash.py
@@ -308,6 +308,7 @@ class DASHStream(Stream):
         """
 
         manifest, mpd_params = cls.fetch_manifest(session, url_or_manifest, **kwargs)
+        passthrough_encrypted = session.options.get("stream-passthrough-encrypted")
 
         try:
             mpd = cls.parse_mpd(manifest, mpd_params)
@@ -333,15 +334,23 @@ class DASHStream(Stream):
 
         # Search for suitable video and audio representations
         for aset in period_selection.adaptationSets:
-            if aset.contentProtections:
+            if aset.contentProtections and not passthrough_encrypted:
                 raise PluginError(f"{source} is protected by DRM")
             for rep in aset.representations:
-                if rep.contentProtections:
+                if rep.contentProtections and not passthrough_encrypted:
                     raise PluginError(f"{source} is protected by DRM")
                 if rep.mimeType.startswith("video"):
                     video.append(rep)
                 elif rep.mimeType.startswith("audio"):  # pragma: no branch
                     audio.append(rep)
+
+        if passthrough_encrypted:
+            is_encrypted = any(
+                aset.contentProtections or any(rep.contentProtections for rep in aset.representations)
+                for aset in period_selection.adaptationSets
+            )
+            if is_encrypted:
+                log.warning(f"{source} is protected by DRM and won't be decrypted.")
 
         if not video:
             video.append(None)
--- a/src/streamlink/stream/dash/dash.py
+++ b/src/streamlink/stream/dash/dash.py
@@ -349,8 +349,8 @@ class DASHStream(Stream):
                 aset.contentProtections or any(rep.contentProtections for rep in aset.representations)
                 for aset in period_selection.adaptationSets
             )
-            if is_encrypted:
-                log.warning(f"{source} is protected by DRM and won't be decrypted.")
+            if is_encrypted:  # pragma: no branch
+                log.warning(f"{source} is protected by DRM and won't be decrypted")
 
         if not video:
             video.append(None)
--- a/src/streamlink/stream/hls/hls.py
+++ b/src/streamlink/stream/hls/hls.py
@@ -410,7 +410,7 @@ class HLSStreamWorker(SegmentedStreamWorker[HLSSegment, Response]):
         if self.sequence < 0 and first_segment.key and first_segment.key.method != "NONE":
             log.debug("Segments in this playlist are encrypted")
             if self.passthrough_encrypted:
-                log.warning(f"The stream content is encrypted with '{first_segment.key.method}' and won't be decrypted.")
+                log.warning("The stream content is encrypted with '%s' and won't be decrypted", first_segment.key.method)
 
         self.playlist_changed = [s.num for s in self.playlist_segments] != [s.num for s in segments]
         self.playlist_segments = segments
--- a/src/streamlink/stream/hls/hls.py
+++ b/src/streamlink/stream/hls/hls.py
@@ -95,6 +95,7 @@ class HLSStreamWriter(SegmentedStreamWriter[HLSSegment, Response]):
         if ignore_names:
             segments = "|".join(map(re.escape, ignore_names))
             self.ignore_names = re.compile(segments, re.IGNORECASE)
+        self.passthrough_encrypted = options.get("stream-passthrough-encrypted")
 
     @staticmethod
     def num_to_iv(n: int) -> bytes:
@@ -260,7 +261,8 @@ class HLSStreamWriter(SegmentedStreamWriter[HLSSegment, Response]):
         # TODO: Rewrite HLSSegment, HLSStreamWriter and HLSStreamWorker based on independent initialization section segments,
         #       similar to the DASH implementation
         key = segment.map.key if is_map and segment.map else segment.key
-        if key and key.method != "NONE":
+
+        if key and key.method != "NONE" and not self.passthrough_encrypted:
             try:
                 decryptor = self.create_decryptor(key, segment.num)
             except (StreamError, ValueError) as err:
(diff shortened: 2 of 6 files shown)


bastimeyer (changes requested):
Thanks for the PR.

I'm sorry, but there are several things which I don't agree with.

First off, the option names:
Streamlink has merged several options over the past couple of major version releases that were stream-specific but which had a single/common meaning. Having multiple options for the same thing would re-introduce this kind of nonsense. "check-formats" is also not the right name for checking whether content can be decrypted or not. This name suggests playback capabilities, which Streamlink has no control over anyway, since it's just a stream transport mechanism. You are also not adding any CLI arguments (or relevant tests).

Then about the functionality itself:
There is a reason why Streamlink outright rejects content which it can't decrypt, and that is that we explicitly don't support DRM schemes (wrt decryption-key acquirement), as you can read in the contribution guidelines. There are different kinds of encryption schemes, and some of them mainly exist for DRM reasons. These schemes also vary between MPEG-2 transport streams and MPEG-4 containers.

The HLS protocol defines two different encryption schemes, and both have metadata for key acquirement.

1. `AES-128`, where the whole segment data is encrypted by a specific key. That key can change for every individual segment and thus may not be constant. Apart from that, the decryptor's internal state is reset on each segment and the (optional) initialization vector is applied anew. Outputting raw non-decrypted data will lead to 100% garbage output data, unless you know the exact length of each segment of the resulting concatenated output that you'd then have to split apart again for decrypting the chunks, and you also know the keys and IV data of each of the original segments. If substreams need to be muxed into a single output stream, FFMPEG will fail to do so.
2. `SAMPLE-AES`, where data is encrypted on the media-sample level within a specific format, which varies between different formats (fMP4 containers use the "Common Encryption" cbcs scheme, MPEG-2 transport streams or packed audio streams use HLS-specific encryption schemes). Streamlink currently does not implement this because segment data is not being demuxed or otherwise processed. Considering that this is mostly meant for DRM content, there hasn't been any interest in implementing it.

In DASH, segment-based encryption does not exist and everything is sample-based, depending on the format used (both MPEG-4 containers and MPEG-2 TS). There are different kinds of encryption schemes the specification mentions, but it's also not limited to what's mentioned. The decryption keys have to be obtained by the client on its own and it's almost exclusively used for DRM content.

----

I believe the main reason why you're submitting this PR is that you want to decrypt DASH content which you already have the key for, which can be done by FFMPEG via its own implementation, as the encryption is sample-based.

There already is a fork of Streamlink which adds a CLI argument for passing the necessary key data to FFMPEG. As explained, since Streamlink doesn't want to be involved with DRM schemes, we don't want to add such session options / CLI arguments.

However, a generic session option / CLI argument which ignores encrypted data, just like you've proposed in your PR, would be fine IMO, so you can pipe the output into FFMPEG and set the decryption key data yourself, if it's possible.

So if you want this to be included, then

1. There must only be one session option and respective CLI argument
2. The option/argument must be named appropriately
3. Streamlink should log a warning that the stream content is encrypted and won't be decrypted
4. There should be tests for that case

src/streamlink/stream/hls/hls.py
```diff
@@ -89,12 +89,15 @@ def __init__(self, *args, **kwargs) -> None:
...
-
         self.ignore_names: re.Pattern | None = None
         ignore_names = {*options.get("hls-segment-ignore-names")}
         if ignore_names:
             segments = "|".join(map(re.escape, ignore_names))
             self.ignore_names = re.compile(segments, re.IGNORECASE)
+        passthrough_encrypted = options.get("stream-passthrough-encrypted")
+        if passthrough_encrypted:
+            self.passthrough_encrypted_logged = False
+            self.passthrough_encrypted = passthrough_encrypted
```
bastimeyer: No conditional class attributes

src/streamlink/stream/hls/hls.py
```diff
@@ -89,12 +89,15 @@ def __init__(self, *args, **kwargs) -> None:
         self.key_uri: str | None = None
         self.key_uri_override = options.get("hls-segment-key-uri")
         self.stream_data = options.get("hls-segment-stream-data")
-
```
bastimeyer, line 92: Please don't remove whitespace randomly

src/streamlink/stream/hls/hls.py
```diff
@@ -89,12 +89,15 @@ def __init__(self, *args, **kwargs) -> None:
...
         self.ignore_names: re.Pattern | None = None
         ignore_names = {*options.get("hls-segment-ignore-names")}
         if ignore_names:
             segments = "|".join(map(re.escape, ignore_names))
             self.ignore_names = re.compile(segments, re.IGNORECASE)
+        passthrough_encrypted = options.get("stream-passthrough-encrypted")
+        if passthrough_encrypted:
+            self.passthrough_encrypted_logged = False
```
bastimeyer: If you're adding a class attribute for suppressing additional log messages, then at least make it a private one.

src/streamlink/stream/hls/hls.py
```diff
@@ -260,7 +263,12 @@ def _write(self, segment: HLSSegment, result: Response, is_map: bool):
         # TODO: Rewrite HLSSegment, HLSStreamWriter and HLSStreamWorker based on independent initialization section segments,
         #       similar to the DASH implementation
         key = segment.map.key if is_map and segment.map else segment.key
-        if key and key.method != "NONE":
+
+        if key and key.method != "NONE" and self.passthrough_encrypted and not self.passthrough_encrypted_logged:
+            self.passthrough_encrypted_logged = True
+            log.warning(f"The stream content is encrypted with '{key.method}' and won't be decrypted.")
+
+        if key and key.method != "NONE" and not self.passthrough_encrypted:
```
bastimeyer: Please don't use redundant checks.

```suggestion
        if key and key.method != "NONE":
            if self.passthrough_encrypted:
                if not self.passthrough_encrypted_logged:
                    self.passthrough_encrypted_logged = True
                    log.warning(f"The stream content is encrypted with '{key.method}' and won't be decrypted.")
                self.reader.buffer.write(result.content)
                return

```

I'm not sure though if that's a good spot to log this message. We already have a log message for encrypted segments in the worker thread in `process_segments()`. Considering that we can check whether it's the first time the playlist is loaded and parsed (`sequence < 0`), it's much better to log it here (and also fix the repeated encrypted segments message).

src/streamlink/stream/dash/dash.py
```diff
@@ -334,10 +335,16 @@ def parse_manifest(
...
+                    log.warning(f"{source} is protected by DRM and won't be decrypted")
+                else:
+                    raise PluginError(f"{source} is protected by DRM")
             for rep in aset.representations:
                 if rep.contentProtections:
-                    raise PluginError(f"{source} is protected by DRM")
+                    if passthrough_encrypted:
+                        log.warning(f"{source} is protected by DRM and won't be decrypted")
+                    else:
+                        raise PluginError(f"{source} is protected by DRM")
```
bastimeyer: This logs for every adaptation set or representation, which is unnecessary. Similar to the HLS implementation, it should only log once. You should prevent the PluginError from being raised with the session option value being true and then check for such adaptation sets and representations again afterwards in an `any(...)` call and then log appropriately.

src/streamlink_cli/argparser.py
```diff
@@ -1059,6 +1059,18 @@ def build_parser():
...
+    transport.add_argument(
+        "--stream-passthrough-encrypted",
+        action="store_true",
+        default=None,
+        help="""
+            Pass through data from encrypted streams without decryption or encryption checks.
+
+            This applies to DASH and HLS streams, and will likely result in garbage output.
+
+            Default is None
```
bastimeyer: There's a dot missing, or the regex for formatting the default value won't match:
https://github.com/streamlink/streamlink/blob/8.3.0/docs/sphinxext/ext_argparse.py#L31

bastimeyer (changes requested):

jordandalley:
Hi @bastimeyer,

Thanks for your thorough response earlier. I really appreciate the time you took to outline your case and how I can improve the PR.

Total disclaimer - development is not my day job - I'm still learning.

I've updated the PR again, fingers crossed all good now :)

Regards,
Jordan.

bastimeyer (commented):
Thanks. The changes are looking good now.

I'll have another look later today and will add missing tests to your PR branch after rebasing it. Then I think we can merge this.

bastimeyer (approved):
The PR should be good now... I've added the necessary tests for full test coverage. I will think about the changes once more tomorrow and then probably merge.

Please note that you've submitted the PR from your fork's master branch, which means I had to force-push onto your master branch in order to clean up the PR. If you're going to submit another PR in the future, you'll have to fix your fork's master branch once again (which you had to do anyway - don't commit to master yourself).

jordandalley:
Thanks @bastimeyer,

Noted re the PR

I appreciate you giving me the opportunity to contribute to your project. I'm a long-time fan of Streamlink.

Keep up the great work :)

Jordan.

Issue #6903, opened 2026-04-24
Specifying interface by name doesn't work on macOS (tests broken too)


sandydoo:
### Checklist

- [x] This is a bug report and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [x] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [x] [I have checked the list of open and recently closed bug reports](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22bug%22)
- [x] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Collaboration

- [x] [I will provide feedback should a pull request be opened with a fix for the problem](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-request-feedback)

### Streamlink version

[cli][info] Your Streamlink version (8.3.0) is up to date!

### Description

The new `--interface` option sets the `SO_BINDTODEVICE` socket option when used with an interface name, but this option is not available on macOS. There's an `IP_BOUND_IF` that you could likely use.

My main issue is that the following tests are broken on macOS:

```
test_set_interface[unix-iface]
test_set_interface[unix-iface-prefix]
test_set_interface[unix-ifhost-prefix]
test_set_interface[unix-ifhost-prefix-invalid]
```

It would be great to either gate the tests behind a flag or implement `IP_BOUND_IF` support.

### Debug/trace log

```text
[12:42:25.786431][MainThread][cli][debug] OS:         macOS 26.4.1
[12:42:25.786723][MainThread][cli][debug] Python:     3.13.12
[12:42:25.786778][MainThread][cli][debug] OpenSSL:    OpenSSL 3.6.1 27 Jan 2026
[12:42:25.786816][MainThread][cli][debug] Streamlink: 8.3.0
[12:42:25.786850][MainThread][cli][debug] Dependencies:
[12:42:25.788182][MainThread][cli][debug]  certifi: 2026.1.4
[12:42:25.789582][MainThread][cli][debug]  isodate: 0.7.2
[12:42:25.790253][MainThread][cli][debug]  lxml: 6.0.2
[12:42:25.790646][MainThread][cli][debug]  pycountry: 26.2.16
[12:42:25.791175][MainThread][cli][debug]  pycryptodome: 3.23.0
[...]

PR #6906, opened 2026-04-24
ci.github: add macos-latest runners

bastimeyer:
See #6903 

No idea about the reliability of macOS runners...

`macos-latest` are the ARM64-based ones:
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for--private-repositories

Let's see if there are any compatibility issues. Adding macOS runners also means that we now have 18 test runners, which might be a bit too much. I'll think about only adding one macOS test runner using the latest Python version, since we only need to hit certain code paths (without mocking) that are macOS-specific and not Python-version-specific on macOS.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -3,7 +3,7 @@ codecov:
     require_ci_to_pass: true
     # wait until at all test runners have uploaded a report (see the test job's build matrix)
     # otherwise, coverage failures may be shown while some reports are still missing
-    after_n_builds: 12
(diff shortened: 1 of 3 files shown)


bastimeyer:
This is already annoying since GH can't start all runners at once due to the concurrency limitation for public/free repos.

bastimeyer:
The fails are expected due to the broken --interface implementation

PR #6936, opened 2026-05-28
ci.github: pin all actions and add dependabot config

bastimeyer:
This switches from rolling tags to explicit SHA1 hashes for all actions, not just certain third-party actions. All hashes include a trailing comment with the targeted tag.

This also adds a dependabot config file for automated monthly PRs, to avoid having to update manually. Third-party actions will only be updated on major/minor version bumps and official actions will be grouped into one PR, unless it's a major version bump.

- https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference
- https://github.com/dependabot/dependabot-core/pull/5951

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,24 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    commit-message:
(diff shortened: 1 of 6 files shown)


bastimeyer:
upload-artifact and download-artifact were both bumped to a new major version, but this shouldn't affect the current setup

PR #6937, opened 2026-05-29
build: bump lxml to >=4.6.4,<8

bastimeyer:
https://github.com/lxml/lxml/blob/lxml-7.0.0a1/CHANGES.txt

No breaking changes API-wise in the alpha 1 release of lxml 7. Tests are all passing.

Going to merge this once they've published a finalized release.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ dependencies = [
   "certifi",
   "exceptiongroup ; python_version<'3.11'",
   "isodate",
-  "lxml >=4.6.4,<7",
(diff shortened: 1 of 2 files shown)


bastimeyer:
Going to merge this now, so Streamlink 8.5.0 won't cause packaging issues on some distros once lxml 7.0.0 releases. There haven't been any breaking changes in their alpha 3 release and tests ran just fine after building locally. In case there will be breaking changes that are relevant for us, which I doubt, then we'll have to publish fixes, but allowing lxml 7.x now seems like the better idea.

PR #6959, opened 2026-06-18
plugins.douyu: re-introduce plugin with new implementation

goodryb:
## Description

Add a new plugin for douyu.com, a popular Chinese live streaming platform focused on gaming content.

This restores the Douyu plugin that was removed in 2020 (commit c905798a) due to JS engine requirements. The new implementation uses a different API endpoint (getH5PlayV1) with a pure Python MD5-based authentication mechanism — no JS engine required.

## Features

- URL matching: `www.douyu.com/{room_id}`, `www.douyu.com/topic/{room_id}`
- Optional `dyshid` parameter support via URL query string for user-specific DID
- Room metadata extraction (title, author, category) via betard API
- Multi-quality stream support (returns all available qualities, e.g. 9399k, 4000k, 2000k, 900k)
- Custom `stream_weight()` for proper quality ordering
- Referer header for CDN compatibility
- Graceful handling of offline rooms

## Verification

```
$ streamlink --loglevel debug https://www.douyu.com/952595
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/952595
[plugins.douyu][debug] QUALITY_WEIGHTS: {'7805k': 7805, '4000k': 4000, '2000k': 2000, '900k': 900}
Available streams: 900k (worst), 2000k, 4000k, 7805k (best)

$ streamlink --loglevel debug https://www.douyu.com/2082749
[plugins.douyu][info] Streamer is offline
error: No playable streams found on this URL
```

## Tests

```
$ pytest tests/plugins/test_douyu.py -v
16 passed in 0.05s
```

## AI Assistance Disclosure

This plugin was developed with AI assistance (code generation and debugging). The author has full comprehension of all code submitted, including:
- The MD5-based auth algorithm (reversed from douyuEx userscript)
- The API flow (getEncryption → compute auth → getH5PlayV1)
- URL regex patterns and capture groups
- Streamlink plugin architecture conventions

All code has been manually verified and tested.

### DIFF
--- a/src/streamlink/plugins/douyu.py
+++ b/src/streamlink/plugins/douyu.py
@@ -86,8 +86,47 @@ class Douyu(Plugin):
             f = hashlib.md5((f + key).encode("utf-8")).hexdigest()
         return hashlib.md5((f + key + suffix).encode("utf-8")).hexdigest()
 
-    def _request_stream(self, rid: str, rate: int, did: str, enc_data: dict, auth: str) -> dict | None:
+    def _get_encryption(self, did: str) -> dict | None:
+        enc_response = self.session.http.get(
+            self._URL_ENCRYPTION,
+            params={"did": did},
+            schema=validate.Schema(
+                validate.parse_json(),
+                {
+                    "error": int,
+                    "data": validate.any(
+                        None,
+                        {
+                            "key": str,
+                            "rand_str": str,
+                            "enc_time": int,
+                            "enc_data": str,
+                            "is_special": int,
+                        },
+                    ),
+                },
+            ),
+        )
+        if enc_response["error"] != 0 or not enc_response["data"]:
+            return None
+        return enc_response["data"]
+
+    def _request_stream(self, rid: str, rate: int, did: str) -> dict | None:
+        enc_data = self._get_encryption(did)
+        if not enc_data:
+            log.error("Failed to get encryption parameters")
+            return None
+
         ts = int(time.time())
+        auth = self._compute_auth(
+            rid=rid,
+            ts=ts,
+            key=enc_data["key"],
+            rand_str=enc_data["rand_str"],
+            enc_time=enc_data["enc_time"],
+            is_special=enc_data["is_special"],
+        )
+
         play_data = self.session.http.post(
             self._URL_PLAY.format(rid=rid),
             data={
@@ -134,43 +173,7 @@ class Douyu(Plugin):
             log.info("Streamer is offline")
             return
 
-        enc_response = self.session.http.get(
-            self._URL_ENCRYPTION,
-            params={"did": did},
-            schema=validate.Schema(
-                validate.parse_json(),
-                {
-                    "error": int,
-                    "data": validate.any(
-                        None,
-                        {
-                            "key": str,
-                            "rand_str": str,
-                            "enc_time": int,
-                            "enc_data": str,
-                            "is_special": int,
-                        },
-                    ),
-                },
-            ),
-        )
-
-        if enc_response["error"] != 0 or not enc_response["data"]:
-            log.error("Failed to get encryption parameters")
-            return
-
-        enc_data = enc_response["data"]
-        ts = int(time.time())
-        auth = self._compute_auth(
-            rid=rid,
-            ts=ts,
-            key=enc_data["key"],
-            rand_str=enc_data["rand_str"],
-            enc_time=enc_data["enc_time"],
-            is_special=enc_data["is_special"],
-        )
-
-        first_data = self._request_stream(rid, 0, did, enc_data, auth)
+        first_data = self._request_stream(rid, 0, did)
         if not first_data:
             log.error("Failed to get stream URL")
             return
@@ -205,7 +208,7 @@ class Douyu(Plugin):
             if rate == 0:
                 data = first_data
             else:
-                data = self._request_stream(rid, rate, did, enc_data, auth)
+                data = self._request_stream(rid, rate, did)
 
             if data:
                 url = f"{data['rtmp_url']}/{data['rtmp_live']}"
--- a/src/streamlink/plugins/douyu.py
+++ b/src/streamlink/plugins/douyu.py
@@ -1,0 +1,237 @@
+"""
+$description 斗鱼直播 - 以游戏直播为主的弹幕式互动直播平台
+$url www.douyu.com
+$type live
+"""
+
+import hashlib
+import logging
+import re
+import time
+
+from streamlink.exceptions import NoStreamsError, PluginError
+from streamlink.plugin import Plugin, pluginargument, pluginmatcher
+from streamlink.plugin.api import validate
+from streamlink.stream.http import HTTPStream
+
+
+log = logging.getLogger(__name__)
+
+
+@pluginmatcher(
+    re.compile(
+        r"https?://(?:www\.)?douyu\.com/(?:topic/)?(?P<rid>\d+)(?:\?.*dyshid=(?P<dyshid>[^&]+))?",
+    ),
+)
+@pluginargument(
+    "rate",
+    metavar="RATE",
+    default="0",
+    help="Stream quality rate: 0=best (default), 1=smooth, 2=HD, 3=super HD, 4=4Mbps",
+)
+class Douyu(Plugin):
+    _URL_ENCRYPTION = "https://www.douyu.com/wgapi/livenc/liveweb/websec/getEncryption"
+    _URL_PLAY = "https://www.douyu.com/lapi/live/getH5PlayV1/{rid}"
+
+    DID = "10000000000000000000000000001501"
+
+    def _get_real_rid(self, rid: str) -> str:
+        """Get real room ID (URL room ID may differ from actual rid)"""
+        res = self.session.http.get(f"https://www.douyu.com/{rid}")
+
+        # Method 1: Extract from $ROOM.room_id
+        match = re.search(r"\$ROOM\.room_id\s*=\s*(\d+)", res.text)
+        if match:
+            return match.group(1)
+
+        # Method 2: Extract from roomID
+        match = re.search(r"roomID[\":\s]+(\d+)", res.text)
+        if match:
+            return match.group(1)
+
+        # Method 3: Extract from canonical link
+        match = re.search(r'<link[^>]+rel=["\']canonical["\'][^>]+href=["\'][^"\']*/(\d+)', res.text)
+        if match:
+            return match.group(1)
+
+        # Method 4: Return input if it's already a number
+        if rid.isdigit():
+            return rid
+
+        raise PluginError(f"Cannot get real room ID: {rid}")
+
+    def _get_room_metadata(self, rid: str):
+        """Get room metadata (streamer name, title, category, etc.)"""
+        try:
+            data = self.session.http.get(
+                f"https://www.douyu.com/betard/{rid}",
+                schema=validate.Schema(
+                    validate.parse_json(),
+                    {
+                        "room": {
+                            "room_id": int,
+                            "room_name": str,
+                            "nickname": str,
+                            "cate_name": str,
+                            "show_status": int,  # 1=live, 2=offline
+                        },
+                    },
+                    validate.get("room"),
+                ),
+            )
+            self.id = str(data["room_id"])
+            self.title = data["room_name"]
+            self.author = data["nickname"]
+            self.category = data["cate_name"]
+            return data["show_status"] == 1
+        except Exception:
+            log.debug("Failed to get room metadata, continuing with stream extraction")
+            return True  # Assume live, let subsequent steps determine
+
+    @staticmethod
+    def _compute_auth(rid: str, did: str, ts: int, key: str, rand_str: str, enc_time: int, is_special: int) -> str:
+        """
+        Compute auth signature parameter.
+
+        Algorithm from douyuEx userscript:
+        1. If is_special != 1, append rid + ts as suffix
+        2. Start from rand_str, iterate enc_time times: md5(f + key)
+        3. Final: md5(f + key + suffix) is the auth
+        """
+        suffix = "" if is_special == 1 else f"{rid}{ts}"
+        f = rand_str
+        for _ in range(enc_time):
+            f = hashlib.md5((f + key).encode("utf-8")).hexdigest()
+        return hashlib.md5((f + key + suffix).encode("utf-8")).hexdigest()
+
+    def _get_stream_url(self, rid: str, rate: str, did: str | None = None) -> str | None:
+        """
+        Get actual live stream URL.
+
+        Process:
+        1. GET getEncryption to get encryption parameters
+        2. Compute auth signature
+        3. POST getH5PlayV1 to get stream URL
+        """
+        did = did or self.DID
+
+        # Step 1: Get encryption parameters
+        enc_data = self.session.http.get(
+            self._URL_ENCRYPTION,
+            params={"did": did},
+            schema=validate.Schema(
+                validate.parse_json(),
+                {
+                    "error": int,
+                    "data": validate.any(
+                        None,
+                        {
+                            "key": str,
+                            "rand_str": str,
+                            "enc_time": int,
+                            "enc_data": str,
+                            "is_special": int,
+                        },
+                    ),
+                },
+            ),
+        )
+
+        if enc_data["error"] != 0 or not enc_data["data"]:
+            log.error("Failed to get encryption parameters")
+            return None
+
+        d = enc_data["data"]
+        ts = int(time.time())
+
+        # Step 2: Compute auth
+        auth = self._compute_auth(
+            rid=rid,
+            did=did,
+            ts=ts,
+            key=d["key"],
+            rand_str=d["rand_str"],
+            enc_time=d["enc_time"],
+            is_special=d["is_special"],
+        )
+
+        # Step 3: Request stream URL
+        play_data = self.session.http.post(
+            self._URL_PLAY.format(rid=rid),
+            data={
+                "enc_data": d["enc_data"],
+                "tt": str(ts),
+                "did": did,
+                "auth": auth,
+                "cdn": "",
+                "rate": rate,
+                "hevc": "0",
+                "fa": "0",
+                "ive": "0",
+            },
+            headers={"Content-Type": "application/x-www-form-urlencoded"},
+            schema=validate.Schema(
+                validate.parse_json(),
+                {
+                    "error": int,
+                    "data": validate.any(
+                        None,
+                        {
+                            "rtmp_url": str,
+                            "rtmp_live": str,
+                        },
+                    ),
+                },
+            ),
+        )
+
+        if play_data["error"] != 0 or not play_data["data"]:
+            error_code = play_data["error"]
+            if error_code == 102:
+                log.error("Room does not exist")
+            elif error_code == 104:
+                log.info("Room is offline")
+            else:
+                log.error(f"Failed to get stream URL (error={error_code})")
+            return None
+
+        url = f"{play_data['data']['rtmp_url']}/{play_data['data']['rtmp_live']}"
+        return url
+
+    def _get_streams(self):
+        rid_input = self.match.group("rid")
+        # Extract user DID from URL if available
+        user_did = self.match.groupdict().get("dyshid")
+
+        # Get real room ID
+        real_rid = self._get_real_rid(rid_input)
+        log.debug(f"Real room ID: {real_rid}")
+
+        # Get room metadata
+        is_live = self._get_room_metadata(real_rid)
+        if not is_live:
+            log.info("Streamer is offline")
+            raise NoStreamsError
+
+        # Get stream quality parameter
+        rate = self.get_option("rate")
+
+        # Get stream URL (prefer user DID if available)
+        stream_url = self._get_stream_url(real_rid, rate, did=user_did)
+        if not stream_url:
+            raise NoStreamsError
+
+        log.debug(f"Stream URL: {stream_url}")
+
+        # Douyu CDN requires Referer header
+        yield (
+            "live",
+            HTTPStream(
+                self.session,
+                stream_url,
+                headers={"Referer": "https://www.douyu.com/"},
+            ),
+        )
+
+
+__plugin__ = Douyu
--- a/src/streamlink/plugins/douyu.py
+++ b/src/streamlink/plugins/douyu.py
@@ -19,6 +19,7 @@ from urllib.parse import urlparse
 
 from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
+from streamlink.session.http import TLSSecLevel1Adapter
 from streamlink.stream.http import HTTPStream
 from streamlink.utils.parse import parse_qsd
 
@@ -173,6 +174,11 @@ class Douyu(Plugin):
         if not first_data:
             log.error("Failed to get stream URL")
             return
+
+        # Mount TLSSecLevel1Adapter for CDN domains that may not be compatible
+        # with OpenSSL 3.0's default security level (SECLEVEL=2)
+        cdn_netloc = urlparse(first_data["rtmp_url"]).netloc
+        self.session.http.mount(f"https://{cdn_netloc}", TLSSecLevel1Adapter())
 
         multirates = first_data.get("multirates", [])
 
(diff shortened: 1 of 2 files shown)


src/streamlink/plugins/douyu.py
```diff
@@ -0,0 +1,237 @@
+"""
+$description 斗鱼直播 - 以游戏直播为主的弹幕式互动直播平台
```
bastimeyer: English only. The description should be a short text about the site and who it is run by.

src/streamlink/plugins/douyu.py
```diff
@@ -0,0 +1,237 @@
...
+    re.compile(
+        r"https?://(?:www\.)?douyu\.com/(?:topic/)?(?P<rid>\d+)(?:\?.*dyshid=(?P<dyshid>[^&]+))?",
+    ),
+)
+@pluginargument(
+    "rate",
+    metavar="RATE",
+    default="0",
+    help="Stream quality rate: 0=best (default), 1=smooth, 2=HD, 3=super HD, 4=4Mbps",
+)
```
bastimeyer: This is not how Streamlink works. A plugin must return a collection of stream qualities for a particular stream. The respective quality is then chosen by the CLI implementation, not by any custom plugin arguments. If you have different non-standardized stream names, then you must implement a custom quality weight function.

Refactor `_get_stream_url` and return all (relevant) stream qualities. Or just return the best quality to avoid having to make unnecessary HTTP requests.

src/streamlink/plugins/douyu.py
```diff
@@ -0,0 +1,237 @@
...
+class Douyu(Plugin):
+    _URL_ENCRYPTION = "https://www.douyu.com/wgapi/livenc/liveweb/websec/getEncryption"
+    _URL_PLAY = "https://www.douyu.com/lapi/live/getH5PlayV1/{rid}"
+
+    DID = "10000000000000000000000000001501"
+
+    def _get_real_rid(self, rid: str) -> str:
+        """Get real room ID (URL room ID may differ from actual rid)"""
```
bastimeyer: All of these comments, not just this one, are superfluous. In the entire plugin. They don't provide any additional information that would be useful in order to understand what's going on.

src/streamlink/plugins/douyu.py
```diff
@@ -0,0 +1,237 @@
...
+            return match.group(1)
+
+        # Method 3: Extract from canonical link
+        match = re.search(r'<link[^>]+rel=["\']canonical["\'][^>]+href=["\'][^"\']*/(\d+)', res.text)
+        if match:
+            return match.group(1)
+
+        # Method 4: Return input if it's already a number
+        if rid.isdigit():
+            return rid
```
bastimeyer: Why does this check come last when `rid` is the input of this method? All channels I can see on their website have numeric IDs, which means that unless there's some "URL ID" -> "room ID" mapping, any of the regexes above are unneeded.

src/streamlink/plugins/douyu.py
```diff
@@ -0,0 +1,237 @@
...
+        if match:
+            return match.group(1)
+
+        # Method 2: Extract from roomID
+        match = re.search(r"roomID[\":\s]+(\d+)", res.text)
+        if match:
+            return match.group(1)
+
+        # Method 3: Extract from canonical link
+        match = re.search(r'<link[^>]+rel=["\']canonical["\'][^>]+href=["\'][^"\']*/(\d+)', res.text)
```
bastimeyer: Regexes should never be used for reading data from XML/HTML responses. The HTTP request should instead have a validation schema where the response is parsed as HTML and then an XPath query should be done.

src/streamlink/plugins/douyu.py
```diff
@@ -0,0 +1,237 @@
...
+    _URL_PLAY = "https://www.douyu.com/lapi/live/getH5PlayV1/{rid}"
+
+    DID = "10000000000000000000000000001501"
+
+    def _get_real_rid(self, rid: str) -> str:
+        """Get real room ID (URL room ID may differ from actual rid)"""
+        res = self.session.http.get(f"https://www.douyu.com/{rid}")
+
+        # Method 1: Extract from $ROOM.room_id
+        match = re.search(r"\$ROOM\.room_id\s*=\s*(\d+)", res.text)
```
bastimeyer: Please provide a link where this regex is needed. I have only seen channels/rooms where method 2 was needed, from the DOM rehydration data.

src/streamlink/plugins/douyu.py
```diff
@@ -0,0 +1,237 @@
...
+        if match:
+            return match.group(1)
+
+        # Method 2: Extract from roomID
+        match = re.search(r"roomID[\":\s]+(\d+)", res.text)
```
bastimeyer: Parsing the room ID like this is a bit "wonky", but it should work. Otherwise, we'll need a far more sophisticated extraction logic, but for a single value like this, it's fine.

src/streamlink/plugins/douyu.py
```diff
@@ -0,0 +1,237 @@
...
+                },
+            ),
+        )
+
+        if play_data["error"] != 0 or not play_data["data"]:
+            error_code = play_data["error"]
+            if error_code == 102:
+                log.error("Room does not exist")
+            elif error_code == 104:
+                log.info("Room is offline")
```
bastimeyer: It'd be better to have these constants as class attributes instead of literals within the function.

src/streamlink/plugins/douyu.py
```diff
@@ -0,0 +1,237 @@
...
+from streamlink.plugin.api import validate
+from streamlink.stream.http import HTTPStream
+
+
+log = logging.getLogger(__name__)
+
+
+@pluginmatcher(
+    re.compile(
+        r"https?://(?:www\.)?douyu\.com/(?:topic/)?(?P<rid>\d+)(?:\?.*dyshid=(?P<dyshid>[^&]+))?",
```
bastimeyer: Query string parameters should not be part of the URL matcher.

Do this instead:
```py
from streamlink.utils.parse import parse_qsd

params = parse_qsd(urlparse(self.url).query)
params.get("dyshid", "")
```

src/streamlink/plugins/douyu.py
```diff
@@ -0,0 +1,237 @@
...
+            },
+            headers={"Content-Type": "application/x-www-form-urlencoded"},
+            schema=validate.Schema(
+                validate.parse_json(),
+                {
+                    "error": int,
+                    "data": validate.any(
+                        None,
+                        {
+                            "rtmp_url": str,
```
bastimeyer: Considering that you're concatenating this value with `rtmp_live` down below, it'd be better to have `validate.url(scheme="https")` here, to ensure that it's a valid URL.

src/streamlink/plugins/douyu.py
```diff
@@ -0,0 +1,237 @@
...
+
+        # Get real room ID
+        real_rid = self._get_real_rid(rid_input)
+        log.debug(f"Real room ID: {real_rid}")
+
+        # Get room metadata
+        is_live = self._get_room_metadata(real_rid)
+        if not is_live:
+            log.info("Streamer is offline")
+            raise NoStreamsError
```
bastimeyer: `return None` instead of `raise NoStreamsError`.

`NoStreamsError` is okay, but it's usually used for having an exit mechanism in a deeper call stack.

tests/plugins/test_douyu.py
```diff
@@ -0,0 +1,26 @@
...
+        ("https://www.douyu.com/288016", {"rid": "288016"}),
+        ("https://www.douyu.com/topic/12345", {"rid": "12345"}),
+        ("https://www.douyu.com/952595?dyshid=abc123", {"rid": "952595", "dyshid": "abc123"}),
+    ]
+
+    should_not_match = [
+        "https://www.douyu.com/",
+        "https://www.douyu.com/member/cp",
+        "https://www.douyu.com/abc_def",
+        "https://example.com/12345",
```
bastimeyer: Generic non-matching URLs are already part of each plugin can-handle-url test.

bastimeyer (changes requested):
Thanks for the PR.

A plugin for this site already existed until 2020 when it got removed. There were several plugin requests on the issue tracker since then which apparently were closed because of certain JS restrictions at the time. Adding the plugin back is therefore fine.

However, the plugin previously was called `douyutv`, not `douyu`. I think it would be better to use the old name again.

```
$ git --no-pager log --oneline --grep douyu
c905798a plugins.douyutv: removed
e6db4113 Merge pull request #1109 from fozzysec/douyu_17jul
a31e0b64 update douyutv.py to support multiple rates by steven7851
357dc4e7 fix douyutv plugin by using new API
5cc7793f Merge pull request #1091 from steven7851/patch-12
51be89c8 plugins.douyutv: use douyu open API
38e50b24 Merge pull request #1060 from steven7851/patch-10
d1a17e21 plugins.douyutv: update post data
a277c62c [douyu] temporary fix by revert to previously commit (#1015)
fb81b65e plugins.douyutv: support vod (#706)
b89d4a0a Merge pull request #671 from fozzysec/douyu_new_api
1a34b26c update douyutv plugin, use new API
a2e5cc34 fix douyutv issue #637 (#666)
bc9733d2 plugins.douyutv: fix room id regex (#514)
d5428a79 Update plugin.douyutv
885ef257 update douyu
b52c77a6 plugins.douyutv: Use new api.
46302bac Updated douyutv.com plugin
7fc6afb4 plugins.douyutv: Add support for transcodes.
9e38b3ce Merge pull request #469 from nixxquality/douyutv
820a9cbd plugins.douyutv: Added
```

Before this PR can get merged, a couple of things need to be fixed. I've annotated all the issues I've found.

goodryb:
Hi @bastimeyer, thank you so much for the thorough review and detailed feedback! I really appreciate the time you took to go through the code.

I've addressed all the issues you pointed out in commit 9ff8524. Here's a summary of the changes:

### Changes made

- **Multi-quality streams**: Removed `--douyu-rate` plugin argument. The plugin now returns all available stream qualities (e.g. `9399k`, `4000k`, `2000k`, `900k`) with a custom `stream_weight()` implementation, following the pattern used in the Huya plugin.
- **URL matcher**: Removed `dyshid` from the regex. Now using `parse_qsd(urlparse(self.url).query)` to extract the optional DID parameter.
- **Deleted `_get_real_rid`**: All Douyu rooms use purely numeric IDs, so this function was unnecessary. This also resolves the HTML regex parsing concern.
- **Removed all superfluous comments**: The code should now be self-explanatory.
- **Constants as class attributes**: Extracted `HEVC`, `FA`, `IVE` and added `_URL_BETARD`.
- **`validate.url(scheme="https")`** for `rtmp_url`.
- **`return` instead of `raise NoStreamsError`**.
- **Test cleanup**: Removed generic non-matching URLs and `dyshid`-specific test cases.
- **Added `$metadata` declarations**: `id`, `author`, `category`, `title`.

### Additional fixes found during testing

- The `betard` API returns `cate_name: null` for some rooms, so I made it optional in the schema with `validate.any(None, str)`.
- The `getH5PlayV1` API returns `data: ""` (empty string) for offline rooms instead of `null`, so I added empty string as a valid schema value.

### About the plugin name

I noticed you suggested using the old name `douyutv`. After some consideration, I'd like to keep it as `douyu` because:
- The current domain is `www.douyu.com`
- `www.douyutv.com` now automatically redirects to `www.douyu.com`
- Other plugins in the project are typically named after their domain (e.g. `huya`, `bilibili`, `youtube`)

I believe `douyu` is more consistent with the current naming convention, but I'm happy to change it if you think otherwise.

### Verification

```
$ streamlink --loglevel debug https://www.douyu.com/952595
Available streams: 900k (worst), 2000k, 4000k, 7805k (best)

$ streamlink --loglevel debug https://www.douyu.com/2082749
[plugins.douyu][info] Streamer is offline
error: No playable streams found on this URL

$ pytest tests/plugins/test_douyu.py -v
16 passed in 0.05s
```

Please let me know if there's anything else that needs adjustment. Thanks again!

bastimeyer (approved):
Thanks for the update. Looking good...

> `www.douyutv.com` now automatically redirects to `www.douyu.com`

Ok, so this sounds like they've rebranded or at least changed their URL. I have no idea why the old plugin was named like this, but since there's the redirection now, it indeed doesn't make sense using the old name.

----

One issue I've noticed while testing a couple of streams is that sometimes streams can't be opened because of a 403 HTTP response.

This particular stream made problems where it was only accessible like 1 out of 10 times:

```
$ streamlink https://www.douyu.com/138243 best
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
```

```
$ streamlink https://www.douyu.com/138243 best
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
[cli][info] Available streams: 900k (worst), 2000k, 4000k, 15184k (best)
[cli][info] Opening stream: 15184k (http)
[cli][info] Player closed
[cli][info] Stream ended
```

On their website however, it's working perfectly fine.

Could there be an issue with the authentication?

bastimeyer:
```
$ while ! streamlink https://www.douyu.com/138243; do sleep 1; done
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
error: Unable to open URL: https://www.douyu.com/lapi/live/getH5PlayV1/138243 (403 Client Error: Forbidden for url: https://www.douyu.com/lapi/live/getH5PlayV1/138243)
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/138243
Available streams: 900k (worst), 2000k, 4000k, 15184k (best)
```

I think there's an issue when requesting multiple stream URLs using the same auth token.

goodryb:
Thanks for the review!

**Regarding the `douyu` naming:**

Agreed, since `www.douyutv.com` now redirects to `www.douyu.com`, using `douyu` as the plugin name makes sense.

**Regarding the intermittent 403 error on room 138243:**

I was able to reproduce this issue locally. After investigation, I found two separate problems:

1. **SSL handshake failure** — Some CDN nodes (e.g., `edgesrv.com:8443`) are not compatible with OpenSSL 3.0's default security level (SECLEVEL=2), causing SSL handshake failures when downloading streams in `--player` mode. I've fixed this by mounting a `TLSSecLevel1Adapter` for the CDN domain, similar to what the `filmon` plugin does.

2. **API 403 rate limiting** — The `getH5PlayV1` endpoint appears to have rate limiting that can reject requests with a 403 response. Using a real device ID (`did`) instead of a hardcoded fallback value reduces the chance of being rate-limited. Users can pass a real `did` via the `dyshid` URL parameter (e.g., `https://www.douyu.com/138243?dyshid=<real_did>`), which the plugin already supports via `parse_qsd`. Without it, the plugin falls back to a static default value.

With the SSL fix and a real `did`, I've confirmed that stream extraction and VLC playback both work correctly.

goodryb:
Apologies for the previous comment — I incorrectly attributed the 403 errors to API rate limiting when it was actually a bug in the plugin code. I also realize now that I pushed the last commit before properly addressing the logs you shared with the `while` loop test.

After further investigation, I found the real root cause: the `auth` token was computed with one timestamp, but the subsequent POST request to `getH5PlayV1` used a *different* (newer) timestamp for the `tt` field. This mismatch between the signed `auth` and the actual `tt` value caused the API to reject the request as invalid, resulting in the intermittent 403 errors.

The fix restructures `_request_stream` to be self-contained — each call now fetches fresh encryption parameters, computes the `auth` token, and sends the POST request all within the same method, ensuring that `enc_data`, `auth`, and `tt` are always consistent.

With this fix, the 403 issue on room 138243 is fully resolved in my local testing (previously ~40% success rate, now 100% across multiple attempts).

Sorry again for the premature conclusion in my last comment.

bastimeyer:
No reason to apologize. This is all fine. Thanks for taking the time investigating the issue thoroughly.

Also, using the `TLSSecLevel1Adapter` is much better than defaulting to HTTP.

The PR branch is looking good and all streams appear to be working fine. The only thing that is a bit unfortunate is that the API requests have to be made consecutively, so it's a bit slow fetching all available stream qualities, but this is a limitation of Streamlink's synchronous HTTP session (unless a thread pool is spawned, but this would be a ridiculous overkill for a plugin).

Going to merge now. Thanks again for the plugin re-introduction and re-implementation, @goodryb.

PR #6922, opened 2026-05-10
plugins.youtube: rewrite plugin, solve n-challenge

fl-Henry:
- [x] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [x] [I have read the rules about AI-assisted contributions](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#ai-assisted-contributions)

----

The submitted solution adds support for YouTube's n-parameter challenge that's required to access HLS streams. The old plugin still works as a fallback if anything goes wrong.

## What changed

Added two solvers for n-challenges:
- DenoSolver - uses deno JS runtime for solving n-challenges
- CDPSolver - solves n-challenges using Chrome DevTools Protocol

Both use the yt_dlp_ejs package to solve the tokens.

Added extractors for handling different YouTube pages:
- `VideoExtractor` for `/watch?v=` URLs
- `LiveExtractor` for `/@channel/live` URLs  
- `StreamsExtractor` for `/@channel/streams` URLs

Added `--youtube-stream` CLI option to pick which stream from the `/streams` page to play. Options are `first`, `last`, `popular` (default), or a number like `1`, `2`, `3`.

Fixed _data_video_id to work with YouTube's new initial JSON format. Still works with the old format too.

If the new code fails for any reason, it falls back to the original plugin automatically.

## Requirements

Deno is optional. If it's not installed, the plugin uses CDP instead.

PR #6976, opened 2026-07-05
build: remove redundant env markers in uv.lock

bastimeyer:
Changed in uv 0.11.25

----

Syncing the lockfile next time without this format change commit would have the changes included as a side-effect, which isn't great...

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -631,7 +631,7 @@ name = "exceptiongroup"
 version = "1.3.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "typing-extensions", marker = "python_full_version < '3.11'" },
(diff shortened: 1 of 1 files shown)


bastimeyer:
No idea what's going wrong. A uv version mismatch?
My local uv version has removed the markers after trying to upgrade the lockfile after adding the "youtube" dependencies extra with `yt_dlp_ejs` for #6922 and re-syncing it, so I wanted to exclude the unrelated changes and have them in a separate commit.
Will investigate this later...

bastimeyer:
lockfile export bug fixed in uv 0.11.27

bastimeyer:
This will have to be re-applied, because I just merged dependabot PRs which included the env markers due to dependabot still using an older version of uv.

PR #6995, opened 2026-07-18
Add Clippiti and Lurkiti information to thirdparty.rst

tarzasai:

----

Adding my two Streamlink powered GUI apps to the Third Party Applications list. Both open source and cross-platforms.

### DIFF
--- a/docs/thirdparty.rst
+++ b/docs/thirdparty.rst
@@ -76,4 +76,30 @@ OBS-Streamlink
   embedding streams directly as a scene source. This is useful for broadcasters of
   multi-platform live streams and for live commentators.
 
+Clippiti
+---------
+
+.. image:: https://raw.githubusercontent.com/tarzasai/Clippiti/main/doc/media/main-window.png
+  :align: center
+  :alt: Clippiti screenshot with OSD help and pinned toolbar
+
+:Description: Cross-platforms livestream player with clipping and recording
+:OS: :fab:`windows;fa-xl` :fab:`apple;fa-xl` :fab:`linux;fa-xl`
+:Author: `tarzasai <https://github.com/tarzasai>`_
+:Website: https://github.com/tarzasai/clippiti
+:Info: A PyQt6 desktop app that opens livestreams with Streamlink, keeps a rolling local buffer, and lets you take snapshots or save both live and past moments as clips (ts or mp4). Quick install from PyPi.
+
+Lurkiti
+---------
+
+.. image:: https://raw.githubusercontent.com/tarzasai/Lurkiti/main/media/settings1.png
+  :align: center
+  :alt: Lurkiti screenshot of the settings window
+
+:Description: A system-tray app to monitors and launch livestreams with Streamlink
+:OS: :fab:`windows;fa-xl` :fab:`apple;fa-xl` :fab:`linux;fa-xl`
+:Author: `tarzasai <https://github.com/tarzasai>`_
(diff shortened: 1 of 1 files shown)


bastimeyer (commented):
Thanks for the PR.

While I appreciate that Streamlink is used as the base for new projects, I'm a little bit hesitant merging this. Both of your projects don't appear to have any users yet, and one of them also appears to be brand new. Considering this, I feel like we should wait a bit before listing them here, because I am not sure if your projects will be kept maintained in the future. And yes, I am aware that there are projects on there that seem to have been abandoned. Those will have to be removed... I haven't noticed this until now.

Also, nowadays I unfortunately have to ask... Was AI/LLM coding involved? From looking at the code, I don't believe so, which is good, because we don't want to advertise vibe-coded projects.

To give you a bit of feedback on your projects: it looks like you're using the streamlink CLI instead of using Streamlink's Python API. For a Python project itself, this choice is a bit weird, as you're unnecessarily running two Python interpreters. You're therefore also executing it twice, once for reading stream data from stdout and once for reading stream metadata as JSON from stdout.

Regarding your changes, please split this up into two commits. Image URLs must always be static and immutable, meaning you must use a commit ID instead of a branch or tag name. Images should also be neutral, so showing any kind of stream content, especially people on camera, is an absolute no-go. Please also respect the max line length of the docs.

tarzasai:
Thanks for the comment @bastimeyer.

> While I appreciate that Streamlink is used as the base for new projects, I'm a little bit hesitant merging this. Both of your projects don't appear to have any users yet, and one of them also appears to be brand new. [...]

Yes they are quite new and I don't think they have users yet (except me of course). Perhaps somebody will find them on PyPi. TBH I don't know where else I could promote them... the Streamlink subreddit seems abandoned and the Twitch ones only care about streamers drama. Anyway while I would be happy if somebody liked these apps, I made them for me.

> Also, nowadays I unfortunately have to ask... Was AI/LLM coding involved? From looking at the code, I don't believe so, which is good, because we don't want to advertise vibe-coded projects.

To generate unit tests and other boring stuff yes, after 30 years of software development I'm happy to delegate those things to a machine :)
But seriously, I don't do vibe coding.

> To give you a bit of feedback on your projects: it looks like you're using the streamlink CLI instead of using Streamlink's Python API. For a Python project itself, this choice is a bit weird, as you're unnecessarily running two Python interpreters. You're therefore also executing it twice, once for reading stream data from stdout and once for reading stream metadata as JSON from stdout.

Now I feel like an idiot because I just didn't think to do that. I guess I was too happy that Clippiti was finally working. I'll fix both.

> Regarding your changes, please split this up into two commits.

Sorry you mean 2 PRs?

> Image URLs must always be static and immutable, meaning you must use a commit ID instead of a branch or tag name.

Sure.

> Images should also be neutral, so showing any kind of stream content, especially people on camera, is an absolute no-go. Please also respect the max line length of the docs.

Is it ok if I blur the video area inside the window? The player is minimalistic so there's literally nothing else to show. Otherwise I can submit a PR for Clippiti without screenshot.

bastimeyer:
> Sorry you mean 2 PRs?

No, two commits on a single PR. Otherwise, you'd have to rebase and fix conflicts after one got merged. But if you want to do two PRs, I don't mind either.

> Is it ok if I blur the video area inside the window?

Just choose something neutral.

tarzasai:
Thanks @bastimeyer, I will open another PR after fixing the apps to use the api.

PR #6997, opened 2026-07-20
Add Lurkiti and Clippiti apps to the third party list.

tarzasai:

----

Re-proposing my apps in 2 separate commits as requested - sorry for the different title styles, you would think that Github suggests consistent titles, but it doesn't.

I also updated both the app to create a better Streamlink session, even more in Clippiti as it was spawning a separate process instead of using the Streamlink python module. A big thank you to @bastimeyer for pointing it out.

### DIFF
--- a/docs/thirdparty.rst
+++ b/docs/thirdparty.rst
@@ -95,4 +95,21 @@ Lurkiti
   per-stream quality, notification, Streamlink and player option overrides.
   Quick install from PyPi.
 
+Clippiti
+--------
+
+.. image:: https://raw.githubusercontent.com/tarzasai/clippiti/v0.1.16/doc/media/main-window.png
+  :align: center
+  :alt: Clippiti main window with OSD help and pinned toolbar
+
+:Description: Livestream player with a rolling buffer, clipping, and recording
+:Type: GUI (graphical user interface)
+:OS: :fab:`windows;fa-xl` :fab:`apple;fa-xl` :fab:`linux;fa-xl`
+:Author: `tarzasai <https://github.com/tarzasai>`_
+:Website: https://github.com/tarzasai/clippiti
+:Info: Clippiti is a PyQt6 desktop player that opens a livestream and keeps a
+  rolling local HLS buffer, letting you take snapshots, save past moments as
+  clips, and record with optional auto-remux to MP4. It uses the Streamlink
+  Python API for stream resolution and ffmpeg for buffering and export.
+
 .. content list end
--- a/docs/thirdparty.rst
+++ b/docs/thirdparty.rst
@@ -79,7 +79,7 @@ OBS-Streamlink
 Lurkiti
 -------
 
-.. image:: https://raw.githubusercontent.com/tarzasai/Lurkiti/v1.0.22/media/settings1.png
+.. image:: https://github.com/user-attachments/assets/25ae6d3c-f2d1-4303-b686-d9d703d38661
   :align: center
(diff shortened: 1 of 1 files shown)


docs/thirdparty.rst
```diff
@@ -76,4 +76,40 @@ OBS-Streamlink
   embedding streams directly as a scene source. This is useful for broadcasters of
   multi-platform live streams and for live commentators.
 
+Lurkiti
+-------
+
+.. image:: https://raw.githubusercontent.com/tarzasai/Lurkiti/v1.0.22/media/settings1.png
```
bastimeyer: Tag names can be replaced. I asked for static/immutable URLs in my previous review. I'm not saying that you will replace this and hijack our docs, but in theory this is possible. Use the commit ID of that tag, or better, don't commit any media files to your git repo at all, because you will carry this binary data forever, even when removing or replacing it later on. Just git things... Paste the image to your repo's issue tracker and use GitHub's generated link. You don't even have to submit the post.

docs/thirdparty.rst
```diff
@@ -76,4 +76,40 @@ OBS-Streamlink
...
+:Info: Lurkiti is a system-tray app that monitors a list of livestreams (Twitch,
+  YouTube and any other Streamlink-supported site) and notifies you when they go
+  live. Streams open in your preferred player (mpv, VLC, Clippiti, etc.) with
+  per-stream quality, notification, Streamlink and player option overrides.
+  Quick install from PyPi.
+
+Clippiti
+--------
+
+.. image:: https://raw.githubusercontent.com/tarzasai/clippiti/v0.1.16/doc/media/main-window.png
```
bastimeyer: Same...

Also not neutral. There's advertisement in the title. Sorry, but I'm intentionally very picky with this.

bastimeyer (commented):
Thanks for the update.

> Re-proposing my apps in 2 separate commits as requested

You could've just updated the other PR branch with a force push instead of opening a new PR. But it's fine. If I'm going to merge this, I will have to rebase this anyway, because of the commit messages.

> I also updated both the app to create a better Streamlink session

I had a look, and just as a quick heads-up, the `streamlink_cli` module is technically not a public interface. I was talking about the Streamlink API in my previous comment. Streamlink's CLI sits on top the `streamlink` module, just like any other client can and should. But considering that you're using the same CLI args, it probably makes sense importing this - not sure. I haven't looked that deep into the code, just a bit.

----

As said earlier, let's wait a tiny bit before merging, because I don't want to add some fresh projects which don't have many users yet. But yes, I totally understand that this probably the best way for your projects to get some new users, and I want to see third-party projects to grow too.

tarzasai:
> I had a look, and just as a quick heads-up, the `streamlink_cli` module is technically not a public interface.

I know, but the only alternative I could see was to map all the Streamlink arguments (and maybe the plugins' too) in my code. Not a realistic alternative. But I'll try to find another way.

> As said earlier, let's wait a tiny bit before merging,

Sure, no prob.

docs/thirdparty.rst
tarzasai: > Paste the image to your repo's issue tracker and use GitHub's generated link.

aah, I didn't think about that. Thanks for the hint.
Is it ok to add a commit to replace the images?

docs/thirdparty.rst
tarzasai: I'll use a screenshot of the settings window then.

PR #7003, opened 2026-07-22
plugins.douyu: no assert_hostname in HTTPAdapter

### LINKED ISSUE #7001: plugins.douyu: CDN redirects to underscore-containing domain causing SSL certificate validation error
### Checklist

- [x] This is a [plugin issue](https://streamlink.github.io/plugins.html) and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [x] [I have checked the list of open and recently closed plugin issues](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22plugin+issue%22)
- [x] [I have checked the commit log of the master branch](https://github.com/streamlink/streamlink/commits/master)

### Collaboration

- [x] [I will provide feedback should a pull request be opened with a fix for the plugin](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-request-feedback)

### Streamlink version

the master branch (last commit 488f18f)

### Description

First of all, thanks  for bring plugin douyu back.

I have reviewed the entire discussion thread of #6959 and noticed that several CDN-related issues have already been identified and fixed during the review process. However, I have encountered a new issue that was not covered in that PR.

In some cases, the CDN used by Douyu will redirect the request to a domain name that contains an underscore (_).

Due to the way OpenSSL validates domain names, underscores are considered illegal characters in DNS names (as per RFC 1034/1035 and the SSL/TLS specification). As a result, when Streamlink attempts to establish a secure connection to such a redirected domain, an SSL certificate validation error is thrown, and the stream cannot be fetched.

### Debug/trace log

```text
streamlink --loglevel=debug --player mpv https://www.douyu.com/9999 best
NoTagError: `git describe --long --dirty --always --tags` could not find a tag
[cli][debug] OS:         Linux-7.1.3-arch1-3-x86_64-with-glibc2.43
[cli][debug] Python:     3.14.6
[cli][debug] OpenSSL:    OpenSSL 3.6.3 9 Jun 2026
[cli][debug] Streamlink: 0.0.0+unknown
[cli][debug] Dependencies:
[cli][debug]  certifi: 2026.6.17
[cli][debug]  isodate: 0.7.2
[cli][debug]  lxml: 6.1.1
[cli][debug]  pycountry: 26.2.16
[cli][debug]  pycryptodome: 3.23.0
[cli][debug]  PySocks: 1.7.1
[cli][debug]  requests: 2.34.2
[cli][debug]  trio: 0.33.0
[cli][debug]  trio-websocket: 0.12.2
[cli][debug]  urllib3: 2.7.0
[cli][debug]  websocket-client: 1.9.0
[cli][debug] Arguments:
[cli][debug]  url=https://www.douyu.com/9999
[cli][debug]  stream=['best']
[cli][debug]  --player=mpv
[cli][info] Found matching plugin douyu for URL https://www.douyu.com/9999
[plugins.douyu][debug] QUALITY_WEIGHTS: {'14460k': 14460, '8000k': 8000, '4000k': 4000, '2000k': 2000, '900k': 900}
[cli][info] Available streams: 900k (worst), 2000k, 4000k, 8000k, 14460k (best)
[cli][info] Opening stream: 14460k (http)
[cli][error] Try 1/1: Could not open stream <HTTPStream ['http', 'https://huos1a.douyucdn2.cn/live/9999rMfVyoBTjV0x.flv?wsAuth=f55c93cb4782193de0425cfb325450a5&token=web-h5-0-9999-26339a83b5a3e114f52f1ad798419f40934aa30a01cf244c&logo=0&expire=0&did=10000000000000000000000000001501&pt=2&st=0&sid=434117773&mcid2=0&origin=dy&fcdn=hs&fo=0&mix=0&isp=']> (Could not open stream: Unable to open URL: https://huos1a.douyucdn2.cn/live/9999rMfVyoBTjV0x.flv?wsAuth=f55c93cb4782193de0425cfb325450a5&token=web-h5-0-9999-26339a83b5a3e114f52f1ad798419f40934aa30a01cf244c&logo=0&expire=0&did=10000000000000000000000000001501&pt=2&st=0&sid=434117773&mcid2=0&origin=dy&fcdn=hs&fo=0&mix=0&isp= (HTTPSConnectionPool(host='2408_8748_a100_5_8000_0_b00_20.bytefcdnrd.com', port=443): Max retries exceeded with url: /live/9999rMfVyoBTjV0x.flv?302_type=cold_aggr&_session_id=1040171713.n.cn-cn2509g1.44933_2755660&cac=1&cb_retry=0&did=10000000000000000000000000001501&domain=huos1a.douyucdn2.cn&expire=0&fcdn=hs&fo=0&fp_user_url=https%3A%2F%2Fhuos1a.douyucdn2.cn%2Flive%2F9999rMfVyoBTjV0x.flv%3FwsAuth%3Df55c93cb4782193de0425cfb325450a5%26token%3Dweb-h5-0-9999-26339a83b5a3e114f52f1ad798419f40934aa30a01cf244c%26logo%3D0%26expire%3D0%26did%3D10000000000000000000000000001501%26pt%3D2%26st%3D0%26sid%3D434117773%26mcid2%3D0%26origin%3Ddy%26fcdn%3Dhs%26fo%3D0%26mix%3D0%26isp%3D&isp=&logo=0&manage_ip=&mcid2=0&mix=0&node_id=&origin=dy&pro_type=http2&pt=2&redirect_from=pod.cn-cn2509g1.44933.nss&redirect_from_ip=2408%3A871a%3A1101%3A1%3A8000%3A0%3Ab00%3A27&redirect_to=fc.cn-cn1731g1&redirect_to_ip=2408%3A8748%3Aa100%3A5%3A8000%3A0%3Ab00%3A20&sid=434117773&st=0&token=web-h5-0-9999-26339a83b5a3e114f52f1ad798419f40934aa30a01cf244c&vhost=huosa.douyucdn2.cn&wsAuth=f55c93cb4782193de0425cfb325450a5 (Caused by SSLError(SSLCertVerificationError(1, "[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Hostname mismatch, certificate is not valid for '2408_8748_a100_5_8000_0_b00_20.bytefcdnrd.com'. (_ssl.c:1082)")))))
error: Could not open stream <HTTPStream ['http', 'https://huos1a.douyucdn2.cn/live/9999rMfVyoBTjV0x.flv?wsAuth=f55c93cb4782193de0425cfb325450a5&token=web-h5-0-9999-26339a83b5a3e114f52f1ad798419f40934aa30a01cf244c&logo=0&expire=0&did=10000000000000000000000000001501&pt=2&st=0&sid=434117773&mcid2=0&origin=dy&fcdn=hs&fo=0&mix=0&isp=']>, tried 1 times, exiting
```

hxzhao527:
- [X] [I have read the rules about AI-assisted contributions](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#ai-assisted-contributions)

----

fix #7001 

set `kwargs["assert_hostname"] = False` to disable domain check.

### DIFF
--- a/src/streamlink/plugins/douyu.py
+++ b/src/streamlink/plugins/douyu.py
@@ -19,12 +19,28 @@ from urllib.parse import urlparse
 
 from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
-from streamlink.session.http import TLSPartialVerifyAdapter
+from streamlink.session.http import TLSSecLevel1Adapter
 from streamlink.stream.http import HTTPStream
 from streamlink.utils.parse import parse_qsd
 
 
 log = logging.getLogger(__name__)
+
+
+class TLSPartialVerifyAdapter(TLSSecLevel1Adapter):
+    """
+    The base class SSLContextAdapter resets ssl_context.check_hostname to True when sending a request, so overriding get_ssl_context does not take effect.
+
+    Instead, assert_hostname is used to influence urllib3's internal decision logic,
+    ultimately achieving the effect of check_hostname=False to skip certificate hostname verification,
+    while still retaining certificate chain-related validation."
+    """
+
+    def init_poolmanager(self, *args, **kwargs):
+        kwargs["ssl_context"] = self.get_ssl_context()
+        kwargs["assert_hostname"] = False
+        return super().init_poolmanager(*args, **kwargs)
+
 
 @pluginmatcher(
     re.compile(
--- a/src/streamlink/plugins/douyu.py
+++ b/src/streamlink/plugins/douyu.py
@@ -37,7 +37,6 @@ class TLSPartialVerifyAdapter(TLSSecLevel1Adapter):
     """
 
     def init_poolmanager(self, *args, **kwargs):
-        kwargs["ssl_context"] = self.get_ssl_context()
         kwargs["assert_hostname"] = False
         return super().init_poolmanager(*args, **kwargs)
 
--- a/src/streamlink/plugins/douyu.py
+++ b/src/streamlink/plugins/douyu.py
@@ -29,7 +29,8 @@ log = logging.getLogger(__name__)
 
 class TLSPartialVerifyAdapter(TLSSecLevel1Adapter):
     """
-    The base class SSLContextAdapter resets ssl_context.check_hostname to True when sending a request, so overriding get_ssl_context does not take effect.
+    The base class SSLContextAdapter resets ssl_context.check_hostname to True when sending a request,
+    so overriding get_ssl_context does not take effect.
 
     Instead, assert_hostname is used to influence urllib3's internal decision logic,
     ultimately achieving the effect of check_hostname=False to skip certificate hostname verification,
@@ -192,7 +193,8 @@ class Douyu(Plugin):
             log.error("Failed to get stream URL")
             return
 
-        # Mount TLSPartialVerifyAdapter for CDN domains that may not be compatible with OpenSSL 3.0's default security level (SECLEVEL=2)
+        # Mount TLSPartialVerifyAdapter for CDN domains that may not be compatible
+        # with OpenSSL 3.0's default security level (SECLEVEL=2)
         # or domains with underscore
         self.session.http.mount("https://", TLSPartialVerifyAdapter())
 
--- a/src/streamlink/plugins/douyu.py
+++ b/src/streamlink/plugins/douyu.py
@@ -19,13 +19,12 @@ from urllib.parse import urlparse
 
 from streamlink.plugin import Plugin, pluginmatcher
 from streamlink.plugin.api import validate
-from streamlink.session.http import TLSSecLevel1Adapter
+from streamlink.session.http import TLSPartialVerifyAdapter
 from streamlink.stream.http import HTTPStream
 from streamlink.utils.parse import parse_qsd
 
 
 log = logging.getLogger(__name__)
-
 
 @pluginmatcher(
     re.compile(
@@ -178,10 +177,9 @@ class Douyu(Plugin):
             log.error("Failed to get stream URL")
             return
 
-        # Mount TLSSecLevel1Adapter for CDN domains that may not be compatible
-        # with OpenSSL 3.0's default security level (SECLEVEL=2)
-        cdn_netloc = urlparse(first_data["rtmp_url"]).netloc
-        self.session.http.mount(f"https://{cdn_netloc}", TLSSecLevel1Adapter())
+        # Mount TLSPartialVerifyAdapter for CDN domains that may not be compatible with OpenSSL 3.0's default security level (SECLEVEL=2)
+        # or domains with underscore
+        self.session.http.mount("https://", TLSPartialVerifyAdapter())
 
         multirates = first_data.get("multirates", [])
 
--- a/src/streamlink/session/http.py
+++ b/src/streamlink/session/http.py
@@ -415,5 +415,16 @@ class TLSSecLevel1Adapter(SSLContextAdapter):
 
         return ctx
 
+class TLSPartialVerifyAdapter(TLSSecLevel1Adapter):
+    def get_ssl_context(self) -> ssl.SSLContext:
+        ctx = super().get_ssl_context()
+        # not work, will be set to True in create_urllib3_context because verify_mode == ssl.CERT_REQUIRED
+        # ctx.check_hostname = False 
+        ctx.verify_mode = ssl.CERT_REQUIRED
+        return ctx
+    def init_poolmanager(self, *args, **kwargs):
+        kwargs["ssl_context"] = self.get_ssl_context()
+        kwargs["assert_hostname"] = False
+        return super().init_poolmanager(*args, **kwargs)
 
-__all__ = ["HTTPSession", "SSLContextAdapter", "TLSNoDHAdapter", "TLSSecLevel1Adapter"]
+__all__ = ["HTTPSession", "SSLContextAdapter", "TLSNoDHAdapter", "TLSSecLevel1Adapter", "TLSPartialVerifyAdapter"]
--- a/src/streamlink/session/http.py
+++ b/src/streamlink/session/http.py
@@ -415,16 +415,5 @@ class TLSSecLevel1Adapter(SSLContextAdapter):
 
         return ctx
 
-class TLSPartialVerifyAdapter(TLSSecLevel1Adapter):
-    def get_ssl_context(self) -> ssl.SSLContext:
-        ctx = super().get_ssl_context()
-        # not work, will be set to True in create_urllib3_context because verify_mode == ssl.CERT_REQUIRED
-        # ctx.check_hostname = False 
-        ctx.verify_mode = ssl.CERT_REQUIRED
-        return ctx
-    def init_poolmanager(self, *args, **kwargs):
-        kwargs["ssl_context"] = self.get_ssl_context()
-        kwargs["assert_hostname"] = False
-        return super().init_poolmanager(*args, **kwargs)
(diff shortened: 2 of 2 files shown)


src/streamlink/session/http.py
```diff
@@ -415,5 +415,16 @@ def get_ssl_context(self) -> ssl.SSLContext:
...
+    def get_ssl_context(self) -> ssl.SSLContext:
+        ctx = super().get_ssl_context()
+        # not work, will be set to True in create_urllib3_context because verify_mode == ssl.CERT_REQUIRED
+        # ctx.check_hostname = False 
+        ctx.verify_mode = ssl.CERT_REQUIRED
+        return ctx
+    def init_poolmanager(self, *args, **kwargs):
+        kwargs["ssl_context"] = self.get_ssl_context()
+        kwargs["assert_hostname"] = False
+        return super().init_poolmanager(*args, **kwargs)
```
bastimeyer: This whole adapter class doesn't belong here, as it's not being reused. If it's indeed needed, then it should be part of the plugin itself.

Reusable adapters would also need tests.

src/streamlink/session/http.py
```diff
@@ -415,5 +415,16 @@ def get_ssl_context(self) -> ssl.SSLContext:
...
+class TLSPartialVerifyAdapter(TLSSecLevel1Adapter):
+    def get_ssl_context(self) -> ssl.SSLContext:
+        ctx = super().get_ssl_context()
+        # not work, will be set to True in create_urllib3_context because verify_mode == ssl.CERT_REQUIRED
+        # ctx.check_hostname = False 
+        ctx.verify_mode = ssl.CERT_REQUIRED
+        return ctx
+    def init_poolmanager(self, *args, **kwargs):
+        kwargs["ssl_context"] = self.get_ssl_context()
+        kwargs["assert_hostname"] = False
```
bastimeyer: `assert_hostname` as the poolmanager keyword is just a wrapper keyword for setting `context.check_hostname`, so the `init_poolmanager()` override is not necessary.

- https://github.com/urllib3/urllib3/blob/2.7.0/src/urllib3/connection.py#L811
- https://github.com/urllib3/urllib3/blob/2.7.0/src/urllib3/connection.py#L939-L951
- https://docs.python.org/3/library/ssl.html#ssl.SSLContext.check_hostname

src/streamlink/session/http.py
```diff
@@ -415,5 +415,16 @@ def get_ssl_context(self) -> ssl.SSLContext:
 
         return ctx
 
+class TLSPartialVerifyAdapter(TLSSecLevel1Adapter):
+    def get_ssl_context(self) -> ssl.SSLContext:
+        ctx = super().get_ssl_context()
+        # not work, will be set to True in create_urllib3_context because verify_mode == ssl.CERT_REQUIRED
+        # ctx.check_hostname = False 
+        ctx.verify_mode = ssl.CERT_REQUIRED
```
bastimeyer: The `verify_mode` attr should default to `ssl.CERT_REQUIRED`, as we don't set urllib3's `cert_reqs` keyword when creating the context.

urllib3's `create_urllib3_context()` function includes this code at the bottom, so I'm not sure if this was the reason why you added it, since the `context.verify_mode` property has a setter in Python's ssl module, and I'm not sure what this is about...

```py
    # The order of the below lines setting verify_mode and check_hostname
    # matter due to safe-guards SSLContext has to prevent an SSLContext with
    # check_hostname=True, verify_mode=NONE/OPTIONAL.
    # We always set 'check_hostname=False' for pyOpenSSL so we rely on our own
    # 'ssl.match_hostname()' implementation.
    if cert_reqs == ssl.CERT_REQUIRED and not IS_PYOPENSSL:
        context.verify_mode = cert_reqs
        context.check_hostname = True
    else:
        context.check_hostname = False
        context.verify_mode = cert_reqs
```

hxzhao527: I finally found where `check_hostname` is set to `True`. At first I thought it was set inside `create_urllib3_context`, but after debugging, I discovered that it's actually caused by the `send` implementation of the base `class SSLContextAdapter`

https://github.com/streamlink/streamlink/blob/488f18fe2585685d25bf6b5ef0665a5d9e2faa70/src/streamlink/session/http.py#L386-L394

I will clean `get_ssl_context` override, and just keep `kwargs["assert_hostname"] = False`

src/streamlink/session/http.py
hxzhao527: .venv/lib/python3.14/site-packages/urllib3/connection.py

`assert_hostname` and `check_hostname` are related, but if we only set `check_hostname` to `False`, it will be reset to `True` by the current `SSLContextAdapter`. However, if we explicitly set `assert_hostname` to `False`, then when `HTTPSConnection` executes connect, it will set `check_hostname` back to `False` again, ultimately resulting in no certificate verification.

```python
    # In some cases, we want to verify hostnames ourselves
    if (
        # `ssl` can't verify fingerprints or alternate hostnames
        assert_fingerprint
        or assert_hostname
        # assert_hostname can be set to False to disable hostname checking
        or assert_hostname is False
        # We still support OpenSSL 1.0.2, which prevents us from verifying
        # hostnames easily: https://github.com/pyca/pyopenssl/pull/933
        or ssl_.IS_PYOPENSSL
        or not ssl_.HAS_NEVER_CHECK_COMMON_NAME
    ):
        context.check_hostname = False
```

https://github.com/urllib3/urllib3/blob/6a8536b6e1b0cbe69ae801337c599951e527ae29/src/urllib3/connection.py#L989

src/streamlink/session/http.py
hxzhao527: Got it, I will move it to plugin inside.

src/streamlink/plugins/douyu.py
```diff
@@ -27,6 +27,21 @@
...
+    """
+    The base class SSLContextAdapter resets ssl_context.check_hostname to True when sending a request, so overriding get_ssl_context does not take effect.
+
+    Instead, assert_hostname is used to influence urllib3's internal decision logic,
+    ultimately achieving the effect of check_hostname=False to skip certificate hostname verification,
+    while still retaining certificate chain-related validation."
+    """
+
+    def init_poolmanager(self, *args, **kwargs):
+        kwargs["ssl_context"] = self.get_ssl_context()
```
bastimeyer: `TLSSecLevel1Adapter` inherits from `SSLContextAdapter`, which already calls `self.get_ssl_context()` in its `init_poolmanager()` method. All you need to do is set the `assert_hostname` kwarg here, otherwise the context object is built twice.

src/streamlink/session/http.py
bastimeyer: This was added in #6205 due to incompatibilities between `--http-no-ssl-verify` and `--http-disable-dh`. The PR has a bit more context in its description.

It's a bit difficult remembering these things... I should've been a bit more thorough in my earlier comment(s).

Considering that you want to keep the cert verification, but only disable the hostname comparison, simply setting `verify=False` via the respective Streamlink session option (`http-ssl-verify`) unfortunately won't work, so the custom adapter will be needed, I guess.

bastimeyer (commented):
Thanks for the update and the further clarification in the issue thread.

src/streamlink/session/http.py
hxzhao527: As shown in the logs you posted in #7001, the certificate mismatch does not always get triggered. Therefore, blindly using `--http-no-ssl-verify` is somewhat inappropriate, and for such known cases, requiring users to add a CLI parameter at runtime is not very convenient either.

Therefore, customizing the adapter seems to be the most appropriate approach.

bastimeyer:
Thanks

PR #7042, opened 2026-08-12
build: bump setuptools to 84.0.0

bastimeyer:
Closes #7038
Closes #7040


bastimeyer:
No breaking changes for us. Also checked the sdist + wheels. Looking fine...

PR #7075, opened 2026-09-02
stream.dash: fix t attr in SegmentTimeline

winklemad:

----

`SegmentTimeline.segments` only honoured an `<S>` element's explicit `@t` when the running time was still `0`:

```python
t = 0
for tsegment in self.timeline_segments:
    if t == 0 and tsegment.t is not None:
        t = tsegment.t
    ...
```

So `@t` was applied on the first segment only. In a `SegmentTimeline`, a later `<S t="...">` sets a new start time to signal a **gap / discontinuity** (per ISO/IEC 23009-1). That start time was ignored and the segment was placed at the accumulated time instead, producing wrong `$Time$` values — and therefore requests for the wrong segments.

Example: `<S t="0" d="4000" r="1"/>` followed by `<S t="10000" d="4000"/>` should produce `$Time$` = 0, 4000, **10000**, but produced 0, 4000, **8000**.

### Fix

Apply `@t` whenever an `<S>` carries it, not only on the first segment. Contiguous timelines (no `@t` on later segments) are unaffected.

### Test

Added `test_segments_timeline_gap` with a fixture MPD (`dash/test_timeline_gap.mpd`). On `master` the third segment is `?t=8000`; with this change it is `?t=10000`. The existing DASH segment tests stay green and ruff is clean.

### AI-assisted contributions

I used an AI assistant (Claude Code) to help find this and draft the test. I verified it myself — reproduced the wrong `$Time$` red→green through the real `MPD` parser, ran the DASH test suite, and confirmed ruff passes. I understand the change and take responsibility for it.

### DIFF
--- a/src/streamlink/stream/dash/manifest.py
+++ b/src/streamlink/stream/dash/manifest.py
@@ -1041,7 +1041,9 @@ class SegmentTimeline(MPDNode):
     def segments(self) -> Iterator[TimelineSegment]:
         t = 0
         for tsegment in self.timeline_segments:
-            if t == 0 and tsegment.t is not None:
+            # An explicit @t overrides the running time on any segment, not just
+            # the first one, which is how a SegmentTimeline signals a gap.
+            if tsegment.t is not None:
                 t = tsegment.t
             # check the start time from MPD
             for _ in range(tsegment.r + 1):
--- a/tests/resources/dash/test_timeline_gap.mpd
+++ b/tests/resources/dash/test_timeline_gap.mpd
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<MPD type="static" xmlns="urn:mpeg:dash:schema:mpd:2011"
+  minBufferTime="PT2S" mediaPresentationDuration="PT18S"
+  profiles="urn:mpeg:dash:profile:isoff-live:2011">
+  <Period id="0" start="PT0S">
+    <AdaptationSet id="0" mimeType="video/mp4" contentType="video">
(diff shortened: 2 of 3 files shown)


bastimeyer (approved):
Thanks for the PR.

It's a valid bug, even if discontinuities are not supported by Streamlink. The MPD must be followed correctly, or invalid segment requests will be made, causing actual errors.

The fix and unit test are looking fine...

----

ISO/IEC 23009-1:2022(E) Section 5.3.9.6.2:

> | Element or Attribute Name | Use | Description |
> | :-- | --- | :-- |
> | **SegmentTimeline** | | specifies the Segment timeline information |
> | **S** | 0..N | specifies Segment start time and duration for a contiguous sequence of segments or identical durations, referred to as series in the following.<br><br>Note: The **S** elements are ordered in sequence of increasing values of the attribute `@t`. |
> | `@t` | O | this value of this attribute minus the value of the `@presentationTimeOffset` specifies the MPD start time, in `@timescale` units, of the first Segment in the series. The MPD start time is relative to the beginning of the Period.<br><br>The value of this attribute shall be equal to or greater than the sum of the previous **S** element earliest presentation time and the sum of the contiguous Segment durations.<br><br>If the value of the attribute is greater than what is expressed by the previous **S** element, it expresses discontinuities in the timeline.<br><br>If not present, then the value shall be assumed to be zero for the first **S** element and for the subsequent **S** elements, the value shall be assumed to be the sum of the previous **S** element's earliest presentation time and contiguous duration. \[i.e. previous `S@t + @d * (@r + 1)`]. |

Issue #6472, opened 2025-03-16
stream.hls: sync first segments in muxed streams


bastimeyer:
Streamlink's HLS implementation currently doesn't sync substreams in muxed streams, meaning that there can be a time-offset at the start of the substreams, depending on the first access time of the involved media playlists, which can lead to either missing video or missing audio data at the beginning of the output stream until FFmpeg is able to sync the video/audio tracks, as it doesn't discard any video/audio track data of the earlier/longer substream(s).

The `HLSStreamWorker` threads need to synchronize and agree on a common `EXT-X-PROGRAM-DATE-TIME` value of the first segment that will be written to the buffer. `EXT-X-PROGRAM-DATE-TIME` is optional, but if it exists in a media playlist, then it must exist in all media playlists of their respective multivariant/master playlist.
https://datatracker.ietf.org/doc/html/rfc8216#section-6.2.4

The DASH implementation already synchronizes the first segments, but here, it's trivial.
https://github.com/streamlink/streamlink/blob/7.1.3/src/streamlink/stream/dash/dash.py#L123-L124

Resolving this issue might depend on #6341

PR #7083, opened 2026-09-03
stream.hls: add support for packed audio streams

### LINKED ISSUE #4721: stream.hls: packed audio support
Streamlink currently does not support "packed audio" HLS streams:
https://datatracker.ietf.org/doc/html/rfc8216#section-3.4

> A Packed Audio Segment contains encoded audio samples and ID3 tags
> that are simply packed together with minimal framing and no per-
> sample timestamps.  Supported Packed Audio formats are Advanced Audio
> Coding (AAC) with Audio Data Transport Stream (ADTS) framing
> [[ISO_13818_7](https://datatracker.ietf.org/doc/html/rfc8216#ref-ISO_13818_7)], MP3 [[ISO_13818_3](https://datatracker.ietf.org/doc/html/rfc8216#ref-ISO_13818_3)], AC-3 [[AC_3](https://datatracker.ietf.org/doc/html/rfc8216#ref-AC_3)], and Enhanced AC-3
> [[AC_3](https://datatracker.ietf.org/doc/html/rfc8216#ref-AC_3)].
>
> A Packed Audio Segment has no Media Initialization Section.
>
> Each Packed Audio Segment MUST signal the timestamp of its first
> sample with an [ID3](https://datatracker.ietf.org/doc/html/rfc8216#ref-ID3) Private frame (PRIV) tag [ID3] at the beginning of
> the segment.  The ID3 PRIV owner identifier MUST be
> "com.apple.streaming.transportStreamTimestamp".  The ID3 payload MUST
> be a 33-bit MPEG-2 Program Elementary Stream timestamp expressed as a
> big-endian eight-octet number, with the upper 31 bits set to zero.
> Clients SHOULD NOT play Packed Audio Segments without this ID3 tag.

----

Simply concatenating the segments and piping the data to FFmpeg does not work and the audio will be desynced (in addition to a long output delay), so additional stream output logic is required.

Example streams:
- https://mcdn.daserste.de/daserste/de/master.m3u8
- https://mcdn.daserste.de/daserste/de/master_audio1_128.m3u8

#4687, #4703, #3534

----

Implementing packed audio streams means

1. Detecting the format of the media stream. There's no metadata, so only file extensions of the segment URL path can be used.
   Related: https://github.com/streamlink/streamlink/issues/791#issuecomment-913222776
2. Implementing an ID3 parser. Unfortunately, there's no maintained package on PyPI for that.
   Spec: https://id3.org/id3v2.4.0-structure
3. Parsing and removing the ID3 data from the output stream.
4. Applying the parsed `com.apple.streaming.transportStreamTimestamp` timestamp to the first sample of the audio segment.

1-3 is fairly trivial, but 4 is the big question mark.

When opening packed audio streams directly, setting the timestamp is irrelevant, so just removing the ID3 data works fine. However, when muxing the video and audio streams, the audio timestamp is required, otherwise there will be an audio desync. Another problem implementation-wise is that parsing the timestamp can only be done after opening the audio stream, and at this point the FFmpeg process has already been spawned by the `MuxedHLSStream`, and its argv been set, so the timestamp is missing (also no idea how to make FFmpeg use the timestamp).

bastimeyer:
Resolves #4721 
Enabled by #7082, #7081, #7059

This adds support for HLS packed audio streams.

- [x] First commits adds the ID3v2 `PRIV` frame parser for `com.apple.streaming.transportStreamTimestamp`.
- [x] Second commit refactors buffer write calls, so both decrypted and non-encrypted segments can be parsed
- [x] Third commit adds the `packed_audio_timestamp` attribute to `HLSStream` and makes the `HLSStreamWriter` parse (and discard) the ID3v2 tags at the beginning of each segment. If no tags were found, then no further attempts will be made to parse in follow-up segments. This avoids unnecessary work on regular streams.
- [x] Fourth commit updates the `HLSMuxedStream` and makes it set the `itsoffset` option with the parsed audio stream timestamp offsets. To do that, it first needs to wait for all substreams to have data in their buffers. When data is available, this means that the ID3v2 metadata was parsed, so it can then properly be set in the FFmpeg argv. If data was not available in time, then no `itsoffset` option value can be set for those particular substreams, but the `open()` call won't fail because of that.

----

**One important note:** (see #6472)

Since there's no segment synchronization between the individual `HLSStream`s in a `HLSMuxedStream` (unlike in Streamlink's DASH implementation where there's only one data manifest), this means that depending on the access time of each media playlist, the first downloaded segment might not match ones from other playlists.

This means that the beginning of the overall stream output might miss data from the other sub-streams, either the video or the audio stream. FFmpeg will be fed with the data of all streams immediately, but as it receives the correct PTS values for packed audio streams, the streams will be aligned properly. The end result is that there will be no audio for a short time at the beginning if the video stream started eariler, or if the audio stream started earlier than the video stream, the player will show a black screen or will wait with its output.

However, even if #6472 were resolved, syncing HLS segments accross media playlists (if program date time metadata is available) doesn't guarantee that the video+audio streams are fully/correctly aligned. There might still be a slight offset, still causing a tiny unnoticable delay with the packed audio stream implementation.

### DIFF
--- a/src/streamlink/stream/hls/hls.py
+++ b/src/streamlink/stream/hls/hls.py
@@ -328,7 +328,7 @@ class HLSStreamWriter(SegmentedStreamWriter[HLSSegment, Response]):
             # we defer the buffer writes by one read call and apply the unpad call only to the last read call.
             decrypted_chunk = decryptor.decrypt(encrypted_chunk)
             chunk = unpad(decrypted_chunk, AES.block_size, style="pkcs7")
-            self.reader.buffer.write(chunk)
+            self._write_into_buffer(iter([chunk]))
         except ValueError as err:
             log.error(f"Error while decrypting segment {segment.num}: {err}")
             return False
@@ -337,13 +337,17 @@ class HLSStreamWriter(SegmentedStreamWriter[HLSSegment, Response]):
 
     def _write_plain(self, segment: HLSSegment, response: Response) -> bool:
         try:
-            for chunk in self.iter_segment_content(segment, response):
-                self.reader.buffer.write(chunk)
+            self._write_into_buffer(self.iter_segment_content(segment, response))
         except (ChunkedEncodingError, ContentDecodingError, ConnectionError) as err:
             log.error(f"Download of segment {segment.num} failed: {err}")
             return False
 
         return True
+
+    def _write_into_buffer(self, iterator: Iterator[bytes]) -> None:
+        buffer = self.reader.buffer
+        for chunk in iterator:
+            buffer.write(chunk)
 
     # noinspection PyMethodMayBeStatic
     def get_segment_content(self, segment: HLSSegment, response: Response) -> bytes:
--- a/src/streamlink/stream/hls/hls.py
+++ b/src/streamlink/stream/hls/hls.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 import re
 import struct
 import warnings
+from dataclasses import dataclass
 from datetime import timedelta
+from struct import unpack
 from typing import TYPE_CHECKING, Any, ClassVar, Literal, TypeVar
 from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


jordandalley:
Woo! Nice work @bastimeyer !!

Issue #6899, opened 2026-04-17
cli.argparser: replace `"store_{true,false}"` and custom `boolean` actions with `argparse.BooleanOptionalAction`


bastimeyer:
### Checklist

- [x] This is a feature request and not [a different kind of issue](https://github.com/streamlink/streamlink/issues/new/choose)
- [x] [I have read the contribution guidelines](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#contributing-to-streamlink)
- [x] [I have checked the list of open and recently closed feature requests](https://github.com/streamlink/streamlink/issues?q=is%3Aissue+label%3A%22feature+request%22)

### Collaboration

- [x] My request is genuine and in the best interest of the project
- [x] [I will provide feedback should a pull request be opened with a new feature implementation](https://github.com/streamlink/streamlink/blob/master/CONTRIBUTING.md#pull-request-feedback)

### Description

The stdlib's `argparse` module has added the [`BooleanOptionalAction`](https://docs.python.org/3/library/argparse.html#argparse.BooleanOptionalAction) in Python 3.9 which automatically adds `--no-*` CLI arguments for boolean arguments. Previously, this could only be achieved by adding duplicate negated args with the same `dest`, so this was never done.

Streamlink has lots of boolean CLI arguments which can't be unset once set, which is annoying, especially when they are set in the config file(s). For example the various stream transport arguments like `--player-http` and related ones. Replacing the `action="store_true"` / `action="store_false"` arguments with `BooleanOptionalAction` fixes that.

We also already have a couple of explicit `--no-*` arguments like `--no-plugin-sideloading` or `--no-plugin-cache` which would need to be modified accordingly.

And then there's our [custom `boolean` action](https://github.com/streamlink/streamlink/blob/8.3.0/src/streamlink/utils/args.py#L19-L23) which accepts `{yes,true,1,on,no,false,0,off}` values. Some boolean args use this action, but lots of others don't. It's a wild and inconsistent mix that should be fixed eventually.

Considering that the custom `boolean` arg accepts an argument value while `BooleanOptionalAction` or `store_{true,false}` do not (separate args), this will result in a breaking change if we change those arguments.

It's not important to fix the inconsistent CLI args now, but this should be kept in mind, so I'm opening this issue as a reminder now.

There are also a couple of args which could use some shorthand argument names, like `-H` for `--http-header` for example.

PR #7124, opened 2026-09-18
cli.argparser: add negative HTTP player options

dexhunter:

----

Add `--no-player-http`, `--no-player-continuous-http`, and `--no-player-external-http` so the command line can turn off HTTP player options enabled in configuration files. The three options use `argparse.BooleanOptionalAction` with explicit `False` defaults, preserving their existing default behavior.

This covers the three HTTP player options discussed in #6899. The remaining boolean-option conversions are outside this change.

Validation on Python 3.12: 187 parser/configuration tests, Ruff checks and formatting, ty, and mypy passed. HTML and man-page builds and Bash/Fish/Zsh completion generation passed; all three negative options appear in the generated outputs. The shallow checkout used the fallback version because it had no release tags.

[Weco validation trajectory](https://dashboard.weco.ai/share/v6v_RBX1qqk9MQZLbSpb5VKITK4hkP15)

AI assistance was used for the implementation, regression tests and this description.


bastimeyer:
Sorry, I am going to close this. For the following reasons:

1. We don't accept automatically generated and thus inauthentic PRs in order to "farm contributions" that are not meant to genuinely help out FOSS projects. Personally, I find generated PRs insulting to every project maintainer. This is a waste of my time reviewing this because of point 2.

2. This is not even half finished ([See "AI-assisted contributions - Functionality and intent"](https://github.com/streamlink/streamlink/blob/8.6.1/CONTRIBUTING.md#ai-assisted-contributions)).

   #6899 explicitly mentions that there are many more such CLI arguments. `--player-http` and related ones were just a few examples. Changing only a tiny subset of the problematic CLI args doesn't make any sense, as the result would be even more inconsistent, now with a wild mixture of `store_{true,false}`, our old custom `boolean` type, explicit `--no-*` or `--*-no-*` args, and now a few args with the `BooleanOptionalAction`.
   
   Just as an example, the `--{,no-}player-external-http` args and the `--player-external-http-continuous` arg right below it:
   https://streamlink--7124.org.readthedocs.build/cli.html#cmdoption-player-external-http

   The linked issue was opened as a reminder to **fix the inconsistencies**. Fixing these inconsistencies is not trivial, as each arg needs to be carefully checked and evaluated, and some args will need a deprecation phase first if they have to be renamed, e.g. `--player-no-close` or `--ffmpeg-no-validation`.

3. The added tests check lots of argparse logic which is not relevant. A simple positive+negative+default check would've sufficed, without positional or config file stuff, etc.

   There's also code branching involved in a try-except block, which doesn't belong in any tests, as tests must always have linear code execution, hence why we require 100% code coverage in our tests to guard against this.

   The reason why your LLM has added this try-except block seems to be the `ArgumentParser.exit_on_error` attr which defaults to `True`, but this should be disabled in the `parser` module-scope fixture anyway. `ArgumentParser.exit_on_error` support was added in Python 3.9. Streamlink's CLI code hasn't been updated since the EOL of Python 3.8 with the disabling of the default behavior of exiting on argument parsing error. Something else on the TO-DO list.

PR #7127, opened 2026-09-20
utils.args: add new Boolean argparse action

### LINKED ISSUE #6899: cli.argparser: replace `"store_{true,false}"` and custom `boolean` actions with `argparse.BooleanOptionalAction`

bastimeyer:
Resolves #6899 
Closes #7124

Not yet finished. Opening this as a draft. Will write a full description later. Done for now...


bastimeyer:
This will need at least a couple of more checks before merging.

The results however can be seen here:
https://streamlink--7127.org.readthedocs.build/cli.html#command-line-usage

With the deprecations listed here:
https://streamlink--7127.org.readthedocs.build/deprecations.html

